PR #10, opened 2016-03-10
Adds unit test support

aforcier:
Added unit test support with an example test for the SiteStore, using Robolectric.

Unit tests can be run separately from our regression tests, using `gradlew testDebug`, and are run by Travis.

Also fixed Travis and updated the README.


maxme:
Should we run `./gradlew test` on travis?

maxme:
:shipit:

PR #15, opened 2016-03-11
SiteStore implementation

aforcier:
Fills out SiteStore, mirroring all the methods for the `accounts` table of `WordPressDB`.

Not ready for merge yet.

Remaining:
- [x] The `getBlogsBy(...)` methods of `WordPressDB` are too generic, find all its usages and provide less generic methods for each so we can drop `getBlogsBy(...)` (or at least not have to expose it anymore)
- [x] Add unit tests
- [x] Add regression tests
- [x] Offload some of the SQL methods to `SiteSqlUtils`
- [x] Add `HIDE_SITES` and `SHOW_SITES` actions
- [x] Add any other actions that require more than SELECT queries
- [x] Return rows affected for UPDATE/DELETE queries and propagate through `onChanged()`

<h4>Implementation/Import details:</h4>
<h5>Methods implemented:</h5>
* `addBlog(Blog blog)` > already implemented as `SiteSqlUtils.insertOrUpdateSite(site)`
* `saveBlog(Blog blog)` > already implemented as `SiteSqlUtils.insertOrUpdateSite(site)`

* `instantiateBlogByLocalId(int localId)` > `getSiteByLocalId(int id)`

* `getNumBlogs()` > `getSitesCount()`
* `hasAnyJetpackBlogs()` > `hasJetpackSite()`
* `getNumDotComBlogs()` > `getDotComSitesCount()`
* `getVisibleBlogs()` > `getVisibleDotComSites()`
* `getNumVisibleBlogs()` > `getVisibleDotComSitesCount()`

* `setAllDotComBlogsVisibility(boolean visible)` > `setAllDotComSitesVisibility(boolean visible)`
* `setDotComBlogsVisibility(int id, boolean visible)` > `setDotComSiteVisibilityByLocalId(int id, boolean visible)`
* `isDotComBlogVisible(int blogId)` > `isDotComSiteVisibleByLocalId(int id)`

* `isBlogInDatabase(int blogId, String xmlRpcUrl)` > `hasSiteWithSiteIdAndXmlRpcUrl(int siteId, String xmlRpcUrl)`
* `isLocalBlogIdInDatabase(int localBlogId)` > `hasSiteWithSiteId(int id)`

* `isCurrentUserAdminOfRemoteBlogId(long remoteBlogId)` > `isCurrentUserAdminOfSiteId(int siteId)`

* `deleteBlog(Context ctx, int id)` > `deleteSite(SiteModel site)`
* `deleteWordPressComBlogs(Context ctx)` > handled by `SiteAction.DELETE_WPCOM_SITES`

<h5>Methods implemented reliant on the Jetpack WPCOM site id</h5>
* `getLocalTableBlogIdForJetpackRemoteID(int remoteBlogId, String xmlRpcUrl)` > `getLocalIdForJetpackRemoteId(int jetpackRemoteId, String xmlRpcUrl)`
* `getLocalTableBlogIdForRemoteBlogId(int remoteBlogId)` > `getLocalIdForRemoteSiteId(int siteId)`
* `getLocalTableBlogIdForRemoteBlogIdAndXmlRpcUrl(int remoteBlogId, String xmlRpcUrl)` > `getLocalIdForRemoteSiteIdAndXmlRpcUrl(int siteId, String xmlRpcUrl)`
* `getRemoteBlogIdForLocalTableBlogId(int localBlogId)` > `getSiteIdForLocalId(int id)`
* `isRemoteBlogIdDotComOrJetpack(int remoteBlogId)` > `hasDotComOrJetpackSiteWithSiteId(int siteId)`
* `getBlogForDotComBlogId(String dotComBlogId)` > `getSiteByDotComSiteId(String dotComSiteId)`

<h5>Methods not implemented for special reasons (future work required):</h5>
* `dangerouslyDeleteAllContent()`: this clears all tables, not only sites - we'll probably be implementing this at a higher level later on
* `deleteQuickPressShortcutsForLocalTableBlogId(Context ctx, int blogId)`: this acts on a different Store, added TODOs about handling this when a site is deleted

<h5>Methods not implemented because they're not necessary:</h5>
* `removeDotComCredentials()` // added to clear credentials from older DB versions (March 2015)
* `updateWPComCredentials(String username, String password)`: this is still called on `saveBlog()` in `WordPressDB`, but we won't be using this going forward (should have been removed at the same time as `removeDotComCredentials()`)
* `updateDotcomFlag()` // added when .COM flag identification changed to fix any mis-labelled sites (https://github.com/wordpress-mobile/WordPress-Android/pull/3548)
* `migrateWPComAccount()` // added for migration when global .COM account support was added: https://github.com/wordpress-mobile/WordPress-Android/commit/efccc0b6b7ef2305b0e9a98eb7c0c1fae5d09f8a
* `migratePasswords()` // added a very long time ago, migration-only
* `getWPCOMBlogID()` // unused, very old method
* `getVisibleDotComBlogs()` // unused, was added for Reader reblogging but that feature has been dropped: https://github.com/wordpress-mobile/WordPress-Android/pull/2836. In any case, performed the same function as `getVisibleBlogs()` (now performed by `getVisibleDotComSites()`), since we only support hiding .com sites
* `getAllBlogsIDs()` // unused
* `hasDotOrgBlogForUsernameAndUrl(String username, String url)` // unused, was used by SignInFragment and was removed here: https://github.com/wordpress-mobile/WordPress-Android/commit/b91838bf69aea8bd66359963a3125515ce89ec06
* `loadStatsLogin(int id)` // unused


WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
```diff
@@ -103,6 +424,17 @@ public void onAction(Action action) {
                 // TODO: check for WP-REST-API plugin and use it here
                 mSiteXMLRPCClient.pullSite(site);
             }
+        } else if (actionType == SiteAction.DELETE_SITE) {
```
maxme: I'm not sure about the word `DELETE`, `REMOVE` seems better to me. 

Related to this PR that call the `sites/delete` endpoint: https://github.com/wordpress-mobile/WordPress-Stores-Android/pull/2

WordPressStores/src/main/java/org/wordpress/android/stores/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -78,10 +78,15 @@ public void onErrorResponse(VolleyError error) {
 
     private SiteModel siteResponseToSiteModel(SiteWPComRestResponse from) {
         SiteModel site = new SiteModel();
-        site.setSiteId(from.ID);
+        if (from.jetpack) {
+            site.setDotComIdForJetpack(from.ID);
+        } else {
+            site.setSiteId(from.ID);
+        }
```
maxme: Now, I wonder if we should use: `dotOrgSiteId` and `siteId` instead of `dotComIdForJetpack` and `siteId` seems cleaner because `siteId` and `dotComIdForJetpack` will be used as the same id for REST calls when `dotOrgSiteId` will only be used on XMLRPC calls.

WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
aforcier: Agreed, fixed in https://github.com/wordpress-mobile/WordPress-Stores-Android/commit/43c34da815877ebc4a7c93c30609fb77a89c1636.

WordPressStores/src/main/java/org/wordpress/android/stores/network/rest/wpcom/site/SiteRestClient.java
aforcier: I've come to the same conclusion. We have cases like [`getBlogsBy()` from `WordPressDB`](https://github.com/wordpress-mobile/WordPress-Android/blob/ad5b5fa949b81a503727b6d069375f0a678764ca/WordPress/src/main/java/org/wordpress/android/WordPressDB.java#L540-L559), which selects based on certain criteria, but by default ignores any Jetpack sites that were only downloaded over the REST API.

Right now, on WPAndroid, we filter those out by checking that they're not .COM and have a blank password; I want to switch to instead checking that they are Jetpack AND they have no `siteId`. With the change you suggested this would be much clearer: instead, they are Jetpack sites with no `dotOrgSiteId`, and it's much less ambiguous what that means.

aforcier:
Updated.

* Use `siteId`/`dotOrgSiteId` instead of `siteId`/`dotComIdForJetpack`
* Used 'Remove Site' instead of 'Delete Site' language
* Added methods to complete coverage of `getBlogsBy` from `WordPressDB`, allowing us not to implement those methods directly as they're too generic

`getBlogsBy` migration:
* `getBlogsBy(null, extraFields)` > `getSites()`
* `getBlogsBy("dotcomFlag=1", extraFields)` > `getDotComSites()`
* `getBlogsBy("isHidden=0", extraFields)` > `getVisibleSites()`
* `getBlogsBy("isHidden=0 AND dotcomFlag=1", extraFields)` > `getVisibleDotComSites()`
* `getBlogsBy("dotcomFlag=0 OR (isHidden=0 AND dotcomFlag=1) ", extraFields)` > `getVisibleSites()`

WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
```diff
@@ -55,27 +71,328 @@ public void onRegister() {
...
+    /**
+     * Checks whether the store contains a self-hosted site matching the given (remote) site id and XML-RPC URL.
+     */
+    public boolean hasDotOrgSiteWithSiteIdAndXmlRpcUrl(long dotOrgSiteId, String xmlRpcUrl) {
         return WellSql.select(SiteModel.class)
-                .where().equals(SiteModelTable.IS_WPCOM, 0).endWhere()
-                .getAsCursor().getCount() != 0;
+                .where().beginGroup()
+                .equals(SiteModelTable.DOT_ORG_SITE_ID, dotOrgSiteId)
+                .equals(SiteModelTable.X_MLRPC_URL, xmlRpcUrl)
```
maxme: `X_MLRPC_URL` -> `XMLRPC_URL `

WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
```diff
@@ -55,27 +71,328 @@ public void onRegister() {
...
+     * Returns the number of visible .COM sites.
+     */
+    public int getVisibleDotComSitesCount() {
+        return getVisibleDotComSites().size();
+    }
+
+    /**
+     * Sets the visibility of all .COM sites to the given value.
+     */
+    public void setAllDotComSitesVisibility(boolean visible) {
```
maxme: Should be private, the store should be responsible for changing site states.

WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
```diff
@@ -55,27 +71,328 @@ public void onRegister() {
...
+                        cv.put(SiteModelTable.IS_VISIBLE, item);
+                        return cv;
+                    }
+                }).execute();
+    }
+
+    /**
+     * Sets the visibility of the .COM site with the given (local) id to the given value.
+     */
+    public void setDotComSiteVisibilityByLocalId(int id, boolean visible) {
```
maxme: Should be private too

WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
```diff
@@ -111,4 +450,10 @@ private void createOrUpdateSites(SitesModel sites) {
             SiteSqlUtils.insertOrUpdateSite(site);
         }
     }
+
+    private void deleteSites(List<SiteModel> sites) {
```
maxme: not a big deal since it's internal only but we could rename this method to `removeSites`

maxme:
We should decide which one to use: `self hosted` vs `dot org` and `wpcom` vs `dotcom` and probably use checkstyle to enforce them

WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
maxme: Actually, renaming `mXMLRpcUrl` to `mXmlRpcUrl` (which is the correct naming) fixes the name.

aforcier: That seems to be how WellSql wants to generate the name for that field in the table. We can use `mXmlrpcUrl` in the `SiteModel` to end up with `XMLRPC_URL` in the `SiteModelTable`.

WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
aforcier: Fixed by dropping this method and relying on the single-site method.

WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
aforcier: Fixed (moved to utils file, call from `SiteStore` is now in a private method).

WordPressStores/src/main/java/org/wordpress/android/stores/store/SiteStore.java
maxme: No, let's use `mXmlRpcUrl`,  `XML_RPC_URL` is good enough, note we can force the field name by adding `@Column(name = "XMLRPC_URL")`

Note about naming: https://source.android.com/source/code-style.html#follow-field-naming-conventions see section "Treat Acronyms as Words"

aforcier: Ah, your second comment hadn't loaded for me when I made my comment. This is fixed in https://github.com/wordpress-mobile/WordPress-Stores-Android/commit/557ac1f30ca426a46777d704bfcdd0fd45f8e963.

example/src/androidTest/java/org/wordpress/android/stores/release/ReleaseStack_SiteTest.java
```diff
@@ -153,16 +155,44 @@ public void testSelfHostedHTTPAuthFetchSites2() throws InterruptedException {
...
+    public void testSelfHostedFetchAndDeleteSite() throws InterruptedException {
+        RefreshSitesXMLRPCPayload payload = new RefreshSitesXMLRPCPayload();
+        payload.username = BuildConfig.TEST_WPORG_USERNAME_SH_SIMPLE;
+        payload.password = BuildConfig.TEST_WPORG_PASSWORD_SH_SIMPLE;
+        payload.xmlrpcEndpoint = BuildConfig.TEST_WPORG_URL_SH_SIMPLE;
+        mExpectedEvent = TEST_EVENTS.SITE_CHANGED;
+        mDispatcher.dispatch(SiteAction.FETCH_SITES_XMLRPC, payload);
+        mCountDownLatch = new CountDownLatch(1);
+        // Wait for a network response / onChanged event
+        mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
maxme: We should add an `assert` on `await` return, test was failing with a weird Exception due to a timeout.

aforcier: Fixed in https://github.com/wordpress-mobile/WordPress-Stores-Android/commit/7f30f2f0f7dd5a3de917bfc63dbc88468cc91af4.

aforcier:
Added regression tests and fully implemented `LOGOUT_WPCOM` (was `REMOVE_WPCOM_SITES`_. The last main item left is to propagate rows affected for update/delete operations, and update affected tests. Waiting for #21 to be merged, which fixes WellSql internals allowing this to be done easily.

aforcier:
Merged in `develop` (`1.0.6 WellSql`) and updated to return rows affected for UPDATE/DELETE queries and propagate through onChanged().

Ready for another pass @maxme!

maxme:
:shipit:

PR #19, opened 2016-03-12
Support self signed SSL certificates

maxme:
Support self signed SSL certificates using a `MemorizingTrustManager`. 

I first wanted to use: https://github.com/ge0rg/MemorizingTrustManager - but it doesn't fit our needs and they're mixing UI code and KeyStore management which is probably a bad idea. I just kept the name ;).

This was branched from `issue/9-http-auth-support2` so it's better to wait for https://github.com/wordpress-mobile/WordPress-Stores-Android/pull/12 to be merged before reviewing this one.


maxme:
There is one thing I don't like: I have to keep the state of the [last error](https://github.com/wordpress-mobile/WordPress-Stores-Android/blob/issue/13-self-signed-ssl/WordPressStores/src/main/java/org/wordpress/android/stores/network/MemorizingTrustManager.java#L146) because there is no way to get the certificate when I encounter the exception for the first time.

example/src/androidTest/java/org/wordpress/android/stores/release/ReleaseStack_SiteTest.java
```diff
@@ -168,8 +179,11 @@ public void onAuthenticationChanged(OnAuthenticationChanged event) {
             AppLog.i(T.TESTS, "error " + event.authError);
             if (event.authError == AuthError.HTTP_AUTH_ERROR) {
                 assertEquals(mExpectedEvent, TEST_EVENTS.HTTP_AUTH_ERROR);
-                mCountDownLatch.countDown();
+            }
+            if (event.authError == AuthError.INVALID_SSL_CERTIFICATE) {
+                assertEquals(mExpectedEvent, TEST_EVENTS.INVALID_SSL_CERTIFICATE);
```
aforcier: Missed this in the HTTP auth PR, but we should be using `assertEquals(TEST_EVENTS.EVENT, mExpectedEvent)` rather than the other way around. I suppose `mExpectedEvent` naming is a little confusing, maybe `mNextEvent` would convey it better, but it's not a big issue.

aforcier:
Getting a crash on login, when I accept the certificate, on a fresh app install:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'void java.security.KeyStore.setCertificateEntry(java.lang.String, java.security.cert.Certificate)' on a null object reference
	at org.wordpress.android.stores.network.MemorizingTrustManager.storeCert(MemorizingTrustManager.java:131)
	at org.wordpress.android.stores.network.MemorizingTrustManager.storeLastFailure(MemorizingTrustManager.java:126)
	at org.wordpress.android.stores.example.MainExampleActivity$5.onClick(MainExampleActivity.java:163)
	at org.wordpress.android.stores.example.SSLWarningDialog$1.onClick(SSLWarningDialog.java:58)
	at android.support.v7.app.AlertController$ButtonHandler.handleMessage(AlertController.java:157)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:135)
	at android.app.ActivityThread.main(ActivityThread.java:5254)
	at java.lang.reflect.Method.invoke(Method.java:372)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
````

Doesn't happen if I re-open the app and try to login again.

File: WordPressStores/src/main/java/org/wordpress/android/stores/network/MemorizingTrustManager.java [hunk 2]
```diff
@@ -0,0 +1,162 @@
...
+    private Context mContext;
+
+    public MemorizingTrustManager(Context appContext) {
+        mContext = appContext;
+        try {
+            mLocalKeyStore = loadTrustStore();
+        } catch (FileNotFoundException e) {
+            // Init the key store for the first time
+            try {
+                initLocalKeyStoreFile();
```
aforcier: `mLocalKeyStore` needs to be initialized by this call; otherwise, the app crashes if you attempt to log into a self-signed SSL site on the first run of the app (this causes the crash in my other comment).

maxme:
Ready for second round.

aforcier:
:shipit:

PR #22, opened 2016-03-15
Fix maven artifact publishing

maxme:
This also makes the package almost ready to be published in bintray repo.


aforcier:
Looks good, lint is still happy (or no less unhappy) with `minSdk` 14. :shipit:

PR #28, opened 2016-03-17
AccountStore sign out v2

maxme:


tonyr59h:
Some of my commits in there, but I'm gonna assume if there was any issue it would have been mentioned in the OP or previous PR ;) and :shipit:

Issue #26, opened 2016-03-16
Improve XML-RPC endpoint discovery (including redirects handling)


aforcier:
Currently, `HTTP`>`HTTPS` (presumably also `HTTP`>`HTTP`) redirects are followed, but there's an issue when `POST`ing to `XML-RPC`: the `POST` isn't re-submitted, and a `GET` request is made instead, which results in a `405` error.

This is following HTTP guidelines, which try to limit cases of submitting multiple `POST`s, or submitting `POST`s to other URLs, without the user's explicit consent.

On WPAndroid we probe for the `rsd` file with `GET` first. We should do the same for Stores, as we improve XML-RPC endpoint discovery in general (see #4).

maxme:
We should reuse the [discovery code from wpandroid ](https://github.com/wordpress-mobile/WordPress-Android/blob/b34e383dfdbdb420afc848005778bf6313abc351/WordPress/src/main/java/org/xmlrpc/android/XMLRPCUtils.java#L316)

maxme:
Make sure to add a "real life" redirect test to `tests.properties` for instance http://bia.is/wpbeta redirects https://bia.is/wpbeta

aforcier:
Fixed in #33.

PR #33, opened 2016-03-18
#26: Improve self hosted endpoint discovery

aforcier:
Addresses #26.

Imports XML-RPC discovery methods from WPAndroid, wrapping them in a `SelfHostedEndpointFinder` class which is run at the sign in stage (maintaining provisions previously present for future WP-API implementation and discovery).

To-do list:
* [x] Import recent changes to XML-RPC discovery from WPAndroid
* [x] Tidy up error reporting
* [x] Test discovery with SSL (partially done: works for actually logging in, but I'm having issues with regression tests)
* [x] <s>Test</s> Fix discovery with HTTP auth
* [x] Test all site types against URLs without a protocol
* [x] Test all site types against URLs with an invalid protocol (e.g. `hppt://`)
* [x] Correctly handle being given a `*/xmlrpc.php` endpoint URL by the user, with an HTTP>HTTPS redirect (not handled by WPAndroid at the time this PR was opened)
* [x] Test that if a `*.wordpress.com` URL is given on login the user is correctly logged into WordPress.com instead
* [ ] Add tests for sites requiring 'advanced' discovery (e.g. rsd lookup)
* [x] Improve recognition of invalid URLs (e.g. something like `notaurl&*@` should be recognized early as not a real URL, and not go through the entire discovery process before being rejected as currently happens)


example/src/main/java/org/wordpress/android/stores/example/MainExampleActivity.java
```diff
@@ -227,11 +227,20 @@ private void signInAction(final String username, final String password, final St
         if (TextUtils.isEmpty(url)) {
             wpcomFetchSites(username, password);
         } else {
-            SelfHostedDiscoveryUtils.discoverSelfHostedEndPoint(url, new DiscoveryCallback() {
+            mSelfhostedPayload = new RefreshSitesXMLRPCPayload();
```
aforcier: I'm wondering if we should rename `RefreshSitesXMLRPCPayload` to something more generic since it's now used both for refreshing sites and for discovery/login.

aforcier:
Note: with this PR `tests.properties` needs to be updated locally. Before, we had:

```
TEST_WPORG_URL_SH_SIMPLE = http://something.com/xmlrpc.php
```

Now, it should be:

```
TEST_WPORG_URL_SH_SIMPLE = http://something.com
TEST_WPORG_URL_SH_SIMPLE_ENDPOINT = http://something.com/xmlrpc.php
```

The site/login tests now rely on the `_ENDPOINT` URLs since they're bypassing discovery, while discovery tests will be using the non-`_ENDPOINT` URLs.

example/src/androidTest/java/org/wordpress/android/stores/release/ReleaseStack_DiscoveryTest.java
```diff
@@ -0,0 +1,395 @@
...
+                    }
+                });
+        // Wait for a network response / onChanged event
+        assertEquals(true, mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
+    }
+
+    public void testWordPressCOMUrlFetchSites() throws InterruptedException {
+        final RefreshSitesXMLRPCPayload payload = new RefreshSitesXMLRPCPayload();
+        payload.username = BuildConfig.TEST_WPCOM_USERNAME;
+        payload.password = BuildConfig.TEST_WPCOM_PASSWORD;
```
maxme: Did you mean: TEST_WPCOM_USERNAME_TEST1 and TEST_WPCOM_PASSWORD_TEST1?

example/src/androidTest/java/org/wordpress/android/stores/release/ReleaseStack_DiscoveryTest.java
```diff
@@ -0,0 +1,395 @@
...
+                        mDispatcher.dispatch(SiteActionBuilder.newFetchSitesXmlRpcAction(payload));
+                    }
+                });
+
+        // Wait for a network response / onChanged event
+        assertEquals(true, mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
+    }
+
+    private static String addBadProtocolToUrl(String url) {
+        return "hppt://" + UrlUtils.removeScheme(url);
```
maxme: ❤️

example/src/androidTest/java/org/wordpress/android/stores/release/ReleaseStack_DiscoveryTest.java
aforcier: I did! I had some leftover properties in my `gradle.properties` and didn't notice that. Fixed in 937f62f.

example/tests.properties-example
```diff
@@ -7,30 +7,42 @@ TEST_WPCOM_BAD_PASSWORD = FIXME
 TEST_WPORG_USERNAME_SH_SIMPLE = FIXME
 TEST_WPORG_PASSWORD_SH_SIMPLE = FIXME
 TEST_WPORG_URL_SH_SIMPLE = FIXME
+TEST_WPORG_URL_SH_SIMPLE_ENDPOINT = FIXME
 
 # Simple Self hosted HTTP - contributor role
 TEST_WPORG_USERNAME_SH_SIMPLE_CONTRIB = FIXME
 TEST_WPORG_PASSWORD_SH_SIMPLE_CONTRIB = FIXME
 TEST_WPORG_URL_SH_SIMPLE_CONTRIB = FIXME
```
maxme: `TEST_WPORG_URL_SH_SIMPLE_CONTRIB` is not used anymore

example/tests.properties-example
```diff
@@ -7,30 +7,42 @@ TEST_WPCOM_BAD_PASSWORD = FIXME
...
+TEST_WPORG_URL_SH_SIMPLE_CONTRIB_ENDPOINT = FIXME
 
 # Multi site Self Hosted
 TEST_WPORG_USERNAME_SH_MULTISITE = FIXME
 TEST_WPORG_PASSWORD_SH_MULTISITE = FIXME
 TEST_WPORG_URL_SH_MULTISITE = FIXME
```
maxme: `TEST_WPORG_URL_SH_MULTISITE` is not used anymore

example/tests.properties-example
aforcier: Removed in 2203ca6, along with `TEST_WPORG_URL_SH_MULTISITE`.

example/src/main/java/org/wordpress/android/stores/example/MainExampleActivity.java
```diff
@@ -227,11 +227,20 @@ private void signInAction(final String username, final String password, final St
...
+                public void onError(DiscoveryError error, String lastEndpoint) {
+                    if (error == DiscoveryError.WORDPRESS_COM_SITE) {
                         wpcomFetchSites(username, password);
+                    } else if (error == DiscoveryError.HTTP_AUTH_REQUIRED) {
+                        showHTTPAuthDialog(lastEndpoint);
+                    } else if (error == DiscoveryError.ERRONEOUS_SSL_CERTIFICATE) {
+                        mSelfhostedPayload.xmlrpcEndpoint = lastEndpoint;
+                        showSSLWarningDialog(mMemorizingTrustManager.getLastFailure().toString());
                     }
                     AppLog.e(T.API, "Discover error: " + error);
```
maxme: can we also log the error using `prependToLog`?

aforcier: Done in eb33fd1!

aforcier:
Updated to use two new `AuthenticationAction`s, `DISCOVER_ENDPOINT` and `DISCOVERY_RESULT`, instead of callback methods for discovery. Also switched to using an `AsyncTask` instead of a `Thread` for discovery, to force the dispatched action to be called from the calling thread (in the example app, the UI thread).

aforcier:
Did a little refactor in https://github.com/wordpress-mobile/WordPress-Stores-Android/pull/33/commits/e43f6f80c67571ec66a4fb72a10f24c395bdacf2, I think `AccountStore.onDiscoveryCompleted()` works better as two classes, distinguishing a successful discovery event from a failed one.

I thought of doing the same to `DiscoveryResultPayload`, which would mean having two `AuthenticationAction`s, `DISCOVERY_SUCCEEDED` and `DISCOVERY_FAILED`, instead of `DISCOVERY_RESULT`. I'm not sure the extra complexity of doing this would be worth it, considering this is an 'internal' event (the API consumer never has to send or receive `DISCOVERY_RESULT`). It will also complicate the [AsyncTask](https://github.com/wordpress-mobile/WordPress-Stores-Android/pull/33/files#diff-8068974e9e871f8a9c64f997c5e113d1R93).

aforcier:
Merged in event bus changes, dropped the AsyncTask and went back to a Thread for discovery, and updated the tests. Also ran the linter and did some cleanup in SelfHostedEndpointFinder.

Ready for another look @maxme 🎉

maxme:
Filled a new issue https://github.com/wordpress-mobile/WordPress-Stores-Android/issues/62

maxme:
:shipit: :shipit: :shipit:

PR #34, opened 2016-03-18
Two step login for .com

### LINKED ISSUE #23: 2FA support for wordpress.com
See how it's done in [wpandroid](https://github.com/wordpress-mobile/WordPress-Android/blob/2a57008f7db7414a0ead50c7f1895878a906bda5/WordPress/src/main/java/org/wordpress/android/ui/accounts/helpers/LoginWPCom.java#L61-L69)

tonyr59h:
Fixes #23 

Adds support for .com accounts to login with 2FA enabled. In the example, if detected, another EditText will be shown for the authentication code.

Before merge:
- [x] Unit tests...eh, perhaps later
- [x] Integration tests


maxme:
Renaming done

maxme:
I'm not sure to understand why we need `redirectUri` here, can we drop it?

maxme:
I think something went wrong during a merge, we should only have url ending by `_ENDPOINT` in `ReleaseStack_SiteTest` (other URLs are used in Discovery tests).

tonyr59h:
> I'm not sure to understand why we need redirectUri here, can we drop it?

I included it when porting over some WPAndroid code that used it. It's unused though, removing it should be fine.

tonyr59h:
> I'm not sure to understand why we need redirectUri here, can we drop it?

Dropped in 15d02bdbd73146f2b37ba9a7cc123549e406ea33

example/src/main/java/org/wordpress/android/fluxc/example/MainExampleActivity.java
```diff
@@ -66,6 +65,10 @@
     // Would be great to not have to keep this state, but it makes HTTPAuth and self signed SSL management easier
     private RefreshSitesXMLRPCPayload mSelfhostedPayload;
 
+    // used for 2fa
+    private String mLastUsername;
```
maxme: I think we should remove `mLastUsername` and `mLastPassword` and follow the same way we di for `RefreshSitesXMLRPCPayload`: keep the state of the `AuthenticatePayload`.

tonyr59h: Probably cleaner that way 👍 

> Would be great to not have to keep this state, but it makes HTTPAuth and self signed SSL management easier

Along the lines of this comment, would it be feasible to include payloads in `OnChange` events? That would mean emitters/consumers wouldn't have to track state so much.

maxme:
:shipit:

PR #35, opened 2016-03-30
Issue #20: Attach payload information to actions

### LINKED ISSUE #20: Attach payload information to actions
With the way Actions are currently structured, there's no simple way to know when calling `Dispatcher.createAction` if a certain action needs a payload attached, and what type the payload should be.

One idea would be to create an annotation that generates a `ThingActionBuilder` (working title) for every item in a`ThingAction` enum. The annotation declares whether the item needs to be accompanied by a payload, and what type that payload should take. Then, instead of calling `Dispatcher.createAction(ThingAction.FETCH_THINGS, arbitraryPayload)`, we would have to call something like `Dispatcher.createAction(ThingActionBuilder.createFetchThingsAction(stronglyTypedPayload))`.

Before sinking time into this, we could try manually creating `ThingActionBuilder` and make sure it would actually make things simpler/clearer.

aforcier:
Fixes #20, by using an annotation processor to parse annotated actions and generate builder classes with methods specific to those actions.

For example,

```java
@ActionEnum
public enum AuthenticationAction implements IAction {
    @Action(payloadType = AccountStore.AuthenticatePayload.class)
    AUTHENTICATE,
}
```

results in:
```java
public final class AuthenticationActionBuilder extends ActionBuilder {
	public static Action<AccountStore.AuthenticatePayload> generateAuthenticateAction(AccountStore.AuthenticatePayload payload) {
		return new Action<>(AuthenticationAction.AUTHENTICATE, payload);
	}
}
```

<h4>TODO:</h4>
* [x] Add annotations to `AccountAction` and drop `AccountActionBuilder`
* [x] Add Javadoc comments
* [x] ~~Update the README with Action annotation usage~~


aforcier:
Ready for review now @maxme!

I changed my mind about adding something to the README about the action annotations, as I think it might be too much detail for a README and I feel their usage is clear enough by example. Let me know if you disagree 😉.

stores-processor/src/main/java/org/wordpress/android/stores/processor/CodeGenerationUtils.java
```diff
@@ -0,0 +1,21 @@
+package org.wordpress.android.stores.processor;
+
+public class CodeGenerationUtils {
+
+    public static String getActionBuilderMethodName(AnnotatedAction annotatedAction) {
+        return "generate" + CodeGenerationUtils.toCamelCase(annotatedAction.getActionName()) + "Action";
```
maxme: Nitpick but IMO `AuthenticationActionBuilder.generateAuthenticateAction(payload)`

maybe `AuthenticationActionBuilder.authenticate(payload)` or eventually `AuthenticationAction.buildAuthenticate(payload)` (we will have to rename the `XAction` enums)

aforcier:
Updated to use a `ThingBuilder.newSomethingAction` convention instead of `ThingBuilder.generateSomethingAction` for brevity, as discussed on Slack.

maxme:
Pouet

PR #44, opened 2016-06-29
Make the library buildable by jitpack

maxme:


aforcier:
Everything still works, tests run, and I see warnings if OAUTH.APP.ID and SECRET are unset and if no `tests.properties` has been set. Also travis is happy 🎉.

:shipit:

PR #47, opened 2016-07-01
Rename `LOGOUT_WPCOM` to `REMOVE_WPCOM_SITES`

maxme:
`LOGOUT_WPCOM` was a bad name because: 1. we have a `SIGNOUT` action in AccountAction 2. It's not obvious what happen in the SiteStore when we dispatch it.

`REMOVE_WPCOM_SITES` seems like a better name.


example/src/test/java/org/wordpress/android/stores/SiteStoreUnitTest.java
```diff
@@ -281,7 +281,7 @@ public void testGetRestApiSites() {
         SiteSqlUtils.insertOrUpdateSite(jetpackSiteOverDotOrg);
         SiteSqlUtils.insertOrUpdateSite(jetpackSiteOverRestOnly);
 
-        List<SiteModel> restApiSites = SiteSqlUtils.getAllRestApiSites();
+        List<SiteModel> restApiSites = SiteSqlUtils.getAllWPComSites();
```
aforcier: Might as well rename the entire method to `testGetWPComSites`, and `restApiSites` to `wpcomSites`,  to completely eliminate `restApiSite`-type names from the codebase.

maxme: Names updated in  	c44e769

aforcier:
:shipit:

maxme:
Thanks for d2ebe38 !

PR #49, opened 2016-07-04
Update "Post Settings" network response

maxme:
The "Post Settings" network call response depends of what was changed in the request. For instance if you update `display_name`, the response will only contains the updated `display_name`, that makes things a bit more complicated with Gson response, because we don't know which field are updated+empty and which fields are just not updated. Solution is to use a generic HashMap (what we wanted to avoid :() and only update the field(s) contained in the response.


WordPressStores/src/main/java/org/wordpress/android/stores/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -34,6 +35,15 @@ public AccountRestPayload(AccountModel account, VolleyError error) {
         public AccountModel account;
     }
 
+    public static class AccountPostResponsePayload implements Payload {
```
aforcier: Would we extend `AccountPostResponsePayload` in the future to handle things like the response to POSTs to `/me/two_step/` and `/me/connected-applications`? Otherwise I'm wondering if `AccountPostSettingsResponsePayload` would be more accurate.

WordPressStores/src/main/java/org/wordpress/android/stores/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -149,6 +161,30 @@ private AccountModel responseToAccountSettingsModel(AccountSettingsResponse from
         accountSettings.setNewEmail(from.new_user_email);
         accountSettings.setPendingEmailChange(from.user_email_change_pending);
         accountSettings.setWebAddress(from.user_URL);
+        accountSettings.setDisplayName(from.display_name);
         return accountSettings;
     }
+
+    public static boolean updateAccountModelFromGenericResponse(AccountModel accountModel, Map<String, Object> from) {
```
aforcier: A few of the keys in this list exist in GET `/me/settings/` but not in POST, so they wouldn't be sent as a response to a [POST `/me/settings/` call](https://developer.wordpress.com/docs/api/1.1/post/me/settings/). I diffed the two, here are the differences:

In GET but not in POST:

```
user_login
new_user_email
avatar_URL
two_step_enabled
two_step_sms_disabled
```

In POST but not in GET:
```
password (always an empty string in the response to a POST action)
```

maxme:
Checked the documentation of GET /me, GET /me/setting and POST /me/settings and cleaned things up in e356c9d - sorry for the ugly diff, I reordered field names the same way they're ordered in the documentation.

aforcier:
:shipit:

PR #50, opened 2016-07-04
Save default account indexed by a predefined local id

maxme:
Pretty bad bug: the default account wasn't always saved with the default id. (It's `1` as default id, since the account is inserted just after table creation, so the bug doesn't always show up).


aforcier:
:shipit:

PR #51, opened 2016-07-05
Stores are now Singletons and are auto registered on instantiation

maxme:
This allows us to drop registration from activities/fragments


aforcier:
Integration tests fail with:

```
 java.lang.IllegalArgumentException: Object already registered.
        at com.squareup.otto.Bus.register(Bus.java:222)
```

We need to drop `mDispatcher.register(mSomeStore)` from the tests also, like you did for `MainExampleActivity`.

aforcier:
Tests are working now, :shipit:!

PR #54, opened 2016-07-06
Fix #29: CREATE_NEW_USER action

### LINKED ISSUE #29: AccountStore: VALIDATE_NEW_USER and NEW_USER actions
To let new users signup, see https://github.com/wordpress-mobile/WordPress-Android/blob/23d7d289e2c0b39d70008a6bddad8ce182eeb814/WordPress/src/main/java/org/wordpress/android/ui/accounts/helpers/CreateUserAndBlog.java#L112-L150

maxme:
Fix #29: new action: CREATE_NEW_USER  with a dry run mode to validate parameters before creation

Also updated the example app so it will be easier to add new buttons and merge changes (simple linear layout + scrollview)


WordPressStores/src/main/java/org/wordpress/android/stores/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -124,6 +145,104 @@ public void onErrorResponse(VolleyError error) {
...
+                payload.errorMessage = (String) errorObj.get("message");
+            } catch (JSONException e) {
+                // Do nothing (keep default error)
+            }
+        }
+        return payload;
+    }
+
+    private NewUserError errorStringToErrorType(String error) {
+        if (error.equals("username_only_lowercase_letters_and_numbers")) {
```
aforcier: This is nitpicky, but since the enum values are the same as the error strings but in uppercase, why not modify `NewUserError` with a `fromString`-type method, returning `GENERIC_ERROR`, if the string doesn't match an ENUM? Then we could drop this whole method.

maxme: Yes that's a good idea

WordPressStores/src/main/java/org/wordpress/android/stores/action/AccountAction.java
```diff
@@ -10,8 +10,6 @@
     POSTED_SETTINGS, // response received from Account Settings post
     UPDATE, // update in-memory and persisted Account in AccountStore
     SIGN_OUT, // delete persisted Account, reset in-memory Account, delete access token
-    VALIDATE_NEW_ACCOUNT, // validate new account informations
-    VALIDATED_NEW_ACCOUNT, // validate new account response
-    CREATE_NEW_ACCOUNT, // create a new account
+    CREATE_NEW_ACCOUNT, // create a new account (can be used to validate the account before creating it)
```
aforcier: 👍

WordPressStores/src/main/java/org/wordpress/android/stores/network/rest/wpcom/account/AccountRestClient.java
maxme: Changed in d8dee99

aforcier:
Looks good, :shipit:!

PR #59, opened 2016-07-08
Cleanup authentication errors and auth error parsing

maxme:
This will allow us to finish the sign in screen wiring in wpandroid.


WordPressStores/src/main/java/org/wordpress/android/stores/store/AccountStore.java
```diff
@@ -88,6 +89,43 @@ public NewAccountPayload(@NonNull String username, @NonNull String password, @No
...
+        UNSUPPORTED_RESPONSE_TYPE,
+        UNKNOWN_TOKEN,
+
+        // From response's "message" field - sadly... (be careful with i18n)
+        INCORRECT_USERNAME_OR_PASSWORD,
+
+        // .org specifics
+        INVALID_SSL_CERTIFICATE,
+        HTTP_AUTH_ERROR,
+        NOT_AUTHENTICATED,
```
maxme: I wonder if we should add an "INVALID_URL" here for .org sites and have all sign in errors (kind of) centralized here.

maxme:
Ready for review.

WordPressStores/src/main/java/org/wordpress/android/stores/store/AccountStore.java
maxme: Not seeing any usage of `INVALID_URL`, so no we won't add it here since this kind of errors will be caught by the Discovery tool.

aforcier:
:shipit:

PR #61, opened 2016-07-13
Replace Otto with EventBus v3 (which supports thread selection on delivery)

maxme:


aforcier:
The integration tests are broken, they need to be updated with the new EventBus import.

maxme:
Test fixed in 
	dec203e

aforcier:
Nothing seems to be broken, the thread selection seems to be working as expected, and the tests are running with no appreciable speed difference from before. :shipit:

PR #64, opened 2016-07-18
Fix #63: drop `AccountStore.isSignedIn` method

### LINKED ISSUE #63: AccountStore.isSignedIn() only looks at .com visible sites
AccountStore.isSignedIn() should return true if user has an auth token OR a visible site (.org or .com).

maxme:
Fix #63: drop `AccountStore.isSignedIn` method, this should be managed by the app or eventually by an utility method at the library level but not by the AccountStore itself.


aforcier:
:shipit:

PR #72, opened 2016-07-19
fix #71: Make SiteModel serializable

### LINKED ISSUE #71: Make SiteModel serializable
So we can pass it from activities to activities / fragments instead of an abstract "local blog id".

maxme:
fix #71: Make SiteModel serializable


aforcier:
Good call, :shipit:

PR #73, opened 2016-07-19
Fix #68: Add timezone to SiteModel and update the value from REST or XMLRPC request

### LINKED ISSUE #68: Add timezone to SiteModel
And fetch it from `getOptions` call. It's used by stats [here](https://github.com/wordpress-mobile/WordPress-Android/blob/92fc6cdd9ad8b21a44f434df939a5614456c961c/WordPress/src/main/java/org/wordpress/android/ui/stats/StatsUtils.java#L139-L156).

maxme:
Fix #68: I'm not sure why, but that value is always the empty string on .com REST


aforcier:
I can confirm that the `timezone` value seems to always be the empty string for .com, that's odd.

Anyway, I checked that the option names used are correct for both XML-RPC and .com REST, and everything's working. :shipit:

PR #74, opened 2016-07-19
Fix #70: add plan id/shortname to sitemodel and fetch it from site(s) update

### LINKED ISSUE #70: Add Plans in SiteModel / SiteStore
See https://github.com/wordpress-mobile/WordPress-Android/blob/1389eb2df5c0eea8e6913c547790c517a5d99652/WordPress/src/main/java/org/wordpress/android/models/Blog.java#L539-L553

maxme:
Fix #70: add plan id/shortname to sitemodel and fetch it from site(s) update

### DIFF
--- a/WordPressStores/src/main/java/org/wordpress/android/stores/network/rest/wpcom/site/SiteWPComRestResponse.java
+++ b/WordPressStores/src/main/java/org/wordpress/android/stores/network/rest/wpcom/site/SiteWPComRestResponse.java
@@ -16,6 +16,11 @@ public class SiteWPComRestResponse implements Payload, Response {
         public String admin_url;
     }
 
+    public class Plan {
+        public long product_id;
+        public String product_name_short;
+    }
+
     public int ID;
     public String URL;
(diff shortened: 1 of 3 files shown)


WordPressStores/src/main/java/org/wordpress/android/stores/network/rest/wpcom/site/SiteWPComRestResponse.java
```diff
@@ -16,13 +16,19 @@
         public String admin_url;
     }
 
+    public class Plan {
+        public long product_id;
+        public String product_name_short;
```
aforcier: Any reason for omitting `product_name`? It seems to be [used by WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/blob/f201cfd97280622815ff9c1803e76685282f40f8/WordPress/src/main/java/org/wordpress/android/ui/plans/PlanFragment.java#L98).

aforcier:
:shipit:

PR #75, opened 2016-07-19
Add site capabilities to wpcom sites

maxme:
I was hesitant to add all capabilities flat like this, it's ugly. Introducing another model and a 1-to-1 binding will make things more complicated for one only purpose: clean the site model.


aforcier:
I agree it's ugly but I agree with you on the simplicity factor. One suggestion, how about s/IsCapability/HasCapability/? It should make it more readable.

maxme:
`develop` merged in

aforcier:
:shipit:

PR #77, opened 2016-07-26
PostStore implementation

aforcier:
Implements PostStore (#38).

TODO:
* [x] Implement all POSTS table methods from `WordPressDB` in `PostStore` (or `PostSqlUtils`)
    * [x] `blogHasLocalChanges` - dropped, unused
    * [x] `deleteAllPostsForLocalTableBlogId` - dropped for now as we determined it's not necessary (performed when removing a site or logging out from .com)
    * [x] `deletePost` - is now the `REMOVE_POST` action, and implemented as `PostSqlUtils.deletePost(PostModel)`
    * [x] `deleteUploadedLocalDrafts` - dropped, migration method
    * [x] `deleteUploadedPosts`	 - is now `PostSqlUtils.deleteUploadedPostsForSite`, and is handled internally by the `PostStore` when fetching completes
    * [x] `getPostForLocalTablePostId` - is now `PostStore.getPostByLocalPostId()`
    * [x] `getPostsListPosts` - not implemented, for a first iteration WPAndroid will do a normal `getPostsForSite()` and will convert the resulting `List<PostModel>` into a `PostsListPostList` object
    * [x] `getUploadedCountInBlog	` - is now `PostStore.getUploadedPostsCountForSite()`
    * [x] `savePost` - is now the `UPDATE_POST` action, and implemented as `PostSqlUtils.insertOrUpdatePost(PostModel, true)` (used to create a new post)
    * [x] `savePosts` - is now the `FETCH_POSTS` action, and implemented as `PostSqlUtils.insertOrUpdatePost(PostModel, false)` (used from the `PostUpdateService` and `updateSinglePost` in the WPAndroid app)
    * [x] `updatePost` - is now the `UPDATE_POST` action, and implemented as `PostSqlUtils.insertOrUpdatePost(PostModel, true)` (used when saving posts from the editor or uploading them)
    * [x] `postHasLocalChanges` - dropped, its only usage was refactored into the insertOrUpdatePost methods
* [x] Add XML-RPC support
    * [x] Fetching posts
    * [x] Uploading new posts
    * [x] Editing posts
    * [x] Deleting posts
    * [x] Error handling
    * [x] Migrate from `metaWeblog` XML-RPC methods to `wp`
    * [x] ~~Handle uploading post media (replicate all `PostUploadService` behavior)~~ (media post uploads will be handled by `MediaStore`, with the post-parsing part happening in WPAndroid rather than within the `PostStore`'s network clients)
        * [x] ~~Link up to future `MediaStore`~~
        * [x] ~~Create `onChanged` methods for API call results where appropriate, and implement in the example app (main example: `onUploadProgress`, with status updates whenever a new media item is being uploaded)~~
        * [x] ~~Implement some form of multi-part `Request`, allowing for progress updates for media uploads~~
* [x] Add REST API support
    * [x] Fetching posts
    * [x] Uploading new posts
    * [x] Editing posts
    * [x] Deleting posts
    * [x] Error handling
* [x] Add integration tests with live sites
    * [x] XML-RPC
        * [x] Uploading new posts
        * [x] Uploading edits to existing posts
        * [x] Reverting local changes to uploaded posts
        * [x] Saving local changes to posts (both local drafts and locally changed uploaded posts)
        * [x] Fetching multiple remote posts (with `fetchMore`)
        * [x] Deleting remote posts
        * [x] Uploading posts with all optional params set, pulling the uploaded post, and confirming all params
        * [x] Error handling (uploading post with invalid featured image ID, attempting to fetch a post that doesn't exist, XML-RPC errors, etc.)
        * [x] Posts != Pages
    * [x] .COM REST
        * [x] Uploading new posts
        * [x] Uploading edits to existing posts
        * [x] Reverting local changes to uploaded posts
        * [x] Saving local changes to posts (both local drafts and locally changed uploaded posts)
        * [x] Fetching multiple remote posts (with `fetchMore`)
        * [x] Deleting remote posts
        * [x] Uploading posts with all optional params set, pulling the uploaded post, and confirming all params
        * [x] Error handling (attempting to fetch a post that doesn't exist, etc.)
        * [x] Posts != Pages
* [x] Add unit tests
    * [x] `PostModel` unit tests
    * [x] `PostStore` unit tests
* [x] Odds and ends
    * [x] Support tags and categories, storing as little as possible in the `PostModel`, and consolidate different requirements between XML-RPC and REST API (these differences were minimized by switching from `metaWeblog` to `wp` XML-RPC calls: now both requests only need the category/tag IDs to update the remote post)
    * [x] Consolidate date handling differences between XML-RPC and REST API

<h4>Notes</h4>

A few fields from WPAndroid's `Post` object were dropped:
* `mChangedFromLocalToPublished` - it's only used for analytics, and it's up to the host activity to keep track of posts that are newly published so it can send an analytics event when the upload completes
* `mediaPaths` - no longer used
* `quickPostType` - it's very old, it was used for setting a custom header for analytics purposes, redundant with the rich analytics we currently have

I added instrumentation tests with live sites, but they're fairly basic since they're supposed to work for any site. When we have a set of shared sites we use for tests, we can expand these tests with better coverage. Opened an issue here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/121.


maxme:
Renaming done.

aforcier:
Ready for review @maxme!

aforcier:
Note about XML-RPC error handling:

Currently all XML-RPC errors from post requests are being reported to the `PostStore` as generic, though the passed error message will have the correct (localized) string from the response. 

There are two kinds of errors we can get from post requests: 403s and 404s (they're all listed in [PostXMLRPCClient](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/b0e34f8d05a1f681ff35a1674bdb8104b41b3102/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/post/PostXMLRPCClient.java).

We need to solve https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/129 to correctly recognize `403` errors.

`404` errors have a similar problem. 'Real' `404` errors should be distinguishable from `404` `faultCode`s from XML-RPC, since a real `404` won't have an XML response with a `faultCode` and `faultString`. All the same, I left `404`s generic as well for now, since there's one case where this approach won't work: editing a post can give two `404` errors: 'Invalid post ID.' and 'Invalid attachment ID.'. I'm adding this note to #129 as we'll likely fix both `403` and `404` at once.

README.md
```diff
@@ -40,9 +40,11 @@ Each store should have a corresponding enum defining actions for that store. For
 Action naming guide:
 
     FETCH_X - request data from the server
+    PUSH_X - send data to the server
```
tonyr59h: I've got some renaming to do 🏃 💨

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCOM.java
```diff
@@ -0,0 +1,591 @@
...
+
+    private void createNewPost() throws InterruptedException {
+        // Instantiate new post
+        mNextEvent = TEST_EVENTS.POST_INSTANTIATED;
+        mCountDownLatch = new CountDownLatch(1);
+
+        InstantiatePostPayload initPayload = new InstantiatePostPayload(mSite, false);
+        mDispatcher.dispatch(PostActionBuilder.newInstantiatePostAction(initPayload));
+
+        assertEquals(true, mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
```
maxme: Nice, I was wondering how these methods could be called synchronously in tests. I guess we'll need multiple `mCountDownLatch` soon in the tests in case we broadcast multiple actions at once.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -0,0 +1,449 @@
...
+
+    @Override
+    public void onRegister() {
+        AppLog.d(AppLog.T.API, "PostStore onRegister");
+    }
+
+    /**
+     * Returns all posts and pages in the store as a {@link PostModel} list.
+     */
+    public List<PostModel> getPosts() {
```
maxme: I'm not sure how these methods (getPosts, getPostsCursor, getPostsCount, ... without a site parameters) could be used.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -0,0 +1,449 @@
...
+        if (result.isEmpty()) {
+            return null;
+        } else {
+            return result.get(0);
+        }
+    }
+
+    @Subscribe(threadMode = ThreadMode.ASYNC)
+    @Override
+    public void onAction(Action action) {
```
maxme: Can we split `onAction` in smaller private methods (we should do it in SiteStore probably), it's hard to read. And we should probably use a switch/case without a default case, so we could be sure all actions are treated.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -0,0 +1,449 @@
...
+        } else if (actionType == PostAction.DELETE_POST) {
+            RemotePostPayload payload = (RemotePostPayload) action.getPayload();
+            if (payload.site.isWPCom()) {
+                mPostRestClient.deletePost(payload.post, payload.site);
+            } else {
+                // TODO: check for WP-REST-API plugin and use it here
+                mPostXMLRPCClient.deletePost(payload.post, payload.site);
+            }
+        } else if (actionType == PostAction.DELETED_POST) {
+            // Handle any necessary changes to post status in the db here
```
maxme: Humm, what about removing the post from the DB here? (or dispatching  `REMOVE_POST`) so the app doesn't have to "refresh" the post list or call remove.

maxme: Note: If we decide to not do that, then we should update the example app and dispatch REMOVE_POST in OnPostChanged, so if we tap the "Delete Post" button several time in the example it won't do weird things.

aforcier: My thinking was that we should keep these decoupled so it's possible to 'undo' a delete action (especially if deleting a trashed post, or a post on a site without trash enabled, since we can't get that back but we can at least re-upload it from the persisted copy).

aforcier: Although now that I'm looking at it again, we should probably update the db with the new post status if it has changed to `TRASHED` on `DELETE_POST`.

example/src/main/java/org/wordpress/android/fluxc/example/MainExampleActivity.java
```diff
@@ -432,4 +483,36 @@ public void onNewSiteCreated(OnNewSiteCreated event) {
     public void onSiteRemoved(OnSiteRemoved event) {
         mUpdateFirstSite.setEnabled(mSiteStore.hasSite());
     }
+
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    public void onPostChanged(OnPostChanged event) {
```
maxme: Can we add something like:

```java
if (event.isError()) {
  prependToLog("Error from: " + event.causeOfChange + " - error: " + event.error.type);
  return;
}
```
```

File: example/src/main/java/org/wordpress/android/fluxc/example/MainExampleActivity.java [hunk 7]
```diff
@@ -432,4 +483,36 @@ public void onNewSiteCreated(OnNewSiteCreated event) {
...
+        if (!mPostStore.getPosts().isEmpty()) {
+            if (event.causeOfChange.equals(PostAction.FETCH_POSTS) ||
+                    event.causeOfChange.equals(PostAction.FETCH_PAGES)) {
+                SiteModel firstSite = mSiteStore.getSites().get(0);
+                prependToLog("Fetched " + event.rowsAffected + " posts from: " + firstSite.getName());
+            }
```
maxme: And here, something like:

```java
if (event.causeOfChange.equals(PostAction.DELETE_POST)) {
    prependToLog("Post deleted!");
}
```

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
maxme: > My thinking was that we should keep these decoupled so it's possible to 'undo' a delete action (especially if deleting a trashed post, or a post on a site without trash enabled, since we can't get that back but we can at least re-upload it from the persisted copy).

Well, the undo won't really done like this in the app. We'll "fake" the action (like hide the item in a list view) and dispatch the action when the undo is not possible anymore (for example when the user tap the back button)

I agree it's usually better to keep action decoupled, but in that case I don't see any good use cases.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCOM.java
aforcier: I think in some cases we can probably get away with collecting all the actions under one `CountDownLatch` like @tonyr59h's doing in the [MediaStore tests](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/122/files#diff-00663fe439202f9492c9288e6a5b3a42R310).

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
aforcier: They're basically only there for the tests to use (though the situations they would catch are a little extreme, e.g. mixing up a parameter somewhere and saving a post using its post ID as a site ID, or something). Perhaps that's not worth exposing them to the client; I can drop them if you think so.

maxme: Yes, I would prefer to have them in a TestUtils class or we could mark them private and use reflection to test them.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -0,0 +1,449 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class PostStore extends Store {
+    public static final int NUM_POSTS_PER_FETCH = 20;
+
+    public static class FetchPostsPayload extends Payload {
+        public SiteModel site;
```
maxme: It's probably too early for this, but we might want to add other parameters here, like the post status (If the app needs to filter by draft/published/trashed status for instance). We should fill a new issue for this.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
aforcier: > we should probably use a switch/case without a default case, so we could be sure all actions are handled

Good point, this will also help ensure that stores don't accidentally try to handle actions that don't belong to them, since that would require a separate check or a new switch (as we'll have to do for `AccountStore`, for example, which handles `AccountAction`s and `AuthenticationAction`s.

aforcier:
TODO:

* [ ] Don't expose getPosts(), getPostsCursor(), getPostsCount(), etc. in the API
* [ ] Split up `onAction` and use a `switch`
* [ ] Dispatch `REMOVE_POST` from `DELETED_POST` (when 'deleted' post's status is `DELETED`, otherwise update the db to reflected `TRASH` status)
* [ ] Add more `OnPostChanged` logging to `MainExampleActivity`
* [ ] File issue for accepting filters in `FetchPostsPayload`

maxme:
> (when 'deleted' post's status is DELETED, otherwise update the db to reflected TRASH status)

I'm not even sure about that, unless we had a `status` parameter to `getPosts()`.

aforcier:
> I'm not even sure about that, unless we had a `status` parameter to `getPosts()`

You're right, I'll add a note to the issue for that to mention the `trash` subtlety.

example/src/main/java/org/wordpress/android/fluxc/example/MainExampleActivity.java
```diff
@@ -505,7 +519,7 @@ public void onPostChanged(OnPostChanged event) {
     public void onPostInstantiated(OnPostInstantiated event) {
         PostModel examplePost = event.post;
         examplePost.setTitle("From example activity");
-        examplePost.setContent("Hi there, I'm a post from WordPress-{Placeholder Name}-Android!");
+        examplePost.setContent("Hi there, I'm a post from FluxC!");
```
maxme: +1

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -290,50 +290,65 @@ public PostModel getPostByLocalPostId(long localId) {
...
+            case DELETE_POST:
+                deletePost((RemotePostPayload) action.getPayload());
+                break;
+            case DELETED_POST:
+                handleDeletePostCompleted((RemotePostPayload) action.getPayload());
+                break;
+            case REMOVE_POST:
+                PostSqlUtils.deletePost((PostModel) action.getPayload());
+                break;
+        }
```
maxme: 🏆 I'll do the same for Site and Account stores

maxme:
:shipit:

PR #88, opened 2016-07-27
Adding error enum to OnAccountChanged event and emitting as needed

tonyr59h:


maxme:
Renaming done.

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
```diff
@@ -141,6 +143,24 @@ public static AuthenticationError fromString(String string) {
         }
     }
 
+    public enum AccountError {
+        ACCOUNT_FETCH_ERROR,
+        SETTINGS_FETCH_ERROR,
+        SETTINGS_POST_ERROR,
+        GENERIC_ERROR;
+
+        public static AccountError fromString(String string) {
```
maxme: We don't need this method in `AccountError`, we don't read the error from the API.

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
```diff
@@ -239,12 +259,16 @@ public void onAction(Action action) {
             if (!checkError(data, "Error fetching Account via REST API (/me)")) {
                 mAccount.copyAccountAttributes(data.account);
                 updateDefaultAccount(mAccount, AccountAction.FETCH_ACCOUNT);
+            } else {
+                emitAccountChangeError(AccountError.ACCOUNT_FETCH_ERROR);
```
maxme: I wonder if we should attach more info the the event. I think that's ok for a first pass. We could revisit later, we also have to manage generic errors somehow (timeout, network error).

tonyr59h: I agree that there should be more info with these sorts of events but with our error handling in flux at the moment (four open issues to improve error handling) this should do for now.

maxme:
:shipit:

PR #94, opened 2016-07-28
Fix #93: Use ThreadMode.ASYNC for onAction methods

### LINKED ISSUE #93: Use "Background" or "Async" thread mode for onAction()
When we switched from Otto bus to EventBus in #61 - I thought the default mode was "Background" but it's actually "Posting". We might want to run all our actions in one (or multiple) background thread(s).

aforcier:
This ensures that all action handling is performed on worker threads.

cc @maxme


maxme:
:shipit:

PR #87, opened 2016-07-27
fix #86: add xmlrpc url to .com sites

maxme:
fix #86: add xmlrpc url to .com sites

PR #104, opened 2016-08-04
fix #103: get the XMLRPC endpoint from the /me/sites response

### LINKED ISSUE #103: Get the XMLRPC endpoint from the REST response
We can drop it later, that would be a good test to see if we only use REST calls on .com/jetpack sites - but for now, we need it for compatibility.

maxme:
fix #103: get the XMLRPC endpoint from the /me/sites response


maxme:
Dup #87

PR #112, opened 2016-08-09
REST API endpoint builder

aforcier:
Changed `WPCOMREST` from an `enum` to a builder with nested classes.

This lets us do things like call `WPCOMREST.sites.site(56).posts.post(45).getEndpoint()` when requesting post `45` from site `56`. Full use cases demonstrated in [WPComEndpointTest](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/abbd1e76145995eb1c763a3da4c724d39d21a702/example/src/test/java/org/wordpress/android/fluxc/WPComEndpointTest.java).

The `@Endpoint` annotation doesn't have any programmatic use currently - I added that to make it easier to pick out the implemented endpoints from the nested classes in `WPCOMREST`. It also has the benefit of showing up in the endpoint's Javadoc comment with the full API string for reference.

cc @maxme

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/WPComEndpointTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/WPComEndpointTest.java
@@ -1,0 +1,38 @@
+package org.wordpress.android.fluxc;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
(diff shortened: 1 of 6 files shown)


maxme:
TIL about `@Documented`, I love how it feels to use it, but that's not obvious to add a new endpoint there.

:shipit: and we'll see if we like it/hate it when we add new endpoints (or if we have a generator).

PR #116, opened 2016-08-10
Update WellSql libs to 1.0.8

maxme:
I wanted to add a new `closeDb()` method somewhere in FluxC, the only place I found that could make sense is `WellSqlConfig`. But since any app using FluxC could directly call `WellSql.closeDb()`, I don't think that's necessary (Documentation / Warning is more suitable, but too early IMO).

### DIFF
--- a/fluxc/build.gradle
+++ b/fluxc/build.gradle
@@ -50,8 +50,8 @@ dependencies {
     }
 
     // Custom WellSql version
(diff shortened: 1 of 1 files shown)


aforcier:
Everything still works! 🎉

No test issues, hopefully we never see tests fail due to `the connection pool has been closed` again!

:shipit:

PR #119, opened 2016-08-11
Set BaseRequest timeout to 30 seconds

aforcier:
Updated our `BaseRequest` to timeout at 30 seconds instead of the default 2.5 seconds. The default was causing issues with longer-running requests such as post creation over WP.COM REST.

I also added a method for disabling retries, and updated new account and new site creation to never attempt to retry.

cc @maxme

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
@@ -4,7 +4,6 @@ import android.util.Base64;
 
 import com.android.volley.DefaultRetryPolicy;
 import com.android.volley.Response.ErrorListener;
-import com.android.volley.RetryPolicy;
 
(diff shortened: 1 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
```diff
@@ -15,13 +17,18 @@
     }
 
     private static final String USER_AGENT_HEADER = "User-Agent";
+    private static final int DEFAULT_REQUEST_TIMEOUT = 30000;
+
     protected OnAuthFailedListener mOnAuthFailedListener;
+    protected RetryPolicy mRetryPolicy;
```
maxme: hum, not sure how this works, should we just drop that field and use `setRetryPolicy()` instead of `mRetryPolicy = ...`?

maxme:
:shipit: 

@tonyr59h  make sure to call `request.disableRetries();` when creating new Media or on upload.

PR #126, opened 2016-08-16
Add EndpointProcessor for generating endpoint classes

aforcier:
As an expansion on https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/112, this PR aims to simplify the "back-end" of REST endpoints within FluxC.

Adding a new WP.COM REST API works as follows:

1. Add the plain text endpoint to the [list](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/c82261c17f5b6d004ea9f421963787a55f3c4054/fluxc/src/main/tools/wp-com-endpoints.txt)
2. Rebuild the project
3. Find endpoint class and method/field added to the generated `WPCOMREST` file, usable through the 'builder' pattern added in #112

I also added generation for XML-RPC endpoints so that they're grouped together, though that case is far simpler and wouldn't be worth a generator on its own. The process works the same, just add a new endpoint to the [XMLRPC file](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/c82261c17f5b6d004ea9f421963787a55f3c4054/fluxc/src/main/tools/xmlrpc-endpoints.txt).

<h3>Some caveats:</h3>

1. The WP.COM REST generator can support `WP-API` endpoints when we add support for that, but will need some modifications to handle their style of defining variable endpoints (`posts/<id>/` instead of .COM's `posts/$post_ID/`)
2. The WP.COM REST builder doesn't support endpoints of the type `/sites/$site/posts/slug:$post_slug` yet
3. The above endpoint also illustrates one more limitation: the generator assumes all variable endpoints (`$`) accept a `long` value, but they can also be a `String` (perhaps other types as well)

I'm not sure how likely we are to include any of the unsupported endpoints, so I left out support for now. We can expand the generator in the future if it becomes necessary.


fluxc-processor/src/main/java/org/wordpress/android/fluxc/processor/XMLRPCPoet.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Locale;
+
+import javax.lang.model.element.Modifier;
+
+public class XMLRPCPoet {
```
maxme: https://www.youtube.com/watch?v=njos57IJf-0

maxme:
This is great, much much simpler than editing `WPCOMREST` ourselves.

I tested by adding `/read/feed/$feed_url_or_id` to the `wp-com-endpoints.txt`:

* `WPCOMREST` was not regenerated, I've absolutely no idea if and how we can force the generation on file modified (Makefile-like) or if we can force generation. This is not a big deal, we won't add new endpoint every 10 minutes and if you know you need to clean/build, it's pretty easy to regenerate it.

* The name of the generated endpoint method was cut: `WPCOMREST.read.feed.feed_url_or(1)`, notice the `feed_url_or` instead of `feed_url_or_id`, not sure if we should clean the endpoint [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/46971a277413a1a883f793e4749e2d64b1b25979/fluxc-annotations/src/main/java/org/wordpress/android/fluxc/annotations/endpoint/EndpointNode.java#L67-L67) (or just drop the `_id` part). Also, `feed_url_or_id` is probably one of his kind. This is acceptable IMO.

We should add a few lines in the [README.md](README.md) to explain `WPCOMREST` is generated and you need to edit `fluxc/src/main/tools/*txt` files to add new endpoints.

aforcier:
Filed a new issue assembling all the endpoint generator improvements together (including the case you mentioned): https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/128

aforcier:
`README` updated in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/126/commits/a1f85f94b6b2d44580f14a1c2e00557bdbbc5c75.

maxme:
:shipit:

PR #130, opened 2016-08-19
Error handling changes

maxme:
Replace Volley `deliverError` by `deliverBaseNetworkError` in our Requests sub classes.

One improvement we could do: move the `OnAuthFailedListener` to the BaseRequest and rename it to something like `OnLowLevelNetworkButHighPriorityErrorListener`, because some network errors need special handling (ssl certs, http auth, unauth)

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
@@ -9,7 +9,6 @@ import com.android.volley.NetworkError;
 import com.android.volley.NoConnectionError;
 import com.android.volley.ParseError;
 import com.android.volley.Request;
-import com.android.volley.ServerError;
 import com.android.volley.TimeoutError;
 import com.android.volley.VolleyError;
 
@@ -165,11 +164,6 @@ public abstract class BaseRequest<T> extends Request<T> {
         // Parse Error
         if (volleyError instanceof ParseError) {
             return new BaseNetworkError(GenericErrorType.PARSE_ERROR, volleyError);
-        }
-
-        // Parse Error
-        if (volleyError instanceof ServerError) {
-            return new BaseNetworkError(GenericErrorType.SERVER_ERROR, volleyError);
         }
 
         // Null networkResponse? Can't get more infos
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
@@ -43,6 +43,12 @@ public abstract class BaseRequest<T> extends Request<T> {
         public String message;
         public VolleyError volleyError;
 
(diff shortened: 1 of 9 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequest.java
```diff
@@ -50,26 +57,29 @@ public BaseNetworkError deliverBaseNetworkError(@NonNull BaseNetworkError error)
...
 
-            // TODO: we could use GSON here
             JSONObject jsonObject;
             try {
                 jsonObject = new JSONObject(jsonString);
             } catch (JSONException e) {
                 jsonObject = new JSONObject();
             }
+            String apiError = jsonObject.optString("error", "");
+            String apiMessage = jsonObject.optString("error_description", "");
```
aforcier: At least for post errors, this key should be `message`, not `error_description`:
```
{"error":"unknown_post","message":"Unknown post"}
```
I'm not sure where `error_description` came from, do we have inconsistent error responses from the API?

fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
```diff
@@ -43,6 +43,12 @@
         public String message;
         public VolleyError volleyError;
 
+        public BaseNetworkError(@NonNull BaseNetworkError error) {
+            this.message = error.message;
+            this.error = error.error;
```
aforcier: Maybe we should rename `BaseNetworkError.error` to `errorType` or `type`, and avoid these `error.error` calls.

maxme:
Ready for another look

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequest.java
```diff
@@ -99,16 +99,13 @@ public BaseNetworkError deliverBaseNetworkError(@NonNull BaseNetworkError error)
...
+                error.type = GenericErrorType.AUTHORIZATION_REQUIRED; // Augmented error
                 payload.error.type = AuthenticationErrorType.AUTHORIZATION_REQUIRED;
             } else if (xmlrpcFault.getFaultCode() == 403) {
-                // Augmented error
-                error.type = GenericErrorType.NOT_AUTHENTICATED;
-                // Not authenticated
+                error.type = GenericErrorType.NOT_AUTHENTICATED; // Augmented error
                 payload.error.type = AuthenticationErrorType.NOT_AUTHENTICATED;
             }
+            error.message = xmlrpcFault.getMessage();
```
aforcier: This will set the message to `Invalid post ID. [Code: 404]` in the case of an post request with invalid post ID, as that's how `XMLRPCFault.getMessage()` is setup. Do we still need to append the error code?

aforcier:
:shipit:

PR #137, opened 2016-08-24
MediaStore EndpointProcessor tweaks

aforcier:
Makes a few improvements to the `EndpointProcessor` that came up with @tonyr59h's work on the `MediaStore`.

1. Fixed handling of the `../media/$media_ID/` WP.COM REST endpoint, which previously didn't work as the generator attempted to add a `MediaEndpoint` class inside a `MediaEndpoint` class (it will now call the inner class `MediaItemEndpoint`)
2. Added support for aliases to the XML-RPC endpoint generator, allowing us to define `EDIT_MEDIA` and `DELETE_MEDIA` and have them point to `wp.editMedia` and `wp.deleteMedia`, for improved readability

When the rule from point 1 is applied, the generated method is also changed: `item()` instead of `media()`, since we felt `WPCOMREST.sites.site(siteId).media.item(mediaId)` reads better than `WPCOMREST.sites.site(siteId).media.media(mediaId)` (see tests for usage examples).

### DIFF
--- a/fluxc-processor/src/main/java/org/wordpress/android/fluxc/processor/EndpointProcessor.java
+++ b/fluxc-processor/src/main/java/org/wordpress/android/fluxc/processor/EndpointProcessor.java
@@ -11,6 +11,10 @@ import org.wordpress.android.fluxc.annotations.endpoint.WPComEndpoint;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import javax.annotation.processing.AbstractProcessor;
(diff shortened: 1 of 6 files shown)


fluxc-processor/src/main/java/org/wordpress/android/fluxc/processor/EndpointProcessor.java
```diff
@@ -26,6 +30,18 @@
     private static final String WPCOMREST_ENDPOINT_FILE = "fluxc/src/main/tools/wp-com-endpoints.txt";
     private static final String XMLRPC_ENDPOINT_FILE = "fluxc/src/main/tools/xmlrpc-endpoints.txt";
 
+    private static final Map<String, List<String>> XML_RPC_ALIASES;
```
tonyr59h: <3 Thanks

tonyr59h:
:shipit:

PR #138, opened 2016-08-24
Fix #136: ClassCastException in AccountRestClient

### LINKED ISSUE #136: ClassCastException in AccountRestClient
When changing the primary site for a .com account (using WPAndroid), I get this crash:

```
...
Caused by: java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Long
  at org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.updateAccountModelFromPostSettingsResponse(AccountRestClient.java:254)
  at org.wordpress.android.fluxc.store.AccountStore.onAction(AccountStore.java:281)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113) 
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588) 
  at java.lang.Thread.run(Thread.java:818) 
```

aforcier:
The underlying problem is that when posting settings to an account (`POST_SETTINGS`) we don't use the `AccountResponse` custom response object but instead let Gson do its own interpretation and give us a  `Map<String, Object>`. The trap is that Gson will save any JSON number as a `Double` unless told otherwise, as that's the most permissive number object - and our straight cast to `(long)` for the `primary_site_ID` threw an exception.

I opted for the simple fix of fixing the cast rather than using the specific `AccountResponse` response object for posting settings (or keeping track of which settings were set in `params` in the first place). That's a valid option too, but we will probably have to jump through some hoops to make sure we don't think a value was returned empty when it actually wasn't returned at all (we might wrongly set the 'about me' to the empty string, as an example).

I also made a few other tweaks to posting settings and added some tests.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/AccountModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/AccountModel.java
@@ -19,7 +19,7 @@ public class AccountModel extends Payload implements Identifiable {
     @Column private String mDisplayName;
     @Column private String mProfileUrl; // profile_URL
     @Column private String mAvatarUrl; // avatar_URL
-    @Column private long mPrimaryBlogId;
+    @Column private long mPrimarySiteId;
     @Column private int mSiteCount;
     @Column private int mVisibleSiteCount;
     @Column private String mEmail;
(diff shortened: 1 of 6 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/AccountModel.java
```diff
@@ -19,7 +19,7 @@
     @Column private String mDisplayName;
     @Column private String mProfileUrl; // profile_URL
     @Column private String mAvatarUrl; // avatar_URL
-    @Column private long mPrimaryBlogId;
+    @Column private long mPrimarySiteId;
```
maxme: 👍

maxme:
> The trap is that Gson will save any JSON number as a Double unless told otherwise, as that's the most permissive number object - and our straight cast to (long) for the primary_site_ID threw an exception.

Sometimes, I wonder if we should use another JSON library, GSON is great when the response is known and fixed, but it complicates and makes thing more error prone in other cases.

maxme:
:shipit:

PR #151, opened 2016-08-30
Fix #150: Make PostModel serializable

### LINKED ISSUE #150: PostModel should implement Serializable

aforcier:
Fixes #150. `PostModel` needs to be serializable so it can be saved and restored in [`EditPostActivity`](https://github.com/wordpress-mobile/WordPress-Android/blob/6290960d4419ca2745d5109cf71ed25f95d85786/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L417).


maxme:
:shipit:

Issue #145, opened 2016-08-26
Add more params to FetchPostPayload


aforcier:
Add more params to `FetchPostsPayload`, allowing us to request posts by status, for example.

#### Note about trashed/deleted posts

Once we allow fetching posts by status, we might want to revisit our post deletion implementation. Currently all deleted posts are removed from the db, even if they are in fact trashed. This is because `fetchPosts` with default settings won't return trashed posts from the server, so as far as the app is concerned they may as well be deleted permanently.

This changes if we're keeping track of trashed posts, and we should update the post status on delete instead of removing it from the db (unless the new status is `deleted`).

ref: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/77#discussion_r76405907

PR #154, opened 2016-08-30
Fix #147: Fetch all post statuses (except trashed) by default from WP.com REST

### LINKED ISSUE #147: Request posts by all statuses from REST (except trash)
The [`posts` REST endpoint](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/) by default only fetches published posts, while XML-RPC will fetches all statuses except trashed posts. The `PostStore` should specify all those post types by default in the request.

aforcier:
By default, the WP.com REST API only returns published posts. By requesting draft/pending/private/publish/future posts, we get the same results as the XML-RPC wp.getPosts default (which is what WPAndroid already uses).

For now there's no way to request anything but the default values, but in the future with https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/145 we'll be able to request specific post statuses (or groups of statuses) from WPAndroid.

#### Note about XML-RPC defaults

I opted not to add a default array for XML-RPC. By default, XML-RPC will return posts of any status except `trash`. The `future` status is omitted from the result of `wp.getPostStatusList`, but if we specify `post_status=publish,private,pending,draft` in the `wp.getPosts` request, we don't get Scheduled posts in the response. It seems that the default request implicitly includes Scheduled posts, though those posts have `post_status` set to `publish` in the response.

We can get the same as the default response if we request posts with `post_status=publish,private,pending,draft,future` (i.e., the same defaults as WP.com REST), but I'm hesitant to change the request when it will have no benefits and might introduce unexpected bugs (different behavior on earlier WP versions or custom installations, perhaps).

cc @maxme

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/post/PostStatus.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/post/PostStatus.java
@@ -4,6 +4,7 @@ import org.wordpress.android.fluxc.model.PostModel;
 import org.wordpress.android.fluxc.utils.DateTimeUtils;
 
 import java.util.Date;
+import java.util.List;
 
 public enum PostStatus {
     UNKNOWN,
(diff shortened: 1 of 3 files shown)


maxme:
:shipit:

PR #157, opened 2016-09-01
Refactor if/elseif to switch in `onAction` event handlers

maxme:
Refactor if/elseif to switch in `onAction` event handlers. I was sure there was a `EnumSwitchStatementWhichMissesCases`  checkstyle rule, but it's actually a check in the IDEA inspection tool only:

screen shot 2016-09-01 at 11 29 33

I  added the [FallThrough](http://checkstyle.sourceforge.net/config_coding.html#FallThrough) check in 53bbb2b to make sure we don't have a missing break in there.


aforcier:
Looks good! :shipit:

PR #159, opened 2016-09-01
Fix #158: Incorrect post statuses being sent in WP.com REST requests

### LINKED ISSUE #158: PostStatus toString not being called properly
Post statuses aren't being converted to their String value when posts are requested over WP.com REST:

`https://public-api.wordpress.com/rest/v1.1/sites/12345678/posts/?status=DRAFT%2CPENDING%2CPRIVATE%2CPUBLISHED%2CSCHEDULED&number=20&type=page&context=edit&locale=fr_FR`

(original ref: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/154)

aforcier:
Fixes #158. The problem was that [this call](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/154/files#diff-5232274a2e1be4691bbced897ec8bf1eR85) assumed `PostStatus.toString()` had been implemented, but it was actually calling `Enum.toString()`.

I replaced the static `PostStatus.toString(PostStatus)` method with an instance method, which fixes the problem. We'll need to make a few changes WPAndroid-side, which will be handled in the PostStore branch (which is replacing WPAndroid's version of `PostStatus` with this FluxC one).

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/post/PostStatus.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/post/PostStatus.java
@@ -15,6 +15,25 @@ public enum PostStatus {
     TRASHED,
     SCHEDULED; // NOTE: Only recognized for .com REST posts - XML-RPC returns scheduled posts with status 'publish'
 
+    public String toString() {
+        switch (this) {
+            case PUBLISHED:
+                return "publish";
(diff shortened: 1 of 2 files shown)


maxme:
Looks good:

screen shot 2016-09-02 at 09 42 34

:shipit:

PR #168, opened 2016-09-09
Replace DateTimeUtils with wputils version

aforcier:
Replaces our local `DateTimeUtils` with a version shared by WPAndroid in wputils (added in `v1.14`).

cc @maxme


maxme:
:shipit:

PR #170, opened 2016-09-24
Fix #167: Order posts returned from db by local draft status and date created

### LINKED ISSUE #167: PostStore should sort returned posts by draft status + date created
Just as we're doing in [`WordPressDB`](https://github.com/wordpress-mobile/WordPress-Android/blob/61ba7d2df6d231aa1e657c677c33545c497d4e5a/WordPress/src/main/java/org/wordpress/android/WordPressDB.java#L1191).

Note: For the most part the default sorting works okay, especially on WP.com REST which seems to return posts in descending order of date created already. It breaks for:

1. Scheduled posts on XML-RPC (XML-RPC returns by id, so scheduled posts won't float to the top as they currently do on WPAndroid, and already do for WP.com REST).

2. XML-RPC posts that have had their 'date created' modified.

2. Local drafts, when first returning to the post list after creating one. The next fetch from the server will delete existing posts and replace them, incrementing their local IDs and causing them to be listed below drafts (which don't get deleted in the update from the server).

aforcier:
Fixes #167. When selecting posts from the db, they're now returned sorted first by local draft status (local drafts float to the top) and then by date created (just like [`WordPressDB`](https://github.com/wordpress-mobile/WordPress-Android/blob/61ba7d2df6d231aa1e657c677c33545c497d4e5a/WordPress/src/main/java/org/wordpress/android/WordPressDB.java#L1191)).

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/post/PostStoreUnitTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/post/PostStoreUnitTest.java
@@ -19,7 +19,12 @@ import org.wordpress.android.fluxc.network.xmlrpc.post.PostXMLRPCClient;
 import org.wordpress.android.fluxc.persistence.PostSqlUtils;
 import org.wordpress.android.fluxc.persistence.WellSqlConfig;
 import org.wordpress.android.fluxc.store.PostStore;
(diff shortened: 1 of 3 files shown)


maxme:
:shipit:

PR #172, opened 2016-09-27
Fix #171: Emit OnPostUploaded on failed post uploads

### LINKED ISSUE #171: PostStore should fire a OnPostUploaded event even for failed uploads
Currently, the `PostStore` [fires an OnPostChanged event](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/3d41dc0585d469e5e9f0e8b4acf614b7df4b954b/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java#L406-L409) instead.

aforcier:
Fixes #171, emitting `OnPostUploaded` instead of `OnPostChanged` when a post upload fails. This is more consistent with our general pattern of emitting expected events (even if the initial action was not successful), especially since the post was not 'changed'.

'OnPostUploaded` is also better suited for this task since it includes a reference to the post that failed to upload, which we'll need WPAndroid-side when handling failed uploads.


maxme:
:shipit:

PR #174, opened 2016-09-30
Issue #173: Make PostModel cloneable

### LINKED ISSUE #173: Make PostModel cloneable
When editing, we pass `EditPostActivity` a `PostModel`, and we need to create a duplicate copy to check against for edits throughout the lifetime of the activity.

The pre-FluxC implementation [requests the post from the db twice](https://github.com/wordpress-mobile/WordPress-Android/blob/4d4ad90a7fe7b9c89220dbe8b4395f14b6cb764c/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L267), but instead we can make it possible to clone `PostModel`.

aforcier:
Fixes #173, making `PostModel` cloneable.

In order to make `PostModel` cloneable, the mutable `BaseNetworkError` field inherited from its `Payload` superclass needed to also be 'cloned', which should be handled by `Payload` (not by `PostModel`).

I intended to make `Payload` itself cloneable, but this risks `clone()` being called on an instance of a subclass that hasn't properly set up cloning for its own non-mutable fields.

Instead:

- `Payload` is now abstract since we don't intend for any pure `Payload` objects to exist, and so never needs to be cloned itself (so, it also doesn't implement `Cloneable`)
- `Payload` provides a safe `clone()` method its subclasses can use
- Any model (or other `Payload` subclass) that wants to be cloneable needs to explicitly implement `Cloneable` and override `clone()` in the usual way

This got a bit convoluted, but I still think this is the best way to allow the client (in this case, [`EditPostActivity`](https://github.com/wordpress-mobile/WordPress-Android/blob/4d4ad90a7fe7b9c89220dbe8b4395f14b6cb764c/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L267)) to make an independent copy of a model it already has without making an extra call to a store.

We could add a copy constructor so we could call `mClonedPost = new PostModel(mOriginalPost)`, but we don't want clients to instantiate models directly.

cc @maxme

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/PayloadTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/PayloadTest.java
@@ -1,0 +1,51 @@
+package org.wordpress.android.fluxc;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+import org.wordpress.android.fluxc.network.BaseRequest;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNull;
+
+@RunWith(RobolectricTestRunner.class)
+public class PayloadTest {
+
+    private class CloneablePayload extends Payload implements Cloneable {
+        @Override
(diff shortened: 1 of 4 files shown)


maxme:
:shipit:

Issue #128, opened 2016-08-18
REST endpoint generator improvements


aforcier:
- [x] <s>Support endpoints of type `/sites/$site/posts/slug:$post_slug`</s> (addressed in #217)
- [x] <s>Fix stripping of `_id` from `/read/feed/$feed_url_or_id` (or create a new rule that catches this case) ref: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/126#issuecomment-240675471</s> (addressed in #383)
- [x] <s>Allow setting types when defining endpoints in the text file, e.g. the above endpoint should create two methods:</s> (addressed in #181)
    ```
    /read/feed/$feed_url_or_id#long, /read/feed/$feed_url_or_id#String
    ```

aforcier:
All of the improvements have been added!

PR #181, opened 2016-10-08
Support typed endpoints in the endpoint processor

aforcier:
Addresses item 3 in https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/128:

> Allow setting types when defining endpoints in the text file

With this change, you can specify a type for a variable endpoint when adding it to [`wp-com-endpoints.txt`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/324af44323c2d784ec7efad6584b31907e52194f/fluxc/src/main/tools/wp-com-endpoints.txt).

For example,

```java
/sites/$site/taxonomies/$taxonomy#String/terms/
```

will make the generated `taxonomy()` method accept a `String` (instead of the default `long`).

Though we don't need it yet, I also added support for multiple types, e.g.:

```java
/sites/$site#long,String
```

will create two methods, accepting `long` and `String` values for `sites.site()`.

This will come in handy if we ever need to support domains for the [`$site` endpoint](https://developer.wordpress.com/docs/api/1.2/get/sites/%24site/). Also for the `/read/feed/$feed_url_or_id` endpoint which is mentioned in #128 (but needs an additional fix for the `_id` part which we'll address separately).

cc @maxme

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/EndpointNodeTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/EndpointNodeTest.java
@@ -13,7 +13,7 @@ import java.io.IOException;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 @RunWith(RobolectricTestRunner.class)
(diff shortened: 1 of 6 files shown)


maxme:
:shipit:

PR #183, opened 2016-10-12
Allow media uploading on sites with http auth

maxme:
Reuse HTTPAuthManager and UserAgent in the MediaXMLRPCClient to decorate the okhttp request and allow media uploading on sites with http auth.


maxme:
cc @tonyr59h @oguzkocer

tonyr59h:
:shipit:

PR #184, opened 2016-10-13
Reuse UserAgent in the MediaRESTClient to decorate the okhttp request used for upload

maxme:
Reuse UserAgent in the MediaRESTClient to decorate the okhttp request used for upload

I forgot this in #183


tonyr59h:
Looks good, tests pass :shipit:

PR #187, opened 2016-10-14
Allow date changes for published WP.com posts

aforcier:
Fixes as issue where it was impossible to change the published date for a post over the REST API (this also made it impossible to change status from `'publish'` to `'future'`).

To test, the new WPCOM test added in 8da752a should fail when checking out that commit (while the XML-RPC one works). Checking out the branch should fix the test.

cc @maxme


maxme:
:shipit:

Issue #189, opened 2016-10-18
Support WP.com POST requests with arbitrary JSON 


aforcier:
Our current [`WPComGsonRequest` setup](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/9e535369eb7dfef24c844e90443b3c634d3c4bde/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequest.java#L32) only supports a `String -> String` map of values for the request body.

We haven't needed to send JSON objects or arrays yet, but we'll need to support that in order to send [location data for posts](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/188). We can do this by overriding `Request.getBody()` and formatting a JSON request using Gson.

aforcier:
Fixed in #192.

PR #192, opened 2016-10-19
Issue #189: Support WP.com POST requests with arbitrary JSON

aforcier:
Addresses #189. Modifies `GsonRequest` to support generating a JSON body from a `Map<String, Object>`, and updates `WPComGsonRequest` with two static factory methods (one for `GET`, and one for `POST`).

I'd like to remove [this method](https://github.com/wordpress-mobile/WordPress-FluxC-Android/compare/issue/189-wpcom-complex-post-requests?expand=1#diff-8aec461b9ea3f73c8e15c4b4a6231f9bR36) entirely, and rely only on factory methods for all calls to `WPComGsonRequest` (this means refactoring all our current REST calls). If we need to support methods other than `POST` or `GET` in the future for WP.com we can expand on this. What do you think @maxme?

#### Aside:

I experimented with using a builder rather than a factory method, but I'm not sure how well it suits us since all but one of the params are required, so the only 'optional' (`setBody()` or `setUrlParams()`) ends up buried under all the listener callback code:

```java
final WPComGsonRequest<PostWPComRestResponse> request = new WPComGsonRequest.WPComGsonRequestBuilder<>(
		Method.POST, url, PostWPComRestResponse.class,
		new Listener<PostWPComRestResponse>() {
			@Override
			public void onResponse(PostWPComRestResponse response) {
				// TODO: Handle response
			}
		},
		new BaseErrorListener() {
			@Override
			public void onErrorResponse(@NonNull BaseNetworkError error) {
				// TODO: Handle error
			}
		})
		.setBody(body)
		.build();

request.addQueryParameter("context", "edit");

request.disableRetries();
add(request);
```

(either `.setBody()` or `.setParams()` would need to be called to access `.build()`)

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/utils/RequestQueryParametersTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/utils/RequestQueryParametersTest.java
@@ -18,7 +18,7 @@ public class RequestQueryParametersTest {
     public void testBaseRequestAddQueryParameters() {
         String baseUrl = "https://public-api.wordpress.com/rest/v1.1/sites/56/posts/";
 
-        WPComGsonRequest wpComGsonRequest = new WPComGsonRequest<Object>(Method.HEAD, baseUrl, null, null, null, null);
+        WPComGsonRequest<Object> wpComGsonRequest = WPComGsonRequest.buildGetRequest(baseUrl, null, null, null, null);
 
         wpComGsonRequest.addQueryParameter("type", "post");
         assertEquals(baseUrl + "?type=post", wpComGsonRequest.getUrl());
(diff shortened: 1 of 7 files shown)


maxme:
> I'd like to remove this method entirely, and rely only on builder methods for all calls to WPComGsonRequest (this means refactoring all our current REST calls). If we need to support methods other than POST or GET in the future for WP.com we can expand on this. What do you think @maxme?

Yes, that looks good, refactoring should be easy.

aforcier:
> I'd like to remove [this method](https://github.com/wordpress-mobile/WordPress-FluxC-Android/compare/issue/189-wpcom-complex-post-requests?expand=1#diff-8aec461b9ea3f73c8e15c4b4a6231f9bR36) entirely, and rely only on factory methods for all calls to WPComGsonRequest (this means refactoring all our current REST calls).

Done in 191be31.

maxme:
:shipit: after fixing  checkstyle error

PR #195, opened 2016-10-19
Issue #188: Fix location for posts

### LINKED ISSUE #188: Fix location for posts
There are a few issues to address with locations for posts.

#### PostLocation + PostModel + wellsql

For both APIs,, the `PostLocation` object and the way we're using it in the `PostModel` needs a few tweaks.

Since wellsql doesn't support arbitrary Java objects as columns, we're not using `PostLocation` directly, but instead store latitude and longitude and have some helper methods to set and get them in the form of a `PostLocation`.

This has a few side effects, e.g. we can't just null the `PostLocation` to signify no coordinates, and we need to be able to construct a `PostLocation` object even with invalid coordinates.

Tweaking this is all that's needed to fully support post location for self-hosted sites.

#### REST API, location metadata, and nested JSON requests

The .com REST API specifically poses some further complications:

1. Our current [`WPComGsonRequest` setup](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/9e535369eb7dfef24c844e90443b3c634d3c4bde/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequest.java#L32) only supports a `String -> String` map of values for the request body. We haven't needed to send JSON objects or arrays yet, but we'll need to support that in order to send the metadata array containing the location data to the API. We can do this by overriding `Request.getBody()` and formatting a JSON request using Gson. (#189 -> #192)
2. The .com REST API has a more [complex approach](https://developer.wordpress.com/2013/07/18/metadata-geotagging-and-the-rest-api/) than XML-RPC to setting and unsetting the location data, and we're probably going to have to make some additions to the `PostModel` to handle this.

aforcier:
Fixes #188 - several issues with post location across XML-RPC and WP.com.

#### Summary of the changes:

* Avoid the need to null `PostLocation` to unset it; instead, get `mLatitude` and `mLongitude` in `PostModel` to reflect an invalid location properly
* Since we need to know the difference between a local removal of post location and a post with no post location to begin with, `mLatitude` and `mLongitude` get a special value when the post's location has been unset (this is similar to what we do for [featured images](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5a081753d02c8336ccc26c9315dfc99bee6db603/fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java#L433-L435))
* The `PostRestClient` now handles adding/removing location data when uploading posts

To test, check out the tests in 708c610 - they should fail. Checking out 8149e48 should fix the XML-RPC tests, and finally checking out the branch should green both sets of tests.

cc @maxme

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
@@ -375,7 +375,7 @@ public class ReleaseStack_PostTestWPCom extends ReleaseStack_Base {
         assertEquals(false, mPost.hasLocation());
 
         // 2. Modify the post, setting some location data
-        mPost.setPostLocation(EXAMPLE_LATITUDE, EXAMPLE_LONGITUDE);
+        mPost.setLocation(EXAMPLE_LATITUDE, EXAMPLE_LONGITUDE);
         mPost.setIsLocallyChanged(true);
 
         uploadPost(mPost);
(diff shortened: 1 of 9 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
```diff
@@ -40,6 +40,8 @@
 
     private static final String POST_DEFAULT_TITLE = "PostTestWPCom base post";
     private static final String POST_DEFAULT_DESCRIPTION = "Hi there, I'm a post from FluxC!";
+    private static final double EXAMPLE_LATITUDE = 44.8378;
+    private static final double EXAMPLE_LONGITUDE = 0.5792;
```
maxme: I first thought this was a vineyard, but Google Maps says it's not.

maxme:
:shipit:

PR #198, opened 2016-10-20
Issue #197: Delay OnPostUploaded until post is synced

### LINKED ISSUE #197: Delay OnPostUploaded until post is synced
Currently, when a XML-RPC post upload completes, the sequence of events is:

1. `PUSHED_POST` is emitted by the client
2. `PostStore` picks up `PUSHED_POST`, and:
  1. Emits `OnPostUploaded`
  2. Marks the post as uploaded locally
  3. Requests a fresh copy of the post from the server
3. A new copy of the post is fetched from the server, and `FETCHED_POST` is emitted by the client
4. `PostStore` picks up `FETCHED_POST`, and:
  1. Updates the local copy of the post
  2. Emits `OnPostChanged(causeOfChange = UPDATE_POST)`

The process for WP.com REST is simpler, since the API returns the new post in the response to the upload call (XML-RPC only returns the remote ID of the post for a new post, or `1` for a successful edit of a post).

1. `PUSHED_POST` is emitted by the client
2. `PostStore` picks up `PUSHED_POST`, and:
  1. Emits `OnPostUploaded`
  2. Updates the local copy of the post
  3. Emits `OnPostChanged(causeOfChange = UPDATE_POST)`

There are a few problems with this distinction:
1. FluxC is supposed to abstract away the XML-RPC vs WP.com REST (vs WP-API) differences wherever possible - emitted events should reflect the same internal state of the store regardless of the API
2. More specifically, when we're listening for `OnPostUploaded`, the version of the post we have hasn't been synced with the server yet for XML-RPC, and we're missing any changes made server-side (e.g., post location metadata)
3. For extra safety, we should not emit `OnPostUploaded` until the local db has been updated, in case the client requests the post from the db before FluxC has updated it (this affects WP.com REST too)

The new sequence for XML-RPC should be:

1. `PUSHED_POST` is emitted by the client
2. `PostStore` picks up `PUSHED_POST`, and:
  1. Marks the post as uploaded locally
  2. Requests a fresh copy of the post from the server
3. A new copy of the post is fetched from the server, and `FETCHED_POST` is emitted by the client
4. `PostStore` picks up `FETCHED_POST`, and:
  1. Updates the local copy of the post
  2. Emits `OnPostChanged(causeOfChange = UPDATE_POST)`
  3. Emits `OnPostUploaded`

and for WP.com REST:

1. `PUSHED_POST` is emitted by the client
2. `PostStore` picks up `PUSHED_POST`, and:
  1. Updates the local copy of the post
  2. Emits `OnPostChanged(causeOfChange = UPDATE_POST)`
  3. Emits `OnPostUploaded`

aforcier:
Fixes #197, delaying the emission of `OnPostUploaded` until the db has been updated with the server's latest version of the uploaded post.

This means:

1. For XML-RPC, the `PostStore` will wait for a second server call, which fetches the uploaded post from the server, and store the result in the db before emitting `OnPostUploaded`
2. For WP.com, the `PostStore` will save the post to the db first, then emit `OnPostUploaded`

(#197 has a full breakdown of the change and the motivation behind it)

cc @maxme

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
@@ -53,6 +53,7 @@ public class ReleaseStack_PostTestWPCom extends ReleaseStack_Base {
         NONE,
         SITE_CHANGED,
         POST_INSTANTIATED,
+        POST_UPLOADED,
         POST_UPDATED,
         POSTS_FETCHED,
         PAGES_FETCHED,
@@ -667,9 +668,12 @@ public class ReleaseStack_PostTestWPCom extends ReleaseStack_Base {
             }
             return;
         }
+        assertEquals(TEST_EVENTS.POST_UPLOADED, mNextEvent);
         assertEquals(false, event.post.isLocalDraft());
         assertEquals(false, event.post.isLocallyChanged());
         assertNotSame(0, event.post.getRemotePostId());
+
+        mCountDownLatch.countDown();
     }
 
     private void setupPostAttributes() {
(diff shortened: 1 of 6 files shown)


maxme:
:shipit: 

But now I wonder if we should rename `s/event.causeOfChange/event.origin/`

Issue #53, opened 2016-07-06
Actions should and the Dispatcher should only be used for external calls


maxme:
We should use other events (not actions) and eventually a different bus for "wpstores confined actions" for instance:

`FETCHED_SETTINGS`, `FETCHED_ACCOUNT`, `POSTED_SETTINGS`, etc... (generally looks like callback not actions)

maxme:
We won't change this until first release.

PR #200, opened 2016-10-21
Group actions together by type

aforcier:
Groups actions together into remote actions, remote responses, and local actions, following the style we're using for [`MediaStore`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/6ebb6642f930a9efe790db56acdf55f784917d18/fluxc/src/main/java/org/wordpress/android/fluxc/action/MediaAction.java).

The 'remote responses' are the actions we'd like to move to separate, internal-only events, so they're not exposed to FluxC clients. (https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/53)

Also renamed some `pull` methods to `fetch` and a `post` method to `push` for consistency across all network methods.


maxme:
:shipit:

PR #215, opened 2016-11-02
Fix #206 faster app startup

### LINKED ISSUE #206: TrustManager.getAcceptedIssuers() call is super slow and could slow the app start
See the profiling report [here](https://nimbledroid.com/my_apps/org.wordpress.android?p=40ZFrmyvIsJW63#Summary). We should find a way to start it asynchronously (like we did in wpandroid).

I wonder if we could delay network client instantiation with the first network call, could be tricky with Dagger.

maxme:
Fix #206 faster app startup:

* Return an empty list when `getAcceptedIssuers()` is called (original code takes ~1200ms)
* Wrap the keyStore init (~400ms) in a FutureTask, so the injection itself is pretty fast (that won't make the time to first network call faster but that's not really a problem)

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/MemorizingTrustManager.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/MemorizingTrustManager.java
@@ -154,7 +154,9 @@ public class MemorizingTrustManager implements X509TrustManager {
     }
 
     public X509Certificate[] getAcceptedIssuers() {
-        return mDefaultTrustManager.getAcceptedIssuers();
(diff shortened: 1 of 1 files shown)


maxme:
Startup time went down - https://nimbledroid.com/my_apps/org.wordpress.android?p=42TSjecwsbF6Mp#Summary - but still not great - from 2.5 secs to 1.8 secs. I can probably move the loadStore call a FutureTask as well.

maxme:
Now down to 1.6secs similar to wpandroid 6.0 - https://nimbledroid.com/my_apps/org.wordpress.android?p=42TuQYwhVRsqKg#Summary

aforcier:
:shipit:

PR #217, opened 2016-11-02
Support mixed endpoints in the endpoint processor

aforcier:
Addresses item 1 in #128:

> Support endpoints of type `/sites/$site/posts/slug:$post_slug`

I'm implementing this a bit earlier than expected since TaxonomyStore will need the [`/sites/$site/taxonomies/$taxonomy/terms/slug:$slug/` endpoint](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/taxonomies/%24taxonomy/terms/slug:%24slug/).

Here are some extra test cases I checked against to verify other possible configurations for mixed endpoints:

```
/other:$otherThing/
/place:$place/unknown/
/magic/pony:$pony/new/
-----------
assertEquals("/other:42/", WPCOMREST.other(42).getEndpoint());
assertEquals("/place:42/unknown/", WPCOMREST.place(42).unknown.getEndpoint());
assertEquals("/magic/pony:42/new/", WPCOMREST.magic.pony(42).new_.getEndpoint());
```

(Those + the real endpoints cover all permutations of root/non-root, and children/no-children - as well as checking that both `String` and `long` are supported for mixed endpoints.)

cc @maxme


maxme:
:shipit:

PR #222, opened 2016-11-09
Support imports for Javadoc comments in checkstyle

aforcier:
Updated our checkstyle config to allow imports that are only used by Javadocs (previously, this would trigger an 'unused import' warning). I also added the imports in a few places where they were needed by Javadocs.


maxme:
:shipit:

PR #225, opened 2016-11-15
WP-API Endpoint support

aforcier:
Updates the `EndpointProcessor` to generate `WPAPI.java` from a list of WP-API endpoints for future use by FluxC.

PR #226, opened 2016-11-15
WP-API Discovery

aforcier:
Adds the essentials of WP-API discovery for self-hosted sites.

Note: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/225 should be merged first.

I added a [flag](https://github.com/wordpress-mobile/WordPress-FluxC-Android/compare/feature/wp-api-endpoints...feature/wp-api-discovery?expand=1#diff-cb79c1b7b7312c51e664f6a444213f99R1) for WP-API support so that these changes don't get in the way until we fully support WP-API. The WP-API tests also check for the feature flag to run.

Next steps (I'll be opening a new issue with these):

* Parse the [`authorization`](http://v2.wp-api.org/guide/discovery/#authentication-discovery) object from the base endpoint request and attempt login using Oauth1 or Basic auth accordingly
* Store the WP-API endpoint URL in `SiteModel`
* Make sure sites without "pretty permalinks" are supported (their WP-API endpoint will look like `http://example.com/?rest_route=/`)
* Add URL validation and HTTP/HTTPS neutrality (always check both no matter what URL is given)
* Support HTTP AUTH and self-signed SSL (can mostly re-use existing code from XML-RPC)
* Support multi-site installations

cc @maxme


maxme:
Note: we should add a WP instance for "Self hosted - WP-API missing v2 endpoints" on our test server.

fluxc/build.gradle
```diff
@@ -35,6 +35,27 @@ android {
...
+    File propertiesFile = file("gradle.properties")
+    if (propertiesFile.exists()) {
+        gradleProperties.load(new FileInputStream(propertiesFile))
+    } else {
+        // Load defaults
+        gradleProperties.load(new FileInputStream(file("gradle.properties-example")))
+    }
+    gradleProperties.any { property ->
+        if (property.value.equals("true") || property.value.equals("false")) {
+            buildType.buildConfigField "boolean", property.key.replace("wp.", "").replace(".", "_").toUpperCase(),
```
maxme: I think we can drop the "wp." prefix (or replace it with "fluxc.")

aforcier: Done in 5cbd7dd.

maxme:
:shipit:

PR #229, opened 2016-11-16
Issue #223: Set comment_status when uploading XML-RPC posts

### LINKED ISSUE #223: Posts published over XML-RPC always have comments closed
This needs more research, but for some reason all posts created by FluxC have comments disabled. We don't specify the `comment_status` in the request, so XML-RPC should default to the blog default. For some reason this isn't happening, and comments are disabled no matter what the blog default is.

Note: This seems to be working correctly for `metaWeblog.newPost`, which is why we would not have noticed this before.

aforcier:
Fixes #223. It seems XML-RPC's `wp.newPost` will always create posts with comments disabled, regardless of the default setting for the site.

To fix this, I added `default_comment_status` to the options we retrieve with `wp.getOptions`, and use that value for the post's `comment_status` param on every new upload.

In the future we could add a toggle for this in the post settings screen, so it can be changed for individual posts like on the web.

There's one small hiccup with this approach: if we publish a post without having called `wp.getOptions` beforehand we won't have the real default comment status for that site. Comments are enabled by default on first install, so I opted for the default value if we haven't called `wp.getOptions` to be `open`.

To test (on self-hosted site):
1. Make sure comments are enabled by default for the test site (`Settings` -> `Discussion` - > `Allow people to post comments on new articles` in wp-admin)
2. Publish a new post with FluxC (tests or example app) without this branch
3. Notice that the post has comments disabled
4. Switch to this branch and try again
5. The new post should have comments enabled
6. Try editing an existing post - the comment status shouldn't be changed by the edit

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
@@ -31,6 +31,7 @@ public class SiteModel extends Payload implements Identifiable, Serializable {
     @Column private boolean mIsWPCom;
     @Column private boolean mIsAdmin;
     @Column private boolean mIsFeaturedImageSupported;
+    @Column private String mDefaultCommentStatus = "open";
     @Column private String mTimezone;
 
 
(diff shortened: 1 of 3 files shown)


maxme:
:shipit:

PR #230, opened 2016-11-17
Emit OnAuthenticationChanged on manual token update

maxme:
We need this for magic link sign in (also it's kind of a rule for every actions to have a "response" onChanged event).


aforcier:
:shipit:

PR #186, opened 2016-10-14
Feature/40 comment store

maxme:
This PR add a new CommentStore with the following features:

# Features
* [x] Instantiate comment (local)
* [x] Update comment (local)
* [x] Remove single comment (local)
* [x] Remove all comments for a site (local)
* [x] Fetch comments for site (supporting number/offset) (REST & XMLRPC)
* [x] Fetch single comment (refresh) (REST & XMLRPC)
* [x] Create comment for post (REST & XMLRPC)
* [x] Create reply for comment (REST & XMLRPC)
* [x] Edit comment (update/moderate) (REST & XMLRPC)
* [x] Delete comment  (REST & XMLRPC)

# Tests
## Local tests
* [x] Instantiate comment
* [x] Update comment
* [x] Remove comment
* [x] Remove all comments for a site
* [x] Get number of comments in DB for a site

## REST
* [x] Fetch comments REST
* [x] Fetch comment REST
* [x] Create comment REST
* [x] Create comment duplicate REST
* [x] Create Reply comment REST
* [x] Push comment REST
* [x] Push invalid comment REST
* [x] Delete Once comment REST
* [x] Delete Twice comment REST

## XMLRPC
* [x] Fetch comments XMLRPC
* [x] Fetch comment XMLRPC
* [x] Create comment XMLRPC
* [x] Create comment duplicate XMLRPC
* [x] Create Reply comment XMLRPC
* [x] Push comment XMLRPC
* [x] Push invalid comment XMLRPC
* [x] Delete Once comment XMLRPC
* [x] Delete Twice comment XMLRPC

PR #231, opened 2016-11-17
Use base test classes

aforcier:
Some more tweaks to our instrumentation tests.

Note: #234, which fixes a few tests broken since #214 should be merged and the changed tests re-checked in this branch before merging.

### Changes

* Extended the use of the base classes (`ReleaseStack_WPComBase` and `ReleaseStack_XMLRPCBase`, initially introduced with @maxme's [CommentStore work](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/186)) to Posts and Media, offloading authentication and site setup
* Updated the base classes to use a static site object, so we're only making one site fetch call per test class (instead of one per individual test method)
* Moved the globally used `Dispatcher` and `CountDownLatch` to `ReleaseStack_Base` so account, site and discovery tests can be simplified a little as well


aforcier:
Rebased `develop` in now that #214 has been merged.

All tests should be passing now, except for a few Media tests (not caused by this branch - these will be addressed in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/233).

Ready for review!

PR #234, opened 2016-11-19
Issue #232: Fix XML-RPC site tests

### LINKED ISSUE #232: Handle certain network errors differently
There are some issues with HTTP Auth errors (and probably others) not being handled by stores correctly. Found while reviewing #214. For example, `SiteStore` [doesn't recognize](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/site/SiteXMLRPCClient.java#L80) auth errors when fetching sites and ends up performing an unnecessary update.

The test cases in `ReleaseStack_SiteTestXMLRPC` are currently failing because of this issue and should be a good verify a fix.

aforcier:
Fixes #232, repairing two tests in `ReleaseStack_SiteTestXMLRPC` that were broken by test changes in #214 (`testXMLRPCHTTPAuthFetchSites()` and `testXMLRPCSelfSignedSSLFetchSites()`).

With stricter identification of errors in tests to avoid false positives, we now need to be more specific with the site tests. The issue was that SSL and HTTP AUTH failures generate *two* `OnChanged` events, but the tests expected one.

When an SSL or HTTP AUTH error is encountered, an `OnAuthenticationChanged` event is always dispatched with the error. The failed action, however, can also dispatch its own event (here, `OnSiteChanged`), reporting a `GENERIC_ERROR`.

The fix was to specify that there are two errors expected in those tests, and allow `OnSiteChanged` to count down the latch if the tests are expecting an HTTP AUTH or SSL error.

@tonyr59h and I had some discussion about the deeper implications here. I'm not sure that `SiteErrorType`, for example, needs to contain an `HTTP_AUTH_ERROR` and `INVALID_SSL_CERTIFICATE` error, since it's not up to `OnSiteChanged` to handle this - this is a deeper problem with the site itself that needs to be handled in `OnAuthenticationChanged`.

If this is the case, then it seems good enough that `OnSiteChanged` reports a `GENERIC_ERROR`, signaling to the UI to stop showing loading screens/animations, while `OnAuthenticationChanged` takes care of prompting the HTTP AUTH screen/SSL popup. cc @maxme on whether this makes sense or not.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestXMLRPC.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestXMLRPC.java
@@ -108,6 +108,8 @@ public class ReleaseStack_SiteTestXMLRPC extends ReleaseStack_Base {
     }
 
     public void testXMLRPCSelfSignedSSLFetchSites() throws InterruptedException {
+        mMemorizingTrustManager.clearLocalTrustStore();
+
         RefreshSitesXMLRPCPayload payload = new RefreshSitesXMLRPCPayload();
         payload.username = BuildConfig.TEST_WPORG_USERNAME_SH_SELFSIGNED_SSL;
         payload.password = BuildConfig.TEST_WPORG_PASSWORD_SH_SELFSIGNED_SSL;
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestXMLRPC.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestXMLRPC.java
@@ -226,7 +226,7 @@ public class ReleaseStack_SiteTestXMLRPC extends ReleaseStack_Base {
     @SuppressWarnings("unused")
     @Subscribe
     public void onSiteChanged(OnSiteChanged event) {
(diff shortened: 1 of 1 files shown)


maxme:
> I'm not sure that SiteErrorType, for example, needs to contain an HTTP_AUTH_ERROR and INVALID_SSL_CERTIFICATE error, since it's not up to OnSiteChanged to handle this - this is a deeper problem with the site itself that needs to be handled in OnAuthenticationChanged.
> 
> If this is the case, then it seems good enough that OnSiteChanged reports a GENERIC_ERROR, signaling to the UI to stop showing loading screens/animations, while OnAuthenticationChanged takes care of prompting the HTTP AUTH screen/SSL popup. cc @maxme on whether this makes sense or not.

Yes this was the idea: for certain types of error we want to broadcast an event so the app can handle it in a single place. This error must be relayed by the OnChanged event so the app knows something went wrong and can eventually cancel the next step, stop the progress bar, etc. So in that case, we must broadcat a `OnAuthenticationChanged`, and send a `OnSiteChanged` event with a `GENERIC_ERROR`. We should probably replace this `OnAuthenticationChanged` by a generic error event (could be named `AuthenticationErrorEvent`) to avoid confusion between OnChanged events and broadcasted errors.

For now I only identified one type of this generic error: authentication errors, can happen during any kind of request if the user changed his password or removed his app token, and must be handled by the app in a "please enter your password" kind of dialog.

maxme:
:shipit:

PR #241, opened 2016-11-28
Only use selfHostedId in XMLRPC clients

maxme:
Also added a shell check in travis in https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/b2f09a7ba593ac8b3e9aa7c5a760a20051ce07e8 (not super proud of this, we can drop it if it reports false positive)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ allprojects {
     }
 
     checkstyle {
-        toolVersion = '6.9'
+        toolVersion = '7.3'
         configFile file("${project.rootDir}/config/checkstyle.xml")
     }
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -14,7 +14,7 @@ allprojects {
     }
 
     checkstyle {
-        toolVersion = '6.9'
+        toolVersion = '7.3'
```
tonyr59h: Just curious, any specifics for this version bump?

tonyr59h:
Good catch on the wrong site ID! :shipit:

PR #242, opened 2016-11-30
Updated getter "getCommentsQueryForSite" to take multiple statuses as parameters

maxme:
Updated getter "getCommentsQueryForSite" to take multiple statuses as parameters

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
@@ -101,18 +101,7 @@ public class CommentSqlUtils {
 
         // Check if statuses contains ALL
         if (!Arrays.asList(statuses).contains(CommentStatus.ALL)) {
-            selectQueryBuilder = selectQueryBuilder.beginGroup();
-            int i = 0;
-            for (CommentStatus status : statuses) {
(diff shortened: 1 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
```diff
@@ -97,27 +98,40 @@ public static CommentModel getCommentBySiteAndRemoteId(SiteModel site, long remo
...
+            for (CommentStatus status : statuses) {
+                // Insert an "OR" operand between status equalities.
+                if (i != 0) {
+                    selectQueryBuilder = selectQueryBuilder.or();
+                }
+                i += 1;
+                // Actual status check
+                selectQueryBuilder = selectQueryBuilder.equals(CommentModelTable.STATUS, status.toString());
+            }
+            selectQueryBuilder = selectQueryBuilder.endGroup();
```
aforcier: I think we can improve readability here by converting the `CommentStatus` array to a `String` array and using `ConditionClauseBuilder.isIn(CommentModelTable.STATUS, theStringArray)`.

maxme: Good point, I forgot about the `isIn()` method! This will be much cleaner.

aforcier:
:shipit:

PR #221, opened 2016-11-04
TaxonomyStore implementation

aforcier:
Addresses #160, adding a TaxonomyStore.

Work in progress!

### Notes:
* I added some provisions for supporting generic taxonomies (e.g. `TaxonomyModel`), but no taxonomies other than categories and tags are actually supported at this time
* Editing and deleting terms is unsupported for now, since we don't currently offer those operations in WPAndroid

### Features
- [x] Instantiate a term (local)
- [x] Update a term (local)
- [x] Remove all categories or tags or terms for a site (local)
- [x] Fetch all categories for site (REST & XMLRPC)
- [x] Fetch all tags for site (REST & XMLRPC)
- [x] Fetch all terms for generic taxonomy for site (REST & XMLRPC)
- [x] Fetch single category/tag/term (refresh) (REST & XMLRPC)
- [x] Create category/tag/term (REST & XMLRPC)
- [x] Retrieve full details on categories or tags given a `PostModel` (local)

#### Local tests
- [x] Term equality
- [x] Term insertion/retrieval
- [x] Separation of categories/tags/terms
- [x] Clearing all terms for a taxonomy from the db
#### REST
- [x] Fetch categories REST
- [x] Fetch tags REST
- [x] Fetch terms for taxonomy REST
- [x] Attempt to fetch terms for invalid taxonomy REST
- [x] Fetch single term by slug REST
- [x] Create category/tag/term REST
- [x] Attempt to upload term for invalid taxonomy REST
- [x] Attempt to upload a duplicate term REST
#### XML-RPC
- [x] Fetch categories XML-RPC
- [x] Fetch tags XML-RPC
- [x] Fetch terms for taxonomy XML-RPC
- [x] Attempt to fetch terms for invalid taxonomy XML-RPC
- [x] Fetch single term by ID XML-RPC
- [x] Create category/tag/term XML-RPC
- [x] Attempt to upload term for invalid taxonomy XML-RPC
- [x] Attempt to upload a duplicate term XML-RPC

PR #243, opened 2016-11-30
Use ASYNC thread mode for OnCommentInstantiated

aforcier:
This fixes an [issue](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/221#discussion_r89929535) raised by @tonyr59h for the taxonomies demo that also applies to comments.

Without this patch, attempting to create a new comment in the example app will cause a crash, as the `CountDownLatch` times out. `ThreadMode.ASYNC` needs to be used to prevent the main thread from being blocked.

cc @maxme


maxme:
:shipit:

PR #245, opened 2016-11-30
Simplify boolean asserts

aforcier:
Eliminates all remaining instances of `assertEquals(true, X)` and `assertEquals(false, Y)`.


maxme:
Looks good, can't find any other match.
:shipit:

PR #246, opened 2016-12-01
Emit OnPostChanged event when deleting a local post

### LINKED ISSUE #240: REMOVE_POST should emit an OnChanged event
Similarly to the change for `CommentStore` in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/239, the `PostStore` should emit an `OnPostChanged` event when acting on `REMOVE_POST`, as the operation is async and the caller might need to be notified when the post has been removed.

aforcier:
Fixes #240, emitting an `OnPostChanged` event in response to `REMOVE_POST` actions.


tonyr59h:
:shipit:

PR #248, opened 2016-12-02
Fix #247: Check if the email address is _false_ and init the email field with the empty string instead

### LINKED ISSUE #247: Comments email address can be the boolean "false" when coming from the REST API

maxme:
Fix #247: Check if the email address is _false_ and init the email field with the empty string instead

Would be cleaner to check the `response.author.email` type instead, but we can't and "false" is an invalid email address anyway.


tonyr59h:
:shipit:

PR #249, opened 2016-12-07
Add asc+desc parameters to some getters

maxme:
Note: I wanted to add the same parameter to all getters returning a `List<PostModel>` in the `PostStore`, but I noticed local drafts always come first and I want to avoid having a `orderByAscendingPublishingDateExceptForLocalDraftThatComesFirst` parameter ;)

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_CommentTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_CommentTestWPCom.java
@@ -1,5 +1,7 @@
 package org.wordpress.android.fluxc.release;
 
+import com.yarolegovich.wellsql.SelectQuery;
+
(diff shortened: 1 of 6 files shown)


aforcier:
The instrumentation tests need to be updated to use the new param for `getCommentsForSite()` calls - they're currently broken.

aforcier:
> Note: I wanted to add the same parameter to all getters returning a List<PostModel> in the PostStore, but I noticed local drafts always come first and I want to avoid having a orderByAscendingPublishingDateExceptForLocalDraftThatComesFirst parameter ;)

Yes that's a good point, I think it makes sense not to generalize this for now.

maxme:
> The instrumentation tests need to be updated to use the new param for getCommentsForSite() calls - they're currently broken.

Fixed in  e110be8

aforcier:
:shipit:

PR #250, opened 2016-12-08
Add comment local ids to OnCommentChanged events

maxme:


fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
```diff
@@ -399,6 +407,7 @@ private void pushComment(RemoteCommentPayload payload) {
         if (payload.comment == null) {
             OnCommentChanged event = new OnCommentChanged(0);
             event.causeOfChange = CommentAction.PUSH_COMMENT;
+            event.changedCommentsLocalIds.add(payload.comment.getId());
```
aforcier: This will throw an NPE as `payload.comment` is null here.

aforcier:
It's possible for the `CommentModel` to be null in most if not all of these cases - we should be null-checking it, or using `@NonNull` where possible.

maxme:
> It's possible for the CommentModel to be null in most if not all of these cases - we should be null-checking it, or using @NonNull where possible.

Fixed in  	5ceb77d - I wasn't able to make the `CommentModel` `@NonNull` in `RemoteCommentPayload` because it can be used to pull a comment by its site+remoteId only.

fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
```diff
@@ -309,6 +307,7 @@ private void updateComment(CommentModel payload) {
             rowsAffected = CommentSqlUtils.insertOrUpdateComment(payload);
         }
         OnCommentChanged event = new OnCommentChanged(rowsAffected);
+        event.changedCommentsLocalIds.add(payload.getId());
```
aforcier: This should probably be `null` checked as well, since an `UPDATE_COMMENT` action can legitimately be called from a FluxC client (and be `null`). This applies to the `payload.isError()` check in this same method.

fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
```diff
@@ -317,12 +316,14 @@ private void removeComment(CommentModel payload) {
         int rowsAffected = CommentSqlUtils.removeComment(payload);
         OnCommentChanged event = new OnCommentChanged(rowsAffected);
         event.causeOfChange = CommentAction.REMOVE_COMMENT;
+        event.changedCommentsLocalIds.add(payload.getId());
```
aforcier: Same as above comment, `payload` could be `null`.

fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
maxme: Hmmm, I'm not sure about null checking all payloads in all stores. I don't know if we should add `@NonNull` to our actions.

aforcier:
:shipit:

PR #256, opened 2016-12-19
Remove old comments when we fetch new ones

maxme:
Remove old comments when we fetch new ones


maxme:
@aforcier merged with `develop` and fixed a test.

aforcier:
This works overall, but replying/liking comments in the example app is broken. This looks to be due to its use of `INSTANTIATE_COMMENT`, and needs a fix similar to d510372.

aforcier:
:shipit:

PR #263, opened 2017-01-05
fix #261 local draft change date

### LINKED ISSUE #261: Port the changes for Pending Drafts in wpandroid to the post store
Reference: https://github.com/wordpress-mobile/WordPress-Android/pull/4835

maxme:
fix #261 local draft change date


aforcier:
We should make the same changes in d46b632 to the [XML-RPC version of this test](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/11761e629cc1700a11ff8dd6b40de3e5c4f64789/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestXMLRPC.java#L153).

It's a little redundant to have both tests since there are no network events, and we should maybe group 'local only' tests together at the top of `ReleaseStack_PostTestWPCom.java ` (and delete the XML-RPC one), but while both tests exist they should be the same.

Mind you, the `PostModel`s in the two tests are implicitly different since their local site IDs point to different types of site, and maybe it's worth keeping both just to keep testing parity in local draft handling.

maxme:
Added the same test in PostTestXMLRPC in 4ac9ab0

Note: when this is merged we must update the jitpack dependency in wpandroid `feature/fluxc-integration`.

aforcier:
:shipit:

PR #259, opened 2016-12-23
Issue #258: Restrict term lookup from list by site

aforcier:
Fixes #258, limiting the results of `getTermsFromRemoteIdList()` to the given site.

PR #267, opened 2017-01-09
Issue #265: Track post tags by name instead of by ID

### LINKED ISSUE #265: Use tag names instead of IDs in PostModel
Using tag names will make for an easier initial integration with WPAndroid, as the ID-based implementation will add overhead without providing any benefits (until we add support for tag management to WPAndroid).

aforcier:
Fixes #265. ~#259 should be merged first.~

This changes tag handling to use tag names instead of IDs, which is how WPAndroid currently works with tags. This means we won't be storing tags in `TaxonomyStore` for now, though I left the provisions for that in as we might change this in the future.

As a point of interest, both XML-RPC and the .com REST API allow uploading posts with a mix of names and IDs for categories and tags. This might be useful to us if we implement something like ID-based tag management but still want/need to maintain current tags for a post as names.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
@@ -1,5 +1,6 @@
 package org.wordpress.android.fluxc.release;
 
+import org.apache.commons.lang.RandomStringUtils;
 import org.greenrobot.eventbus.Subscribe;
 import org.wordpress.android.fluxc.TestUtils;
 import org.wordpress.android.fluxc.generated.PostActionBuilder;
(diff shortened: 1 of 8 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
```diff
@@ -304,6 +305,11 @@ public void testFullFeaturedPostUpload() throws InterruptedException {
         categoryIds.add((long) 1);
         mPost.setCategoryIdList(categoryIds);
 
+        List<String> tags = new ArrayList<>(2);
+        tags.add("fluxc");
+        tags.add("generated-" + RandomStringUtils.randomAlphanumeric(8));
```
maxme: TIL about `RandomStringUtils`

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -168,21 +168,21 @@ public void setExcerpt(String excerpt) {
...
 
-    public String getTagIds() {
-        return StringUtils.notNullStr(mTagIds);
+    public String getTagNames() {
+        return StringUtils.notNullStr(mTagNames);
     }
 
-    public void setTagIds(String tagIds) {
-        mTagIds = tagIds;
+    public void setTagNames(String tags) {
```
maxme: I wish we could make these private :/

maxme:
Looks good, I also tried to edit a remote post and remove all tags (using `setTagNameList(new ArrayList<String>());`).

:shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
aforcier: Yeah, me too :/

PR #271, opened 2017-01-09
Add methods for obtaining a TermModel given a name

aforcier:
Adds methods for getting a `TermModel` given the name of a term. This is not yet needed by WPAndroid, but might be needed in the future to e.g. obtain `TermModel`s given tag names from a `PostModel` (see #267).

Note: This PR was originally intended to be used to interface with the site settings component in WPAndroid. Site settings [provides the default category name](https://github.com/wordpress-mobile/WordPress-Android/blob/8b3a35d2db44fd840e02fb121a9771bea01331c1/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L259), which we need to turn into a remote term ID. I found a better way of doing this, so this PR is now unnecessary, but the methods used might still be needed in the future.


maxme:
Looks good, :shipit: - at some point we should add `@Nullable` to sql util methods.

PR #275, opened 2017-01-11
Tweak uploading posts with no categories/tags over XML-RPC

aforcier:
Introduces a few fixes to handling terms (categories and tags) during XML-RPC post uploads.

* Separates the logic for first-time publishing VS editing a post for readability
* Fixes an issue where it wasn't possible to remove all tags from a post that previously had tags (we need to use the `terms` param instead of the `terms_names` param in this specific case)
* Fixes an issue that prevented category-less first-time published posts from receiving the site's default category automatically server-side (this is how it works in wp-admin, and also in pre-FluxC WPAndroid)
* Re-restrict the login to posts only (not pages)


maxme:
:shipit:

PR #277, opened 2017-01-11
Issue #276: Request maximum number of terms from WP.com REST

### LINKED ISSUE #276: Increase limit of terms fetched over REST API to 1000
To improve things until we add pagination (https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/264), we may as well request the maximum allowed number of terms when fetching them from the WP.com REST API. We're currently using the default, 100, and the limit is 1000.

XML-RPC's `wp.getTerms` will return all available terms by default, so we don't need to make any changes there yet.

aforcier:
Fixes #276, fetching as many terms at a time as possible from the WP.com REST API ([1000](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/taxonomies/%24taxonomy/terms/)). The previous limit (the default value of the `number` param) was 100.


maxme:
:shipit:

Issue #121, opened 2016-08-12
Add post tests for specific sites


aforcier:
Once we have some shared testing sites, we can add some tests that expect a specific setup. For example:
- [x] ~~Verify that we're getting the exact right number of posts after fetchMore~~ (https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/121#issuecomment-274575722)
- [x] Test with multiple categories and tags (they already have to exist on the site) (done with TaxonomyStore implementation)
- [x] Test setting and unsetting existing images as featured (#279)

Some of these tests (featured image, categories/tags) could be setup automatically once we have `MediaStore` and `TaxonomyStore`, but it would be nice to avoid having to mix stores in tests like that.

aforcier:
> Verify that we're getting the exact right number of posts after `fetchMore`

I don't think this is necessary anymore.

Our generic fetch more tests ([XML-RPC](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/f7fa9ea89c99693f22d3e910b15fd08cbd032679/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestXMLRPC.java#L271-L299), [WP.com REST](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java#L280-L308)) assume a site with over 20 posts, and confirm that we end up with more than 20 posts after the fetch. I think this is good enough.

aforcier:
Closing this as all cases have either been implemented or are unnecessary.

PR #279, opened 2017-01-12
Fixes and tests for featured images

aforcier:
Adds tests for setting and unsetting featured images for posts over WP.com API and XML-RPC (item 3 in #121).

Also fixes featured images fro WP.com (thanks tests!). It turned out that the wrong param was used when pushing featured image IDs to the server. There was also a separate issue with expecting `id` instead of `ID` for the `post_thumbnail` object from the server (which meant the featured image ID wasn't being stored at all) which is now fixed.


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
```diff
@@ -314,6 +315,9 @@ public void testFullFeaturedPostUpload() throws InterruptedException {
         tags.add("generated-" + RandomStringUtils.randomAlphanumeric(8));
         mPost.setTagNameList(tags);
 
+        long featuredImageId = Long.valueOf(BuildConfig.TEST_WPCOM_IMAGE_IDS_TEST1);
```
maxme: This breaks the tests for me because `TEST_WPCOM_IMAGE_IDS_TEST1` is used in MediaTest ([here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/a57bc575b6191fbf2db0a9a6ee9ccfd32a50c3c5/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestWPCom.java#L78-L85)) as a coma separated list of IDs.

maxme: It can be used with this value for instance: `TEST_WPCOM_IMAGE_IDS_TEST1 = 1,3,493`

aforcier: Fixed in 5889ce4.

maxme:
:shipit:

PR #280, opened 2017-01-13
New IS_WPCOM_URL Site action

maxme:
That checks if the url is accessible via the WordPress.com REST API (likely a wordpress.com hosted site or a jetpack site or "test.com").

Known issue: test.com

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/SignedOutActionsFragment.java
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/SignedOutActionsFragment.java
@@ -18,9 +18,11 @@ import org.wordpress.android.fluxc.Dispatcher;
 import org.wordpress.android.fluxc.example.ThreeEditTextDialog.Listener;
 import org.wordpress.android.fluxc.generated.AccountActionBuilder;
 import org.wordpress.android.fluxc.generated.AuthenticationActionBuilder;
+import org.wordpress.android.fluxc.generated.SiteActionBuilder;
 import org.wordpress.android.fluxc.store.AccountStore;
 import org.wordpress.android.fluxc.store.AccountStore.NewAccountPayload;
 import org.wordpress.android.fluxc.store.AccountStore.OnNewUserCreated;
+import org.wordpress.android.fluxc.store.SiteStore.OnURLChecked;
 
 import javax.inject.Inject;
 
@@ -51,6 +53,13 @@ public class SignedOutActionsFragment extends Fragment {
             @Override
             public void onClick(View v) {
                 showMagicLinkDialog();
+            }
+        });
+
+        view.findViewById(R.id.check_url_wpcom).setOnClickListener(new OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                showCheckWPComUrlDialog();
             }
         });
         return view;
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -684,6 +699,21 @@ private void handleDeletedSite(DeleteSiteResponsePayload payload) {
...
     }
 
+    private void checkUrlIsWPCom(String payload) {
+        mSiteRestClient.checkUrlIsWPCom(payload);
+    }
+
+    private void handleCheckedIsWPComUrl(IsWPComResponsePayload payload) {
+        OnURLChecked event = new OnURLChecked(payload.url);
+        if (payload.isError()) {
+            // Return invalid site for all errors (this endpoint is not documented and seems a bit drunk).
```
aforcier: 😂

example/src/main/java/org/wordpress/android/fluxc/example/SignedOutActionsFragment.java
```diff
@@ -84,6 +93,21 @@ public void onClick(String username, String email, String password) {
         newFragment.show(ft, "dialog");
     }
 
+    private void showCheckWPComUrlDialog() {
+        AlertDialog.Builder alert = new AlertDialog.Builder(getActivity());
+        final EditText editText = new EditText(getActivity());
+        editText.setSingleLine();
+        alert.setMessage("Check if the following URL is wpcom or jetpack (and can be contacted via wpcom credentials):");
```
aforcier: This line is over 120 chars (and checkstyle is complaining).

maxme: fixed in  	fed262d

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
aforcier: Just a note for posterity (I don't think the comment needs to change).

The endpoint _is_ documented: the `/sites/$site/` endpoint is [supposed to accept an ID or domain for `$site`](https://developer.wordpress.com/docs/api/1.2/get/sites/%24site/), but the results are definitely irregular and I get the feeling the `domain` functionality may not be used very often.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -69,6 +69,14 @@ public DeleteSiteResponsePayload() {
         public DeleteSiteError error;
     }
 
+    public static class IsWPComResponsePayload extends Payload {
+        public IsWPComResponsePayload() {
+        }
+        public String url;
+        public BaseNetworkError error;
```
aforcier: It's unnecessary to add `error` here since the default `Payload` `error` is already a `BaseNetworkError`.

If we keep this, we actually need to also override `Payload.isError()`, otherwise it will always return `false`. The two `Payloads` defined above are making this error, I'll open an issue to fix them (and any more like them).

aforcier: Opened #281 to track other cases of this (or come up with a better, safer solution).

aforcier:
:shipit:

PR #291, opened 2017-01-25
Trigger a OnMediaUploaded event when there is a media upload error before the actual network call

maxme:
Trigger a OnMediaUploaded event when there is a media upload error before the actual network call


tonyr59h:
:shipit:

PR #285, opened 2017-01-20
Fix #284 crash adding jetpack site

maxme:
Fix #284 

At least 2 cases to test in the example:

1. Jetpack -> self hosted
    * Clear app data
    * Sign in a .com account with a Jetpack site
    * Try to add that Jetpack site as a self hosted site
    * You should get a "DUPLICATE_SITE" error
1. Self hosted -> Jetpack (with the same account / sites)
    * Clear app data
    * Add that Jetpack site as a self hosted site
    * Sign in a .com account
    * You should get the same number of sites, and the previous self hosted site is now considered a Jetpack

Not easy to test automatically without predefined sites or complex mocks.

I'm not a big fan of [checking for the XMLRPC endpoint ](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/8aad65ddc0fa7d458702792aa0df8e47df9bcc18/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java#L87)as a "key", but I don't think we can find anything better.

Issue #101, opened 2016-08-03
New Full stack test with a Jetpack multisite instance


maxme:
We should add 3 new tests to `ReleaseStack_SiteTestWPCom`:
- A wpcom account with a single Jetpack site connected
- A wpcom account with multiple Jetpack sites connected
- A wpcom account with a Jetpack multisites connected

aforcier:
We should add similar tests to `ReleaseStack_SiteTestXMLRPC` (pending https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/207):

* A self-hosted site without Jetpack (check `isJetpack` etc statuses)
* A self-hosted site with Jetpack installed but not connected
* A self-hosted site with Jetpack installed and connected
 * Connected account is the one currently logged into
 * Connected account is not the one currently logged into

aforcier:
Added in #292.

Issue #207, opened 2016-10-28
Jetpack handling over XML-RPC


aforcier:
This issue includes Jetpack-related considerations we still need to address. This can probably be split off into sub-issues when we have our approach cleared up.
#### Self-hosted site Jetpack status

The main status a self-hosted site can have with respect to Jetpack and the current account are:
1. Jetpack is not installed
2. Jetpack installed, but not connected to any account
3. Jetpack installed, and connected, and the current WP.com account logged into has access
4. Jetpack installed, and connected, but the current WP.com account doesn’t have access
5. Edge cases like Jetpack is connected but we're unable to reach the site (Jetpack errors)

...
#### WPAndroid UI considerations

From a UI perspective, these four Jetpack cases should be handled WPAndroid-side as follows (when the user navigates to a REST-only part of the app, e.g. Stats):
1. Show a dialog: You must install Jetpack to see your stats in the apps (this site uses XMLRPC API)
2. Show a dialog: You must complete your Jetpack connection (this site uses XMLRPC API)
3. Here are your stats (this site uses REST API)
4. Show a dialog:  Sorry, we don't support multi WP.com connections (this site uses XMLRPC API) (if possible, tell them which WP.com account it is connected to)
#### SiteModel

We need to think about how much to represent in SiteModel, and how much to leave up to the host application.

There's currently an issue where self-hosted sites without Jetpack installed get [flagged](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5e52ac8cb074b1213e573ff46b2208154cb4f89f/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/site/SiteXMLRPCClient.java#L224) as `isJetpack = true`, and we're unable to distinguish the case of 'Jetpack not installed' from 'Jetpack installed but not connected'.

The `siteId` logic should probably be:

| Jetpack Status | `jetpack_client_id` value (`wp.getOptions`) | siteId should be | isJetpack |
| --- | --- | --- | --- |
| not installed | N/A (not returned) | -1 | false |
| installed, but not connected | 0 | 0 | ? |
| installed and connected | 123456 (the real .com site ID) | 123456 (the real .com site ID) | true? |
##### Questions
1. Do we set `SiteModel.isJetpack = true` if Jetpack is installed but not connected? Or should we add a new method `SiteModel.isJetpackConnected()`?
[...]

PR #292, opened 2017-01-25
Issue #101: Jetpack live site tests

aforcier:
Addresses #101, adding some live tests for Jetpack sites.

* [x] A WP.com account with a single Jetpack site connected
* [x] A WP.com account with multiple Jetpack sites connected
* [x] A WP.com account with a Jetpack multisites connected
* [x] A self-hosted site without Jetpack (check isJetpack etc statuses)
* [x] A self-hosted site with Jetpack installed but not connected
* [x] A self-hosted site with Jetpack installed and connected
  * [x] User is not signed into WP.com
  * [x] Connected account is the one currently logged into (should throw `DUPLICATE_SITE` error: #285)
  * [x] Connected account is not the one currently logged into
* [x] A WP.com account with a Jetpack site that the user has already added as a self-hosted site (the self-hosted Jetpack site should be 'upgraded' to a WP.com Jetpack site: #285)

Also added a test for adding multiple self-hosted sites, including adding the same site twice (we didn't have a specific test for this before).


aforcier:
The test in b26f3d7 brings up an edge case we should be aware of: if a user is signed into WP.com, and then adds a site over XML-RPC that is Jetpack-connected _but to a different WP.com account_, it is added to the store but is flagged as `isJetpack() == true`.

We can tell sites like these because if the user is signed into WP.com, any site that matches `!isWPCom() && isJetpack()` should be a Jetpack self-hosted site linked to a different WP.com account. But we might have a surprise, for example, if we list Jetpack sites based on `mSiteStore.getJetpackSites()`.

This case is mentioned in https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/207 - I'm not sure it's critical enough to address right away, but something we should probably think about (and guard against in the meantime).

aforcier:
Ready for review - cc @maxme.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java
```diff
@@ -0,0 +1,365 @@
...
+        assertEquals(1, mSiteStore.getSelfHostedSitesCount());
+        assertEquals(0, mSiteStore.getJetpackSitesCount());
+
+        SiteModel site = mSiteStore.getSites().get(0);
+
+        assertFalse(site.isJetpack());
+        assertEquals(0, site.getSiteId());
+    }
+
+    public void testXMLRPCJetpackSiteFetch() throws InterruptedException {
```
maxme: Can we rename this one to `testXMLRPCJetpackConnectedSiteFetch`?

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java
```diff
@@ -0,0 +1,365 @@
...
+        // Wait for a network response / onChanged event
+        assertTrue(mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
+
+        // Expect no DB changes
+        assertEquals(1, mSiteStore.getSitesCount());
+        assertTrue(mSiteStore.hasJetpackSite());
+        assertTrue(mSiteStore.hasWPComSite());
+        assertFalse(mSiteStore.hasSelfHostedSite());
+
+        signOutWPCom();
```
maxme: +1

maxme:
Tests look good, great you tested the sign out in `testXMLRPCJetpackToWPComDuplicateSiteFetch`, I didn't think about it but yes the "upgraded" site should be removed when the user signs out

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java
aforcier: Done in 628c94f.

aforcier:
Ready to go @maxme!

maxme:
:shipit:

PR #296, opened 2017-01-26
Issue #289: Simplify endpoint discovery

### LINKED ISSUE #289: Remove username and password requirement from discovery process
The `SelfHostedEndpointFinder` [requires a username and password](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/f7fa9ea89c99693f22d3e910b15fd08cbd032679/fluxc/src/main/java/org/wordpress/android/fluxc/network/discovery/SelfHostedEndpointFinder.java#L100), but it's actually unnecessary - this is a remnant from WPAndroid's counterpart, which required the _HTTP username/password_ for HTTP AUTH-enabled sites. In FluxC we're handling this separately with the `HTTPAuthManager`, and can drop these fields completely.

This also means that the [`DISCOVER_ENDPOINT` action](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/f7fa9ea89c99693f22d3e910b15fd08cbd032679/fluxc/src/main/java/org/wordpress/android/fluxc/action/AuthenticationAction.java#L18-L19) can be simplified to only require a `String` payload for the URL instead of a `RefreshSitesXMLRPCPayload`.

We hadn't noticed this before because in all our use cases we already have the username and password when we attempt discovery. If a `null` username/password are passed, the discovery process fails, but it shouldn't - `system.listMethods` does not require authentication. Work on the now-famous [Pioupiou project](https://github.com/wordpress-mobile/Pioupiou/), with a URL-first login flow, uncovered this oversight (thanks @mzorz!).

aforcier:
Fixes #289, removing unnecessary username and password requirements for starting the endpoint discovery process.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_DiscoveryTest.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_DiscoveryTest.java
@@ -51,7 +51,9 @@ public class ReleaseStack_DiscoveryTest extends ReleaseStack_Base {
     }
 
     private TestEvents mNextEvent;
-    private RefreshSitesXMLRPCPayload mPayload;
+    private String mUrl;
+    private String mUsername;
+    private String mPassword;
 
     @Override
     protected void setUp() throws Exception {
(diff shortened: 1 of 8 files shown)


maxme:
I was worried this would break discovery for sites with http auth, but patch and tests looks good :shipit:

PR #298, opened 2017-01-30
FluxC image loader

maxme:
FluxC image loader that extends Volley Image loader and add authentication for WordPress.com and HTTP Auth self hosted.


tonyr59h:
Working as expected in WPAndroid and the example app :shipit:

PR #302, opened 2017-02-01
Fix #207: new isJetpackInstalled and isJetpackConnected getters

### LINKED ISSUE #207: Jetpack handling over XML-RPC

maxme:
Fix #207: new isJetpackInstalled and isJetpackConnected getters


fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/site/SiteXMLRPCClient.java
```diff
@@ -186,6 +186,41 @@ private long string2Long(String s, long defvalue) {
...
 
+    private void setJetpackStatus(Map<?, ?> siteOptions, SiteModel oldModel) {
+        // * Jetpack not installed: field "jetpack_client_id" not included in the response
+        // * Jetpack installed but not activated: field "jetpack_client_id" not included in the response
+        // * Jetpack installed, activated but not connected: field "jetpack_client_id" included
+        //   and is "0"
+        // * Jetpack installed, activated and connected: field "jetpack_client_id" included and is correctly
+        //   set to wpcom unique id eg. "1234"
+
+        long jetpackClientId = string2Long(getOption(siteOptions, JETPACK_CLIENT_ID_KEY, String.class), -1);
```
aforcier: The return of `-1`! 🛐

maxme: Hmm, actually that makes me think we should only call ` oldModel.setSiteId(jetpackClientId)` - `if (jetpackClientId != 0 && jetpackClientId != -1)`

aforcier:
May I suggest merging https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/292 first? That way we can also update those connected tests and add another case for the not installed vs not connected distinction.

There are already two tests (and live sites) in place that are currently identical, but can be distinct by checking connected/installed status:

https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/292/files#diff-7c1171e090ce28ea081691618b384679R118

https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/292/files#diff-7c1171e090ce28ea081691618b384679R158

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/site/SiteXMLRPCClient.java
```diff
@@ -186,6 +186,46 @@ private long string2Long(String s, long defvalue) {
...
+            oldModel.setIsJetpackConnected(false);
+        }
+
+        // jetpack_client_id is set then it's a Jetpack connected site
+        if (jetpackClientId != 0 && jetpackClientId != -1) {
+            oldModel.setIsJetpackInstalled(true);
+            oldModel.setIsJetpackConnected(true);
+            // We also set the isWPCom flag, since this site should be available by the WPCOM REST API
+            // TODO: check for jetpack options and turn this off if the REST API is disabled
+            oldModel.setIsWPCom(true);
```
aforcier: Isn't this dangerous? What if a user adds a self-hosted Jetpack site connected to a WP.com account other than the one logged into? Then things like `SiteStore.getWPComSites()` no longer mean what we might think they mean.

maxme: Yes it's probably dangerous in the case of mixed accounts, we can keep that false but I'd also set this flag false for other Jetpack sites coming from the wpcom REST call to fetch sites. In the wpandroid, for certains features (stats/plans/themes for instance) we should check `if the site isWPCom or (isJetpackConnected and accessibleViaAccount)`.

We should also make it clear that this doesn't mean "accessible via the WPCom REST API" but means "pure WPCom site".

aforcier: That sounds sensible 👍. Though wouldn't we set `accessibleViaAccount` true for 'pure' WPCom sites - in which case we could just use `isAccessibleViaAccount()` for stats/plans/themes/etc?

It sounds like what we want is to rename `isWPCom` to `isAccessibleViaAccount` (which is basically what `isWPCom` means in its current use), and add a `isWPCom` field to mark pure WPCom sites internally. It's not clear to me that we would even need to export a `isWPCom()` method. As far as WPAndroid is concerned, it can filter out REST API sites using `isAccessibleViaAccount`, and further check `isJetpackConnected` if it needs to distinguish between 'pure' and Jetpack WP.com REST sites.

(All of this would, of course, become a bit more complicated if we decided to support multiple accounts in the future.)

maxme: Actually `accessibleViaAccount` would be a utility method `accessibleViaAccount(AccountModel account, SiteModel model)` (or eventually only works with the default account) not a flag in the SiteModel.

We could rename `isWPCom` to `isAccessibleViaWPComREST` - but this is becoming a mess. I think you're right and if we want to have a long term vision, we should use `isWPCom` (for pure wpcom sites only), `isJetpack` (for jetpack sites), and make sure we use both in wpandroid and we don't assume `isWPCom` means `isAccessibleViaWPComREST`.

Then we could eventually add `accessibleViaAccount(AccountModel account, SiteModel model)`  to fix issues with jetpack sites not connected to the wpcom account in the app.

aforcier: Sounds good - I think we'll need `accessibleViaAccount(AccountModel account, SiteModel site)` right away though, since we'll still need that to tell Jetpack sites we pulled over the REST API from self-hosted Jetpack sites.

aforcier:
The changes made in 8a6a919 also need to be reflected in `ReleaseStack_SiteTestJetpack`.

aforcier:
The `REMOVE_WPCOM_SITES` action no longer clears Jetpack sites that were fetched over WP.com.

(To demonstrate, try running [`testWPComSiteFetchAndLogout()`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/71db5e468d0eda974323ed53673e092c835ccf72/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestWPCom.java#L54) on a WP.com account with a Jetpack site.)

maxme:
Noticed other things when integrating this:

* `getVisibleWPComSites` should be changed (and renamed) to `getVisibleWPComAndJetpackSites`
* I think we'll never use `getWPComSites` in wpandroid, but always do something like `getWPComSites().addAll(getJetpackSites())`, so we should probably write a getter `getWPComAndJetpackSites`

maxme:
This one should be ready for another round, along with the [integration counter part](https://github.com/wordpress-mobile/WordPress-Android/pull/5044).

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java
```diff
@@ -260,19 +264,20 @@ public void testWPComToXMLRPCJetpackDifferentAccountsSiteFetch() throws Interrup
         fetchSite(selfHostedSite);
 
         assertEquals(wpComSiteCount + 1, mSiteStore.getSitesCount());
+        // The site is connected to a different wpcom account but we don't make that difference yet.
         assertEquals(wpComSiteCount, mSiteStore.getWPComSitesCount());
```
aforcier: Looks like I made a mistake when I wrote this test - we're using `TEST_WPCOM_USERNAME_TEST1` as a WP.com account with no Jetpack sites - but that's not mentioned in `tests.properties`, and actually the site I'm using _does_ have a Jetpack site, making this line fail.

I propose that we correct `wpComSiteCount` to use `getWPComSitesCount()`, and track total sites separately for the later comparison. And just remove the 'with no Jetpack sites' part of the comment:

```java
public void testWPComToXMLRPCJetpackDifferentAccountsSiteFetch() throws InterruptedException {
  // Authenticate as WP.com user
  authenticateWPComAndFetchSites(BuildConfig.TEST_WPCOM_USERNAME_TEST1,
      BuildConfig.TEST_WPCOM_PASSWORD_TEST1);

  int wpComSiteCount = mSiteStore.getWPComSitesCount();
  int totalSiteCount = mSiteStore.getSitesCount();

  // Add a Jetpack-connected site as self-hosted (connected to a different WP.com account than the one above)
  fetchSitesXMLRPC(BuildConfig.TEST_WPORG_USERNAME_SINGLE_JETPACK_ONLY,
      BuildConfig.TEST_WPORG_PASSWORD_SINGLE_JETPACK_ONLY,
      BuildConfig.TEST_WPORG_URL_SINGLE_JETPACK_ONLY_ENDPOINT);

  // Fetch site details (including Jetpack status)
  SiteModel selfHostedSite = mSiteStore.getSelfHostedSites().get(0);
  fetchSite(selfHostedSite);

  assertEquals(totalSiteCount + 1, mSiteStore.getSitesCount());
...
```

maxme: Modified in 5ad69769

example/src/test/java/org/wordpress/android/fluxc/utils/SiteUtils.java
```diff
@@ -19,7 +19,7 @@ public static SiteModel generateSelfHostedNonJPSite() {
         SiteModel example = new SiteModel();
         example.setSelfHostedSiteId(6);
         example.setIsWPCom(false);
-        example.setIsJetpack(false);
+        example.setIsJetpackConnected(false);
```
aforcier: Let's do the same explicit config we do for `generateSelfHostedSiteFutureJetpack()` below and explicitly `setIsJetpackInstalled(false)` here too.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java
aforcier: @maxme can we updated the top comment for `testWPComToXMLRPCJetpackDifferentAccountsSiteFetch()` to not say 'with no Jetpack sites' anymore?

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -344,7 +359,7 @@ public int getWPComSitesCount() {
      */
     @NonNull
     public List<SiteModel> getWPComSiteByNameOrUrlMatching(@NonNull String searchString) {
-        return SiteSqlUtils.getAllSitesMatchingUrlOrName(searchString);
+        return SiteSqlUtils.getSitesMatchingUrlOrName(searchString);
```
aforcier: Hmm, this method and the one above it don't make sense - I guess they're suppose to be reversed. And let's rename the `SiteSqlUtils.getAllSitesMatchingUrlOrNameWith()` method to `getSitesMatchingUrlOrNameWith()` since 'All' is dropped from all the other methods in `SiteSqlUtils`.

maxme: Fixed in  	6187887 - should be ready for the merge 🤞

aforcier:
:shipit:

PR #305, opened 2017-02-03
Update original media after upload & Introduce instantiateMediaModel

maxme:


maxme:
Note: I'll probably update the PostStore with an instantiate method when this is merged.

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -289,6 +290,12 @@ public void onRegister() {
     // Getters
     //
 
+    public MediaModel instantiateMediaModel() {
+        MediaModel media = new MediaModel();
+        WellSql.insert(media).asSingleTransaction(true).execute();
```
aforcier: Here we need to account for the fact that the insertion can fail, and this will set the id of `media` to `-1`:

https://github.com/wordpress-mobile/wellsql/blob/28502aa3e2bcb72d2f770cf7a395297cc8019fe2/wellsql/src/main/java/com/yarolegovich/wellsql/InsertQuery.java#L61

(see the behavior of `SQLiteDatabase.insert()` called from that line for the full picture)

I don't see an elegant way to do this beside returning a `null` `MediaModel` (and annotating + documenting that null checking needs to be performed for `instantiateXModel()`-type methods.

maxme: We're not checking for insert error anywhere else (in wpandroid we call or in the fluxc), so I don't think this is an important issue, I'll add a check for  `id == -1` and return null (but I won't add any null check now and see if this ever happen and how we can fail with dignity)

aforcier: > We're not checking for insert error anywhere else (in wpandroid or in fluxc), so I don't think this is an important issue

That's true - I felt it was worth accounting for in this case since the point of having a `MediaStore.instantiateMediaModel()` method instead of just using `new MediaModel()` is that it's meant to be db-connected; so it should sound the alarm if it's not.

👍 to leaving out the `null` check for now and seeing whether this becomes a problem (I doubt it too).

aforcier:
This is currently breaking unit tests (`MediaSqlUtilsTest` and `MediaStoreTest`).

maxme:
> This is currently breaking unit tests (MediaSqlUtilsTest and MediaStoreTest).

Tests fixed in  	ac02638

aforcier:
:shipit:

PR #311, opened 2017-02-08
Use the empty body when we get the null body in a POST, PUT or PATCH method

maxme:
We're using the raw volley request queue in wpandroid, some request are incorrectly passing a null body to a POST call, this patch makes sure null body will be ignored.


aforcier:
:shipit:

PR #312, opened 2017-02-09
Make sure to call the REST API if the site is JetpackConnected

maxme:
I checked all the `isWPCom()` usages.


aforcier:
:shipit:

PR #313, opened 2017-02-09
Issue #307: Add support for Automated Transfers

### LINKED ISSUE #307: Add tests for pure WP.com sites becoming Jetpack sites (AT)
Hard to set up live site tests for this, but we should be able to mock something up (as well as run manual tests on real sites).

Some unit tests on `SiteModel` state would also be useful.

aforcier:
Fixes #307.

1. Modifies `insertOrUpdateSite()` to consider sites the same if they have the same WP.com site ID

    The reason for this is that AT sites might begin as `*.wordpress.com` sites, but they will definitely have a TLD after AT, because having a domain is a requirement for AT. If we force a match with ID and URL, we end up with a duplicate site after AT.

    Incidentally, this should fix a case I don't think we considered before - a user changing the primary domain for their WordPress.com site would end up with a duplicate.

2. Adds a unit test to simulate AT (a WordPress.com site becoming a Jetpack site)

    I decided against a mock network test for now - I think the unit test + some manual checks of a transition provide the same value a mock test would in this case.

3. Adds `SiteModel.isAutomatedTransfer()` so we can identify ATs in the app

cc @maxme

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
@@ -439,6 +439,7 @@ public class SiteStoreUnitTest {
         automatedTransferSite.setIsJetpackInstalled(true);
         automatedTransferSite.setIsJetpackConnected(true);
         automatedTransferSite.setIsWPCom(false);
+        automatedTransferSite.setIsAutomatedTransfer(true);
 
         SiteSqlUtils.insertOrUpdateSite(automatedTransferSite);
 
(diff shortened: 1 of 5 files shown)


maxme:
Can we add a comment on top of `insertOrUpdateSite`, something like:
```
* Exists in the DB already and matches by local id (simple update) -> UPDATE
* Exists in the DB is a Jetpack or WordPress site and matches by remote id (site id) -> UPDATE
* Exists in the DB is a pure self hosted and matches by remote id (site id) + url -> UPDATE
* Exists in the DB matches its XMLRPC_URL was not a Jetpack site but is now a Jetpack site -> UPDATE
* Exists in the DB matches its XMLRPC_URL -> THROW a DuplicateSiteException
* Not matching any previous cases -> INSERT
```

Also we have to update our debug messages like this https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/7338ccc6206f9b7341c4071549b3b228651056a0

aforcier:
> Can we add a comment on top of `insertOrUpdateSite`

Done in 23c8363.

> Also we have to update our debug messages like this 7338ccc.

Done in 0812a1b (there was a bug in your so I wrote my own 😛)

maxme:
:shipit:

PR #318, opened 2017-02-16
Fix #317: Check XML-RPC url by https:// or https:// when we insert a new site

### LINKED ISSUE #317: We must compare XML-RPC urls without their scheme when we check for duplicates
Here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/0812a1bca29a82350ece0de4259a85fe55fc7ed5/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java#L107-L107

maxme:
To test:

1. checkout a14ed92 - run tests - 💥 
2. checkout 907c46b - run tests - 🍺

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java
@@ -108,8 +108,10 @@ public class SiteSqlUtils {
 
             siteResult = WellSql.select(SiteModel.class)
                     .where()
+                    .beginGroup()
                     .equals(SiteModelTable.XMLRPC_URL, forcedHttpXmlRpcUrl)
                     .or().equals(SiteModelTable.XMLRPC_URL, forcedHttpsXmlRpcUrl)
+                    .endGroup()
                     .endWhere().getAsModel();
             if (!siteResult.isEmpty()) {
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java
```diff
@@ -102,9 +103,13 @@ public static int insertOrUpdateSite(SiteModel site) throws DuplicateSiteExcepti
         // If the site is a self hosted, maybe it's already in the DB as a Jetpack site, and we don't want to create
         // a duplicate.
         if (siteResult.isEmpty()) {
+            String forcedHttpXmlRpcUrl = "http://" + UrlUtils.removeScheme(site.getXmlRpcUrl());
+            String forcedHttpsXmlRpcUrl = "https://" + UrlUtils.removeScheme(site.getXmlRpcUrl());
+
             siteResult = WellSql.select(SiteModel.class)
                     .where()
-                    .equals(SiteModelTable.XMLRPC_URL, site.getXmlRpcUrl())
+                    .equals(SiteModelTable.XMLRPC_URL, forcedHttpXmlRpcUrl)
```
aforcier: This is a pretty bad nit, but could we wrap this in a group?

It's not necessary here since this is the only match we're making, but it's possible we might add something else later and miss the fact that these won't be run as one group.

aforcier:
:shipit:

PR #320, opened 2017-02-17
Fix #281: Override Payload.isError() where needed

### LINKED ISSUE #281: Make sure Payloads override isError() where necessary
Any time a `Payload` subclass replaces `Payload`'s `error` field with its own, the `Payload.isError()` method needs to be overridden by that subclass.

For example, [`NewSiteResponsePayload`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/11761e629cc1700a11ff8dd6b40de3e5c4f64789/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java#L58-L63) should be overriding `isError()`.

aforcier:
Fixes #281 - in all Payloads that were missing it, `isError()` would always return `false`.

An ideal fix would be to make it impossible for this to happen in the first place. Because of the way we're using `Payload`s, there's not really a neat fix for this with no side effects.

Probably the most effective thing would be to rip out `isError()` entirely, and go back to relying on `error != null` instead. But we can visit that later.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
@@ -47,12 +47,6 @@ public class AccountRestClient extends BaseWPComRestClient {
             this.account = account;
             this.error = error;
         }
-
-        public boolean isError() {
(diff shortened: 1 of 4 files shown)


aforcier:
Closing in favour of #323.

PR #323, opened 2017-02-18
Fix #281 by using reflection in Payload.isError() implementation

### LINKED ISSUE #281: Make sure Payloads override isError() where necessary

maxme:
Fix #281 by using reflection in Payload.isError() implementation.

It's not super clean, but it's less prone to error than https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/320


aforcier:
🥇 :shipit:

PR #325, opened 2017-02-20
Media Upload: Parse upload error

maxme:
Error type itself is kind of useless (`upload_error`), but the message can be useful, at least for debugging.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
@@ -238,7 +238,7 @@ public class MediaRestClient extends BaseWPComRestClient implements ProgressList
                     public void onErrorResponse(@NonNull BaseRequest.BaseNetworkError error) {
                         AppLog.v(T.MEDIA, "VolleyError deleting media (ID=" + media.getMediaId() + "): " + error);
                         MediaErrorType mediaError = MediaErrorType.fromBaseNetworkError(error);
-                        if (mediaError == MediaErrorType.MEDIA_NOT_FOUND) {
+                        if (mediaError == MediaErrorType.NOT_FOUND) {
                             AppLog.i(T.MEDIA, "Attempted to delete media that does not exist remotely.");
                         }
                         notifyMediaDeleted(site, media, new MediaError(mediaError));
@@ -321,7 +321,7 @@ public class MediaRestClient extends BaseWPComRestClient implements ProgressList
     private MediaError parseUploadError(okhttp3.Response response) {
         MediaError mediaError = new MediaError(MediaErrorType.GENERIC_ERROR);
         if (response.code() == 403) {
-            mediaError.type = MediaErrorType.UNAUTHORIZED;
+            mediaError.type = MediaErrorType.AUTHORIZATION_REQUIRED;
         }
         try {
             JSONObject body = new JSONObject(response.body().string());
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
@@ -8,6 +8,9 @@ import com.android.volley.RequestQueue;
 import com.android.volley.Response.Listener;
 import com.google.gson.Gson;
 
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
 import org.wordpress.android.fluxc.Dispatcher;
 import org.wordpress.android.fluxc.generated.MediaActionBuilder;
 import org.wordpress.android.fluxc.generated.endpoint.WPCOMREST;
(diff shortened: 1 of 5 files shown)


aforcier:
Can we get anything useful from the XML-RPC response in case of error? We're similarly ignoring the response body in `MediaXMLRPCClient`.

maxme:
> We're similarly ignoring the response body in MediaXMLRPCClient.

I'll open another issue for this one, since we're not using the normal client for uploads, we also need to parse the xml response from the error which not as trivial as reading the JSON response from the REST call.

maxme:
I'm working on how we handle XMLRPC errors on upload, I found an issue and I'll fix it and fill the error message as well.

maxme:
@aforcier should be ready for review

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
```diff
@@ -69,7 +70,10 @@
     public MediaRestClient(Context appContext, Dispatcher dispatcher, RequestQueue requestQueue,
                            OkHttpClient.Builder okClientBuilder, AccessToken accessToken, UserAgent userAgent) {
         super(appContext, dispatcher, requestQueue, accessToken, userAgent);
-        mOkHttpClient = okClientBuilder.build();
+        mOkHttpClient = okClientBuilder
+                .connectTimeout(BaseRequest.DEFAULT_REQUEST_TIMEOUT, TimeUnit.SECONDS)
+                .readTimeout(BaseRequest.DEFAULT_REQUEST_TIMEOUT, TimeUnit.SECONDS)
```
aforcier: Is there a reason you omitted setting `writeTimeout` (also in the XML-RPC media client), which we also set to the same value as the others in `OkHttpStack.java`?

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
```diff
@@ -163,16 +163,25 @@ public void uploadMedia(final SiteModel site, final MediaModel media) {
...
+                            AppLog.d(T.MEDIA, "media upload successful, local id=" + media.getId());
+                            // We only get the media Id from the response
+                            media.setMediaId(responseMedia.getMediaId());
+                        }
+                        fetchMedia(site, media, true);
+                    } catch (XMLRPCException fault) {
+                        AppLog.w(T.MEDIA, "media upload successful, local id=" + media.getId()
+                                          + " - " + response.message());
+                        MediaError mediaError = getMediaErrorFromXMLRPCException(fault);
+                        notifyMediaProgress(media, 0.f, mediaError);
```
aforcier: Do you know why we're setting the progress to `0` here while the corresponding cases in `MediaRestClient` send progress `1`?

maxme: I think this is a mistake, but there is an issue for this: https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/315

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
maxme: Fixed in 8098f4c0 (also fixed a timeunit issue)

aforcier:
This looks good overall, but I'm seeing one issue in how we're classifying errors.

If the `MediaRestClient` gets a `403` error, it classifies it as `UNAUTHORIZED`. In WPAndroid, if we get an `UNAUTHORIZED` media error, we display `You don't have permission to upload media to the site`, and ignore whatever the real message was.

In this changeset, `MediaXMLRPCClient` will classify a `401` error from the server, which includes a `Sorry, you have used your space allocation.` error, as `UNAUTHORIZED`. And so, we'll ignore the real message from the server, which is more useful that 'You don't have permission'.

So we need to distinguish between these two error types, or rely on `error.message` for `UNAUTHORIZED` too (we currently [don't](https://github.com/wordpress-mobile/WordPress-Android/blob/7dcb452f5c04fa6c945b52fd256cd980a631e19d/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostActivity.java#L796-L805)).

maxme:
I had a look at the XML-RPC server code and the problem is fault code 401 is used for different thing:

Some errors, are correctly `UNAUTHORIZED` (or not allowed) requests, like: `You do not have permission to upload files.`, `'Sorry, you cannot edit this post.` and other are more technical limitation or problems: `Invalid post ID.`. Error messages are localized so we cannot parse them and classify the errors.

Since this is a mess, my opinion is to not classify 401 errors in `UNAUTHORIZED` but keep them as `GENERIC_ERRORS` for the XML-RPC client. Thus in wpandroid, `error.message` will be used.

This won't impact REST responses that should be correctly categorized as UNAUTHORIZED`.

Note: we're using error type `AUTHORIZATION_REQUIRED` in the base request (same meaning as `UNAUTHORIZED`), I'll rename the error type to `UNAUTHORIZED` because that adds even more confusion to the whole thing.

aforcier:
Looks good! Tested with this branch in https://github.com/wordpress-mobile/WordPress-Android/pull/5248, and I'm seeing the `error.message` everywhere I should be 👍.

Waiting for comments in https://github.com/wordpress-mobile/WordPress-Android/pull/5248, and then :shipit:.

PR #330, opened 2017-02-24
Increase timeout for new site creation to 90 seconds

aforcier:
In all my attempts, new site creation (when not a dry run) took over 30 seconds (our default timeout), resulting in an empty error being displayed in WPAndroid (meanwhile the site is actually created and another attempt to create the same site will say it already exists).

The completion time was always less than 60 seconds in my tests, but I'm setting the timeout to 90 seconds for new site creation just to be safe.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
@@ -3,6 +3,7 @@ package org.wordpress.android.fluxc.network.rest.wpcom.site;
 import android.content.Context;
 import android.support.annotation.NonNull;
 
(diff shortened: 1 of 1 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -205,6 +208,8 @@ public void onErrorResponse(@NonNull BaseNetworkError error) {
         );
 
         request.disableRetries();
```
maxme: We can drop `request.disableRetries();` which is just a convenience method that calls `setRetryPolicy`

aforcier: 👍 fixed in bfc2e79.

PR #332, opened 2017-02-24
Add getMediaWithLocalId()

aforcier:
Adds a `getMediaWIthLocalId()` method to `MediaStore` so we can look up media by local ID rather than relying on `getSiteMediaWithId()`, which doesn't help us for local media.


maxme:
:shipit:

PR #335, opened 2017-02-25
Fix #334: limit the rate of _onProgress callback_ calls

### LINKED ISSUE #334: Debounce the UPLOADED_MEDIA events
I tried to upload a very big image (50Mb) on a fast connection and I got something like 500 UPLOADED_MEDIA events per second, this is way too much and we only use this to update the progress bar. 10 events per second is more than enough.

maxme:
And thus reduce the number of UPLOADED_MEDIA events.


aforcier:
Looks good 👍 I wonder if 100ms is even too frequent for progress updates - but we can tweak this in the future if we want.

:shipit:

PR #336, opened 2017-02-27
Add remote id to on new site created events

maxme:
Note: I wanted to use the local id instead of the remote site id, but we don't get the full new created site in the new site response, so we don't add it to the local db. We only get the new site after FETCH_SITES or eventually a FETCH_SITE.


aforcier:
:shipit:

PR #337, opened 2017-03-01
Default sites to visible in SiteModel

aforcier:
The default setting for site visibility in `SiteModel` should be `true` - self-hosted sites shouldn't have to worry about handling this, and WP.com sites already handle it on site fetch.


maxme:
:shipit:

PR #338, opened 2017-03-01
Make OnChanged classes static

aforcier:
Makes all `OnChanged` classes `static`.

This also makes it easier to use cleaner imports for `OnChanged` events (AS will offer replacing `MediaStore.OnMediaListFetched` with `OnMediaListFetched` without needing to manually add that import.


maxme:
Looks good
```
$ ag "class\sOn"|grep -v static|wc -l
0
```

PR #340, opened 2017-03-01
Fix #339: Handle site fetch response for VIP sites

### LINKED ISSUE #339: Deal with 'vip' plans response to FETCH_SITES
VIP sites will have [`product_id = 'vip'` rather than a number](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/bc85f40f958b228cdc2093cb855ae72f08aa7271/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteWPComRestResponse.java#L23), which causes a (handled) `NumberFormatException` and sites are not stored.

aforcier:
Fixes #339. VIP sites return a String `product_id` (`'vip'`) rather than a number. This caused a handled `NumberFormatException`, resulting in no WP.com sites being fetched.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteWPComRestResponse.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteWPComRestResponse.java
@@ -6,8 +6,6 @@ import org.wordpress.android.fluxc.network.Response;
 import java.util.List;
 
 public class SiteWPComRestResponse extends Payload implements Response {
-    public static final long VIP_PLAN_ID = 31337;
-
     public class SitesResponse {
         public List<SiteWPComRestResponse> sites;
     }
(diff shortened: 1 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteWPComRestResponse.java
```diff
@@ -6,6 +6,8 @@
 import java.util.List;
 
 public class SiteWPComRestResponse extends Payload implements Response {
+    public static final long VIP_PLAN_ID = 31337;
```
maxme: I wonder if we should move that to the SiteModel, in case we want to use it in the app, it makes more sense to `site.getProductId() == SiteModel.VIP_PLAN_ID`

maxme:
I confirm this fixes the issue VIP users were getting on sign in.

maxme:
Ready to 🚢 / waiting for @aforcier reply to https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/340#discussion_r103857062

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteWPComRestResponse.java
aforcier: Makes sense, done in 3dde283.

maxme:
:shipit:

PR #343, opened 2017-03-02
Fix DB duplicates for posts and media

aforcier:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/5258 (hopefully). The issue I found is a race condition where a remote fetch reaches the db before the result of an upload does:

1. `PUSH_POST` for a local draft and `FETCH_POSTS` at ~the same time
2. Say the post list fetch completes after the post is uploaded, but before the DB is updated with the remote ID for the post
3. First db call, remote post is added, local ID is `0`, no match with existing post (site ID + remote ID give no match, the uploaded post hasn’t been updated yet with its remote ID) 💥
4. Second db call, already have two entries for the same post, at this point matches by local ID, and updates the remote post ID, leaving the other entry untouched
5. Two entries for the same post, only different by their local ID

Any model that we store 'local-only' versions of will experience this issue - for now that's only posts and media, so I made the fix for media also.

We should make the same change for other models we might store local versions of in the future, but I didn't want to bloat this PR - once merged, I'll open an issue for this. There might be a more generic way to take care of this for things we add in the future too.

To test:
1. Check out ee36d30, `./gradlew testDebug` 💥
2. Check out ae03283, `./gradlew testDebug` 👍
3. Check out f48990d, `./gradlew testDebug` 💥
2. Check out branch, `./gradlew testDebug` 👍

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/MediaSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/MediaSqlUtils.java
@@ -192,20 +192,15 @@ public class MediaSqlUtils {
     public static int insertOrUpdateMedia(MediaModel media) {
         if (media == null) return 0;
 
-        // If the site already exist and has an id, we want to update it.
         List<MediaModel> existingMedia = WellSql.select(MediaModel.class)
                 .where().beginGroup()
                 .equals(MediaModelTable.ID, media.getId())
+                .or()
+                .beginGroup()
+                .equals(MediaModelTable.LOCAL_SITE_ID, media.getLocalSiteId())
+                .equals(MediaModelTable.MEDIA_ID, media.getMediaId())
+                .endGroup()
                 .endGroup().endWhere().getAsModel();
-
-        // Looks like a new media, make sure we don't already have it by site id / media id.
-        if (existingMedia.isEmpty()) {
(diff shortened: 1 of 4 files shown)


aforcier:
cc @maxme since he's the only one able to reproduce this live 😛

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/MediaSqlUtils.java
```diff
@@ -192,26 +192,28 @@ public static MediaModel getMediaWithLocalId(int localMediaId) {
...
-                    .equals(MediaModelTable.MEDIA_ID, media.getMediaId())
-                    .endGroup().endWhere().getAsModel();
-        }
-
         if (existingMedia.isEmpty()) {
             // insert, media item does not exist
             WellSql.insert(media).asSingleTransaction(true).execute();
             return 1;
         } else {
+            if (existingMedia.size() > 1) {
```
maxme: This will break multiple concurrent uploads (or queued/failed uploads) on the same site, they will have the same local_site_id and media_id (0).

aforcier: Good point - fixed in 189ffe0.

maxme:
:shipit:

PR #344, opened 2017-03-04
Store correct timezone data from site options

aforcier:
The `timezone` field we were previously using from site options always seems to be blank - we actually want the `gmt_offset` field, which contains the GMT offset as a number (defaults to `0`).

This is the same format as the `time_zone` field from XML-RPC, and is what current WPAndroid code is expecting from `SiteModel.getTimezone()`.


maxme:
:shipit:

PR #350, opened 2017-03-09
Store site icons URLs

aforcier:
Stores the `icon.img` field from the WP.com REST API `sites` response. This is necessary to display site icons for Jetpack sites (though normal WP.com sites with an icon will also have its URL stored this way).

Also features our first DB migration 🎉 😂 😱 (assuming it's merged).

This completes the backend requirement to fix https://github.com/wordpress-mobile/WordPress-Android/issues/2822 - I'll make the rest of the changes in a WPAndroid PR.

Note: The `FallThrough` warning suppression part (09bc9cd) isn't necessary yet, but will be on the next migration if we use the incrementing switch approach we've been using in `WordPressDB`.

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -205,6 +205,7 @@
         <module name="UpperEll"/>
 
         <module name="FileContentsHolder"/> <!-- Required by comment suppression filters -->
+        <module name="SuppressWarningsHolder"/> <!-- Required by suppression via annotation filter -->
 
         <module name="OneStatementPerLine"/>
 
@@ -222,4 +223,6 @@
       <property name="checkFormat" value="$1"/>
       <property name="influenceFormat" value="1"/>
     </module>
+
+    <module name="SuppressWarningsFilter"/>
 </module>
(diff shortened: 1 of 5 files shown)


maxme:
> The FallThrough warning suppression part (09bc9cd) isn't necessary yet, but will be on the next migration if we use the incrementing switch approach we've been using in WordPressDB.

There is another way to do this: add a `// fall through` comment at the end of the case, like in:
```java
        switch (oldVersion) {
            case 1:
                db.execSQL("alter table SiteModel add ICON_URL text;");
                // fall through
            case 2:
                break;
        }
```

I like how verbose this is, but for this method in particular I think it's better to use the `SuppressWarnings` annotation to keep the future big switch clean.

maxme:
Looks good, I get gravatar URLs for "old" icons and [non gravatar URLs](https://bia.is/wpbeta/wp-content/uploads/2017/03/cropped-beau-petit-poney-1.jpg) for Jetpack sites.

:shipit:

PR #352, opened 2017-03-10
Report parse errors in an OnChanged event

aforcier:
As part of https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/100, this helps ensure we get error reports for Gson parse errors, even though they don't actually cause an app crash.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
@@ -15,6 +15,7 @@ import com.android.volley.TimeoutError;
 import com.android.volley.VolleyError;
 
 import org.wordpress.android.fluxc.store.AccountStore.AuthenticateErrorPayload;
+import org.wordpress.android.fluxc.utils.ErrorUtils.OnParseError;
 import org.wordpress.android.util.AppLog;
 
 import java.util.HashMap;
@@ -32,10 +33,14 @@ public abstract class BaseRequest<T> extends Request<T> {
     public interface BaseErrorListener {
         void onErrorResponse(@NonNull BaseNetworkError error);
     }
+    public interface OnParseErrorListener {
+        void onParseError(OnParseError event);
+    }
 
     private static final String USER_AGENT_HEADER = "User-Agent";
 
     protected OnAuthFailedListener mOnAuthFailedListener;
+    protected OnParseErrorListener mOnParseErrorListener;
     protected final Map<String, String> mHeaders = new HashMap<>(2);
     private BaseErrorListener mErrorListener;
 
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
```diff
@@ -219,6 +228,9 @@ private BaseNetworkError getBaseNetworkError(VolleyError volleyError) {
     @Override
     public final void deliverError(VolleyError volleyError) {
         AppLog.e(AppLog.T.API, "Volley error", volleyError);
+        if (volleyError instanceof ParseError) {
```
maxme: Are we sure this is always an error while parsing the API response here - Can we get this from a malformed HTTP response header? (is this important?)

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
```diff
@@ -56,6 +65,7 @@ protected AccessToken getAccessToken() {
 
     private WPComGsonRequest setRequestAuthParams(WPComGsonRequest request) {
         request.setOnAuthFailedListener(mOnAuthFailedListener);
+        request.setOnParseErrorListener(mOnParseErrorListener);
```
maxme: Should we do the same in `BaseXMLRPCClient` to catch XMLRPC parse errors? (I guess we can get a lot of false possitive due to plugins/wp versions, but that might be interesting at least temporary)

fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
aforcier: > Can we get this from a malformed HTTP response header?

I'm not sure - I think the answer is no. Volley in general seems pretty hard to break by messing with response headers. A malformed static line will result in a `NoConnectionError` rather than a `ParseError`, but not much else I've been able to do to the response headers has caused a `VolleyError`.

maxme: Ok, that's not a blocker anyway and I hope we can get this info from the error message.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
aforcier: Probably worth doing 👍 - I'll add that in.

aforcier:
Added reporting for XML-RPC parsing errors in 2bed1dc.

maxme:
I was able to report this fake problem https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/eccdca6a247e9e6fa1aebe73c2aad65eee75d944

:shipit:

PR #353, opened 2017-03-10
Fix loading images from HTTP Auth sites

aforcier:
Fixes a couple of issues that were preventing images from being downloaded using the `FluxCImageLoader` for HTTP Auth sites.

We should probably eventually take a more refined approach than what I did in e8c1b88. We could probably update the `ROOT_URL` in `HTTPAuthModel` to the site's reported URL once the initial site fetch is done. There isn't an obvious place to do this (it would have to be handled in the clients, probably), and I decided to skip the necessary refactor for now.

The only situation where the current fix won't work is if the user has moved their `xmlrpc.php` file deeper in their site's folder structure (e.g. their site url is http://mysecretsite.site/wp, and they stored their `xmlrpc.php` file in http://mysecretsite.site/cant-find-me/xmlrpc.php). We can't just use the domain, because their site might live in a subfolder, and we have no way of knowing where the true base URL is from the XML-RPC URL alone.

To test:
1. Check out a0850fe, run test 💥
2. Check out the branch, run test 🎉

cc @maxme

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_FluxCImageLoaderTest.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_FluxCImageLoaderTest.java
@@ -78,7 +78,7 @@ public class ReleaseStack_FluxCImageLoaderTest extends ReleaseStack_Base {
                             @Override
                             public void onErrorResponse(VolleyError error) {
                                 AppLog.e(T.TESTS, "Failed to download image: " + error.getMessage());
-                                throw new AssertionError("Image request failed!");
+                                fail("Image request failed!");
                             }
                         }
                 );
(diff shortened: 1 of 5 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_FluxCImageLoaderTest.java
```diff
@@ -0,0 +1,144 @@
...
+                                    return;
+                                }
+                                assertNotNull(response.getBitmap());
+                                mCountDownLatch.countDown();
+                            }
+
+                            @Override
+                            public void onErrorResponse(VolleyError error) {
+                                AppLog.e(T.TESTS, "Failed to download image: " + error.getMessage());
+                                throw new AssertionError("Image request failed!");
```
maxme: `fail("Image request failed!");` looks better

Note: I wonder if we can find a better solution to fail/assert in the junit thread to avoid the "process crashed" error when running tests.

aforcier: > `fail("Image request failed!");` looks better

I agree - I guess we should eventually replace all our `throw new AssertionError`s.

> Note: I wonder if we can find a better solution to fail/assert in the junit thread to avoid the "process crashed" error when running tests.

I'll open an issue so we can look into this in the future.

aforcier: Fixed in 262ddfd, and opened #355.

PR #356, opened 2017-03-13
Clean up featured image handling

aforcier:
Simplifies the upload logic for featured images. The reason `mLastKnownRemoteFeaturedImageId` and `featuredImageHasChanged()` existed was `metaWeblog.editPost`, which threw an error when attempting to set the featured image id to the already existing value.

We switched to `wp.editPost` since I added that to the `PostModel` (it existed in the old `Post` class too). `wp.editPost` doesn't have this issue (and the REST API never has), and so we no longer have to track featured image changes, only the latest value of the featured image id.

#### Note on migration/deprecation
I went back and forth for a while on what to do with the unused `mLastKnownRemoteFeaturedImageId` field. In the old `WordPressDB`, we didn't do migrations for removed fields. They'd stay there inactive in older installations, and would never be created in new installations.

@maxme suggested either doing a migration for our peace of mind, or keeping the field and deprecating the getters/setters. Migrations for removals/renames/type changes are ugly in SQLite, as you have to create a new table (and hardcode a String with its fields and types) and migrate the data into it. So I went with deprecation, and perhaps a future change to `PostModel` that does require a migration can drop `LAST_KNOWN_REMOTE_FEATURED_IMAGE_ID` at the same time.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
@@ -24,7 +24,6 @@ import java.util.List;
 
 @Table
 public class PostModel extends Payload implements Cloneable, Identifiable, Serializable {
-    private static final long FEATURED_IMAGE_INIT_VALUE = -2;
     private static final long LATITUDE_REMOVED_VALUE = 8888;
     private static final long LONGITUDE_REMOVED_VALUE = 8888;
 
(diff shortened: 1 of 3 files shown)


maxme:
cAT tests are passing.

maxme:
> @maxme suggested either doing a migration for our peace of mind, or keeping the field and deprecating the getters/setters.

I think it's fine to only deprecate them because these getters/setters are not used in wpandroid. They were only used internally, so we don't risk a bug in wpandroid due to an incorrect use of one of them.

maxme:
:shipit:

PR #359, opened 2017-03-14
Set IsAvailableResponsePayload type/value on error

aforcier:
Fixes a NPE if an `is-available` request returns an error.


maxme:
:shipit:

PR #364, opened 2017-03-15
Fix #363: Continue inserting/updating site even if we get a DuplicateSite exception

### LINKED ISSUE #363: When inserting or updating batch sites: if one is malformed, we should ignore it and continue

maxme:
Fix #363: Continue inserting/updating site even if we get a DuplicateSite exception

I'm not super fan of the fix (re throw the caught exception at the end of the loop).

1. checkout  20d15e3, `./gradlew test`: 💥 
2. checkout  cdee283,  `./gradlew test`: 🏖 🍸

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
@@ -14,6 +14,7 @@ import org.robolectric.RuntimeEnvironment;
 import org.wordpress.android.fluxc.Dispatcher;
 import org.wordpress.android.fluxc.model.PostFormatModel;
 import org.wordpress.android.fluxc.model.SiteModel;
+import org.wordpress.android.fluxc.model.SitesModel;
 import org.wordpress.android.fluxc.network.rest.wpcom.site.SiteRestClient;
 import org.wordpress.android.fluxc.network.xmlrpc.site.SiteXMLRPCClient;
 import org.wordpress.android.fluxc.persistence.SiteSqlUtils;
@@ -21,6 +22,9 @@ import org.wordpress.android.fluxc.persistence.SiteSqlUtils.DuplicateSiteExcepti
 import org.wordpress.android.fluxc.persistence.WellSqlConfig;
 import org.wordpress.android.fluxc.store.SiteStore;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.List;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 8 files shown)


maxme:
Oh and this doesn't fix the underlying issue. I think we should add a test [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/c8c1b05505221ac2f65e344cb34cffb68587f1ca/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java#L120-L120)  to exclude wpcom sites.

example/src/test/java/org/wordpress/android/fluxc/utils/SiteUtils.java
```diff
@@ -8,10 +8,17 @@
 
 public class SiteUtils {
     public static SiteModel generateWPComSite() {
+        return generateWPComSite(556, "", "", true, true);
+    }
+
+    public static SiteModel generateWPComSite(long remoteId, String url, String xmlRpcUrl, boolean isWPCom,
```
aforcier: `generateWPComSite()` isn't accurate anymore since the method accepts `boolean isWPCom` - maybe just rename to `generateTestSite()`?

maxme: Fixed in 740f6fef - I started to factorize other methods to use it (generateSelfHostedNonJPSite, generateJetpackSiteOverXMLRPC, etc.) but it makes these methods less readable.

maxme:
I'll update the PR to pass the number of rows affected to the error event

maxme:
> I'll update the PR to pass the number of rows affected to the error event

Done in  	bf97c16

example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
```diff
@@ -515,10 +516,62 @@ public void testBatchInsertSiteDuplicateWPCom()
...
+        createOrUpdateSites.setAccessible(true);
+        UpdateSitesResult res = (UpdateSitesResult) createOrUpdateSites.invoke(mSiteStore, sites);
+
+        assertFalse(res.duplicateSiteFound);
+        assertEquals(5, res.rowsAffected);
+        assertEquals(5, mSiteStore.getSitesCount());
+    }
+
+    @Test
+    public void testBatchInsertSiteNoDuplicateWPCom2()
```
aforcier: This test needs a different name - `testSingleInsertSiteDuplicateWPCom` or something.

maxme: Test name updated in  	e65d835

PR #374, opened 2017-03-18
Example app: Add the ability to upload Video

daniloercoli:
This PR adds the ability to upload Videos from the example app.

I thought it was easy as changing the `type` of the picker intent, setting it to ` intent.setType("image/*, video/*");`, instead it seems that the default picker does ignore everything after the first type. 
No matter how you wrote the allowable types. I then added a 2nd button to upload video.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/MediaFragment.java
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/MediaFragment.java
@@ -126,16 +126,17 @@ public class MediaFragment extends Fragment {
             }
         });
 
-        view.findViewById(R.id.upload_media).setOnClickListener(new View.OnClickListener() {
+        view.findViewById(R.id.upload_image).setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View view) {
-                if (mSite == null) {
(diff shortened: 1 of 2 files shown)


maxme:
I think we should just have one "Upload file" button, so we can later support other media upload (audio, pdf). The example app is Developer tool, not "user facing".

daniloercoli:
> The example app is a developer tool, not "user facing".

Yes, and it was my intention, but it seems you can do that with one button, since the picker ignore the 2nd mime-type passed. Wasn't able to find a way to show both videos and images in the default Android picker (Google Photos apps in my case).

maxme:
> it seems you can do that with one button, since the picker ignore the 2nd mime-type passed. Wasn't able to find a way to show both videos and images in the default Android picker (Google Photos apps in my case).

Ok, I thought we could just use "*" as mimetype.

daniloercoli:
Closing this PR since it seems we can use `*` as mime type. It's a ugly solution because the device will show a big list of apps the first time you tap the button, and if you select "remember" you will lose the ability to select audio later without clearing prefs, but this could work in the demo app.

PR #375, opened 2017-03-20
Add nullability annotations to PostModel

aforcier:
Nearly all `PostModel` getters already never return `null` - this PR brings the rest in line and adds annotations.

I think we might want to consider giving all our models this treatment - `PostModel` is special because `Post.java` historically [took the same approach](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/e013c94a64ea0b3139e9eb3f3b38a5a196d7a133/fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java), so all the `StringUtils.notNullStr()` wrappers were carried over in the original re-write.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
@@ -1,6 +1,7 @@
 package org.wordpress.android.fluxc.model;
 
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.text.TextUtils;
 
 import com.yarolegovich.wellsql.core.Identifiable;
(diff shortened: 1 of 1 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -357,33 +355,34 @@ public boolean equals(Object other) {
         PostModel otherPost = (PostModel) other;
 
         return getId() == otherPost.getId() && getLocalSiteId() == otherPost.getLocalSiteId()
-               && getRemoteSiteId() == otherPost.getRemoteSiteId() && getRemotePostId() == otherPost.getRemotePostId()
```
maxme: My AS, reformat this like it was before the patch (`&&` aligned with `getId`) - I wonder if we have a different AS configuration for formatting or if you added the spaces manually?

aforcier: Must be a different AS config as AS formats it to 8 spaces (rather than aligned with the top statement).

maxme:
:shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
aforcier: Future reader: it turned out that @maxme had `Align when multiline` enabled in AS under `Editor` > `Code Style` > `Java` > `Wrapping and Braces` > `Binary expressions`. Default settings won out.

PR #383, opened 2017-03-21
Handle /read/feed/$feed_url_or_id endpoint processing

aforcier:
Addresses item 2 in #128:

> Fix stripping of _id from /read/feed/$feed_url_or_id (or create a new rule that catches this case) ref: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/126#issuecomment-240675471

I ended up just adding the endpoint itself - we'll eventually need it for `ReaderStore` anyway. In the meantime it's useful for testing this PR and as an example of using mixed-type endpoints, which were added in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/181.

### DIFF
--- a/fluxc/src/main/tools/wp-com-endpoints.txt
+++ b/fluxc/src/main/tools/wp-com-endpoints.txt
@@ -1,3 +1,10 @@
+/auth/send-login-email/
+
+/is-available/blog/
+/is-available/domain/
+/is-available/email/
+/is-available/username/
+
 /me/
 /me/settings/
 /me/sites/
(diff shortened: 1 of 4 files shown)


fluxc/src/main/tools/wp-com-endpoints.txt
```diff
@@ -1,41 +1,43 @@
+/auth/send-login-email/
```
maxme: I'm not sure to understand the order here. I think you want to keep them sorted, but I noticed some of them are not correctly sorted, should we just do:

```shell
$ sort fluxc/src/main/tools/wp-com-endpoints.txt  | grep -v "^$" | uniq | tee fluxc/src/main/tools/wp-com-endpoints.txt
```

maxme: (Note: I don't care about the order, I think we should just keep them grouped for readability)

maxme:
Looks good, we're now ready to implement the ReaderStore ;) - I'm not merging yet (not urgent) waiting for a reply to the [nitpicking comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/383/files#r109655407)

fluxc/src/main/tools/wp-com-endpoints.txt
aforcier: My aim was to sort the endpoints within each group (in a `$a-z` order), but keep them logically grouped. So for example, all these make sense to me to group together as site-level endpoints:

```
/sites/
/sites/$site/delete
/sites/$site/exports/start
/sites/$site/post-formats/
/sites/new/
```

I'm not sure that dropping the spaces between the groups makes it more readable 😕

`/sites/$site/comments/$comment_ID/likes/new` is out of alphabetical order though, that one should be fixed if we keep my scheme.

maxme: We don't need to be strict about it, so let's merge it and we will sort it again when our OCD kicks in.

maxme:
:shipit:

PR #387, opened 2017-03-23
Fix offset calculation when fetching more pages

aforcier:
We were using the wrong method when getting the current number of fetched pages, so our offset was never changing as more pages were fetched.


maxme:
:shipit:

PR #388, opened 2017-03-23
Use REST client when cancelling media uploads to Jetpack sites

aforcier:
Noticed a case where we weren't using the WP.com REST client for Jetpack sites.


maxme:
I double checked there was no other cases, `ag isWPCom | grep -v Jetpack` looks good. 
:shipit:

PR #394, opened 2017-03-27
Fix #393: Account for missing home_url in wp.getOptions

### LINKED ISSUE #393: Crash if site is empty in XML-RPC wp.getOptions response
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
       at java.net.URI.parseURI(URI.java:353)
       at java.net.URI.<init>(URI.java:204)
       at org.wordpress.android.fluxc.model.SiteModel.setUrl(SiteModel.java:110)
       at org.wordpress.android.fluxc.network.xmlrpc.site.SiteXMLRPCClient.updateSiteFromOptions(SiteXMLRPCClient.java:228)
       at org.wordpress.android.fluxc.network.xmlrpc.site.SiteXMLRPCClient.access$400(SiteXMLRPCClient.java:30)
       at org.wordpress.android.fluxc.network.xmlrpc.site.SiteXMLRPCClient$3.onResponse(SiteXMLRPCClient.java:103)
       at org.wordpress.android.fluxc.network.xmlrpc.XMLRPCRequest.deliverResponse(XMLRPCRequest.java:52)
```

(via WPAndroid 7.0)

aforcier:
Apparently `wp.getOptions` can be missing `home_url`, and we're ignoring the `@NonNull` annotation in `SiteModel.setUrl()`, causing the crash.

Since we already might have the `home_url` from `wp.getUsersBlogs`, I opted to just not update it if it's null or empty. Even if it's missing from both responses, we mostly use it for display purposes and the app can function without it (I'd like to do a sanity check with `SiteModel.getUrl()` returning `null` in WPAndroid to be sure though).

cc @maxme

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/site/SiteXMLRPCClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/site/SiteXMLRPCClient.java
@@ -223,7 +223,11 @@ public class SiteXMLRPCClient extends BaseXMLRPCClient {
 
     private SiteModel updateSiteFromOptions(Object response, SiteModel oldModel) {
         Map<?, ?> siteOptions = (Map<?, ?>) response;
-        oldModel.setName(getOption(siteOptions, SITE_TITLE_KEY, String.class));
+        String siteTitle = getOption(siteOptions, SITE_TITLE_KEY, String.class);
+        if (!TextUtils.isEmpty(siteTitle)) {
+            oldModel.setName(siteTitle);
+        }
+
         // TODO: set a canonical URL here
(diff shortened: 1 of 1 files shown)


maxme:
> I opted to just not update it if it's null or empty.

Looks good, but I'd do the same for `setName` in `updateSiteFromOptions`. We also get the name from the `wp.getUsersBlogs` response.

aforcier:
> Looks good, but I'd do the same for setName in updateSiteFromOptions. We also get the name from the wp.getUsersBlogs response.

Makes sense, done in e489cac.

maxme:
:shipit:

PR #396, opened 2017-03-28
Add method to search local posts by title

tonyr59h:
This is similar to the [MediaStore's search method](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java#L419). wordpress-mobile/WordPress-Android#62 will be using this to search the post list.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
@@ -275,6 +275,14 @@ public class PostStore extends Store {
         }
     }
 
+    public List<PostModel> searchPostTitles(SiteModel site, String searchTerm) {
+        return PostSqlUtils.searchPostTitles(site, searchTerm, false);
+    }
+
+    public List<PostModel> searchPageTitles(SiteModel site, String searchTerm) {
+        return PostSqlUtils.searchPostTitles(site, searchTerm, true);
+    }
+
     @Subscribe(threadMode = ThreadMode.ASYNC)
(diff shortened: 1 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -275,6 +275,14 @@ public PostModel getPostByLocalPostId(long localId) {
         }
     }
 
+    public List<PostModel> searchPostTitles(SiteModel site, String searchTerm) {
```
catehstn: can there be unit tests for these methods?

tonyr59h: Can do 👍

maxme:
We should wait before merging this.

1. Are we adding local search in the app? I personally think this is confusing and we should only have online search for .com/jetpack users.
2. If yes, are we searching by matching title only?

tonyr59h:
I think searching locally makes sense. If there's a bunch of posts cached locally I'd expect (as a user) that the search feature would be able to filter them without a network connection. This is something we need more input for it seems. cc @mattmiklic 

[The WPAndroid branch](https://github.com/wordpress-mobile/WordPress-Android/tree/issue/62-search-posts-non-searchable) is currently hiding the search field for self-hosted sites. I'll update it to check for Jetpack as well.

tonyr59h:
Closing this in favor of an online search.

PR #399, opened 2017-03-29
Feature/multi uploads singleton okhttpclient

mzorz:
This PR makes the OkHttpClient being used for media uploads a singleton, as recommended in OkHttpClient specs here https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html

> OkHttpClients should be shared
> 
> OkHttp performs best when you create a single OkHttpClient instance and reuse it for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes resources on idle pools.
> 

cc @aforcier


aforcier:
@mzorz the commit history is a bit messed up, there are commits that shouldn't belong here - could you take a look at that?

mzorz:
Closing this in favor of #400

PR #400, opened 2017-03-29
okHttpClient shared instance

mzorz:
Supersedes #399 (bad commit history)

This PR makes the OkHttpClient being used for media uploads a singleton, as recommended in OkHttpClient specs here https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html

> OkHttpClients should be shared
> 
> OkHttp performs best when you create a single OkHttpClient instance and reuse it for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes resources on idle pools.
> 

cc @aforcier

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
@@ -41,7 +41,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.TimeUnit;
 
 import okhttp3.Call;
 import okhttp3.Callback;
@@ -69,13 +68,9 @@ public class MediaRestClient extends BaseWPComRestClient implements ProgressList
     private ConcurrentHashMap<Integer, Call> mCurrentUploadCalls = new ConcurrentHashMap<>();
 
     public MediaRestClient(Context appContext, Dispatcher dispatcher, RequestQueue requestQueue,
-                           OkHttpClient.Builder okClientBuilder, AccessToken accessToken, UserAgent userAgent) {
+                           OkHttpClient okHttpClient, AccessToken accessToken, UserAgent userAgent) {
         super(appContext, dispatcher, requestQueue, accessToken, userAgent);
-        mOkHttpClient = okClientBuilder
-                .connectTimeout(BaseRequest.DEFAULT_REQUEST_TIMEOUT, TimeUnit.MILLISECONDS)
-                .readTimeout(BaseRequest.UPLOAD_REQUEST_READ_TIMEOUT, TimeUnit.MILLISECONDS)
-                .writeTimeout(BaseRequest.DEFAULT_REQUEST_TIMEOUT, TimeUnit.MILLISECONDS)
-                .build();
+        mOkHttpClient = okHttpClient;
     }
 
     @Override
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
@@ -456,7 +456,9 @@ public class MediaRestClient extends BaseWPComRestClient implements ProgressList
         media.setCaption(from.caption);
         media.setDescription(from.description);
         media.setAlt(from.alt);
-        media.setThumbnailUrl(from.thumbnails.thumbnail);
+        if (from.thumbnails != null) {
+            media.setThumbnailUrl(from.thumbnails.thumbnail);
+        }
         media.setHeight(from.height);
         media.setWidth(from.width);
         media.setLength(from.length);
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
@@ -48,7 +48,6 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.TimeUnit;
 
 import okhttp3.Call;
 import okhttp3.Callback;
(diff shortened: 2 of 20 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/SharedOkHttpClientForMedia.java
```diff
@@ -0,0 +1,22 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.media;
+
+import org.wordpress.android.fluxc.network.BaseRequest;
+
+import java.util.concurrent.TimeUnit;
+
+import okhttp3.OkHttpClient;
+
+public class SharedOkHttpClientForMedia {
```
aforcier: I think `MediaOkHttpClient` would be better and fit our naming scheme better in general - and maybe we could also move this up three packages to `org.wordpress.android.fluxc.network`? What do you think @maxme?

Also I wonder if we should just use inject this in the media clients.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/SharedOkHttpClientForMedia.java
```diff
@@ -0,0 +1,22 @@
...
+    private static OkHttpClient mOkHttpClient;
+
+    public static OkHttpClient getOkHttpClientSharedInstance(OkHttpClient.Builder builder) {
```
mzorz: @maxme what do you think about using DI here instead? I've seen you're using DI for `Singleton` elsewhere in FluxC, but we're not sure wether to use it (or not) here. Open to learning advantages 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/SharedOkHttpClientForMedia.java
mzorz: Good call, addressed both moving the package and renaming the class in b026cc0
Waiting for @maxme 's input re: injection

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/SharedOkHttpClientForMedia.java
maxme: Yes, please, it's much easier to use dagger for this.

mzorz: addressed in 0848332

fluxc/src/main/java/org/wordpress/android/fluxc/network/MediaOkHttpClient.java
```diff
@@ -0,0 +1,22 @@
...
+
+import dagger.Module;
+import dagger.Provides;
+import okhttp3.OkHttpClient;
+
+@Module
+public class MediaOkHttpClient {
+    @Provides
+    @Singleton
+    public static OkHttpClient getOkHttpClientSharedInstance(OkHttpClient.Builder builder) {
```
maxme: We should move this to the `ReleaseNetworkModule`. This method shouldn't be static (so we're not tempted to by pass injection).

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
```diff
@@ -71,11 +71,7 @@
     public MediaRestClient(Context appContext, Dispatcher dispatcher, RequestQueue requestQueue,
                            OkHttpClient.Builder okClientBuilder, AccessToken accessToken, UserAgent userAgent) {
         super(appContext, dispatcher, requestQueue, accessToken, userAgent);
-        mOkHttpClient = okClientBuilder
-                .connectTimeout(BaseRequest.DEFAULT_REQUEST_TIMEOUT, TimeUnit.MILLISECONDS)
-                .readTimeout(BaseRequest.UPLOAD_REQUEST_READ_TIMEOUT, TimeUnit.MILLISECONDS)
-                .writeTimeout(BaseRequest.DEFAULT_REQUEST_TIMEOUT, TimeUnit.MILLISECONDS)
-                .build();
+        mOkHttpClient = MediaOkHttpClient.getOkHttpClientSharedInstance(okClientBuilder);
```
maxme: Injection is bypassed here since we just call the static method and thus we don't get a OkHttpClient singleton. We can also drop `OkHttpClient.Builder` since it won't be used anymore.

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
```diff
@@ -65,11 +65,7 @@ public MediaXMLRPCClient(Dispatcher dispatcher, RequestQueue requestQueue, OkHtt
                              AccessToken accessToken, UserAgent userAgent,
                              HTTPAuthManager httpAuthManager) {
         super(dispatcher, requestQueue, accessToken, userAgent, httpAuthManager);
-        mOkHttpClient = okClientBuilder
-                .connectTimeout(BaseRequest.DEFAULT_REQUEST_TIMEOUT, TimeUnit.MILLISECONDS)
-                .readTimeout(BaseRequest.UPLOAD_REQUEST_READ_TIMEOUT, TimeUnit.MILLISECONDS)
-                .writeTimeout(BaseRequest.DEFAULT_REQUEST_TIMEOUT, TimeUnit.MILLISECONDS)
-                .build();
+        mOkHttpClient = MediaOkHttpClient.getOkHttpClientSharedInstance(okClientBuilder);
```
maxme: Same thing here.

maxme:
Checkstyle is still failing, can you rebase and squash everything in one commit here, it could be handy for reverting that patch later.

mzorz:
Here's the squashed commit https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/400/commits/efc00b2f8f6ff4398701b6d0542db1e65abf52f9

mzorz:
Closing in favor of #410

PR #401, opened 2017-03-29
Feature/multi uploads xmlrpc

mzorz:
Provides multiupload-multicancel capabilties and tests on FluxC for XMLRPC.

cc @aforcier

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestXMLRPC.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestXMLRPC.java
@@ -202,7 +202,7 @@ public class ReleaseStack_MediaTestXMLRPC extends ReleaseStack_XMLRPCBase {
     public void testUploadMultipleImages() throws InterruptedException {
         // upload media to guarantee media exists
         mUploadedIds = new ArrayList<>();
-        mNextEvent = ReleaseStack_MediaTestXMLRPC.TestEvents.UPLOADED_MUTIPLE_MEDIA;
+        mNextEvent = TestEvents.UPLOADED_MUTIPLE_MEDIA;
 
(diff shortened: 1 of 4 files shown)


aforcier:
Interesting, the cancellation test fails for me even though it has the [same timeout](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/401/files#diff-89d0e5351b721034c8f85ffb620a4471R441) as the WP.com tests.

It works if I step the wait time down to `100ms`, but I think I'll just use a slightly larger test image in the tests (`9KB` is maybe excessively small 😂). No need to change it.

mzorz:
interesting! For me I had to boost the timeout to 60 seconds just for the single image upload and a ~500kb image I use for testing
These tests might benefit from having a dataset attached to them

aforcier:
> interesting! For me I had to boost the timeout to 60 seconds just for the single image upload and a ~500kb image I use for testing
These tests might benefit from having a dataset attached to them

A set different media for each test to run through 🤔 that might be really neat to have.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestXMLRPC.java
```diff
@@ -42,6 +44,9 @@
         FETCHED_MEDIA,
         DELETED_MEDIA,
         UPLOADED_MEDIA,
+        UPLOADED_MUTIPLE_MEDIA, // these don't exist in FluxC, but are an artifact to wait for all
```
aforcier: `MULTIPLE` is missing an `L` in several places.

mzorz: fixed in a8443cb

aforcier:
:shipit:

PR #405, opened 2017-04-07
Fix #404 and Fix #403: XML encode strings and add back Content-Length header

### LINKED ISSUE #403: Some selfhosted user get a Length Required Error
When they try to upload a media to their self hosted site. We can fix it by replacing the `Transfer-Encoding: chunked` header by `Content-Length: xxx`.

Overriding the `public long contentLength()` method in `XmlrpcUploadRequestBody`should be enough to fix it.

### LINKED ISSUE #404: XMLencode all the things in the media upload request

maxme:
Fix #404 and Fix #403: XML encode strings and add back Content-Length header.

This also fix the `getProgress()` method (was returning invalid number since bytes counted were different from bytes wrote on the socket).

cc @daniloercoli


daniloercoli:
Looks good! :shipit:

PR #406, opened 2017-04-08
Track origin of the site and use XMLRPC if the site is jetpack but was added by XMLRPC 

maxme:
Track origin of the site and use XMLRPC if the site is jetpack but was added by XMLRPC.

This allows us to know if a certain site is Jetpack connected but added via XMLRPC, in that case the user might not want to connect to wpcom and continue via XMLRPC.

If the user later connects to his wpcom account, the same site will be fetched by a WPCOM REST and thus will override previous origin.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
@@ -596,4 +596,33 @@ public class SiteStoreUnitTest {
         }
         assertTrue(duplicate);
     }
+
+    @Test
+    public void testJetpackSelfHostedAndForceXMLRPC() {
+        SiteModel jetpackSite = generateJetpackSiteOverXMLRPC();
+        jetpackSite.setOrigin(SiteModel.ORIGIN_WPCOM_REST);
+        assertTrue(jetpackSite.isUsingWpComRestApi());
+
+        // Force the origin, it should now use XMLRPC instead of REST.
+        jetpackSite.setOrigin(SiteModel.ORIGIN_XMLRPC);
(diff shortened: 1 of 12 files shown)


example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
```diff
@@ -596,4 +596,33 @@ public void testUpdateSiteUniqueConstraintFail() throws DuplicateSiteException {
...
+    @Test
+    public void testDefaultUsageWpComRestApi() {
+        SiteModel wpComSite = generateWPComSite();
+        assertTrue(wpComSite.isUsingWpComRestApi());
+
+        SiteModel jetpack1 = generateJetpackSiteOverRestOnly();
+        assertTrue(jetpack1.isUsingWpComRestApi());
+
+        SiteModel jetpack2 = generateJetpackSiteOverXMLRPC();
+        assertTrue(jetpack1.isUsingWpComRestApi());
```
mzorz: I think you meant to reference `jetpack2` instead of `jetpack1` in this line?

mzorz: Also probably meant to negate the assertion condition?  as in `!jetpack2.isUsingWpComRestApi()`

mzorz: Sorry, better use `assertFalse` than negating  😁

maxme: Copy/Paste 🤕 - yep this must be an `assertFalse` here.

mzorz:
LGMT :shipit:

PR #413, opened 2017-04-14
SiteStore: New domain suggestions action

maxme:
Note: I had to change the GsonRequest a bit in 5dbea5f to allow "generic element" (aka json array) as root response. it's not a huge change since it's only affects the new request used here in this PR - cAT tests are fine.

I didn't want to add connected tests because it's not a fundamental endpoint. No unit tests either because it's a pure network round trip.

To test: open the example app, go to "Signed Out Actions", click "Domain suggestions", have fun.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ allprojects {
     }
 
     checkstyle {
-        toolVersion = '7.3'
+        toolVersion = '7.6.1'
         configFile file("${project.rootDir}/config/checkstyle.xml")
     }
 }
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/DomainSuggestionResponse.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/DomainSuggestionResponse.java
@@ -1,0 +1,13 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.site;
+
+import org.wordpress.android.fluxc.network.Response;
+
+public class DomainSuggestionResponse implements Response {
+    public String cost;
+    public String domain_name;
+    public boolean is_free;
+
+    public int product_id;
+    public String product_slug;
+    public float relevance;
+}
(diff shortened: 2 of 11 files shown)


build.gradle
```diff
@@ -14,7 +14,7 @@ allprojects {
     }
 
     checkstyle {
-        toolVersion = '7.3'
+        toolVersion = '7.6.1'
```
aforcier: Any particular reason for the bump or just routine?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/DomainSuggestionResponse.java
```diff
@@ -0,0 +1,13 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.site;
+
+import org.wordpress.android.fluxc.network.Response;
+
+public class DomainSuggestionResponse implements Response {
+    public String cost;
+    public String domain_name;
+    public boolean is_free;
+
+    public int product_id;
```
aforcier: I wonder if this might at some point be a String and give us a `NumberFormatException` surprise. That was a different kind of `product_id`, but this happened for VIP plans (https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/339).

aforcier:
Aside from my questions/nit, this looks good to me @maxme 👍 Looks like you and `Gson` got along well 🙃

build.gradle
maxme: Routine, I was trying to change the checkstyle config to avoid this ugly trick: `"include_wordpressdot" + "com"` (otherwise it matches `dotcom`) and finally only kept the checkstyle version bump - I doubt this needs its own PR ;)

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/DomainSuggestionResponse.java
maxme: Ah yes, I'll implement the same workaround :crying_panda_face_with_sad_eyes:

maxme: Done in 5a658c5 (actually we're passing the response directly in the OnChanged event, so we don't need any string2long conversion - plus it won't be used in wpandroid).

aforcier:
:shipit:

PR #414, opened 2017-04-17
Add rule enforcing space after inline annotations

aforcier:
Generates a checkstyle error when there's no whitespace after an in-line annotation followed by text.

For example,

```java
public void exampleMethod(@Named("something")Thingy thingy) {
```

will be flagged, as the correct form is:

```java
public void exampleMethod(@Named("something") Thingy thingy) {
```

cc @maxme

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -51,6 +51,12 @@
       <property name="severity" value="error"/>
     </module>
 
+    <module name="RegexpSingleline">
+        <property name="format" value="@[\S]*\(\&quot;.*\&quot;\)[^ \t]"/>
+        <property name="message" value="In-line annotations must be followed by one whitespace"/>
+        <property name="severity" value="error"/>
+    </module>
+
     <!-- Specific to FluxC / WordPress -->
     <module name="RegexpSingleline">
       <property name="format" value="dotcom"/>
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -52,7 +52,7 @@
     </module>
 
     <module name="RegexpSingleline">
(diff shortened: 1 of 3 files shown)


config/checkstyle.xml
```diff
@@ -51,6 +51,12 @@
       <property name="severity" value="error"/>
     </module>
 
+    <module name="RegexpSingleline">
+        <property name="format" value="@[\S]*\(\&quot;.*\&quot;\)[^ \t]"/>
```
maxme: I'm not sure about the `"`, We don't necessary have a quoted string in an annotation.

Like in our Action classes: `@Action(payloadType = String.class)ACTION_X` would be valid.

Note: I wish we could use the AST instead of a regexp, but I doubt we can do that without a custom rule and I really don't spend more time on this.

aforcier: > I'm not sure about the `"`, We don't necessary have a quoted string in an annotation.

Excellent point, I'll generalize it.

> Note: I wish we could use the AST instead of a regexp, but I doubt we can do that without a custom rule and I really don't spend more time on this.

Agreed. That's the rabbit hole I started going down before I decide I could say what I wanted in regexp and save myself the time.

aforcier:
@maxme ready for another pass.

I merged `develop` in and fixed the annotation checkstyle issues there. I also noticed that the regex was generating some false positives and had to be made a bit more precise (see aaebdf9).

...It's almost getting to the point where we need unit tests for our checkstyle.

maxme:
Looks good, I thought the regexp would be simpler but nope: `@RawConstraints({"UNIQUE (ROOT_URL)"})`.

PR #417, opened 2017-04-20
Do not report "broken pipe" errors to end users

daniloercoli:
device-2017-04-19-203218

This PR does check the IOException error message, and hides `broken pipe` detailed message to end users. The error is logged in the app log, and wp-android does use the dafault error message in this case. 

We can provide a more useful error message, `An error happened, please contact the support if persist`, if we think it's a better solution.


daniloercoli:
Will open a new PR with other errors catched.

PR #420, opened 2017-04-20
Timeout, Broken Pipe & Internal server errors catched and reported

daniloercoli:
This PR does check the IOException, and hides `broken pipe` detailed message to end-users. The error is logged in the app log, and wp-android does use the default error message in this case. 

device-2017-04-19-203218

This PR also catches `internal server error` and `timeout` exception and report them to the final app.
- `internal server error` could happen when the PHP  `memory_limit` or `max_execution_time` is reached.
- `timeout` happens often on slow connections. I used Charles proxy and throttled the connection to be a 28k modem.


maxme:
:shipit 
Note: I merged `develop` back and fixed minor conflicts

Issue #423, opened 2017-04-24
Drop wp.getMediaFile call after media upload when possible


aforcier:
It looks like the `wp.uploadFile` response contains all the fields we need as of [WordPress 4.4](https://core.trac.wordpress.org/changeset/34579) - so we can drop the `wp.getMediaItem` call after XML-RPC media uploads complete unless required fields are missing from `wp.uploadFile`.

aforcier:
Done in #424.

PR #424, opened 2017-04-24
Use a single call for XML-RPC media uploads (when able)

aforcier:
Addresses #423. The response of `wp.uploadFile` is checked, and if it contains the required fields (should always be present on WordPress 4.4 and later) it updates the `MediaModel` using those and doesn't call `wp.getMediaItem`.

The client falls back to the old behavior (`wp.uploadFile` + `wp.getMediaItem`) if any fields are missing from the upload response.

Also added a new test site for a WordPress 4.3 site and some tests so we can catch any future regressions.

For reference, this is the core change that modified `wp.uploadFile` to respond similarly to `wp.getMediaItem`: https://core.trac.wordpress.org/changeset/34579

To test:
* Run the old and new upload tests with network sniffing, and verify that modern enough WordPress sites complete the upload with a single call.
* Always return `false` from [`isDeprecatedUploadResponse()`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/compare/feature/423-single-call-xmlrpc-uploads?expand=1#diff-a59c679264e9e17f7dca0a78844f29d2R519) and watch the WordPress 4.3 tests fail


fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
```diff
@@ -57,6 +57,9 @@
 import okhttp3.Request.Builder;
 
 public class MediaXMLRPCClient extends BaseXMLRPCClient implements ProgressListener {
+    private static final String[] REQUIRED_UPLOAD_RESPONSE_FIELDS = {
+            "attachment_id", "parent", "title", "caption", "description", "thumbnail", "date_created_gmt", "link"};
```
maxme: Note: I think we should keep that list as it, simply because I doubt we'll see a lot of bastard WP versions that reply with a non standard response. 

But I guess it's safe to drop "title", "caption", "description" from the required fields.

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
```diff
@@ -167,15 +170,25 @@ public void onResponse(Call call, okhttp3.Response response) throws IOException
...
+                                // `parent, title, caption, description, videopress_shortcode, thumbnail,
+                                // date_created_gmt, link, width, height` fields, so we need to make a fetch for them
+                                // This only applies to WordPress sites running versions older than WordPress 4.4
+                                fetchMedia(site, media, true);
+                            } else {
+                                MediaModel responseMedia = getMediaFromXmlrpcResponse(responseMap);
+                                // Retain local IDs
+                                responseMedia.setId(media.getId());
+                                responseMedia.setLocalSiteId(site.getId());
+                                // TODO: setLocalPostId()
```
maxme: Not sure what is this TODO. What does a local post id mean for a media (can't a media be attached to multiple posts?). Can you add a descriptive comment?

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
```diff
@@ -167,15 +170,25 @@ public void onResponse(Call call, okhttp3.Response response) throws IOException
...
+
+                                notifyMediaUploaded(responseMedia, null);
```
maxme: I think there is a missing `removeCallFromCurrentUploadsMap` here (it was called via `fetchMedia` previously).

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
maxme: BTW, not related to this PR, but I think there is another call missing here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/4099857706c20b8ca92d0721d1ebf253d0eb43bc/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java#L283-L283 cc @mzorz

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
```diff
@@ -167,15 +170,25 @@ public void onResponse(Call call, okhttp3.Response response) throws IOException
                     // HTTP_OK code doesn't mean the upload is successful, XML-RPC API returns code 200 with an
                     // xml field "faultCode" on error.
                     try {
-                        MediaModel responseMedia = getMediaFromUploadResponse(response);
-                        if (responseMedia != null) {
+                        Map responseMap = getMapFromUploadResponse(response);
+                        if (responseMap != null) {
```
maxme: If `responseMap == null`, we don't call `notifyMediaUploaded` (so we won't get an onChangedEvent), we should call `notifyMediaUploaded` with an error and we should also call `removeCallFromCurrentUploadsMap` (I wonder if we should move this to `notifyMediaUploaded` to make sure we don't forget it).

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
aforcier: That was a note for myself for another PR I'm working on, adding a `localPostId()` (to track media attached to local posts after they finish uploading).

It doesn't make much sense as a TODO, I agree. I'll just remove it and keep a note elsewhere.

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
aforcier: You're right - added in de0f97e.

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
```diff
@@ -486,20 +500,29 @@ private MediaError getMediaErrorFromXMLRPCException(XMLRPCException exception) {
...
+    private static Map getMapFromUploadResponse(okhttp3.Response response) throws XMLRPCException {
+        Map responseMap = new HashMap();
         try {
             String data = new String(response.body().bytes(), "UTF-8");
             InputStream is = new ByteArrayInputStream(data.getBytes(Charset.forName("UTF-8")));
-            Object obj = XMLSerializerUtils.deserialize(XMLSerializerUtils.scrubXmlResponse(is));
-            if (obj instanceof Map) {
-                media.setMediaId(MapUtils.getMapLong((Map) obj, "id"));
+            Object responseObject = XMLSerializerUtils.deserialize(XMLSerializerUtils.scrubXmlResponse(is));
+            if (responseObject instanceof Map) {
```
aforcier: Agreed - 5c3baa9.

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
aforcier: > we should call `notifyMediaUploaded` with an error and we should also call `removeCallFromCurrentUploadsMap`

Done in 83e0939.

> (I wonder if we should move this to `notifyMediaUploaded` to make sure we don't forget it)

My thoughts exactly - noticed a few other pre-existing cases where we aren't calling `removeCallFromCurrentUploadsMap()`. Moved in b592064.

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
aforcier: > BTW, not related to this PR, but I think there is another call missing here

I believe you're right, that should check `isFreshUpload` and call `notifyMediaUploaded` instead if `true`.

aforcier: Same [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/4099857706c20b8ca92d0721d1ebf253d0eb43bc/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java#L308).

aforcier: Done in 0b3bce1.

aforcier:
Should be ready for a second pass @maxme 🕺

maxme:
:shipit:

PR #426, opened 2017-04-25
Retain local post ID after media upload

### LINKED ISSUE #422: Differentiate local from remote post IDs in MediaModel
In WPAndroid, we're currently using `MediaModel.mPostId` for both local and remote post IDs. Media about to be uploaded from the editor are given the local post ID of the post they were added to. Remote post IDs.

This is causing a few problems:
1. Media inserted in a post on a WP.com REST API site are uploaded setting the `parent` post ID as the `local` ID of the post, which means they're being attached to unrelated posts (except during rare alignments of the heavens when `local id == remote id`)
(we don't send this param over XML-RPC)
2. The post ID is reset to the one in the server response once the upload completes. This 'works' for WP.com REST sites, since that's still the local post ID, but for self-hosted sites this is `0`

2 currently doesn't seem to be breaking anything. After a media upload completes, we generally already know which post ID we care about (because the editor is open to that post).

Uploading media attached to a post asynchronously presents a problem though, since we need to be able to identify which (local only, in many cases) post a completed media belonged to so we can update that post's contents. Currently, looking up the `postId` of completed media over XML-RPC will always fail.

### Solution

We have a few options here:
1. Add a second, `mLocalPostId`, and only use `mPostId` for the _remote_ post ID (if any) of the media
2. Stop using `mPostId` to ever refer to the remote post id - we don't currently use this anywhere

2 is simplest, but it's conceivable we'll want to know the remote post media is attached to (maybe an expansion of the media browser will allow changing this value). I think we may as well do 1 and be prepared.

aforcier:
Fixes #422.

* Adds a `LOCAL_POST_ID` field to `MediaModel`, leaving `POST_ID` to refer only to the remote post ID (`parent`) of the media item
* Preserves the local post ID after upload, so media can be matched to the (possibly local-only) post they were uploaded in
* Modifies the XML-RPC media upload call to set the remote post ID (`POST_ID`) when set

~~#424 should be merged first, and the base branch updated to `develop`.~~

A small WPAndroid-side PR is needed to fix a usage of `setPostId()` that should be `setLocalPostId()` with this changeset. (see it at https://github.com/wordpress-mobile/WordPress-Android/pull/5745)

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestWPCom.java
@@ -179,6 +179,37 @@ public class ReleaseStack_MediaTestWPCom extends ReleaseStack_WPComBase {
         deleteMedia(testMedia);
     }
 
+    public void testUploadImageAttachedToPost() throws InterruptedException {
+        // Upload media attached to remotely saved post
+        MediaModel testMedia = newMediaModel(BuildConfig.TEST_LOCAL_IMAGE, MediaUtils.MIME_TYPE_IMAGE);
+        testMedia.setLocalPostId(5);
+        testMedia.setPostId(1);
+        mNextEvent = TestEvents.UPLOADED_MEDIA;
+        uploadMedia(testMedia);
+
+        testMedia.setMediaId(mLastUploadedId);
+        MediaModel uploadedMedia = mMediaStore.getSiteMediaWithId(sSite, testMedia.getMediaId());
+        assertNotNull(uploadedMedia);
+        assertEquals(1, uploadedMedia.getPostId());
+
+        mNextEvent = TestEvents.DELETED_MEDIA;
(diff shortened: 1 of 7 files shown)


aforcier:
Opened https://github.com/wordpress-mobile/WordPress-Android/pull/5745 to apply these changes to WPAndroid.

nbradbury:
@aforcier This all looks good to me and I'd be happy to :shipit:, but since this is my first real FluxC review maybe a more experienced fluxxer should also look at it?

aforcier:
> This all looks good to me and I'd be happy to :shipit:, but since this is my first real FluxC review maybe a more experienced fluxxer should also look at it?

Thanks @nbradbury! Sure, let's leave this open, I'll ask someone to take a look tomorrow if no one volunteers 😈

maxme:
Should we update models local post id in `fetchMedia`, `fetchMediaList` and `pushMedia`? 

* Create a "local" media, set it's `localPostId`
* Upload it (and update `localPostId`).
* Fetch it again, localPostId is reset (via `getMediaFromRestResponse`).

PR #429, opened 2017-04-28
Media client lint fixes

aforcier:
Just cleaning up a few linter issues in the media clients - mostly using specific imports, and making a few `null` usages explicit.


maxme:
I wonder if we can enable a rule in checkstyle for this.

aforcier:
> I wonder if we can enable a rule in checkstyle for this.
If not, we could make Travis treat all lint warnings as errors :trollface:

aforcier:
> I wonder if we can enable a rule in checkstyle for this.

Following up on this - as we discussed there isn't a rule for this, and there's also no straightforward way of getting Travis to run Android Studio's inspections. Abandoning standardizing this particular inspection.

So, uh, ready for review @maxme 😉

maxme:
:shipit:

PR #430, opened 2017-04-28
Site store getters renaming

maxme:
`getSelfHostedSite` gradually became another kind of method to get sites that are accessed via XMLRPC, These sites could be Jetpack sites (installed or connected) or pure self hosted sites depending how the user addded the site in the app.

To make these getters less confusing, we renamed these them to getSitesAccessedViaXMLRPC / getSitesAccessedViaWPComRest - they now reflects the data model more accurately.

I also removed `getJetpackSites` and his siblings, it's not used in the app and it's misleading: what's a Jetpack site? a self hosted with Jetpack installed? or Connected? What about Jetpack sites accessed via XMLRPC?

Note: I kept `getLocalIdForSelfHostedSiteIdAndXmlRpcUrl` because it's used in a very specific case (migration) in wpandroid, but once we drop migration code, we can drop that getter too.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
@@ -401,22 +401,23 @@ public class SiteStore extends Store {
     }
 
     /**
-     * Returns all .COM and Jetpack sites in the store.
+     * Returns sites accessed via WPCom REST API (WPCom sites or Jetpack sites connected via WPCom REST API).
      */
     public List<SiteModel> getSitesAccessedViaWPComRest() {
         return SiteSqlUtils.getSitesAccessedViaWPComRest().getAsModel();
     }
 
     /**
-     * Returns the number of .COM and Jetpack sites in the store.
+     * Returns the number of sites accessed via WPCom REST API (WPCom sites or Jetpack sites connected
+     * via WPCom REST API).
      */
     public int getSitesAccessedViaWPComRestCount() {
         return SiteSqlUtils.getSitesAccessedViaWPComRest().getAsCursor().getCount();
     }
 
     /**
-     * Checks whether the store contains at least one site accessed via XMLRPC (self-hosted sites or
-     * Jetpack sites accessed via XMLRPC).
+     * Checks whether the store contains at least one site accessed via WPCom REST API (WPCom sites or Jetpack
+     * sites connected via WPCom REST API).
      */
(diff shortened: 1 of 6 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -403,15 +403,23 @@ public boolean hasSiteWithLocalId(int id) {
     /**
      * Returns all .COM and Jetpack sites in the store.
```
aforcier: We should probably update these comments to be more specific about 'Jetpack'.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -403,15 +403,23 @@ public boolean hasSiteWithLocalId(int id) {
...
      * Returns the number of .COM and Jetpack sites in the store.
      */
-    public int getWPComAndJetpackSitesCount() {
-        return SiteSqlUtils.getWPComAndJetpackSites().getAsCursor().getCount();
+    public int getSitesAccessedViaWPComRestCount() {
+        return SiteSqlUtils.getSitesAccessedViaWPComRest().getAsCursor().getCount();
+    }
+
+    /**
+     * Checks whether the store contains at least one site accessed via XMLRPC (self-hosted sites or
```
aforcier: Copy/paste? 😛

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -447,60 +456,23 @@ public boolean hasWPComSite() {
     /**
      * Returns all self-hosted sites (can't be Jetpack) in the store.
```
aforcier: Here too, 'can't be Jetpack' is unclear now.

aforcier:
What about [this call](https://github.com/wordpress-mobile/WordPress-Android/blob/ea576cc07c9a039984a3aa3cc9f21b863b824600/WordPress/src/main/java/org/wordpress/android/util/AnalyticsUtils.java#L62) to `hasJetpackSite()`?

maxme:
> What about this call to hasJetpackSite()?

We'll replace that one by `getSitesAccessedViaWPComRestCount() - getWPComSitesCount() > 0` and comment it in `AnalyticsUtils`.

maxme:
Updated getter comments, ready for another pass.

aforcier:
:shipit:!

PR #432, opened 2017-05-02
Some lint pickup

aforcier:
Fixes all 'unnecessary fully qualified imports' lint warnings (when paired with #429), drops some redundant null checking, and other small tweaks.


daniloercoli:
👍 :shipit:

PR #436, opened 2017-05-04
Fix #435: Fix page upload and fetch in the XMLRPC client

### LINKED ISSUE #435: Pages are not handled correctly in the Post XMLRPC client.
There are 3 issues:

* We don't set the right post_type when we publish a page.
* We set the wrong post_type when we fetch pages (we force the `setPage(true)` setter).
* There is a test for that case, but it's not fetching the newly uploaded page, so everything seems fine.

maxme:


aforcier:
:shipit:


(_fast, before anyone else notices_)

PR #437, opened 2017-05-04
catching IllegalArgumentException and reconverting to DiscoveryException

mzorz:
... which is more meaningful in this context

cc @tonyr59h


tonyr59h:
:shipit:

PR #438, opened 2017-05-05
WordPress.com fetch + signout collision fix

### LINKED ISSUE #433: Fetch + sign out collision
It's possible to dispatch a `FETCH_SITES` action, immediately followed by a `REMOVE_WPCOM_AND_JETPACK_SITES`. This can result in a successful server request, with the response processed _after_ the `REMOVE_WPCOM_AND_JETPACK_SITES` action was handled. This ends up adding the removed sites right back into the database - but there is no `AccountModel` or access token. From the WordPress app, this looks like the user wasn't actually logged out, as the sites remain in the site list.

There are many possible solutions, most of them pretty involved.

1. Track a 'state' at the dispatcher level, and update/check the state when processing sensitive actions (like fetches, `AUTHENTICATE`, `SIGN_OUT`, `REMOVE_WPCOM_AND_JETPACK_SITES`)
2. Use some kind of middleware to throttle/validate/invalidate actions (e.g. check that we haven't just received a `REMOVE_WPCOM_AND_JETPACK_SITES` before handling `UPDATE_SITES`)
3. Inject the `AccountStore` in the `SiteStore`, and use it to verify that a token exists before inserting WordPress.com sites
4. Use `FOREIGN KEY` constraints in the `SiteModel` table, to map a `SiteModel` field to an `AccountModel` entry
5. Enforce (4) at the insertion level (in `SiteSqlUtils`) instead of using `FOREIGN KEY`

1 and 2 are quite complex. 1 is probably the 'ideal' solution, as we can also defend against other similar collisions (`FETCH_POSTS` and then `DELETE_POST`), but it poses some concurrency problems. For example, it's possible to receive an `UPDATE_SITES` action, validate it against the state in the store, and *then* receive and process a `REMOVE_WPCOM_AND_JETPACK_SITES`. Because we're using `ThreadMode.ASYNC`, those actions can run concurrently, and we have no guarantee that the database calls of one will complete before the other (even though the database is locked, we can't guaranteed which will reach the lock first).

3 is unpalatable, as we've been able to keep our stores nicely separate so far and we shouldn't break that unless we have no choice.

4 would work nicely, but requires some significant work. We could rely on the existing `SiteModelTable.SITE_ID`, but Jetpack sites added as self-hosted also have this field, and they would fail the constraint (as there's no WP.com account in FluxC to link to). There isn't a way to conditionally apply `FOREIGN KEY` constraints. There's a better way to do this that's involved, which I'll open a separate issue for.

5 is the way to go for now. Straightforward enough to implement, and will address the problem, if a little ham-fistedly.

Extra note: It would be better to check for an auth token instead of checking for an `AccountModel`, which is what 4 and 5 do. This same bug can also cause the `AccoutStore` to have an account while having no token (`FETCH_ACCOUNT` and then `SIGN_OUT` immediately). Fixing this account bug is straightforward enough though, and with that we should be able to rely on `no token == no account`.

aforcier:
Fixes #433 (though I'll be opening another issue or two to track longer-term refactors mentioned in that issue).

~~Marked as 'not ready for review' until I have a companion PR with 'nextable' actions up (see below).~~ Companion PR: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/439

The gist here is that before WordPress.com sites are added/updated in the database, we now check whether there is a 'valid' WordPress.com account present. 'Valid' just means a non-zero `USER_ID`. (We'll need to change this behavior if/when we add support for multiple accounts.)

A similar problem can happen if we `FETCH_ACCOUNT` and then logout - we can add an `AccountModel` after it was removed, while we have no token. This PR also prevents accounts from being added if there's no auth token.

A side effect of all this fix is that we can't call `FETCH_ACCOUNT` and `FETCH_SITES` concurrently on login anymore - if `FETCH_SITES` completes first, the insertion of sites in the database will fail because there's no account yet. This ends up making for even more fragmented login code as we have to trigger the next action in the `OnChanged` method the last one called. A second PR adding 'nextable' actions (https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/439) will allow us to run all the login actions in one place, in sequence.

I'm targeting a feature branch on this one because I'd like https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/439 to also be reviewed/merged before this hits `develop` and we have to update the WordPress app.

To test:
1. Check out 0d97656 and run the new site and account connected tests --> 💥
2. Check out 46fcaa6, run account test --> 🎉
3. Check out the branch, run all tests --> 🎉
4. Test signing in to the demo apps

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
@@ -36,13 +36,13 @@ import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
-import static org.wordpress.android.fluxc.utils.SiteUtils.generateJetpackSiteOverRestOnly;
-import static org.wordpress.android.fluxc.utils.SiteUtils.generateJetpackSiteOverXMLRPC;
-import static org.wordpress.android.fluxc.utils.SiteUtils.generatePostFormats;
-import static org.wordpress.android.fluxc.utils.SiteUtils.generateSelfHostedNonJPSite;
-import static org.wordpress.android.fluxc.utils.SiteUtils.generateSelfHostedSiteFutureJetpack;
-import static org.wordpress.android.fluxc.utils.SiteUtils.generateTestSite;
-import static org.wordpress.android.fluxc.utils.SiteUtils.generateWPComSite;
+import static org.wordpress.android.fluxc.site.SiteUtils.generateJetpackSiteOverRestOnly;
+import static org.wordpress.android.fluxc.site.SiteUtils.generateJetpackSiteOverXMLRPC;
+import static org.wordpress.android.fluxc.site.SiteUtils.generatePostFormats;
+import static org.wordpress.android.fluxc.site.SiteUtils.generateSelfHostedNonJPSite;
+import static org.wordpress.android.fluxc.site.SiteUtils.generateSelfHostedSiteFutureJetpack;
+import static org.wordpress.android.fluxc.site.SiteUtils.generateTestSite;
+import static org.wordpress.android.fluxc.site.SiteUtils.generateWPComSite;
 
 @RunWith(RobolectricTestRunner.class)
 public class SiteStoreUnitTest {
--- a/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
@@ -13,6 +13,7 @@ import org.mockito.Mockito;
 import org.robolectric.RobolectricTestRunner;
 import org.robolectric.RuntimeEnvironment;
 import org.wordpress.android.fluxc.Dispatcher;
+import org.wordpress.android.fluxc.WellSqlTestUtils;
 import org.wordpress.android.fluxc.model.PostFormatModel;
 import org.wordpress.android.fluxc.model.SiteModel;
 import org.wordpress.android.fluxc.model.SitesModel;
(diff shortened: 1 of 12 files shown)


example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
```diff
@@ -70,6 +71,8 @@ public void testSimpleInsertionAndRetrieval() {
 
     @Test
     public void testInsertOrUpdateSite() throws DuplicateSiteException {
+        WellSqlTestUtils.setupWordPressComAccount();
```
maxme: Can't we just add this call to the setUp method?

```diff
     public class SiteStoreUnitTest {
         WellSqlConfig config = new WellSqlConfig(appContext);
         WellSql.init(config);
         config.reset();
+        WellSqlTestUtils.setupWordPressComAccount();
     }
```

maxme:
Looks good - just a minor comment in the test suite.

I hope this is the last addition to `insertOrUpdateSite`.

example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
aforcier: We could - I left it out because some of those tests don't require a .com account, and could reveal bugs by failing due to missing WP.com account.

I could add an explicit test for that instead (and start that test by deleting the account added in `setUp()`). What do you think is better?

aforcier:
> I hope this is the last addition to `insertOrUpdateSite`.

Cross my heart.

example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
maxme: No that's fine.

maxme:
:shipit:

PR #441, opened 2017-05-08
Generics cleanup

aforcier:
Refactors some of our generic code, reducing our compiler warnings from `13` to `1`. The final warning is address in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/442.

No major issues fixed - I mostly learned a few things, fixed a few potential traps, and added some documentation.

I had to use `SuppressWarnings` in [one spot](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/441/files#diff-3c2b3bb0a2654c0ed1fab89a38c2dfe5R28), but there's an explanation of why it's safe to do so in that case.

PR #442, opened 2017-05-08
Fix compiler warning in XMLRPCRequest

aforcier:
Fixes our last compiler warning (the rest are in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/441).

I tried out a few approaches for this one, and settled on this. It forces the XML-RPC clients to set a parameter for the `Listener`, and restricts that to only `Object` or `Object[]`. A perfect solution would've been to refactor the `XMLRPCSerializer` to be parameterizable, but I'm unwilling to mess with code we've had for so long just for this.


maxme:
:shipit:

PR #444, opened 2017-05-10
Remove useless DataInputStream and close the file in case of an IOException

maxme:


daniloercoli:
Code looks good. Tested and everything works fine.

PR #447, opened 2017-05-12
Drop neenbedankt apt plugin, use the stock annotationProcessor instead

maxme:
Drop neenbedankt apt plugin, use the stock annotationProcessor instead.

First step to kapt?

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -4,7 +4,6 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.3.0'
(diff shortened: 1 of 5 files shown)


aforcier:
Works great! Out of paranoia I did a clean project clone, switched to this branch, and built + ran tests. All working nicely.

Sidenote: Looks like [this was fixed just after](http://www.littlerobots.nl/blog/Whats-next-for-android-apt/) we attempted to partially Kotlinify and ran into trouble with annotation processing.

:shipit:

PR #373, opened 2017-03-17
Add email_verified field to Account fetch response

tonyr59h:
See wordpress-mobile/WordPress-Android#4600 for details.

This PR updates the Account response handling to store the `email_verified` field, required for referenced issue in WP-Android.

cc @maxme

PR #448, opened 2017-05-12
Parse and store the "default sizes" for each media item

daniloercoli:
On media uploading, WordPress (the server) creates few predefined image files, together with the original file, and gives them available for future usage. It depends on the server configuration, but in vast majority of them we have the following sizes: `thumbnail`, `medium`, `large`, `medium_large`.
Those files are available via XML-RPC, see [wp.getMediaLibrary](https://codex.wordpress.org/XML-RPC_WordPress_API/Media#wp.getMediaItem), so we should parse and store them, together with other mediaItem info.

This new info will be used in wp-android to improve the loading time of WP Media Library. There is no need to load the full pictures, instead `large` (when available), or `medium` should be OK.

~~This is not ready to be reviewed/merged, I just want an heads up to check if I'm doing it right in regards adding new fields to a model (DB?).~~

cc @aforcier @mzorz

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
@@ -61,10 +61,10 @@ public class MediaModel extends Payload implements Identifiable, Serializable {
     // Local only
     @Column private String mUploadState;
 
-    // Other Sizes. Image files only
-    @Column private String mFileNameMediumSize;
-    @Column private String mFileNameMediumLargeSize;
-    @Column private String mFileNameLargeSize;
+    // Other Sizes. Only available for images on self-hosted (xmlrpc layer) sites
+    @Column private String mFileUrlMediumSize;
+    @Column private String mFileUrlMediumLargeSize;
+    @Column private String mFileUrlLargeSize;
 
     //
     // Legacy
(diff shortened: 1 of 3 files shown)


aforcier:
@daniloercoli You'll need to handle a database upgrade for this one - here's an example PR for a db upgrade: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/373

fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
```diff
@@ -61,6 +61,11 @@
     // Local only
     @Column private String mUploadState;
 
+    // Other Sizes. Image files only
```
aforcier: I think this comment could use a bit more detail, to clarify that:
1. These are only for media over XML-RPC
2. They're file names (not URLs) on the server

@nbradbury what do you think about storing these as a full URL instead of a file name? Essentially, doing the work of building the alternate site image URL from the thumbnail URL in the `MediaXMLRPCClient` so we don't have to do it in WPAndroid?

daniloercoli: Note that wpcom does already return the full URL to pictures. (I know we're not using this info for wpcom).

nbradbury: > @nbradbury what do you think about storing these as a full URL instead of a file name?

Makes sense to me. @daniloercoli Do you want to take this over?

daniloercoli:
Ready for another pass.

aforcier:
Looks good! :shipit:

PR #453, opened 2017-05-26
Match by local post ID in getPostMediaWithPath()

aforcier:
Our actual use case for `MediaStore.getPostMediaWithPath()` is to retrieve local media that have been attached to a post. We were doing this lookup by the remote post ID, but we should actually use the local ID of the post (this used to work because local and remote post IDs were conflated in the `MediaModel`, until https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/426).


daniloercoli:
Looks good. Tests passed without issues. :shipit:

Issue #402, opened 2017-04-03
Jetpack - Image size larger than site's max_upload_filesize error message workaround


daniloercoli:
When you try to upload an image through the wp.com API to a Jetpack site and it's too large (larger than the site's `max_upload_filesize`), it will fail with an error that doesn't seem relevant:

`The corresponding HMAC for this file does not match`

<img src="https://cloud.githubusercontent.com/assets/518232/24604891/f6682316-1866-11e7-81e3-1d8a8049fe92.png" width="270" alt="screenshot_20170324-162026">

Ref: https://github.com/Automattic/jetpack/issues/3400

We should try to capture this error here in the library, give a specific error type, and propagate it to the client.

designsimply:
@daniloercoli I tested uploading images and videos on a site set with `upload_max_filesize = 1M` today and found that the image was too big the error message was good:

> Files exceeds the maximum upload size for …

But when the video was too big the error message was very generic:

> We couldn't complete this action.

See https://github.com/Automattic/jetpack/issues/3400#issuecomment-548155082.

designsimply:
I tried re-testing this today and was unable to reproduce the problem using the following steps:

1. Set up a Jetpack site and use php.ini to set `upload_max_filesize = 1M`.
1. In the app, go to My Site > Account > App Settings and enable the "Optimize Images" setting.
1. Go to My Sites > Media.
1. Try to upload an image larger than 1MB.

**Result:** when I try to upload an image larger than the max upload file size set by the server I don't see an HMAC-related error message but I do see an expected error letting me know "File exceeds the maximum upload filesize" and when I try to upload an image that's an allowed size smaller than the max filesize setting on the server it works normally. ([30s](https://cloudup.com/ce70v7TeyKZ))

upload_max_filesize|optimize setting is on|sample file is 2MB|upload fails with appropriate error
---|---|---|---
[...]

PR #454, opened 2017-05-29
Media Uploading - HMAC error on JP connected sites

daniloercoli:
This is a *temporary?* fix for #402

This patch does check the `error` field available in the REST Response, and set the response error type to `REQUEST_TOO_LARGE ` if  `error` is equals to one of the two predefined error messages we get in case of  too large media.

A better fix is probably required server side.


daniloercoli:
cc @SergioEstevao

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
```diff
@@ -373,6 +373,18 @@ private MediaError parseUploadError(Response response) {
...
                 mediaError.message = body.getString("message");
             }
+
+            if (!siteModel.isWPCom()) {
+                // TODO : temporary fix for "big" media uploads on Jetpack connected site
+                // See https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/402
+                // Tried to upload a media that's too large (larger than the site's max_upload_filesize)
+                if (body.has("error")) {
+                    String error = body.getString("error");
+                    if ("invalid_hmac".equals(error) || "unknown".equals(error)) {
```
aforcier: I can see that we get an `unknown : Unknown error communicating with the Jetpack site` error if we exceed both `upload_max_filesize` and `post_max_size`, and I guess that's why you're changing the error for `unknown` too. But I'm not sure that we should make this change for `unknown` errors, as we could get those legitimately for other reasons (we could get the HMAC one too I guess, but that's less likely).

daniloercoli: Yes, I was torn about catching "unknown" errors. I did some tests on JP sites, and there are times where the error reported is `HMAC....`, and other where it's `Unknown`. All depends on the hosting configuration.

Anyway, I removed the "unknown" catching and left the hmac only, that should be safe to catch. 
Ready for another pass @aforcier

aforcier:
Heads up @daniloercoli this is failing due to a support library conflict - merging in current `develop` should fix that up.

aforcier:
:shipit:

PR #461, opened 2017-05-31
Update post featured image when removed

### LINKED ISSUE #460: Featured image cannot be removed
`PostRestClient` needs to set the featured image ID to an empty string in order to remove it from a post, [per the docs.](https://developer.wordpress.com/docs/api/1.2/post/sites/%24site/posts/%24post_ID/)

tonyr59h:
Fix #460 by passing an empty string as the `featured_image` parameter to remove any previous featured image.

Relevant ticket: wordpress-mobile/WordPress-Android#5930

@aforcier would you mind taking a look at this one?

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
@@ -344,7 +344,11 @@ public class PostRestClient extends BaseWPComRestClient {
         params.put("categories", TextUtils.join(",", post.getCategoryIdList()));
         params.put("tags", TextUtils.join(",", post.getTagNameList()));
 
-        params.put("featured_image", post.getFeaturedImageId());
+        if (post.hasFeaturedImage()) {
+            params.put("featured_image", post.getFeaturedImageId());
+        } else {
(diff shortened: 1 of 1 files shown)


tonyr59h:
Just a reminder: Update the FluxC dependency in WPAndroid once this is merged to get it in the next release.

aforcier:
This makes sense, but I was a bit surprised since we have [a test for this](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/9745efa8afca82418c6bb82e592bca8d15e6f2f2/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java#L463-L494) (on both WP.com and XML-RPC).

It turns out sending `featured_image` `0` to the WP.com API has the same effect as sending `""`, which is what the test is doing, but we're actually setting the featured image to `-1` [in WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/blob/bd9ad9b3628afbca399e8b3a4345f97f7e80deca/WordPress/src/main/java/org/wordpress/android/ui/posts/EditPostSettingsFragment.java#L335). `-1` does *not* work.

I think we should follow the docs and merge this PR, but we should also use `0` in `EditPostSettingsFragment`. The `-1` is a holdover from the dark `hasFeaturedImageChanged()` days.

PR #464, opened 2017-06-02
Bump support lib versions

aforcier:
Updates our support libraries to `25.3.1`. Also switches to static versioning for the WP utils lib so that a)we have build predictability and b)Travis doesn't break due to external changes to the utils lib.


maxme:
:shipit:

PR #468, opened 2017-06-13
Don't upload files exceeding max_upload_size (Jetpack)

aforcier:
A continuation of work for https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/402.

Stores `options.max_upload_size` from `/sites/$site` (this field only exists for Jetpack sites), and checks against it before uploading media. Media larger than the file size limit are never uploaded and result in a `REQUEST_TOO_LARGE` error.

This doesn't make us completely safe, as the php memory limit can still be too low and cause a media upload to fail with an `unknown` error. There's no way to get this value at the moment, but hopefully soon.

cc @daniloercoli


daniloercoli:
Code looks good and tested on a JP site of mine! We can merge this PR, I've just a question doing it. How often are options synched from the external site to wpcom?  If the user changes the `max_memory` on the server, this new value could take time before is propagated to wpcom. Worth to check with Jetpack developers?

aforcier:
@daniloercoli as I understand it, the options in the `/sites/$site` endpoint are piped directly from the Jetpack site through dotcom - we're not waiting for anything to sync.

If I change the values of `upload_max_filesize` and `post_max_size` on my site (you mentioned `max_memory` but that won't affect the `max_upload_size` field) and call `/sites/$site/`, it has the new value right away.

daniloercoli:
Ohh cool! :shipit:

PR #474, opened 2017-06-20
Support fetching from /sites/$site/ using a URL

aforcier:
Adds support for calling the `/sites/$site/` endpoint using a URL instead of a WP.com site ID. This call can be made through a new action, `FETCH_WPCOM_SITE_BY_URL`.

This action is intended to be used during login/discovery only, and does not require authentication - it also does not store the resulting `SiteModel` in the database.

There's a pre-existing action, `IS_WPCOM_URL`, which does something very similar (but returns a boolean instead of a `SiteModel`, and has different criteria for returning an error). Once the use of `FETCH_WPCOM_SITE_BY_URL` is finalized and this is merged, I'll revisit that one as we can probably fold it into `FETCH_WPCOM_SITE_BY_URL` and drop some code.

cc @hypest

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestWPCom.java
@@ -13,9 +13,12 @@ import org.wordpress.android.fluxc.store.AccountStore.AuthenticatePayload;
 import org.wordpress.android.fluxc.store.AccountStore.OnAccountChanged;
 import org.wordpress.android.fluxc.store.AccountStore.OnAuthenticationChanged;
 import org.wordpress.android.fluxc.store.SiteStore;
+import org.wordpress.android.fluxc.store.SiteStore.OnConnectSiteInfoChecked;
 import org.wordpress.android.fluxc.store.SiteStore.OnPostFormatsChanged;
 import org.wordpress.android.fluxc.store.SiteStore.OnSiteChanged;
 import org.wordpress.android.fluxc.store.SiteStore.OnSiteRemoved;
+import org.wordpress.android.fluxc.store.SiteStore.OnWPComSiteFetched;
+import org.wordpress.android.fluxc.store.SiteStore.SiteErrorType;
 import org.wordpress.android.util.AppLog;
 import org.wordpress.android.util.AppLog.T;
 
@@ -38,6 +41,8 @@ public class ReleaseStack_SiteTestWPCom extends ReleaseStack_Base {
         POST_FORMATS_CHANGED,
         SITE_REMOVED,
         FETCHED_CONNECT_SITE_INFO,
+        FETCHED_WPCOM_SITE_BY_URL,
+        SITE_FETCH_ERROR
     }
 
     private TestEvents mNextEvent;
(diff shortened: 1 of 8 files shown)


aforcier:
Updated with @hypest's suggestion of reporting an `UNKNOWN_SITE` error instead of `INVALID_SITE` when a URL is valid, but is not a WP.com site.

mzorz:
Been debugging this one and making sense of the new action. Tests pass - also nice to see error paths covered in the tests as well, nice work @aforcier 

One thing that probably exceeds this PR though, but caught my attention nevertheless, is the following:

> This action [...] does not store the resulting SiteModel in the database.

Given that the general expectation about FluxC Store operations is for the store to take care of the `fetching` _and_ `storing`, I suggest that any exceptions like this one has at least a naming convention that suggests in a clear way that this operation has no effects in the actual store (i.e. doesn’t store anything).

Maybe using a “NS” prefix or postfix (acronym for “NoStore”, or maybe something more creative… that doesn’t look like something Apple-related ;) ).
`onNSFetchedConnectSiteInfo()`
`OnNSWPComSiteFetched()`
`FetchNSWPComSiteResponsePayload`
`SiteActionBuilder.newNSFetchedWpcomSiteByUrlAction`
`NS_FETCH_WPCOM_SITE_BY_URL `

OR

`onFetchedConnectSiteInfoNS()`
`OnWPComSiteFetchedNS()`
`FetchWPComSiteResponsePayloadNS`
`SiteActionBuilder.newFetchedWpcomSiteByUrlActionNS`
`FETCH_WPCOM_SITE_BY_URL_NS`

Do we have any other actions that have no effect on the stores? just a brief look throws `checkUrlIsWPCom` for instance. Maybe this is food for another PR, but I think this case is enough to make this distinction visible to the FluxC user so as to make no mistakes and save them debugging time :). 

Likewise, we might (or not) want to do the same with store-only operations that do not affect or require connecting to the server.

Thoughts?

aforcier:
> Do we have any other actions that have no effect on the stores?

We have a few site-based actions that don't modify the store, this is all of them:

```
IS_WPCOM_URL
SUGGEST_DOMAINS
FETCH_CONNECT_SITE_INFO
FETCH_WPCOM_SITE_BY_URL
```

These are incidentally also 'logged out' actions (see https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/492 where it's made explicit by not passing a token for those network requests).

I think distinguishing actions that are 'signed out'/don't affect the store makes sense. I'm not sold yet on using an internal naming convention, but it would be nice to have them distinguished to the FluxC client. e.g. maybe we could have `SiteUnauthedActionBuilder.newSuggestDomainsAction()`.

In the future, we could conceivably have actions that are unauthed but affect the store, or authed actions that never affect the store, so I think this needs some more thought if we're to introduce a reliable naming scheme.

Want to open an issue for this @mzorz?

> Likewise, we might (or not) want to do the same with store-only operations that do not affect or require connecting to the server.

We actually already have a [naming convention for this](https://github.com/wordpress-mobile/WordPress-FluxC-Android#actions):

```
FETCH_X - request data from the server
PUSH_X - send data to the server
UPDATE_X - local change
REMOVE_X - local remove
DELETE_X - request deletion on the server
```

mzorz:
Opened issue for discussion #503

mzorz:
:shipit:

PR #475, opened 2017-06-22
Add utility method for retrieving a post from the database by a given ID

tonyr59h:
Reference:

Moving a helper method added in wordpress-mobile/WordPress-Android#6030 since it's more appropriate in FluxC.

cc @oguzkocer

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java
@@ -79,6 +79,20 @@ public class PostSqlUtils {
                 .getAsModel();
     }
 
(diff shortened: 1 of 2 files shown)


aforcier:
Is this one still necessary if https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/476 is merged? It seems like the only reason for the original lookup by remote post ID was because the local id of the posts in the search results could be `0`.

If that's the case, let's avoid encouraging lookups by remote ID where possible and not add these methods.

tonyr59h:
> Is this one still necessary if #476 is merged?

Not for the search feature, but I assumed this could be useful for future WPAndroid features.

Since @oguzkocer discovered some flaws with pagination interfering with the search feature we've decided to bench these two PR's until we have a clear path forward.

PR #476, opened 2017-06-22
Update local post ID on given post when updating a post in the database

tonyr59h:
This change came about from wordpress-mobile/WordPress-Android#6030 after noticing the search results don't have valid local ID's.

Now the local ID will be set to the existing DB row's ID after updating it. This will ensure that post change events pass posts with valid local ID's.

cc @oguzkocer


tonyr59h:
Closing this until we have more information, see [comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/475#issuecomment-310462393).

PR #457, opened 2017-05-30
Delete cancelled media from the store

aforcier:
Part of the fix for https://github.com/wordpress-mobile/WordPress-Android/issues/5942.

~We had a `MediaModel.UploadState.CANCELED` state, but it was unused. I opted to drop this entirely and just delete media that are cancelled from the store. If we ever want to support retrying cancelled media, we can re-introduce this; for now, I think this is more straightforward.~

Current WPAndroid cancels media when exiting the editor, but we want to keep those around so we can retry the upload. To accommodate this, I opted to support a `delete` flag (`true` by default) in `CANCEL_MEDIA_UPLOAD`. When set to `false`, the media item will still be cancelled but it won't be deleted from the store - instead it will be set to a `FAILED` state.

This PR also kills the unused `MediaModel.UploadState.CANCELED`, and corrects some behavior in media deletion and in ignoring upload errors for already-deleted media.

cc @mzorz

PR #477, opened 2017-06-22
Manage media upload state in FluxC

### LINKED ISSUE #415: Manage media upload state in FluxC
Currently we're [relying on the client](https://github.com/wordpress-mobile/WordPress-Android/blob/efb1248aa8387313c61cc4997dfa4fa99a198b5a/WordPress/src/main/java/org/wordpress/android/ui/media/services/MediaUploadService.java#L125) to change the `UploadState` for uploading media - we should handle this in FluxC instead.

We should also be able to drop `mUploadCancelled` in `MediaModel` ([ref](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/389/files#r108187722)) once we've cleaned up upload state handling.

aforcier:
Fixes #415, moving the responsibility for updating media states (`QUEUED` --> `UPLOADING`, and `UPLOADING` --> `UPLOADED`/`FAILED`) to FluxC.

I also did some smaller cleanup:
* Renamed `UploadState` to `MediaUploadState`
* Renamed `UploadState.DELETE` to `UploadState.DELETING`
* Overloaded `MediaModel.setUploadState()` to accept `MediaUploadState` enums directly as well as strings
* Refactored the `MediaRestClient` and `MediaXMLRPCClient` to make them more consistent and easier to read/compare (thanks @mzorz for [suggesting it](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/457#issuecomment-308172811))

I'll be opening a WPAndroid PR to address the renaming and to drop the now-redundant setting of the media upload state - please wait to merge this until that one is also ready to merge.

cc @mzorz

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
@@ -19,7 +19,7 @@ public class MediaModel extends Payload implements Identifiable, Serializable {
         public static MediaUploadState fromString(String stringState) {
             if (stringState != null) {
                 for (MediaUploadState state: MediaUploadState.values()) {
-                    if (stringState.equalsIgnoreCase(state.name())) {
+                    if (stringState.equalsIgnoreCase(state.toString())) {
                         return state;
                     }
                 }
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
@@ -14,7 +14,18 @@ import java.io.Serializable;
 @Table
 public class MediaModel extends Payload implements Identifiable, Serializable {
     public enum UploadState {
-        QUEUED, UPLOADING, DELETING, DELETED, FAILED, UPLOADED
+        QUEUED, UPLOADING, DELETING, DELETED, FAILED, UPLOADED;
+
+        public static UploadState fromString(String stringState) {
+            if (stringState != null) {
+                for (UploadState state: UploadState.values()) {
+                    if (stringState.equalsIgnoreCase(state.name())) {
+                        return state;
+                    }
+                }
+            }
+            return UPLOADED;
+        }
     }
 
     @PrimaryKey
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
@@ -13,12 +13,12 @@ import java.io.Serializable;
 
 @Table
 public class MediaModel extends Payload implements Identifiable, Serializable {
-    public enum UploadState {
+    public enum MediaUploadState {
         QUEUED, UPLOADING, DELETING, DELETED, FAILED, UPLOADED;
 
-        public static UploadState fromString(String stringState) {
+        public static MediaUploadState fromString(String stringState) {
             if (stringState != null) {
-                for (UploadState state: UploadState.values()) {
+                for (MediaUploadState state: MediaUploadState.values()) {
                     if (stringState.equalsIgnoreCase(state.name())) {
                         return state;
                     }
@@ -318,7 +318,7 @@ public class MediaModel extends Payload implements Identifiable, Serializable {
         return mUploadState;
     }
 
-    public void setUploadState(UploadState uploadState) {
+    public void setUploadState(MediaUploadState uploadState) {
         mUploadState = uploadState.toString();
     }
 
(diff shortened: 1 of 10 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -494,14 +494,14 @@ public MediaModel getMediaForPostWithPath(PostModel postModel, String filePath)
...
-    public List<MediaModel> getMediaForPostWithState(PostModel postModel, UploadState expectedState) {
+    public List<MediaModel> getMediaForPostWithState(PostModel postModel, MediaUploadState expectedState) {
         return MediaSqlUtils.matchPostMedia(postModel.getId(), MediaModelTable.UPLOAD_STATE,
                 expectedState);
     }
 
     public MediaModel getNextSiteMediaToDelete(SiteModel siteModel) {
         List<MediaModel> media = MediaSqlUtils.matchSiteMedia(siteModel,
-                MediaModelTable.UPLOAD_STATE, UploadState.DELETE.toString());
+                MediaModelTable.UPLOAD_STATE, MediaUploadState.DELETING.toString());
```
mzorz: Could it happen that some media in the previous "DELETE" state are in FluxC's database, and then an upgrade to this FluxC version would make this query not match any files?
Also, some other things might happen, should we run a migration upon start, between any  entries with the "DELETE" state and change them to  "DELETING"?

mzorz: Please forgive me if there's something I'm missing about FluxC's internals here 🙇

aforcier: Not missing anything about FluxC internals, it's a very good point. I had meant to mention this in the PR notes but forgot.

Normally, yes, we'd want a migration to update existing db entries. But we actually haven't been using `DELETE` anywhere. It wasn't set at anytime in FluxC (except in tests), and in WPAndroid we were actually using WPAndroid's own `MediaUploadState.DELETING` (converted to a String) (see https://github.com/wordpress-mobile/WordPress-Android/pull/6129).

mzorz: Awesome! nice to see you've already taken care of it :)

fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
```diff
@@ -13,8 +13,19 @@
...
 public class MediaModel extends Payload implements Identifiable, Serializable {
-    public enum UploadState {
-        QUEUED, UPLOADING, DELETE, DELETED, FAILED, UPLOADED
+    public enum MediaUploadState {
+        QUEUED, UPLOADING, DELETING, DELETED, FAILED, UPLOADED;
+
+        public static MediaUploadState fromString(String stringState) {
+            if (stringState != null) {
+                for (MediaUploadState state: MediaUploadState.values()) {
+                    if (stringState.equalsIgnoreCase(state.name())) {
```
mzorz: This one here compares to `name()` but [this other line here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/477/files#diff-f1c0df0705015841092e56ef743c596bR322) normalizes everything to `toString(): 

```
public void setUploadState(MediaUploadState uploadState) {
       mUploadState = uploadState.toString();
[...]
```

Can we use `.name()` everywhere instead of `toString()`?  as `toString()` can be overriden and I think the upload states should be immutable, also now knowing they're saved as Strings in the WellSql database.

aforcier: Ah yes, this one was an omission, it was supposed to use `toString()` as well.

> Can we use .name() everywhere instead of toString()?

I see your argument, and I went back on forth on this too - I think `toString()` is marginally better. Here's my thinking:

The reason we're using `toString()` or `name()` in the client code is to compare one enum with an enum that was converted to a String _in FluxC_ (by some method) and stored in a `MediaModel`. To me it makes sense to use `toString()` by default from the client, because then we're confident that we're reflecting the same usage the developer used when storing the enum as a String in the `MediaModel` in the first place (trusting that, if they implemented `toString()`, then they're using it, and if they didn't, it doesn't matter if they're using `name()` or `toString()` internally).

> the upload state should be immutable

The immutability of the enum string value can be broken just as easily by renaming the enums as by implementing `toString()` - it doesn't feel like `name()` is offering any real protection.

Ultimately, all of this would be moot if WellSql supported enums (and arbitrary Serializables for that matter), but it doesn't 😞. Probably we'll switch to Room at some point and not have to worry about stringification for these state enums anymore.

mzorz: > To me it makes sense to use toString() by default from the client, because then we're confident that we're reflecting the same usage the developer used

You had me there. 👍 
Let's use `toString()` 👍

aforcier: 👍 updated this case to use `toString()` as well in 48adc97.

aforcier:
Ready for another pass @mzorz!

mzorz:
:shipit:

PR #478, opened 2017-06-22
Fix Jetpack media test class

aforcier:
Adds missing `FETCH_ACCOUNT` step to authorization for `ReleaseStack_MediaTestJetpack` (the single test in this class fails without it).

Background: This test was added after https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/438 (since which an `AccountModel` is required before WP.com REST sites can be stored in the database), but before the feature branch that PR was targeting was merged into `develop`, and we missed that this test was broken.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestJetpack.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestJetpack.java
@@ -3,6 +3,7 @@ package org.wordpress.android.fluxc.release;
 import org.greenrobot.eventbus.Subscribe;
 import org.wordpress.android.fluxc.TestUtils;
 import org.wordpress.android.fluxc.example.BuildConfig;
+import org.wordpress.android.fluxc.generated.AccountActionBuilder;
 import org.wordpress.android.fluxc.generated.AuthenticationActionBuilder;
 import org.wordpress.android.fluxc.generated.MediaActionBuilder;
 import org.wordpress.android.fluxc.generated.SiteActionBuilder;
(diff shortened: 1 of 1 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestJetpack.java
```diff
@@ -133,8 +149,8 @@ public void onSiteChanged(SiteStore.OnSiteChanged event) {
 
     @SuppressWarnings("unused")
     @Subscribe
-    public void onSiteRemoved(SiteStore.OnSiteRemoved event) {
-        AppLog.e(AppLog.T.TESTS, "site count " + mSiteStore.getSitesCount());
+    public void onSiteRemoved(OnSiteRemoved event) {
+        AppLog.d(T.TESTS, "site count " + mSiteStore.getSitesCount());
```
mzorz: May I suggest that we also use the same logging types/levels on `ReleaseStacak_SiteTestJetpack.java`? the corresponding `onSiteRemoved` event here has a call `AppLog.e` as well https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java#L337

aforcier: We should probably do one big sweep and refactor all the connected tests for consistency. Would you mind opening an issue and we can address that cleanup separately and all at once?

mzorz:
Made just a[ small comment here ](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/478/files#r123814107) but it should be addressed on some other PR. This one is good to go
:shipit:

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestJetpack.java
mzorz: tracked down in https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/480

PR #483, opened 2017-06-26
Run unit tests on Travis

aforcier:
I'm not sure what changed (updating to Gradle 4.0?), but at some point Travis stopped running our unit tests. This updates the script to correct that.

cc @maxme

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ android:
 
 env:
   global:
+    - MALLOC_ARENA_MAX=2
     - GRADLE_OPTS="-Xmx768m -Xms256m -Xss1m"
     - ANDROID_SDKS=android-14
     - ANDROID_TARGET=android-14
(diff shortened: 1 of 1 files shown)


aforcier:
Got a bit stuck trying to get this to work. Re-enabling the unit tests seems to now cause Travis to run out of memory and kill the test runner, failing the build.:

```
* What went wrong:
Execution failed for task ':example:testDebugUnitTest'.
> Process 'Gradle Test Executor 1' finished with non-zero exit value 137
```

I've tried a few changes to the `GRADLE_OPTS` so far, but I'm not sure Travis is actually using those settings - I'll do some further testing.

(some background: https://discuss.gradle.org/t/gradle-travis-ci/11928)

maxme:
@aforcier in 41a6a22 I used the same environment variable that we use in wpandroid to lower memory usage and travis seems happy, but it seems a bit random and I was maybe lucky this time. I guess we can :shipit: and revert if that happens again.

aforcier:
Nice @maxme, thanks! Yeah let's :shipit: for now see see what happens 👍

PR #484, opened 2017-06-27
fixed log message

mzorz:
We were using `mediaXMLRPCClient` in the log message instead of `mediaRestClient`, sneaked in as part of #477 

cc @aforcier


aforcier:
:shipit:

PR #486, opened 2017-06-28
Tweak usage of DUPLICATE_SITE

### LINKED ISSUE #479: Tweak DUPLICATE_SITE triggering
When signing in to the same self-hosted site twice, we should attempt to quietly update the existing site rather than show a `DUPLICATE_ERROR`, where possible.

Partly, this essentially comes down to being less sensitive of the differences between `wp.getUsersBlogs` and `wp.getOptions`:

1. Site URLs from `wp.getOptions` (stored) and `wp.getUsersBlogs` (newly fetched during sign-in) [don't match](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/8680c825896d7c457f55c86bbb19357260996d0d/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java#L110) - we knew about this possibility already, but it can happen that the only difference is a trailing slash, which we should account for
2. A Jetpack-connected site (that was added as a self-hosted site) will have a `SITE_ID`, but only once `wp.getOptions` have been fetched. When logging back in in this case, the stored site has a `SITE_ID > 0`, but the newly fetched site has `SITE_ID == 0`. [This check fails](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/8680c825896d7c457f55c86bbb19357260996d0d/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java#L109). We can refine this check to account for the `ORIGIN` field we've since added, since all we're worried about is overwriting a WP.com REST Jetpack site with its self-hosted version (or adding them both as duplicates)

aforcier:
Fixes #479, quietly updating existing self-hosted sites when logging into them again, instead of throwing an error.

This is mainly accounting for the differences between `wp.getUsersBlogs` and `wp.getOptions`. If a self-hosted site already exists in the app, we've called `wp.getOptions` on it (using `FETCH_SITE`). If we attempt to sign in to that same site again, we'll be working with the results of `wp.getUsersBlogs` (`FETCH_SITES`), which are different (the site's URL can be different, and `wp.getUsersBlogs` lacks  info about the site's Jetpack status).

Previously, if a self-hosted site was not recognized as already existing in the `SiteStore` based on its URL or its URL + its WP.com site ID, we would do an XML-RPC URL match. This used to throw an error for all but the case where a Jetpack-connected site is replacing a non-Jetpack connected site. This is now less strict, and will permit any self-hosted site to be overwritten.

cc @maxme

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/site/SiteStoreUnitTest.java
@@ -419,6 +419,38 @@ public class SiteStoreUnitTest {
         assertEquals(1, sitesCount);
     }
 
+    @Test
+    public void testInsertDuplicateXmlRpcJetpackSite() throws DuplicateSiteException {
+        SiteModel jetpackXmlRpcSite = generateJetpackSiteOverXMLRPC();
+
+        jetpackXmlRpcSite.setUrl("http://some.url");
+
+        // Insert a Jetpack powered site over XML-RPC
+        SiteSqlUtils.insertOrUpdateSite(jetpackXmlRpcSite);
+
+        // Set up the same site (by URL/XML-RPC URL), but don't identify it as a Jetpack site
+        // This simulates sites resulting from wp.getUsersBlogs, which don't have the site ID and can't be identified
+        // as Jetpack or not (wp.getOptions is the call that returns that information)
+        SiteModel jetpackXmlRpcSite2 = generateSelfHostedNonJPSite();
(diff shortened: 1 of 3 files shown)


maxme:
Looks good and the comment is helpful, nitpick: we should update comment on line 62/63 in `SiteSqlUtils`

aforcier:
> Looks good and the comment is helpful, nitpick: we should update comment on line 62/63 in `SiteSqlUtils`

Oops. Fixed in dbc6e08.

maxme:
:shipit:

Issue #456, opened 2017-05-30
Crash: ClassCastException in SiteXMLRPCClient.updateSiteFromOptions


theck13:
```
Fatal Exception: java.lang.ClassCastException: java.lang.String cannot be cast to java.util.Map
       at org.wordpress.android.fluxc.network.xmlrpc.site.SiteXMLRPCClient.updateSiteFromOptions(SiteXMLRPCClient.java:203)
       at org.wordpress.android.fluxc.network.xmlrpc.site.SiteXMLRPCClient.access$400(SiteXMLRPCClient.java:31)
       at org.wordpress.android.fluxc.network.xmlrpc.site.SiteXMLRPCClient$3.onResponse(SiteXMLRPCClient.java:81)
       at org.wordpress.android.fluxc.network.xmlrpc.XMLRPCRequest.deliverResponse(XMLRPCRequest.java:52)
       at com.android.volley.ExecutorDelivery$ResponseDeliveryRunnable.run(ExecutorDelivery.java:99)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:135)
       at android.app.ActivityThread.main(ActivityThread.java:5310)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:901)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:696)
```

PR #490, opened 2017-06-28
Handle ClassCastExceptions during XML-RPC site fetch

aforcier:
'Fixes' #456. The source of the crash was `wp.getOptions` returning a `String` instead of a `Map`. This probably means an irregular WP setup.

The app will no longer crash, and we'll now log the actual `String` response along with the site URL using `OnUnexpectedError`.

I broadened the fix a bit to apply to any XML-RPC site method that would throw a `ClassCastException`, handling it and logging the response. I also did the same for _any_ XML-RPC call that expects an `Object[]` but receives something else (all our fetch-list-of methods).

TODO:
* [x] Add some mocked `SiteXMLRPCClient` tests

cc @maxme


aforcier:
Updated to resolve a merge conflict.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -213,8 +216,9 @@ public void onResponse(PostFormatsResponse response) {
                     @Override
                     public void onErrorResponse(@NonNull BaseNetworkError error) {
                         FetchedPostFormatsPayload payload = new FetchedPostFormatsPayload(site,
-                                new ArrayList<PostFormatModel>());
-                        payload.error = error;
+                                Collections.<PostFormatModel>emptyList());
```
maxme: TIL

maxme:
:shipit:

PR #491, opened 2017-06-28
Issue/489 media mime fetch

### LINKED ISSUE #489: Support mime_type parameter when requesting media
As mentioned in [this WPAndroid comment](https://github.com/wordpress-mobile/WordPress-Android/pull/6172#issuecomment-311694550), supporting the new media filters will require us to support fetching media with a partial `mime_type` [parameter](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/media/).

nbradbury:
Fixes #489 - Adds a mime type parameter when fetching a media list. Note that both the REST and xmlrpc endpoints support partial matching, ie: we can use "image/" to fetch all images or "video/" to fetch all videos.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestWPCom.java
@@ -35,6 +35,7 @@ public class ReleaseStack_MediaTestWPCom extends ReleaseStack_WPComBase {
         CANCELED_MEDIA,
         DELETED_MEDIA,
         FETCHED_MEDIA_LIST,
+        FETCHED_MEDIA_IMAGE_LIST,
         FETCHED_KNOWN_IMAGES,
         PUSHED_MEDIA,
         REMOVED_MEDIA,
@@ -96,6 +97,14 @@ public class ReleaseStack_MediaTestWPCom extends ReleaseStack_WPComBase {
         mNextEvent = TestEvents.FETCHED_MEDIA_LIST;
         fetchMediaList();
         assertFalse(mMediaStore.getAllSiteMedia(sSite).isEmpty());
+
+        // remove all media again, fetch only images, verify store is not empty and contains only images
+        mNextEvent = TestEvents.FETCHED_MEDIA_IMAGE_LIST;
+        removeAllSiteMedia();
+        fetchMediaImageList();
+        List<MediaModel> mediaList = mMediaStore.getSiteImages(sSite);
+        assertFalse(mediaList.isEmpty());
+        assertTrue(mMediaStore.getSiteMediaCount(sSite) == mediaList.size());
 
         // delete test image
         mNextEvent = TestEvents.DELETED_MEDIA;
@@ -449,7 +458,9 @@ public class ReleaseStack_MediaTestWPCom extends ReleaseStack_WPComBase {
             return;
         }
         if (event.cause == MediaAction.FETCHED_MEDIA_LIST) {
-            assertEquals(TestEvents.FETCHED_MEDIA_LIST, mNextEvent);
+            boolean isMediaListEvent = mNextEvent == TestEvents.FETCHED_MEDIA_LIST
+                    || mNextEvent == TestEvents.FETCHED_MEDIA_IMAGE_LIST;
+            assertTrue(isMediaListEvent);
         } else if (event.cause == MediaAction.FETCH_MEDIA) {
             if (eventHasKnownImages(event)) {
                 assertEquals(TestEvents.FETCHED_KNOWN_IMAGES, mNextEvent);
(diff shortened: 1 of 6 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
```diff
@@ -243,13 +243,16 @@ public void onFailure(Call call, IOException e) {
...
      */
-    public void fetchMediaList(final SiteModel site, final int offset) {
+    public void fetchMediaList(final SiteModel site, final int offset, String mimeType) {
         List<Object> params = getBasicParams(site, null);
         Map<String, Object> queryParams = new HashMap<>();
         queryParams.put("number", MediaStore.NUM_MEDIA_PER_FETCH);
         if (offset > 0) {
             queryParams.put("offset", offset);
         }
+        if (mimeType != null && mimeType.length() > 0) {
```
daniloercoli: Nitpicking - Should we use `TextUtils.isEmpty` instead?

daniloercoli:
Wondering if we need to add 2 new tests here [1](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestWPCom.java#L79), and [2](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestXMLRPC.java#L100). 

Maybe we should add a new test, say `testFetchImageList`, that uploads 1 video and 1 image. Internally calls a new method `fetchMediaList(String mime)`  (instead of the current `fetchMediaList()`) that in order calls your new constructor ` public void fetchMediaList(final SiteModel site, final int offset, String mimeType) {` and verify that we received back only pictures.

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
nbradbury: > Should we use TextUtils.isEmpty instead?

Ha, for some reason I was thinking `isEmpty()` was a WPAndroid method rather than one built into Android. Fixed in f80d073.

nbradbury:
> Wondering if we need to add 2 new tests here 1, and 2.

I added tests in 88af1de which simply clear the media store, fetch only images, and verify that only images are in the store. These are my first FluxC tests so I'm not sure if they were done correctly.

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -615,9 +622,9 @@ private void performFetchMediaList(FetchMediaListPayload payload) {
             offset = MediaSqlUtils.getMediaWithStates(payload.site, list).size();
```
aforcier: It looks like there's an issue here. We calculate the offset for the fetch, but we're not adjusting it according to the mime type - so if we've already fetched ~50 images, and the site only has 2 videos and they're older than those images, we'll always be fetching video offset by 50, which will always return empty.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestWPCom.java
```diff
@@ -97,6 +98,14 @@ public void testFetchMediaList() throws InterruptedException {
         fetchMediaList();
         assertFalse(mMediaStore.getAllSiteMedia(sSite).isEmpty());
 
+        // remove all media again, fetch only images, verify store is not empty and contains only images
+        mNextEvent = TestEvents.FETCHED_MEDIA_IMAGE_LIST;
```
aforcier: This test is failing for me - we need to expect a `TestEvents.REMOVED_MEDIA` here instead, and change it to `TestEvents.FETCHED_MEDIA_IMAGE_LIST` after the below `remoteAllSiteMedia()` line.

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
nbradbury: Thanks for catching that, I hadn't realized how the offset is calculated. I believe that's fixed now by 93fe044.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_MediaTestWPCom.java
nbradbury: I fixed the tests in 699ef79, but I'm unable to verify the `ReleaseStack_MediaTestXMLRPC` fix because the `CANCELED_MEDIA` event is failing for me.

daniloercoli:
Tests seem OK to me. Waiting @aforcier feedback on the offset fix.

aforcier:
Tests look good, and changes look good overall. I also tested WPAndroid-side and the mime requests + offsets are working well (once I made a small fix locally, which I noted on that PR).

:shipit: from me, and nice work @nbradbury!

daniloercoli:
Merging!!

PR #492, opened 2017-06-29
Omit token for 'public' endpoints

aforcier:
Modifies the `FETCH_CONNECT_SITE_INFO`, `FETCH_WPCOM_SITE_BY_URL`, `IS_WPCOM_URL` and `SUGGEST_DOMAINS` actions to always make unauthenticated requests to WP.com (that is to say, not use a token even if it exists).

(`FETCH_WPCOM_SITE_BY_URL` calls `/sites/$siteUrl/`, which is not necessarily 'public', as a token matching the site would retrieve more information from the API - but we're using it exclusively for login/discovery for the foreseeable future.)

@hypest pointed out that if FluxC has an invalid token, requests to these endpoints will return an error, when omitting the token altogether would've given a successful response.

Note: #474 should be merged before this PR and the base branch updated..


hypest:
FYI, I tested the `FETCH_WPCOM_SITE_BY_URL` action with a relevant build from the Login Rework project and works OK both with a valid and an invalid auth token. 👍

mzorz:
:shipit:

PR #494, opened 2017-06-29
Bumping OkHttp version to 3.8.1

mzorz:
This is needed in order to take advantage of some [important bugfixes since 3.4.1](https://github.com/square/okhttp/blob/master/CHANGELOG.md) 

These are of particular interest: 

> 3.6.0
> Fix: Calling disconnect() on a connecting HttpUrlConnection could cause it to retry in an infinite loop! This regression was introduced in OkHttp 2.7.0.

and

> 3.8.1
> Fix: Recover gracefully from stale coalesced connections. We had a bug where connection coalescing (introduced in OkHttp 3.7.0) and stale connection recovery could interact to cause a NoSuchElementException crash in the RouteSelector.

See [related WPAndroid issue ](https://github.com/wordpress-mobile/WordPress-Android/issues/6155)

cc @aforcier

### DIFF
--- a/fluxc/build.gradle
+++ b/fluxc/build.gradle
@@ -73,8 +73,8 @@ dependencies {
 
     // External libs
     compile 'org.greenrobot:eventbus:3.0.0'
-    compile 'com.squareup.okhttp3:okhttp:3.4.1'
-    compile 'com.squareup.okhttp3:okhttp-urlconnection:3.4.1'
+    compile 'com.squareup.okhttp3:okhttp:3.8.1'
+    compile 'com.squareup.okhttp3:okhttp-urlconnection:3.8.1'
     compile 'com.android.volley:volley:1.0.0'
     compile 'com.google.code.gson:gson:2.7'
     compile 'com.google.dagger:dagger:2.0.2'


aforcier:
This looks to be working nicely, but for some reason WP.com media uploads are now failing with Charles hooked up. Looks like they work without it but I'm not 100% sure yet it's a Charles-only problem - I'm looking into this.

aforcier:
Update with my findings so far:

1. Only happens for WP.com (it's working for the SSL self-hosted site I tested)
2. Only happens for media uploads (with or without title/description params)
3. Only happens with Charles, with SSL proxying turned on (SSL proxying off, or using Burp with SSL proxying on both work)
4. Happens consistently when uploading from the connected tests
5. Happens intermittently when uploading from the example app
6. Whatever changed in OkHttp to make Charles unhappy was introduced in [OkHttp `3.6.0`](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-360)

Going to look into this further on Monday.

aforcier:
I can still reproduce the issue with new OkHttp + Charles + SSL Proxying, on every emulator I try (Genymotion and stock, various API versions), completely consistently (when running a connected WP.com test).

I can't reproduce on a Pixel (it was intermittent before, but now the test always passes).

@daniloercoli could not reproduce on any device, and @maxme could not reproduce on an emulator.

Will try a few more things but it doesn't look like this should be a blocker - will resume reviewing this shortly.

aforcier:
I give up 🤷‍♂️ - I'll use a device or a different MITM tool for media tests until I figure out a fix. Does not seem like an issue that would affect actual users (only failing when SSL proxying is on).

aforcier:
:shipit:, nice one @mzorz!

PR #497, opened 2017-06-30
Issue/496 media fetch clear mime type

### LINKED ISSUE #496: Fetching media list shouldn't clear all media
In #491 we changed the media list fetch to support fetching specific partial mime types, but I just discovered that [the media store is cleared](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java#L721) after every fetch before inserting the fetched media. It should instead only clear media matching the requested mime type.

nbradbury:
Fixes #496 - prior to this PR we were clearing the media store after the media list was fetched, this PR changes it to only clear the store of media matching the requested mime type. I also updated `OnMediaListFetched` to include the requested mime type so that WPAndroid would know what mime type was requested.


aforcier:
:shipit:

PR #498, opened 2017-06-30
Media connected test tweaks

aforcier:
Some small tweaks/cleanup for our connected media tests:

* Drop some unused test event matching
* Stricter matching of event cause when receiving an `OnMediaChanged`
* Add some explanatory notes to `tests.properties-example` about the test media files
* Suppress annoying `SparseArray` warnings - it's not an optimization we really need, and iterating through the keys would turn out to be pretty clunky (it lacks a `values()` implementation)

cc @mzorz


mzorz:
Tests pass, also took a closer look into the usage of `TestEvents.FETCHED_MEDIA_LIST` which confused me a bit now that it's only used in the proper `onMediaListFetched()` - but all good!
LGTM
:shipit:

PR #500, opened 2017-07-01
issue/499-request-48-media

### LINKED ISSUE #499: Fetch media 48 at a time rather than 50 at a time
The media library in WPAndroid previously showed two columns, but it recently changed to three columns in portrait and four in landscape. Since we load media 50 at a time, this results in empty cells until "load more" finishes.

To fix this, we need to request a multiple of both 3 and 4, and 48 seems the most reasonable.

screenshot_1498946437

nbradbury:
Fixes #499 - fetches media 48 at a time so it's a multiple of the column count in WPAndroid (3 in portrait, 4 in landscape).

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_FluxCImageLoaderTest.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_FluxCImageLoaderTest.java
@@ -134,7 +134,8 @@ public class ReleaseStack_FluxCImageLoaderTest extends ReleaseStack_Base {
         mNextEvent = TestEvents.FETCHED_MEDIA_LIST;
         mCountDownLatch = new CountDownLatch(1);
 
-        FetchMediaListPayload fetchPayload = new FetchMediaListPayload(site, false);
+        FetchMediaListPayload fetchPayload = new FetchMediaListPayload(
+                site, MediaStore.DEFAULT_NUM_MEDIA_PER_FETCH, false);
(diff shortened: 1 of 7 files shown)


aforcier:
Instead of this, how about we retire this hardcoded value and pass it from the app? Like we do for comments ([WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/blob/cb6ff74f258442e77b624feaf64f0d26535c9c1f/WordPress/src/main/java/org/wordpress/android/ui/comments/CommentsListFragment.java#L543), [FluxC](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/976bf784510232718c81788408ccc77c9653bce9/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java#L399)).

nbradbury:
> Instead of this, how about we retire this hardcoded value and pass it from the app? Like we do for comments

Works for me. I'll get started on this.

nbradbury:
@aforcier As suggested, fetching the media list now accepts a `number` parameter, similar to comments.

aforcier:
Looks great @nbradbury! Tested it in your branch for https://github.com/wordpress-mobile/WordPress-Android/pull/6184 and it's working nicely (if you have local (failed) media that's not a multiple of 3, you still get the blank spaces, but we can't win that one).

Leaving the :shipit: but holding off merging until tomorrow, as this will also need a PR against WPAndroid `develop` to update the existing usages of `FetchMediaListPayload` to include the `number` (as well as it being updated in your own feature branch).

aforcier:
:shipit:

PR #502, opened 2017-07-04
Always populate the mimeType field in FetchMediaListResponsePayload

### LINKED ISSUE #501: Include the requested mime type in case of errors loading media list
We should set the `mimeType` field in the `FetchMediaListResponsePayload` from the network clients even on errors, and passing it through to the `OnMediaListFetched` event in the store.

Ref: https://github.com/wordpress-mobile/WordPress-Android/pull/6184#issuecomment-312864865

daniloercoli:
Fix #501 by including the mimeType of the request in case of errors retrieving the media list.

This info is required in wp-android to check that's the *current* media list request that it's failing.


aforcier:
We'll also need to set the `mimeType` in the `OnMediaListFetched` event [in the `MediaStore`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/8999308505f091a1a9506521d208e904bfdb78e3/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java#L727).

daniloercoli:
:man_facepalming: 

Added in e93b903.

aforcier:
:shipit:

PR #505, opened 2017-07-04
Report media upload connection errors

aforcier:
Catches `ConnectException` and `UnknownHostException` (thrown when a connection is interrupted, and when a connection wasn't there in the first place, respectively) during media uploads, and reports them as a `CONNECTION_ERROR` instead of a `GENERIC_ERROR`.

cc @mzorz

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
@@ -25,7 +25,9 @@ import org.wordpress.android.fluxc.persistence.MediaSqlUtils;
 import org.wordpress.android.util.AppLog;
 
 import java.io.IOException;
+import java.net.ConnectException;
 import java.net.SocketTimeoutException;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


aforcier:
We should refine this in general for all network calls, but media uploads are the most likely to encounter such an error, and must be handled separately anyway (see https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/506).

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -177,6 +179,10 @@ public static MediaError fromIOException(IOException e) {
                 mediaError.type = MediaErrorType.TIMEOUT;
             }
 
+            if (e instanceof ConnectException || e instanceof UnknownHostException) {
+                mediaError.type = MediaErrorType.CONNECTION_ERROR;
```
mzorz: I was expecting to have fine grained control over which error happened (i.e. escalating whatever we got from the networking layer), but somehow this feels the right pragmatic balance. Good call!

mzorz:
:shipit: 🔥

PR #469, opened 2017-06-13
Remove sites from local DB if they aren't returned in FETCH_SITES response

tonyr59h:
cc @aforcier 

This one is a follow-up to wordpress-mobile/WordPress-Android#6052. I've tested it on that branch in WPAndroid with the latest commit from this branch as its dependency.

To test via WPAndroid:
1. Login to a .com account, head to site picker
2. Add a site to your account on the web
3. Refresh the sites list and verify the site is added
4. Delete the site from the web
5. Refresh the sites list and verify it's been removed

You'll also need to make sure local changes allow the site to stay:
1. Login to a .com account, head to site picker
2. Add a site to your account on the web
3. Refresh the sites list and verify the site is added
4. Make some local only changes to a post on the site
5. Refresh the sites list and verify it's not been removed

PR #508, opened 2017-07-04
Split up FETCHED_SITES action

aforcier:
Since https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/469, a successful `FETCHED_SITES` will call [SiteSqlUtils.removeWPComRestSitesAbsentFromList()](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/2ce630682e0205857158cef45d6be3c75c48663a/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java#L763), regardless of whether the fetch was WP.com or XML-RPC.

Currently this doesn't break anything, as we're calling `UPDATE_SITES` instead on success from XML-RPC, but we could conceivably change things in the future and end up removing WP.com sites on XML-RPC fetches.

To prevent this, there are now two distinct actions, and `SiteSqlUtils.removeWPComRestSitesAbsentFromList()` can only be called as a result of `FETCH_SITES` - not `FETCH_SITES_XML_RPC`.

Note: c4e22fc is the main event - the other commit is just regrouping cases and methods in `SiteStore` so things are grouped together a little better.

cc @tonyr59h

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/action/SiteAction.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/action/SiteAction.java
@@ -46,6 +46,8 @@ public enum SiteAction implements IAction {
     // Remote responses
     @Action(payloadType = SitesModel.class)
     FETCHED_SITES,
+    @Action(payloadType = SitesModel.class)
+    FETCHED_SITES_XML_RPC,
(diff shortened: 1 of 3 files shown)


tonyr59h:
Wow, can't believe that slipped into #469! Thanks for the clean-up, looks good :shipit:

Issue #485, opened 2017-06-27
Store and use memory limit site options for Jetpack sites


aforcier:
The upcoming Jetpack 5.1 will include two new fields under `options` in the `/sites/$site/` response: `wp_memory_limit` and `wp_max_memory_limit`. We can use these to guess when media uploads will be too large for the memory limit and prevent them from being uploaded, as we do for `max_upload_size` (https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/468).

aforcier:
Added in #510 🎉

PR #510, opened 2017-07-06
Don't upload files too close to the memory limit (Jetpack) 

aforcier:
Addresses #485. Stores the new `wp_memory_limit` and `wp_max_memory_limit` site options as a single overall memory limit, and prevents uploading media larger than `0.75` times that limit.

cc @daniloercoli could you give this a spin? Also what do you think about a hard `0.75 * memory-limit` as the ceiling for media uploads? I get failures if I upload files with sizes too close to the real memory limit.

**Note:** There's an issue here in that `/me/sites/` and `/sites/$site/` don't return the same response for Jetpack sites. There are a few discrepancies, but relevant to this PR is that `wp_memory_limit` and `wp_max_memory_limit` aren't set for `/me/sites/`. If we make a `FETCH_SITES` call after tha last `FETCH_SITE` call for a site, we'll replace the stored memory limit in the DB with `0`, until the next `FETCH_SITE` call updates it. From WPAndroid, we're most likely to have called `FETCH_SITE` most recently, but this can still be an issue.

I'm going to open an issue to track this, either getting an API fix or excluding certain fields from being updated by a `FETCH_SITES` call here in FluxC.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
@@ -36,6 +36,7 @@ import org.wordpress.android.fluxc.store.SiteStore.SiteError;
 import org.wordpress.android.fluxc.store.SiteStore.SiteErrorType;
 import org.wordpress.android.fluxc.store.SiteStore.SiteVisibility;
 import org.wordpress.android.fluxc.store.SiteStore.SuggestDomainsResponsePayload;
+import org.wordpress.android.util.StringUtils;
 import org.wordpress.android.util.UrlUtils;
 
 import java.net.URI;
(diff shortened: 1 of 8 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -450,12 +451,25 @@ private SiteModel siteResponseToSiteModel(SiteWPComRestResponse from) {
...
+            // Set the memory limit for media uploads on the site. Normally, this is just WP_MAX_MEMORY_LIMIT,
+            // but it's possible for a site to have its php memory_limit > WP_MAX_MEMORY_LIMIT, and have
+            // WP_MEMORY_LIMIT == memory_limit, in which WP_MEMORY_LIMIT reflects the real limit for media uploads.
+            long wpMemoryLimit = StringUtils.stringToLong(from.options.wp_memory_limit);
+            long wpMaxMemoryLimit = StringUtils.stringToLong(from.options.wp_max_memory_limit);
+            if (wpMemoryLimit > 0 || wpMaxMemoryLimit > 0) {
+                // Only update the value if we received one from the server - otherwise, the original value was
+                // probably not set ('false'), but we don't want to overwrite any existing value we stored earlier,
+                // as /me/sites/ and /sites/$site/ can return different responses for this
+                site.setMemoryLimit(Math.max(wpMemoryLimit, wpMaxMemoryLimit));
```
daniloercoli: Should it be the `Min` between those two values?

aforcier: `Max` should be correct. Most of the time, `WP_MAX_MEMORY_LIMIT` > `WP_MEMORY_LIMIT`, and `WP_MAX_MEMORY_LIMIT` is used to decide the limit for processing uploaded media. So in the vast majority of cases, all we want is `WP_MAX_MEMORY_LIMIT`, and it will be the larger of the two.

It's possible for someone to hardcode `WP_MAX_MEMORY_LIMIT` in WordPress to, say, `12MB`, while the `memory_limit` from PHP is `24MB`. Meanwhile, `WP_MEMORY_LIMIT` wasn't hardcoded, and takes on the value of the `memory_limit`. In that case the backend will use `memory_limit` instead of `WP_MAX_MEMORY_LIMIT` because it's larger, and we have that value through `WP_MEMORY_LIMIT`, so`max(WP_MAX_MEMORY_LIMIT, WP_MEMORY_LIMIT)` will give us the correct value.

If the user has hardcoded _both_ `WP_MAX_MEMORY_LIMIT` and `WP_MEMORY_LIMIT` to something lower than `memory_limit`, then the backend will use `memory_limit` and we have no current way of knowing what the real limit is. In that case we might block an upload that could have succeeded - but this case is pretty rare and doesn't seem worth worrying about.

daniloercoli: Thanks for the detailed clarification!
We can :shipit: it!

PR #512, opened 2017-07-06
issue/511-media-video-thumbs

### LINKED ISSUE #511: Parse media thumbnails for videos
Right now we parse the `thumbnails` of the fetch media list response but only look for the `thumbnail` child, which doesn't exist for videos. We should also look for the `fmt_std` child, as the iOS app does, to support video thumbnails.

screen shot 2017-07-06 at 4 06 46 pm

nbradbury:
Fixes #511 - when parsing the `thumbnails` section of the media list response, we now use `fmt_std` as the thumbnail URL when it exists in order to support video thumbnails (which is consistent with iOS).


daniloercoli:
I tested this and works fine on wpcom. Unfortunately we can't do much for self hosted site, since even the back end the video doesn't have a proper thumb.

PR #514, opened 2017-07-12
Issue/513 media smarter update

### LINKED ISSUE #513: Smarter updating of fetched media
Right now when we fetch the media list for a site, we delete all media from the store for that site before inserting the fetched list. This works fine except that it changes the local IDs of existing media, which results in things like [flickering media grids](https://github.com/wordpress-mobile/WordPress-Android/pull/6271) after a fetch.

We should instead preserve the local IDs of existing media.

nbradbury:
Fixes #513 - after fetching a site's media list, rather than delete all existing media before inserting the fetched list we now retain existing media (and preserve their local IDs), clear everything else out, then insert new media.

Note that the biggest benefit of this PR is that local IDs are retained rather than reset after each fetch. This is important because we want to use the local IDs in WPAndroid as the media grid adapter's `getItemId()` to make updating recycler's more efficient and flicker-free.

You can test using [this WPAndroid branch](https://github.com/wordpress-mobile/WordPress-Android/tree/nbradbury/media-refresh-fluxc), which [writes to the log](https://github.com/wordpress-mobile/WordPress-Android/blob/nbradbury/media-refresh-fluxc/WordPress/src/main/java/org/wordpress/android/ui/media/MediaGridAdapter.java#L114) to let us know whether the list has changed. In most cases it will have changed, but it should not have changed when doing a PTR unless new media have been added.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/MediaSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/MediaSqlUtils.java
@@ -3,6 +3,8 @@ package org.wordpress.android.fluxc.persistence;
 import android.text.TextUtils;
 
 import com.wellsql.generated.MediaModelTable;
+import com.yarolegovich.wellsql.ConditionClauseBuilder;
+import com.yarolegovich.wellsql.DeleteQuery;
 import com.yarolegovich.wellsql.SelectQuery;
 import com.yarolegovich.wellsql.WellCursor;
 import com.yarolegovich.wellsql.WellSql;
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/MediaSqlUtils.java
```diff
@@ -359,6 +362,40 @@ public static int deleteAllMedia() {
...
+                return MediaSqlUtils.deleteAllUploadedSiteMedia(site);
+            }
+        }
+
+        List<Integer> idList = new ArrayList<>();
+        for (MediaModel media: mediaList) {
+            idList.add(media.getId());
+        }
+
+        if (!TextUtils.isEmpty(mimeType)) {
```
daniloercoli: Can't this be simplified with something like the following?
```
        ConditionClauseBuilder<DeleteQuery<MediaModel>> clauseBuilder = WellSql.delete(MediaModel.class)
                .where().beginGroup()
                .isNotIn(MediaModelTable.ID, idList)
                .equals(MediaModelTable.LOCAL_SITE_ID, site.getId())
                .equals(MediaModelTable.UPLOAD_STATE, MediaUploadState.UPLOADED.toString());

        if (!TextUtils.isEmpty(mimeType)) {
            clauseBuilder.contains(MediaModelTable.MIME_TYPE, mimeType);
        }

        return clauseBuilder.endGroup().endWhere().execute();
```

nbradbury: Thanks, I was wondering how to do that. Added in d609ad3.

daniloercoli:
Looks good! I've tested both FluxC and WordPress Android without finding any major problem.

Going to merge this now, but pinging @aforcier or @mzorz since they're more experienced than me with FluxC.

daniloercoli:
:shipit:

aforcier:
> Looks good! I've tested both FluxC and WordPress Android without finding any major problem.
> 
> Going to merge this now, but pinging @aforcier or @mzorz since they're more experienced than me with FluxC.

Looks good to me too, nice work @nbradbury!

PR #520, opened 2017-07-18
Fetch full site path in FETCH_WPCOM_SITE_BY_URL

### LINKED ISSUE #518: /sites/$site endpoint implementation only checks for host part of URI
Currently, FluxC only picks up the host part of the requested URI to test against the `/sites/$siteAddress` API endpoint (see https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/c8e6de55afe86befe739df92d592f469727e0766/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java#L330), but a site can exist in a subfolder under that host.

For example, these test sites (http://do.wpmt.co/) won't be correctly identified because of this issue.

aforcier:
Fixes #518. The full URL is now encoded and sent, which means sites in subfolders are now supported by this action.

cc @hypest


aforcier:
> Looks good! If we have a blog setup for it, could we add a test for the successful path as well?

Added a new test site and test in 6e32aa1. It's the same as `TEST_WPORG_URL_SINGLE_JETPACK_ONLY` if you're using our own test sites, but needed an explicit separate entry in case someone's using their own test sites. (It needs to be Jetpack-connected to receive a successful response from `/sites/$siteUrl`.)

> Also, it's not a big deal, but should we expand on why the UnsupportedEncodingException is impossible? For example, I am not very familiar with that and it could be helpful. It's up to you though.

Updated the comment in e135e87 to be a bit less flippant 😉

PR #521, opened 2017-07-20
Add checkstyle rule colons in for-each statements

aforcier:
Enforces a space on both sides of the colon in for-each statements, and updates pre-existing usages that are now in conflict.

The new rule will flag this:
```java
for (Object object: objectList) {
    ...
}
```

And will allow this:
```java
for (Object object : objectList) {
    ...
}
```

(Debate with @maxme over which format to standardize was settled by the Android Studio auto-formatter, which defaults to the whitespace on both sides format.)


maxme:
French typography FTW :shipit:

PR #525, opened 2017-07-21
Fix #524: fix XMLRPC serialization issue

### LINKED ISSUE #524: XMLRPC serializer crashes on some emoji (chars out of the XML v1.0 spec)
Try to upload a post containing `😆` on a self hosted site to repo.

maxme:
Fix #524: fix XMLRPC serialization issue

* `git checkout  fc4af79 && ./gradlew test` 💥 (176 tests completed, 1 failed)
* `git checkout issue/524-fix-serialization-issue && ./gradlew test` 🍸 🍺 🍷

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/utils/XMLSerializerUtilsTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/utils/XMLSerializerUtilsTest.java
@@ -4,6 +4,7 @@ import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.robolectric.RobolectricTestRunner;
+import org.wordpress.android.fluxc.network.xmlrpc.XMLRPCSerializer;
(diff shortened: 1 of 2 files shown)


maxme:
When this is merged, would be great to update the hash dependency in wpandroid `release/7.9`.

aforcier:
We're still going to get an error log every time a post contains an offending emoji [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/445579994301a6f7e50c4ffb6c94e66585699712/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCSerializer.java#L179), should we maybe step that down to a debug log without a stack trace?

aforcier:
Confirmed the fix with the test and with uploading a post containing 😆 before and after 👍.

maxme:
> should we maybe step that down to a debug log without a stack trace?

Yeah good point, same for the other log, I'll make it a warning only.

aforcier:
:shipit:

Issue #519, opened 2017-07-17
Add an UploadStore


aforcier:
There are things it's useful to track specifically for uploading posts and media.

For posts, we could store the overall upload progress, and the error state if we got an upload error (@mzorz's idea), so we can display it in the post list even after an app restart.

For media, again progress/error state could be stored. We could also keep track of media we've already uploaded using this, and prevent re-uploading the same media to the same site (this was idea of @nbradbury's from a while ago).

aforcier:
This has become the `UploadStore` implementation: #533

aforcier:
Added in #533.

Issue #503, opened 2017-07-04
Improve naming for actions


mzorz:
Coming [from here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/474#issuecomment-312880188)

Given that the general expectation about FluxC Store operations is for the store to take care of the `fetching` _and_  `storing`, I suggest that we come up with a naming convention that suggests in a clear way when an operation is an exception to that rule, i.e. has no effects in the actual store (i.e. doesn’t store anything).

As pointed out [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/474#issuecomment-312886070), this is the list of current actions that have no effect in the store:

```
IS_WPCOM_URL
SUGGEST_DOMAINS
FETCH_CONNECT_SITE_INFO
FETCH_WPCOM_SITE_BY_URL
```

The naming convention should also take into account for actions that are unauthed but affect the store, or authed actions that never affect the store.

My suggestion (proposal?) is to add prefixes or suffixes for the [current action naming convention](https://github.com/wordpress-mobile/WordPress-FluxC-Android#actions), like:

```
FETCH_X - request data from the server
PUSH_X - send data to the server
UPDATE_X - local change
REMOVE_X - local remove
DELETE_X - request deletion on the server

NS_FETCH_X - no store op - request data from the server but don't store
NS_PUSH_X - no store op - send data to the server but don't store

NA_<ACTION> - non-authed action

for example:
NA_NS_FETCH_CONNECT_SITE_INFO - non authed, no store op: fetch of site info
```

For the current list, this would be renamed to:
```
NA_NS_IS_WPCOM_URL
NA_NS_SUGGEST_DOMAINS
NA_NS_FETCH_CONNECT_SITE_INFO
NA_NS_FETCH_WPCOM_SITE_BY_URL
```

Thoughts welcome

cc @aforcier @oguzkocer @maxme @tonyr59h

mzorz:
Also, consider media handling and rename `UPLOADED_MEDIA` to make it fit this scheme (see https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/533#discussion_r137307617)

tonyr59h:
> Also, consider media handling and rename UPLOADED_MEDIA to make it fit this scheme

I think it's a good idea to rename it, stick to the common convention. I also like the _idea_ of the suggested prefixes, but they're a bit verbose and the info could be conveyed through documentation.

[...]

PR #533, opened 2017-08-14
Add UploadStore

aforcier:
Addresses #519.

Adds an `UploadStore`, which shadows `PostStore`/`MediaStore` upload events and manages a new set of models, `PostUploadModel` and `MediaUploadModel.` The goal is to have a way to track uploading posts that are waiting for their media to upload inside FluxC, offloading a lot of logic from WPAndroid (which was previously receiving events from FluxC and managing multiple maps and lists). It also allows to add a bunch of tests.

**Currently a work in progress.**

### List of remaining tasks:

* [x] Implement a `CANCEL_POST` action for marking 'registered' posts (posts that are queued to be uploaded but are waiting for their media to finish) as cancelled
* [x] Add more convenience methods needed by WPAndroid:
   * [x] `isPending/Failed/CancelledPost()`
   * [x] `getUploadProgressForMedia()`
   * [x] `getUploadErrorForPost()`
   * [x] ~`setUploadErrorForPost()`~ (turns out to be unnecessary, handled by intercepting `UPDATE_MEDIA` and setting media/posts to failed with an error)
   * [x] `getAllRegisteredPosts()`
* [x] ~Move progress reporting/calculating logic to `UploadStore`~ (Not possible anymore since video optimization progress has been added WPAndroid-side)
* [x] Move `UploadError` logic to `UploadStore`
* [x] Refactor upload `onAction()`/`emitChange()` (see below)
* [x] Clear cancelled/failed posts when new media are uploaded to them
* [x] Potentially intercept `UPDATE_MEDIA` actions to reflect changes to media upload state (and add a `MediaError` when `FAILED`)
* [x] Add DB migration when models are finalized
* [x] Eliminate public methods that return `*UploadModel`s

#### Lingering questions:
* [ ] Possibly add `PostStore.getUploadErrorForPost()` as an alias for the same `UploadStore` method, to avoid requiring an `UploadStore` instance from the caller (which is not necessarily an upload manager)

#### Independent items:
* [x] ~Update the documentation + example app~ --> #563
* [x] ~Possibly inject the `UploadStore` into the `MediaStore`/`PostStore` instead of adding it to the constructor (I'd like to explore this separately from this PR)~ --> #559

#### `emitChange()` concurrency
The `UploadStore` receives actions through `onAction()` (using only the `PostAction` and `MediaAction`s relevant to uploads) just before the rest of the stores do. However, we're using `ThreadMode.ASYNC` for `onAction()` in stores, which does not wait for each `onAction()` method to return before calling the next, but rather will call multiple store `onAction()`s on multiple threads.

This means that, when `MediaStore` emits a `OnMediaUploaded` event (for example), we have no guarantee that `UploadStore` has already finished processing the `UPLOADED_MEDIA` event that generated that `OnMediaUploaded` event in the `MediaStore`. So, if we query the `UploadStore` from inside a `OnMediaUploaded` listener, the `UploadStore` might not be up to date yet. Given that I think the `UploadStore` makes sense as a separate store (rather than packing it all up inside `MediaStore`/`PostStore`), I see at least three possible solutions:

1. **Use `ThreadMode.BACKGROUND`**
This guarantees sequential delivery of the same event (so `UploadStore.onAction()` would return before `MediaStore.onAction()` got called). However, it no longer guarantees that actions will be called on a background thread - if the calling thread is not the main thread, a new thread is not used ([`ThreadMode` ref](http://greenrobot.org/eventbus/documentation/delivery-threads-threadmode/)). This could lead to trouble with how we're using FluxC from WPAndroid, and we should not make such a sweeping change lightly.
2. **Emit `OnUploadChanged` events from the `UploadStore`**
This would mean that anything depending on `UploadStore` state (pretty much, the `UploadService` in WPAndroid) would listen to `OnUploadChanged` instead of `OnPostUploaded`/`OnMediaUploaded`, and be guaranteed consistent `UploadStore` state. However, we sometimes need both `UploadStore` and `MediaStore`/`PostStore` to be in a consistent state, so emitting them independently is risky. It's a little unorthodox, but I think what we can do in this case is subscribe `UploadStore` to `OnPostUploaded`/`OnMediaUploaded` events, and 'repeat them' (but this requires some ugly logic to 'register' events as they come in so we can look for them when processing is done). We could also write some kind of middleware for the `Dispatcher`, that will hold `OnUploadChanged` events until a matching `OnMediaUploaded`/`OnPostUploaded` is emitted. This would work, but it's not easy to implement, and I'm not sure separating this logic from the stores is a good idea.
3. **'MITM' `UPLOADED_MEDIA` and `PUSHED_POST` in `UploadStore`**
We would add `UploadAction.UPLOADED_MEDIA` and `UploadAction.PUSHED_POST`, and call those from the network clients instead of the `Post/MediaAction` ones. The `UploadStore` calls the `Post/MediaAction` versions when it's done with its own processing. We don't need to listen for `OnUploadChanged` anymore in this case, because the original `OnXUploaded` events aren't sent until the `UploadStore` is updated, so we can rely on them.

I think 3 is the best option - it's the least error-prone, and the most straightforward to implement. I added this behavior in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/533/commits/2930653b321a10491cb9c7970c702379a2d3fd05.


aforcier:
This PR is now ready to be reviewed. There are three items left on my list: documentation (which I can work on in parallel to/after the review, `UploadStore` injection (separate PR), and aliasing the `UploadStore.getUploadErrorForX()` methods (which I'm not sure is necessary/a good idea, and welcome some feedback on once the WPAndroid PR is being reviewed).

cc @mzorz 😀

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
```diff
@@ -0,0 +1,110 @@
...
+                + "\"title\":\"\",\"caption\":\"Test Caption\",\"description\":\"\",\"alt\":\"\","
+                + "\"icon\":\"\",\"thumbnails\":{\"thumbnail\":\"https:\\/\\/place.com\\/photo"
+                + ".jpg?w=150\",\"medium\":\"https:\\/\\/place.com\\/photo.jpg?w=300\","
+                + "\"large\":\"https:\\/\\/place.com\\/photo.jpg?w=880\"},\"height\":585,"
+                + "\"width\":880,\"exif\":{\"aperture\":\"0\",\"credit\":\"\",\"camera\":\"\","
+                + "\"caption\":\"\",\"created_timestamp\":\"0\",\"copyright\":\"\","
+                + "\"focal_length\":\"0\",\"iso\":\"0\",\"shutter_speed\":\"0\",\"title\":\"\","
+                + "\"orientation\":\"0\",\"keywords\":[]},\"meta\":{\"links\":{\"self\":\"\","
+                + "\"help\":\"\",\"site\":\"\"}}}]}";
+        return buildResponse(request, responseJson, 200);
```
mzorz: I wonder if we can turn this into a full, readable string into the `BuildConfig` file? So it's easier to maintain as well.

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
```diff
@@ -0,0 +1,110 @@
...
+            return buildPostSuccessResponse(request);
+        } else {
+            throw new IllegalStateException("Interceptor was given a request with no mocks - URL: " + requestUrl);
+        }
+    }
+
+    private static Response buildMediaErrorResponse(Request request) {
+        String responseJson = "{\"error\": \"invalid_token\","
+                + "\"message\": \"The OAuth2 token is invalid.\""
+                + "}";
```
mzorz: same for this string, maybe we could put this into the `tests.properties` file

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
mzorz: sorry I meant the `tests.properties` file

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
```diff
@@ -0,0 +1,110 @@
...
+    private static Response buildPostSuccessResponse(Request request) {
+        String responseJson = "{\"ID\":7970,\"type\":\"post\"}";
+        return buildResponse(request, responseJson, 200);
+    }
+
+    private static Response buildResponse(Request request, final String responseJson, int responseCode) {
+        return new Response.Builder()
+                .request(request)
+                .protocol(Protocol.HTTP_1_1)
+                .message("")
```
mzorz: this line here can maybe be removed?

example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedNetworkModule.java
```diff
@@ -107,16 +117,16 @@ public SiteRestClient provideSiteRestClient(Context appContext, Dispatcher dispa
     @Singleton
     @Provides
     public MediaRestClient provideMediaRestClient(Dispatcher dispatcher, Context appContext,
-                                                  @Named("regular") RequestQueue requestQueue,
-                                                  @Named("regular") OkHttpClient okHttpClient,
+                                                  RequestQueue requestQueue,
```
mzorz: Why did the `@Named("regular")` annotation go away for `RequestQueue` here?

mzorz:
Using an old version of FluxC (2 months? approximately) and then running this branch made it crash with this following stack trace. It didn't crash anymore once I deleted the FluxC example app and re-installed.

```
09-05 13:26:26.467 7792-7808/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: MediaAction-UPLOAD_MEDIA
09-05 13:26:26.508 7792-7865/org.wordpress.android.fluxc.example D/WordPress-MEDIA: starting upload for: 1
09-05 13:26:26.739 7792-7881/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: UploadAction-UPLOADED_MEDIA
09-05 13:26:26.760 7792-7865/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: MediaAction-UPLOADED_MEDIA
09-05 13:26:27.061 7792-7881/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: UploadAction-UPLOADED_MEDIA
09-05 13:26:27.103 7792-7864/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: MediaAction-UPLOADED_MEDIA
09-05 13:26:27.469 7792-7808/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: MediaAction-CANCEL_MEDIA_UPLOAD
09-05 13:26:27.470 7792-7881/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: UploadAction-UPLOADED_MEDIA
09-05 13:26:27.512 7792-7865/org.wordpress.android.fluxc.example D/WordPress-MEDIA: Canceled in-progress upload: /IMG_20170131_155806.jpg
09-05 13:26:27.513 7792-7865/org.wordpress.android.fluxc.example D/WordPress-MEDIA: mediaRestClient: removed id: 1 from current uploads, remaining: 0
09-05 13:26:27.528 7792-7865/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: MediaAction-CANCELED_MEDIA_UPLOAD
09-05 13:26:27.551 7792-7862/org.wordpress.android.fluxc.example E/SQLiteDatabase: Error inserting UPLOAD_STATE=0 ERROR_MESSAGE=null _id=1 PROGRESS=0.05622723 ERROR_TYPE=null
                                                                                   android.database.sqlite.SQLiteConstraintException: FOREIGN KEY constraint failed (code 787)
                                                                                       at android.database.sqlite.SQLiteConnection.nativeExecuteForLastInsertedRowId(Native Method)
                                                                                       at android.database.sqlite.SQLiteConnection.executeForLastInsertedRowId(SQLiteConnection.java:782)
                                                                                       at android.database.sqlite.SQLiteSession.executeForLastInsertedRowId(SQLiteSession.java:788)
                                                                                       at android.database.sqlite.SQLiteStatement.executeInsert(SQLiteStatement.java:86)
                                                                                       at android.database.sqlite.SQLiteDatabase.insertWithOnConflict(SQLiteDatabase.java:1474)
                                                                                       at android.database.sqlite.SQLiteDatabase.insert(SQLiteDatabase.java:1343)
                                                                                       at com.yarolegovich.wellsql.InsertQuery.execute(InsertQuery.java:61)
                                                                                       at org.wordpress.android.fluxc.persistence.UploadSqlUtils.insertOrUpdateMedia(UploadSqlUtils.java:32)
                                                                                       at org.wordpress.android.fluxc.store.UploadStore.handleMediaUploaded(UploadStore.java:279)
                                                                                       at org.wordpress.android.fluxc.store.UploadStore.onUploadAction(UploadStore.java:104)
                                                                                       at org.wordpress.android.fluxc.store.UploadStore.onAction(UploadStore.java:95)
                                                                                       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
                                                                                       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
                                                                                       at java.lang.Thread.run(Thread.java:761)
09-05 13:26:27.578 7792-7862/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: MediaAction-UPLOADED_MEDIA
09-05 13:26:27.603 7792-7808/org.wordpress.android.fluxc.example D/WordPress-API: Dispatching action: MediaAction-UPLOAD_MEDIA
                                                                                  
                                                                                  --------- beginning of crash
09-05 13:26:27.646 7792-7863/org.wordpress.android.fluxc.example E/AndroidRuntime: FATAL EXCEPTION: pool-1-thread-2
                                                                                   Process: org.wordpress.android.fluxc.example, PID: 7792
                                                                                       at org.greenrobot.eventbus.EventBus.handleSubscriberException(EventBus.java:505)
                                                                                       at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:487)
                                                                                       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
                                                                                       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
                                                                                       at java.lang.Thread.run(Thread.java:761)
                                                                                    Caused by: org.greenrobot.eventbus.EventBusException: Invoking subscriber failed
                                                                                       at org.greenrobot.eventbus.EventBus.handleSubscriberException(EventBus.java:505)
                                                                                       at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:487)
                                                                                       at org.greenrobot.eventbus.EventBus.postToSubscription(EventBus.java:416)
                                                                                       at org.greenrobot.eventbus.EventBus.postSingleEventForEventType(EventBus.java:397)
                                                                                       at org.greenrobot.eventbus.EventBus.postSingleEvent(EventBus.java:370)
                                                                                       at org.greenrobot.eventbus.EventBus.post(EventBus.java:251)
                                                                                       at org.wordpress.android.fluxc.Dispatcher.emitChange(Dispatcher.java:41)
                                                                                       at org.wordpress.android.fluxc.store.Store.emitChange(Store.java:31)
                                                                                       at org.wordpress.android.fluxc.store.MediaStore.handleMediaUploaded(MediaStore.java:743)
                                                                                       at org.wordpress.android.fluxc.store.MediaStore.onAction(MediaStore.java:374)
                                                                                       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133) 
                                                                                       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607) 
                                                                                       at java.lang.Thread.run(Thread.java:761) 
                                                                                    Caused by: junit.framework.AssertionFailedError
                                                                                       at junit.framework.Assert.fail(Assert.java:48)
                                                                                       at junit.framework.Assert.assertTrue(Assert.java:20)
                                                                                       at junit.framework.Assert.assertTrue(Assert.java:27)
                                                                                       at org.wordpress.android.fluxc.release.ReleaseStack_UploadTest.onMediaUploaded(ReleaseStack_UploadTest.java:374)
                                                                                       at org.greenrobot.eventbus.EventBus.postToSubscription(EventBus.java:416) 
                                                                                       at org.greenrobot.eventbus.EventBus.postSingleEventForEventType(EventBus.java:397) 
                                                                                       at org.greenrobot.eventbus.EventBus.postSingleEvent(EventBus.java:370) 
                                                                                       at org.greenrobot.eventbus.EventBus.post(EventBus.java:251) 
                                                                                       at org.wordpress.android.fluxc.Dispatcher.emitChange(Dispatcher.java:41) 
                                                                                       at org.wordpress.android.fluxc.store.Store.emitChange(Store.java:31) 
                                                                                       at org.wordpress.android.fluxc.store.MediaStore.handleMediaUploaded(MediaStore.java:743) 
                                                                                       at org.wordpress.android.fluxc.store.MediaStore.onAction(MediaStore.java:374) 
                                                                                       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133) 
                                                                                       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607) 
                                                                                       at java.lang.Thread.run(Thread.java:761) 
09-05 13:26:27.652 7792-7863/org.wordpress.android.fluxc.example I/Process: Sending signal. PID: 7792 SIG: 9
09-05 13:26:27.706 1120-9013/? I/ActivityManager: Process org.wordpress.android.fluxc.example (pid 7792) has died
09-05 13:26:27.706 1120-9013/? D/ActivityManager: cleanUpApplicationRecord -- 7792
09-05 13:26:27.709 1120-9013/? W/ActivityManager: Crash of app org.wordpress.android.fluxc.example running instrumentation ComponentInfo{org.wordpress.android.fluxc.example.test/android.test.InstrumentationTestRunner}
09-05 13:26:27.710 1120-9013/? I/ActivityManager: Force stopping org.wordpress.android.fluxc.example appid=10152 user=0: finished inst
09-05 13:26:27.712 7728-7728/? D/AndroidRuntime: Shutting down VM
09-05 13:26:28.087 7891-7891/? D/AndroidRuntime: >>>>>> START com.android.internal.os.RuntimeInit uid 2000 <<<<<<
```

fluxc/build.gradle
```diff
@@ -64,8 +64,8 @@ dependencies {
     }
 
     // Custom WellSql version
-    compile 'org.wordpress:wellsql:1.1.3'
```
mzorz: are there any particular things updated in WellSql that might affect other parts of FluxC other than the Uploads in particular? Do we need to change it?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
```diff
@@ -161,7 +162,7 @@ public void onResponse(PostWPComRestResponse response) {
                         uploadedPost.setLocalSiteId(site.getId());
 
                         RemotePostPayload payload = new RemotePostPayload(uploadedPost, site);
-                        mDispatcher.dispatch(PostActionBuilder.newPushedPostAction(payload));
+                        mDispatcher.dispatch(UploadActionBuilder.newPushedPostAction(payload));
```
mzorz: wondering whether we should use the same verb for both Media and Posts here, and instead of using `newPushedPostAction` maybe use `newUploadPostAction`

fluxc/src/main/java/org/wordpress/android/fluxc/store/UploadStore.java
```diff
@@ -0,0 +1,413 @@
...
+    private void handleClearMedia(ClearMediaPayload payload) {
+        PostUploadModel postUploadModel = UploadSqlUtils.getPostUploadModelForLocalId(payload.post.getId());
+        if (postUploadModel == null) {
+            return;
+        }
+
+        // Remove media from the list of associated media for the post
+        Set<Integer> associatedMediaIdList = postUploadModel.getAssociatedMediaIdSet();
+        for (MediaModel mediaModel : payload.media) {
+            associatedMediaIdList.remove(mediaModel.getId());
```
mzorz: I think what this method does is it removes the media models indicated in the payload from the Post's associated media list. This also might or might not be the full list, but I think the name `CLEAR_MEDIA` better describes an absolute determinate action on an object, rather than a partial or possible action on a _relationship_, which is more likely what is really happening here. What if we rename the `CLEAR_MEDIA` action and all related methods i.e. `handleClearMedia` to `UNLINK_MEDIA` or `UNASSOCIATE_MEDIA`? or maybe the longer `CLEAR_MEDIA_POST_RELATIONSHIP`? Please feel free to suggest any better english possibilities as well

mzorz:
Those are... quite a bunch of tests @aforcier ! 👀 
Looks good, please see some comments that I left and consider whether the crash I had is something that might happen to a real user or not. Will start reviewing the integration shortly.
Good work!

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
aforcier: A non-null `message` is [required by okhttp](https://github.com/square/okhttp/blob/dcabf48ecc91e005d56b97887beb8eaa33ec7c3d/okhttp/src/main/java/okhttp3/Response.java#L440).

mzorz: TIL! Thank you

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
aforcier: Great suggestion! I made some `.json` file resources instead, I think it's the cleanest way to do this: adbcb2d.

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
aforcier: Same as above, done in adbcb2d.

example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedNetworkModule.java
aforcier: In `ReleaseNetworkModule`we supply [two types of `RequestQueue`s](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/f633b7bfc617a886496d4683160ad56b68d13f7d/fluxc/src/main/java/org/wordpress/android/fluxc/module/ReleaseNetworkModule.java#L57-L71), because they rely on an `OkHttpClient` and there are [two types of those](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/976bf784510232718c81788408ccc77c9653bce9/fluxc/src/main/java/org/wordpress/android/fluxc/module/ReleaseOkHttpClientModule.java#L25-L44). One type of `OkHttpClient` (`@Named("custom-ssl")` has some special configs to deal with self-signed SSL self-hosted sites.

In the `MockedNetworkModule`, we have no reason to supply two different `OkHttpClient`s, because we're mocking the server response anyway.

fluxc/build.gradle
aforcier: Short answer: nothing else should be affected.

Long answer: Here's the [full diff of 1.1.3 --> 1.2.0](https://github.com/yarolegovich/wellsql/compare/a5d59fa9411bc7358bd21bc3ddcd3dd913ee3b99...16f3180f24d10e916dd4b7e478375d2851989309) - pretty much just https://github.com/wordpress-mobile/wellsql/pull/8. The only real change to the library is exposing [this `onConfigure()` method](https://github.com/yarolegovich/wellsql/compare/a5d59fa9411bc7358bd21bc3ddcd3dd913ee3b99...16f3180f24d10e916dd4b7e478375d2851989309#diff-a5107fcf7554ba75740e79a27baa3dafR40), to allow us to turn on `FOREIGN KEY`. We're not using `FOREIGN KEY` anywhere else in FluxC, so we shouldn't see any changes from this.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
aforcier: So the history here is that there are two distinct actions for media: `PUSH_MEDIA` and `UPLOAD_MEDIA`.  `PUSH_MEDIA` is for modifying the data of a media item on the server (title, description, whatever), while of course `UPLOAD_MEDIA` is a unique network action (for FluxC), sending a file object to the server.

Probably we should've gone with `PUSH_MEDIA` for uploading media, and `PUSH_MEDIA_INFO` or something instead of the current `PUSH_MEDIA`. I'm not sure if this would make things clearer or add more confusion at this stage though.

In this sense, it's actually `UPLOADED_MEDIA` that's breaking the 'rules', and not `PUSHED_POST.

If this makes sense to you, I'd suggest maybe opening an issue for a suggested renaming of media actions, and addressing that separately after some discussion. What do you think?

example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedNetworkModule.java
mzorz: Understood, thanks

fluxc/build.gradle
mzorz: Good to know! thank you for pointing out exactly what the changes were, so we make sure it won't affect existing systems 👍

fluxc/src/main/java/org/wordpress/android/fluxc/store/UploadStore.java
aforcier: I was afraid the naming might be unclear but forgot to circle back to it at the end.

I don't have an easy answer to this one. This action not only removes the media from the post's associated media list, but also deletes the media models themselves. In a way, `UNASSOCIATE_MEDIA`, `UNLINK_MEDIA` and `CLEAR_MEDIA_POST_RELATIONSHIP` aren't telling the whole story. On the other hand, `CLEAR_MEDIA` is a bit vague and makes you look up what it actually does.

Maybe `CLEAR_MEDIA_FOR_POST` could be an acceptable compromise?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
mzorz: Seems reasonable - open issue here https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/503, added comment https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/503#issuecomment-327529589

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
mzorz: 👍 thank you!

aforcier:
> Using an old version of FluxC (2 months? approximately) and then running ReleaseStack_UploadTest fails in testCancelImageUpload with this following stack trace. It didn't crash anymore once I deleted the FluxC example app and re-installed.

Looking at the trace, it looks like the crash was a failing assert in the test (a progress of `> 0` was expected and was not found, I believe because there was no `MediaUploadModel` corresponding to that id, as there was a `FOREIGN KEY` violation above, meaning the corresponding `MediaModel` didn't exist). I'm not sure this had anything to do with an older FluxC version, I wonder if it's some kind of rare race condition in the test. I'm going to have to think about this one a little more.

fluxc/src/main/java/org/wordpress/android/fluxc/store/UploadStore.java
mzorz: Yes please let's use `CLEAR_MEDIA_FOR_POST` 👍

aforcier:
> Those are... quite a bunch of tests @aforcier ! 👀

Consider them a <i>test</i>ament to the complexity async really brought to the table.

I hope they didn't test your patience too much.

Okay, I'll stop 😀

fluxc/src/main/java/org/wordpress/android/fluxc/store/UploadStore.java
aforcier: :+1: done in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/533/commits/09b1856008c4ad3d92b48cc3de238aecfc7c2e73.

aforcier:
Merged `develop` in to resolve conflicts, this changed `WellSqlConfig` db migration - heads up that clearing data will be needed.

aforcier:
Okay, I was able to reproduce https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/533#issuecomment-327235683 once.

It looks like what's happening is that `MediaStore` receives and handles a `CANCEL_MEDIA_UPLOAD` action before `UploadStore` has had a chance to finish dealing with the last `UPLOADED_MEDIA`, so it tries to update the `MediaUploadModel` with the last progress. But, the `MediaModel` it's linked to was already deleted by the `CANCEL_MEDIA_UPLOAD` handler.

In this case I think `FOREIGN KEY` is doing what it's supposed to - preventing the creation of a `MediaUploadModel` when the linked `MediaModel` doesn't exist. But the check [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/533/files#diff-651a13752742742351ee12e13e000df3R371) is a bit dumb and doesn't check first a `MediaUploadModel` even exists for the `MediaModel` it received an action for - it's not necessary that there's a non-zero progress at this point.

I think we can update the logic of that check and move on - what are your thoughts @mzorz?

mzorz:
> I think we can update the logic of that check and move on

Nice find @aforcier ! Totally, let's update that check 👍

aforcier:
Upon further research, it looks like some other strange things are doing on in `testCancelImageUpload()`, possibly pointing to an issue with the actual code - I'll look into this some more.

aforcier:
Updated `UploadStore.getUploadErrorForPost()` to check for media with errors even if the post doesn't have a registered `PostUploadModel` (7d55e58).

This handles the case where we have a post with failed media, but never tried to actually upload the post: https://github.com/wordpress-mobile/WordPress-Android/pull/6613#issuecomment-330767263.

maxme:
> Update the documentation

I would also add "Update the example app" because it's also used for documentation. New issue or new PR?

aforcier:
> I would also add "Update the example app" because it's also used for documentation. New issue or new PR?

Sounds good - I'll open an issue for both and work on a new PR for them.

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
```diff
@@ -0,0 +1,120 @@
...
+import okhttp3.Interceptor;
+import okhttp3.MediaType;
+import okhttp3.Protocol;
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+import okio.BufferedSource;
+import okio.Okio;
+
+class FailingInterceptor implements Interceptor {
```
maxme: We should  rename this to a generic name.

maxme:
💯 for mocked upload tests.

example/src/androidTest/java/org/wordpress/android/fluxc/module/FailingInterceptor.java
aforcier: Agreed, done in ad07c58 (it's now `ResponseMockingInterceptor`).

aforcier:
Found a small bug: if a media attached to a post was cancelled (manually, by the user) and the `delete` flag was on, the associated `PostUploadModel` wasn't being marked as `CANCELLED`.

Fixed in dafb310.

maxme:
PR looks good, it would be great to do something with the `testCancelImageUpload` tests (new issue?).

mzorz:
:shipit:

aforcier:
> PR looks good, it would be great to do something with the testCancelImageUpload tests (new issue?).

Agreed - opened https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/564.

maxme:
:shipit:

PR #539, opened 2017-08-25
Share Android Studio configuration files in the repo

maxme:


oguzkocer:
@maxme I am happy to review this, but I am unfamiliar with the topic. Could you add some testing steps, link to relevant docs etc to the PR? If you were working with someone else on this, that's fine too.

.idea/codeStyleSettings.xml
```diff
@@ -0,0 +1,231 @@
...
+            <emptyLine />
+            <package name="javax" withSubpackages="true" static="false" />
+            <emptyLine />
+            <package name="" withSubpackages="true" static="false" />
+            <emptyLine />
+            <package name="" withSubpackages="true" static="true" />
+            <emptyLine />
+          </value>
+        </option>
+        <option name="RIGHT_MARGIN" value="100" />
```
aforcier: Why `100` when our character limit is `120`?

.idea/codeStyleSettings.xml
```diff
@@ -0,0 +1,231 @@
...
+                  </match>
+                  <order>BY_NAME</order>
+                </rule>
+              </section>
+            </rules>
+          </arrangement>
+        </codeStyleSettings>
+      </value>
+    </option>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default (1)" />
```
aforcier: In a clean clone of the project, these settings weren't automatically used - my local default config was used instead (evidenced by the line limit being `120` and not `100`).

I think we need to use

```xml
<option name="USE_PER_PROJECT_SETTINGS" value="true" />
```

instead of `PREFERRED_PROJECT_CODE_STYLE`, to force use of this project file.

maxme:
Fixed comments in 41e25f9 

Also checked what happen if the required plugin is not installed. AS warns on startup:

screen shot 2017-09-01 at 13 35 51

It doesn't prevent compilation or anything else, but the warning comes back every time the user open the project (annoying enough to make the user install it).

aforcier:
> It doesn't prevent compilation or anything else, but the warning comes back every time the user open the project (annoying enough to make the user install it).

That seems annoying enough 👍

aforcier:
Works great! :shipit:

PR #540, opened 2017-08-25
Feature/add social login

theck13:
Add a method for WordPress.com social login, `pushSocialLogin`.  Currently, the only unique error response contains `no_matching_account` which implies the email address passed as a parameter does not match any WordPress.com account.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
@@ -5,9 +5,11 @@ import org.wordpress.android.fluxc.annotations.ActionEnum;
 import org.wordpress.android.fluxc.annotations.action.IAction;
 import org.wordpress.android.fluxc.model.AccountModel;
 import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.AccountPushSettingsResponsePayload;
+import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.AccountPushSocialResponsePayload;
 import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.AccountRestPayload;
 import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.IsAvailableResponsePayload;
 import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.NewAccountResponsePayload;
+import org.wordpress.android.fluxc.store.AccountStore;
 import org.wordpress.android.fluxc.store.AccountStore.NewAccountPayload;
 import org.wordpress.android.fluxc.store.AccountStore.PushAccountSettingsPayload;
 import org.wordpress.android.fluxc.store.AccountStore.UpdateTokenPayload;
@@ -23,6 +25,8 @@ public enum AccountAction implements IAction {
     SEND_VERIFICATION_EMAIL, // request verification email for unverified accounts
     @Action(payloadType = PushAccountSettingsPayload.class)
     PUSH_SETTINGS,          // request saving Account Settings remotely
+    @Action(payloadType = AccountStore.PushSocialLoginPayload.class)
+    PUSH_SOCIAL,            // request social login remotely
     @Action(payloadType = NewAccountPayload.class)
     CREATE_NEW_ACCOUNT,     // create a new account (can be used to validate the account before creating it)
     @Action(payloadType = String.class)
@@ -43,6 +47,8 @@ public enum AccountAction implements IAction {
     SENT_VERIFICATION_EMAIL,
     @Action(payloadType = AccountPushSettingsResponsePayload.class)
     PUSHED_SETTINGS,        // response received from Account Settings post
+    @Action(payloadType = AccountPushSocialResponsePayload.class)
+    PUSHED_SOCIAL,          // response received from social login post
     @Action(payloadType = NewAccountResponsePayload.class)
     CREATED_NEW_ACCOUNT,    // create a new account response
     @Action(payloadType = IsAvailableResponsePayload.class)
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
@@ -12,7 +12,6 @@ import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.
 import org.wordpress.android.fluxc.store.AccountStore.NewAccountPayload;
 import org.wordpress.android.fluxc.store.AccountStore.PushAccountSettingsPayload;
 import org.wordpress.android.fluxc.store.AccountStore.PushSocialLoginPayload;
-import org.wordpress.android.fluxc.store.AccountStore.PushSocialSignupPayload;
 import org.wordpress.android.fluxc.store.AccountStore.UpdateTokenPayload;
 
 @ActionEnum
@@ -28,8 +27,6 @@ public enum AccountAction implements IAction {
     PUSH_SETTINGS,          // request saving Account Settings remotely
     @Action(payloadType = PushSocialLoginPayload.class)
     PUSH_SOCIAL_LOGIN,      // request social login remotely
-    @Action(payloadType = PushSocialSignupPayload.class)
-    PUSH_SOCIAL_SIGNUP,     // request social signup remotely
     @Action(payloadType = NewAccountPayload.class)
     CREATE_NEW_ACCOUNT,     // create a new account (can be used to validate the account before creating it)
     @Action(payloadType = String.class)
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
@@ -27,7 +27,7 @@ public enum AccountAction implements IAction {
     @Action(payloadType = PushAccountSettingsPayload.class)
     PUSH_SETTINGS,          // request saving Account Settings remotely
     @Action(payloadType = PushSocialLoginPayload.class)
-    PUSH_SOCIAL,            // request social login remotely
+    PUSH_SOCIAL_LOGIN,      // request social login remotely
     @Action(payloadType = PushSocialSignupPayload.class)
     PUSH_SOCIAL_SIGNUP,     // request social signup remotely
     @Action(payloadType = NewAccountPayload.class)
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
@@ -9,9 +9,9 @@ import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.
 import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.AccountRestPayload;
 import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.IsAvailableResponsePayload;
 import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.NewAccountResponsePayload;
-import org.wordpress.android.fluxc.store.AccountStore;
 import org.wordpress.android.fluxc.store.AccountStore.NewAccountPayload;
 import org.wordpress.android.fluxc.store.AccountStore.PushAccountSettingsPayload;
+import org.wordpress.android.fluxc.store.AccountStore.PushSocialLoginPayload;
 import org.wordpress.android.fluxc.store.AccountStore.UpdateTokenPayload;
 
 @ActionEnum
@@ -25,7 +25,7 @@ public enum AccountAction implements IAction {
     SEND_VERIFICATION_EMAIL, // request verification email for unverified accounts
     @Action(payloadType = PushAccountSettingsPayload.class)
     PUSH_SETTINGS,          // request saving Account Settings remotely
-    @Action(payloadType = AccountStore.PushSocialLoginPayload.class)
+    @Action(payloadType = PushSocialLoginPayload.class)
     PUSH_SOCIAL,            // request social login remotely
     @Action(payloadType = NewAccountPayload.class)
     CREATE_NEW_ACCOUNT,     // create a new account (can be used to validate the account before creating it)
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
@@ -12,6 +12,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient.
 import org.wordpress.android.fluxc.store.AccountStore.NewAccountPayload;
 import org.wordpress.android.fluxc.store.AccountStore.PushAccountSettingsPayload;
 import org.wordpress.android.fluxc.store.AccountStore.PushSocialLoginPayload;
+import org.wordpress.android.fluxc.store.AccountStore.PushSocialSignupPayload;
 import org.wordpress.android.fluxc.store.AccountStore.UpdateTokenPayload;
 
 @ActionEnum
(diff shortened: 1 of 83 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -188,6 +195,38 @@ public void onErrorResponse(@NonNull BaseNetworkError error) {
...
+     * No HTTP POST call is made if the given parameter map is null or contains no entries.
+     */
+    public void pushSocialLogin(Map<String, Object> body) {
+        if (body == null || body.isEmpty()) return;
+        String url = WPCOMREST.users.social.new_.getUrlV1_1();
+        // Map is used in onResponse since the API response format differs depending on the request.
+        add(WPComGsonRequest.buildPostRequest(url, body, Map.class,
+                new Listener<Map<String, Object>>() {
+                    @Override
+                    public void onResponse(Map<String, Object> response) {
```
maxme: Can you replace `Map<String, Object>` by a less generic `AccountSocialResponse` here? You can get inspiration for `AccountSocialResponse` from [`AccountBoolResponse`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/0be498522a7075ed32bc635664f27e3b1325770e/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountBoolResponse.java)

Would be great to also have a less generic `AccountPushSocialResponsePayload` (for example with `bearerToken` and `username` string fields).

theck13: > Can you replace `Map<String, Object>` by a less generic `AccountSocialResponse` here? You can get inspiration for `AccountSocialResponse` from [`AccountBoolResponse`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/0be498522a7075ed32bc635664f27e3b1325770e/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountBoolResponse.java)

I update the `AccountSocialResponse` class in 9262669a99c5c252e1106111d3943a5b2982417c.

> Would be great to also have a less generic `AccountPushSocialResponsePayload` (for example with `bearerToken` and `username` string fields).

I updated the `AccountPushSocialResponsePayload` class in 97cedd3f89211486a2b5504125c78e99ef71a539.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -55,6 +55,39 @@ public AccountPushSettingsResponsePayload(BaseNetworkError error) {
         public Map<String, Object> settings;
     }
 
+    public static class AccountPushSocialResponsePayload extends Payload {
```
aforcier: Could you use the public variable form we're been using for other `Payload`s here and explicitly set the values (see how `IsAvailableResponsePayload` below is structured/used for example)?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountSocialResponse.java
```diff
@@ -1,8 +1,8 @@
 package org.wordpress.android.fluxc.network.rest.wpcom.account;
 
+import org.json.JSONObject;
 import org.wordpress.android.fluxc.network.Response;
 
 public class AccountSocialResponse implements Response {
-    public String bearer_token;
-    public String username;
+    public JSONObject json;
```
aforcier: The previous form with `bearer_token` and `username` instead of `json` was consistent with other `...Response` classes - was there a reason to switch to this generic form?

fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
```diff
@@ -23,6 +25,8 @@
     SEND_VERIFICATION_EMAIL, // request verification email for unverified accounts
     @Action(payloadType = PushAccountSettingsPayload.class)
     PUSH_SETTINGS,          // request saving Account Settings remotely
+    @Action(payloadType = AccountStore.PushSocialLoginPayload.class)
```
aforcier: Bit of a nit, but could you use the fully qualified import for this so it's consistent with the rest of the payload declarations?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
theck13: Yep!  I updated the variable (3a20337557ca4f0ae9e06861d9603ccd9407dc35) and usage (5cffe28b3a9cd2b46aa3b798c35f0f1988239c05).

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountSocialResponse.java
theck13: The valid response can be either of these two forms.

```
{
    "bearer_token": "smarterthantheaveragebearertoken",
    "username": "yogi"
}
```

```
{
    "error": "no_matching_account",
    "message": "We did not find a WordPress.com account with a matching email address."
}
```

In the second case, assigning the `error` value to `payload.bearer_token` and the `message` value to `payload.username` seemed misleading.  That being said, I'm open to a different implementation if you have one in mind.

fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
theck13: D'oh.  I fixed that in eb6a160ed779b215254177d7aeb295c671f8a4b8.  I feel the same way about fully qualified imports.  Thanks for pointing it out!

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountSocialResponse.java
aforcier: @theck13 it shouldn't be necessary to assign the error/message values to `bearer_token` and `username`. The error case should receive special handling - you can set up a `AccountSocialError` and `AccountSocialErrorType`, and translate the error response from the server into those.

Here's one example: https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5fcc553239e28c3d4fd48343070c508bcdf25678/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/auth/Authenticator.java#L222-L226

(You will also need to overwrite the `error` field for the `AccountPushSocialResponsePayload` to be an `AccountSocialError` ([example](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5fcc553239e28c3d4fd48343070c508bcdf25678/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/auth/Authenticator.java#L70)).)

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
```diff
@@ -70,6 +71,12 @@ public PushAccountSettingsPayload() {
         }
     }
 
+    public static class PushSocialLoginPayload extends Payload {
+        public Map<String, Object> params;
```
aforcier: Looking at the endpoint for this, it looks like it takes these params in the request:

```
'client_id',
'client_secret',
'service',
'signup_flow_name',
'access_token',
'id_token'
```

Why not use the fields explicitly to the `PushSocialLoginPayload`? The `Map<String, Object>` doesn't give any hints as to what params are needed to build this request/action. See the [NewAccountPayload](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/3c2dc961e39b2aba7fe98b37597b640fa7d78203/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java#L73-L85) for an example.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountSocialResponse.java
theck13: I update the `AccountSocialResponse` variables in bcf68db8cf24de72f5bfee549eaecf65043a7f20.  I added the `AccountSocialError` class and the `AccountSocialErrorType` enum in 4b3dc320717cfa62771825110153dfe95b3f24b4.

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
theck13: The endpoint is undergoing updates simultaneously.  Those parameters are not required for _login_, but they are for _signup_.  I updated the social login POST method in 064b52c5ee5b61cd0f16b2d833ae1311508f66c6 and added the social signup POST method in c62d14f77196dd49f22280c19169b2a52659a314 to account for the differences in parameters.

theck13:
[Travis is failing](https://travis-ci.org/wordpress-mobile/WordPress-FluxC-Android/builds/279009152#L2616-L2617) with the following errors.
```
:fluxc:checkstyle[ant:checkstyle] [ERROR] /home/travis/build/wordpress-mobile/WordPress-FluxC-Android/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java:61: Payload should not be used as a raw type [RegexpSingleline]
[ant:checkstyle] [ERROR] /home/travis/build/wordpress-mobile/WordPress-FluxC-Android/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java:78: Payload should not be used as a raw type [RegexpSingleline]
```

Is this a new style check?  I see multiple instances when a payload is used as a raw type including previous commits in this pull request which passed (e.g. [createPushSocialLogin](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/540/commits/064b52c5ee5b61cd0f16b2d833ae1311508f66c6#diff-21a42e5cb9fc1d9531f135b868bc3e3aR550) and [handlePushSocialCompleted](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/540/commits/354b60b05b3f3504e158dd1f0a2baa9edaf009dc#diff-21a42e5cb9fc1d9531f135b868bc3e3aL502)).

aforcier:
> Is this a new style check?  I see multiple instances when a payload is used as a raw type including previous commits in this pull request which passed (e.g. [createPushSocialLogin](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/540/commits/064b52c5ee5b61cd0f16b2d833ae1311508f66c6#diff-21a42e5cb9fc1d9531f135b868bc3e3aR550) and [handlePushSocialCompleted](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/540/commits/354b60b05b3f3504e158dd1f0a2baa9edaf009dc#diff-21a42e5cb9fc1d9531f135b868bc3e3aL502)).

@theck13 Yes that's new (it came in with your latest merge from `develop`) - here's the PR that added it: #548.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/FormRequest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.wordpress.android.fluxc.network.BaseRequest;
+import org.wordpress.android.fluxc.network.rest.wpcom.account.AccountSocialResponse;
+import org.wordpress.android.util.AppLog;
+import org.wordpress.android.util.AppLog.T;
+
+import java.util.Map;
+
+public class FormRequest extends BaseRequest<AccountSocialResponse> {
```
aforcier: Since this only works for `AccountSocialResponse` as written, could we call this class `AccountSocialRequest` and move it to `network/rest/wpcom/account` for now?

I think we could instead make `FormRequest` more generic with some extra work, but I'm not sure at this stage how much of it would ever be reused. I'll think about this one a bit more and follow up with you.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
```diff
@@ -46,12 +47,12 @@ public void onParseError(OnUnexpectedError event) {
         };
     }
 
-    protected Request add(WPComGsonRequest request) {
+    protected Request add(BaseRequest request) {
```
aforcier: I'm not sure about generalizing this to `BaseRequest`, the more restrictive `WPComGsonRequest` makes it a bit clearer what this client is expecting.

How about instead of this adding a pair of new methods:

```java
protected Request addUnauthedRequest(FormRequest request) {
  // Add "locale=xx_XX" query parameter to all request by default
  return addUnauthedRequest(request, true);
}

protected Request addUnauthedRequest(FormRequest request, boolean addLocaleParameter) {
  if (addLocaleParameter) {
    request.addQueryParameter("locale", LanguageUtils.getPatchedCurrentDeviceLanguage(mAppContext));
  }
  return mRequestQueue.add(setRequestAuthParams(request, false));
}
```

(Or possibly `AccountSocialRequest` instead of `FormRequest`, whatever it ends up being.)

Since for now we're only using this for unauthed requests, I think this would be enough, and we can add a pair of `Request add(FormRequest request)` methods in the future if necessary. What do you think?

aforcier: Edited the code block to remove usage of `setRequestAuthParams()`, which didn't make sense (thanks @theck13).

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/FormRequest.java
theck13: I refactored the class name in a2cdf3cd9a051226c3c5b648bbf69faee20c7d57.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
theck13: I reverted and added the methods in bd77c51086e2b6102d9bb85830211a7a6cc80c4f and f26bd0ca5a5ba386e9ef50750513517ba8c4c010.

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
```diff
@@ -70,6 +75,17 @@ public PushAccountSettingsPayload() {
...
     }
 
+    public static class PushSocialLoginPayload extends Payload<BaseNetworkError> {
+        public String idToken;
+        public String service;
+        public String version;
+        public PushSocialLoginPayload(@NonNull String idToken, @NonNull String service, @NonNull String version) {
+            this.idToken = idToken;
+            this.service = service;
+            this.version = version;
```
aforcier: I'm not sure about passing the endpoint version through the payload. Every other endpoint version is not configurable from outside FluxC, and I think we should stay consistent with that (maybe defining a constant in `AccountRestClient` instead). Especially since we don't actually support different versions/responses for the same endpoint currently in FluxC.

aforcier:
I think we should move the newly-renamed `AccountSocialRequest` one more package deep into `fluxc.network.rest.account.AccountSocialRequest` since it can't really be used for anything else in its current form.

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
theck13: I remove the parameter (b657be47c88b199390459a40ad0ed91bf92931d4 and e86d29b070fee3a8b5e1d5bf7350587dce199ef3) and added the constant 2c3a7f25b1bb0121472ce9e258513375cf343bd4.

theck13:
I moved the class into the `account` package in a0987f49ecf8f5762d8e1e09aa2f5bf3d80e20cd.

aforcier:
:shipit: 🍾 👏

PR #541, opened 2017-08-25
Issue/update checkstyle readme

theck13:
Update the ***Setting up Checkstyle*** section to include details about installing the CheckStyle-IDEA plugin through Android Studio and add a note about targeting Checkstyle 8.0 or lower in the ***Compatibility*** subsection of the `README.md` file.


maxme:
Since we recently merged https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/539, @theck13 can you update the README to say something like: 

> [When you import the project in AS it should be set up automatically, if not, ] you can configure the plugin here, blahblahblah

PR #542, opened 2017-08-29
ThemeStore fetch actions

tonyr59h:
First PR for the feature branch. This one adds `ThemeStore` with support for all the fetch actions needed to browse the themes list in WPAndroid. The example app is also updated to make testing a bit easier (hopefully).

Note, I haven't included WellSql DB version migration so you'll have to test with a fresh build to avoid SQL errors. I anticipate the `ThemeModel` will have a few changes before merging into `develop` so I decided to hold off on the migration till the end.

The response objects are overly robust as I just copied everything the API sends us. I'll trim the unnecessary fields as part of the cleanup before final merge.


tonyr59h:
Forgot to cc @oguzkocer

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java
```diff
@@ -0,0 +1,136 @@
...
+                        FetchedThemesPayload payload = new FetchedThemesPayload(null);
+                        payload.themes = ThemeUtils.createThemeListFromWPComResponse(response);
+                        mDispatcher.dispatch(ThemeActionBuilder.newFetchedWpThemesAction(payload));
+                    }
+                }, new BaseRequest.BaseErrorListener() {
+                    @Override
+                    public void onErrorResponse(@NonNull BaseRequest.BaseNetworkError error) {
+                        AppLog.d(AppLog.T.API, "Received error response to WP.com themes fetch request.");
+                        FetchThemesError themeError =
+                                new FetchThemesError(ThemeErrorType.GENERIC_ERROR, "Error fetching WP.com themes");
```
oguzkocer: Are there any specific errors we could get here? Also, why are we setting a custom message here instead of carrying over the network error?

fluxc/src/main/java/org/wordpress/android/fluxc/utils/ThemeUtils.java
```diff
@@ -0,0 +1,72 @@
...
+import org.wordpress.android.fluxc.model.ThemeModel;
+import org.wordpress.android.fluxc.network.rest.wpcom.theme.ThemeJetpackResponse.MultipleJetpackThemesResponse;
+import org.wordpress.android.fluxc.network.rest.wpcom.theme.ThemeJetpackResponse;
+import org.wordpress.android.fluxc.network.rest.wpcom.theme.ThemeWPComResponse;
+import org.wordpress.android.fluxc.network.rest.wpcom.theme.ThemeWPComResponse.MultipleWPComThemesResponse;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ThemeUtils {
```
oguzkocer: Unless I am mistaken, we prefer to keep these methods within the RestClient as they are not shared with any other class. It's up to you, but I'd prefer to be consistent on this.

fluxc/src/main/java/org/wordpress/android/fluxc/utils/ThemeUtils.java
```diff
@@ -0,0 +1,72 @@
...
+        theme.setAuthorUrl(response.author_uri);
+        theme.setVersion(response.version);
+        theme.setActive(response.active);
+        theme.setAutoUpdate(response.autoupdate);
+        theme.setAutoUpdateTranslation(response.autoupdate_translation);
+
+        // the screenshot field in Jetpack responses does not contain a protocol so we'll prepend 'https'
+        String screenshotUrl = response.screenshot;
+        if (!TextUtils.isEmpty(screenshotUrl) && screenshotUrl.startsWith("//")) {
+            screenshotUrl = screenshotUrl + "https";
```
oguzkocer: This looks like append and not prepend.

fluxc/src/main/java/org/wordpress/android/fluxc/utils/ThemeUtils.java
```diff
@@ -0,0 +1,72 @@
...
+        theme.setAuthorName(response.author);
+        theme.setAuthorUrl(response.author_uri);
+        theme.setVersion(response.version);
+        theme.setActive(response.active);
+        theme.setAutoUpdate(response.autoupdate);
+        theme.setAutoUpdateTranslation(response.autoupdate_translation);
+
+        // the screenshot field in Jetpack responses does not contain a protocol so we'll prepend 'https'
+        String screenshotUrl = response.screenshot;
+        if (!TextUtils.isEmpty(screenshotUrl) && screenshotUrl.startsWith("//")) {
```
oguzkocer: If we are going to check for `startsWith("//")`, we probably don't need to check if the string is empty. A null-check might be better, but I realize this is kind of a nitpick :)

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java
```diff
@@ -0,0 +1,136 @@
...
+                    public void onResponse(MultipleJetpackThemesResponse response) {
+                        AppLog.d(AppLog.T.API, "Received response to Jetpack installed themes fetch request.");
+                        List<ThemeModel> themes = ThemeUtils.createThemeListFromJetpackResponse(response);
+                        FetchedThemesPayload payload = new FetchedThemesPayload(site, themes);
+                        mDispatcher.dispatch(ThemeActionBuilder.newFetchedInstalledThemesAction(payload));
+                    }
+                }, new BaseRequest.BaseErrorListener() {
+                    @Override
+                    public void onErrorResponse(@NonNull BaseRequest.BaseNetworkError error) {
+                        FetchThemesError themeError = new FetchThemesError(ThemeErrorType.GENERIC_ERROR,
```
oguzkocer: I have similar concerns about how we handle the error here.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java
```diff
@@ -0,0 +1,136 @@
...
+                                "Error fetching installed themes for Jetpack site.");
+                        FetchedThemesPayload payload = new FetchedThemesPayload(themeError);
+                        mDispatcher.dispatch(ThemeActionBuilder.newFetchedInstalledThemesAction(payload));
+                        AppLog.d(AppLog.T.API, "Received error response to Jetpack installed themes fetch request.");
```
oguzkocer: Should we use `AppLog.e` in these situations?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ThemeSqlUtils.java
```diff
@@ -0,0 +1,77 @@
...
+    public static void removeThemes(@NonNull SiteModel site) {
+        WellSql.delete(ThemeModel.class)
+                .where()
+                .equals(ThemeModelTable.LOCAL_SITE_ID, site.getSiteId())
+                .endWhere().execute();
+    }
+
+    public static void removeThemeWithId(@NonNull ThemeModel theme) {
+        String themeId = theme.getThemeId();
+        if (!TextUtils.isEmpty(themeId)) {
```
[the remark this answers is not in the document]
tonyr59h: Mostly a failsafe because happen. This method is being removed.

fluxc/src/main/java/org/wordpress/android/fluxc/utils/ThemeUtils.java
tonyr59h: D'oh!

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java
tonyr59h: According to [the docs](https://developer.wordpress.com/docs/api/1.1/get/themes/) no errors returned from this endpoint.

fluxc/src/main/java/org/wordpress/android/fluxc/action/ThemeAction.java
```diff
@@ -0,0 +1,31 @@
...
+import org.wordpress.android.fluxc.annotations.action.IAction;
+import org.wordpress.android.fluxc.model.SiteModel;
+import org.wordpress.android.fluxc.store.ThemeStore.FetchedCurrentThemePayload;
+import org.wordpress.android.fluxc.store.ThemeStore.FetchedThemesPayload;
+
+@ActionEnum
+public enum ThemeAction implements IAction {
+    // Remote actions
+    @Action
+    FETCH_WP_THEMES,
```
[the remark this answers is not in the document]
tonyr59h: This action will fetch from the `/themes` endpoint, unassociated with any site. It gets the list of WP.com themes so it seems appropriate to rename.

tonyr59h:
> this is a massive PR, one that we should avoid opening as much as possible.

Yeah, it turned out much bigger than I hoped even after chipping it down to only the fetch calls. Maybe adding an empty store for the first PR would have helped. On the plus side the subsequent PR's will be very manageable in comparison ;)

Thanks for taking the time for such a big review! Everything's been addressed, ready for round 2 😄

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
```diff
@@ -0,0 +1,190 @@
...
+    public void onRegister() {
+        AppLog.d(AppLog.T.API, "ThemeStore onRegister");
+    }
+
+    public List<ThemeModel> getWpThemes() {
+        return ThemeSqlUtils.getThemesForSite(null);
+    }
+
+    public List<ThemeModel> getThemesForSite(@NonNull SiteModel site) {
+        return ThemeSqlUtils.getThemesForSite(site);
```
oguzkocer: Afaik the `@NonNull` annotation is a compile time check (just a warning), so there is no guarantee that a client of this library won't call this method with a `null` site which currently returns all the wp.com themes. My comments about `ThemeSqlUtils.removeThemes(@Nullable SiteModel site) ` applies to `ThemeSqlUtils.getThemesForSite(@Nullable SiteModel site)` as well.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ThemeSqlUtils.java
```diff
@@ -0,0 +1,67 @@
...
+        for (ThemeModel theme : themes) {
+            theme.setLocalSiteId(site.getSiteId());
+        }
+
+        WellSql.insert(themes).execute();
+
+        return themes.size();
+    }
+
+    public static void removeThemes(@Nullable SiteModel site) {
```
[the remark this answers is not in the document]
tonyr59h: > especially since it's a public method (which it shouldn't be)?

I think `removeThemes` needs to be exposed so they can be deleted when self-hosted sites are removed in WPAndroid. That's the case I was thinking of at least.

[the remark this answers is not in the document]
oguzkocer: That's not what I am saying though. I am saying if the `removeThemes` is called with `null` parameter unintentionally because the `SiteModel` is `null`, it deletes wp.com themes. That's not what I'd expect as the consumer of the library.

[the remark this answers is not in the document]
oguzkocer: Also please note that I changed the proposed method `removeThemes` accessibility to `public`. It was not my intention to propose it as private since you wouldn't be able to use it at all in that case.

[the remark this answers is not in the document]
tonyr59h: That makes sense. That's not an issue anymore with the changes with you suggested though, right? If a `null` site is passed it'll end up throwing an exception here `removeThemes(site.getId());`.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ThemeSqlUtils.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class ThemeSqlUtils {
+    public static void insertOrUpdateTheme(@NonNull ThemeModel theme) {
+        List<ThemeModel> existing = WellSql.select(ThemeModel.class)
+                .where()
+                .equals(ThemeModelTable.THEME_ID, theme.getThemeId())
+                .endWhere().getAsModel();
+
+        if (existing.isEmpty()) {
+            WellSql.insert(theme).asSingleTransaction(true).execute();
```
oguzkocer: Since we are inserting a single object, I don't think we need to do it as a single transaction. I could be wrong about this, if that's the case, we should update the `PluginSqlUtils` as well. However, it doesn't seem like we normally use transactions for single objects on other classes either.

Issue #547, opened 2017-08-30
Kotlin Support


aforcier:
This issue tracks progress of adding Kotlin support for FluxC. There are a few blocking items to address before we can add it:

### 1. Payload `error` field overwriting

Kotlin is not happy with our overwriting the `error` field in payloads to use custom errors (see, e.g., [AuthenticateErrorPayload](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/3c2dc961e39b2aba7fe98b37597b640fa7d78203/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java#L57-L65)). Any Kotlin code we wrote that needed to access a payload with an overwritten type would cause a compiler error. This is because the underlying `BaseNetworkError` is exposed to Kotlin through a `getError()` method, so there's a conflict between the base `BaseNetworkError getError()` and the overwriting, custom error.

I think the least painful way to solve this is to stop overwriting the error, and instead let `Payload` accept a generic parameter for the `error` type. We won't be able to default to `BaseNetworkError` anymore (we didn't want other errors to inherit from this), so any `Payload` subclasses not using a custom error type will need to use `extends Payload<BaseNetworkError>`. This is a bit flimsy since there's nothing forcing `Payload` subclasses to parameterize `Payload`, giving those subclasses a useless `error` type. So, we'll enforce parameterizing `Payload` subclasses using a Checkstyle rule.

I'm confident we can find a better fix in the future, especially if we move the `Payload` and its subclasses to Kotlin - but for now this is the most painless way to unblock `Payload` usage in Kotlin code.

#### Tasks:
* [x] Generify `Payload`, and enforce its parameterization when subclassed (PR: #548)

### 2. Linting and Checkstyle

Kotlin for Android is still quite new, and we don't have the same range of options available as we do for Java. [`ktlint`](https://github.com/shyiko/ktlint) looks like a great option, but there are a few problems:

<ol>
<li>No IDEA plugin

Checkstyle's IDEA plugin allows checkstyle violations to be highlighted in the IDE. There's no such plugin yet for `ktlint`, which means we're likely not to notice violations until the commits have been pushed and Travis has run `ktlint` on the commandline and failed the build.

We can reflect some of `ktlint`'s rules in the Android Studio settings, but we can't do it for everything. Using a pre-commit hook for ktlint would probably be a good compromise here.
</li>

<li>Maintaining checkstyle rules in `*.kt` files

[...]

PR #548, opened 2017-08-30
Generify Payload

aforcier:
Updates `Payload` to take a generic parameter, which represents the error type that `Payload` can have. Subclasses of `Payload` should always set the parameter - either to a custom error type or to `BaseNetworkError` (to reproduce the previous default behavior).

This PR also adds a Checkstyle rule to enforce that anything extending `Payload` supplies a parameter.

This is a prerequisite for Kotlin support - see #547 for more background detail.

cc @maxme

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -57,6 +57,12 @@
         <property name="severity" value="error"/>
     </module>
 
+    <module name="RegexpSingleline">
+        <property name="format" value="extends Payload[^&lt;]"/>
+        <property name="message" value="Payload should not be used as a raw type"/>
+        <property name="severity" value="error"/>
+    </module>
+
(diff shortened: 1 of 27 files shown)


config/checkstyle.xml
```diff
@@ -57,6 +57,12 @@
         <property name="severity" value="error"/>
     </module>
 
+    <module name="RegexpSingleline">
+        <property name="format" value="extends Payload[^&lt;]"/>
```
maxme: Good we have a rule for `<' is preceded with whitespace.` ;)

maxme:
Looks good, +1 for removing the `Payload` class, :shipit:

config/checkstyle.xml
aforcier: 😉

PR #557, opened 2017-09-06
Drop name() from IAction

aforcier:
This is a prerequisite for being able to write new `Action` classes in Kotlin - see item 3 in https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/547.

cc @maxme


maxme:
:shipit:

Issue #559, opened 2017-09-07
Inject UploadStore in PostStore and MediaStore


aforcier:
As an enhancement to the `UploadStore` addition (#533), we can inject the `UploadStore` into the `PostStore` and `MediaStore` instead of adding it to the constructor for those classes.

From my research so far, there's a blocker to that, which is that we can't use modules inside FluxC without a `Component`. Setting up a component inside FluxC will cause us to have two graphs when running inside an app, and the `UploadStore` won't actually be hooked up to the same `Dispatcher` as the rest of FluxC.

We can, however, use normal `@Inject` constructor annotation to supply `UploadStore` (we already do this for all stores), and this will allow the `PostStore` and `MediaStore` to inject it.

The problem is that we currently use both `@Inject` constructor annotations and `@Provides` methods (in `ReleaseStoreModule`) for our stores. This is redundant though it hasn't harmed us  so far. In this case, however, since `@Provides` seems to take priority, injecting `UploadStore` within FluxC doesn't actually work (because the library can't see the `@Provides` part of the Dagger graph).

Simply dropping `ReleaseStoreModule` fixes the problem, and I think this is a good solution. If we end up wanting a `MockedStoreModule`, we can make one and use `@Named` to differentiate those stores when injecting from the tests.

mzorz:
Closing via #562

PR #562, opened 2017-09-20
Drop ReleaseStoreModule + Inject UploadStore

aforcier:
Addresses #559.

While working on #559, I noticed that we were using both `@Inject`-annotated constructors and `@Provides` methods (in a module) for stores. This is actually redundant: we only need one or the other.

This didn't actually cause any problems until I attempted to use DI internally in FluxC, to inject the `UploadStore` into `PostStore` and `MediaStore` (instead of adding it to their constructors). Since the module is set up in the `Component`, which is part of the application and not part of FluxC itself, it looks like the dagger graph doesn't include the `UploadStore` dependency, and so it was never injected.

Dropping the `ReleaseStoreModule` and relying only on `@Inject` constructors fixed the issue.

So, this PR:
* Removes `ReleaseStoreModule`, relying on `@Inject` annotated constructors for stores instead
* Uses DI to ensure `UploadStore` is running whenever `PostStore` or `MediaStore` are
* Adds a test to verify that the above injection is working properly

~Note: #533 should be merged first.~ Also, this PR will cause breaking changes to WPAndroid (we'll need to drop the `ReleaseStoreModule` from there).

cc @maxme


maxme:
:shipit:

Issue #563, opened 2017-09-21
Use UploadStore in example app


aforcier:
Add a 'Upload media post' button in the example app that will demonstrate use of the `UploadStore` to register a post and upload it once its pending media complete.

Also update the documentation with the final set of actions/responses for `UploadStore` (#533).

aforcier:
Added in #565.

PR #565, opened 2017-09-25
Add UploadStore demo fragment to example app

aforcier:
Addresses #563. Adds a new `UploadFragment` to the example app that demonstrates (simple) use of the `UploadStore`.

~#533 should be merged first.~

cc @maxme


aforcier:
Noticed that the `.json` files with sample API responses were missing newlines at EoF (`checkstyle` caught it when I set it to run on all filetypes), fixed that up here.

aforcier:
Also [updated the Wiki](https://github.com/wordpress-mobile/WordPress-FluxC-Android/wiki/Actions-and-OnChanged-events/_compare/84aba0599c1afca1096d884570046bff612d892f...86c904abb5340c0ac958ae08e921153db72793e2) with `UploadAction`s.

example/src/main/java/org/wordpress/android/fluxc/example/UploadsFragment.java
```diff
@@ -0,0 +1,225 @@
...
+                if (!isAdded()) {
+                    return;
+                }
+
+                if (mCurrentMediaUpload == null) {
+                    mCancelButton.setEnabled(false);
+                    return;
+                }
+
+                cancelMediaUpload(mSite, mCurrentMediaUpload);
```
maxme: +1, it's great to be able to test cancellation here.

maxme:
I got this crash while testing 2 uploads at the same time

```
10-04 10:46:33.656  3566  3566 E AndroidRuntime: Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int org.wordpress.android.fluxc.model.MediaModel.getId()' on a null object reference
10-04 10:46:33.656  3566  3566 E AndroidRuntime: 	at org.wordpress.android.fluxc.example.UploadsFragment.onMediaUploaded(UploadsFragment.java:153)
10-04 10:46:33.656  3566  3566 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
10-04 10:46:33.656  3566  3566 E AndroidRuntime: 	at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:485)
```

`mCurrentMediaUpload` became `null` after I cancelled the first upload. Not a big deal but we could disable the upload button before allowing a second upload.

aforcier:
> we could disable the upload button before allowing a second upload.

Done in 99d3794.

maxme:
:shipit:

PR #566, opened 2017-09-25
Add Kotlin support

### LINKED ISSUE #547: Kotlin Support

aforcier:
🎉!

Enables Kotlin support, sets up some tooling, and converts a few sample files to Kotlin. Closes #547.

Checklist:
* [x] Enable Kotlin in FluxC and demo projects
* [x] Use `kapt`
* [x] Add `ktlint` and configure it to run globally
* [x] Add `ktlintFormat`
* [x] Make Travis run `ktlint` (example of failed `ktlint` on Travis: [link](https://travis-ci.org/wordpress-mobile/WordPress-FluxC-Android/builds/282692737))
* [x] `checkstyle` + Kotlin (tested and set `checkstyle` to run on all files)
* [x] Add some Kotlin style rules to shared configs
* [x] Convert a few sample classes to Kotlin
* [x] Update the README

I did some rough conversions locally (`SiteStore`, several network clients, action enums, several core classes). I didn't want to commit those as they were a fairly rough (not very idiomatic) conversion, and I only did it to verify things were building and to flush out any issues that would need to be addressed (see #557 for an example of an issue that turned up this way). Overall it looked like there aren't any blockers to add any type of new class in Kotlin once this PR is merged (converting some existing stuff to Kotlin isn't as simple, but that's not a blocker).

I _did_ leave in a small unobtrusive `PostFormatModel` conversion just as a demo of what those classes can look like in Kotlin (_swoon_ `data` classes 😍). I also converted a few core parts of the `example` app.

I converted several more parts of the example app, but I'll add those to a separate PR, where we can have a more leisurely discussion/iteration on style/idiom improvements/conventions without blocking Kotlin support.

I'm targeting a feature branch in case we want to do some follow-up PRs before this hits `develop` - if not we can change the target branch before merging.


example/src/main/java/org/wordpress/android/fluxc/example/AccountFragment.java
```diff
@@ -32,7 +32,7 @@
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        ((ExampleApp) getActivity().getApplication()).component().inject(this);
+        ((ExampleApp) getActivity().getApplication()).getComponent().inject(this);
```
maxme: We should switch to Android injectors at some point

maxme:
* New `PostFormatModel` looks much better than the java version 💯

example/src/main/java/org/wordpress/android/fluxc/example/AccountFragment.java
aforcier: Definitely, I'm actually hoping to do this soon - I'll be following up with you about this one.

maxme: Ok, last time I checked there was a very long discussion about performances of the Android injectors, I'm not sure if it's a big deal or if it's solved.

aforcier:
> We can bump kotlin 1.1.51 and ktlint version 0.9.2 https://github.com/shyiko/ktlint/blob/master/CHANGELOG.md

Done in 89465f8. Better merge fast before they go up again 😉

maxme:
:shipit:

PR #569, opened 2017-09-27
ThemeStore: Add actions to install themes on Jetpack sites

tonyr59h:
Theme activation for both Jetpack and WP.com was added in #552. While any WP.com theme can be activated on WP.com sites without issue, Jetpack sites require the theme to be installed first. This PR adds the necessary install actions and updates the example app to demo and help with testing.

To test:
* Go to the site's wp-admin and check which themes are currently installed
* In the example app...sign into an account with a Jetpack site linked
* Go to the Theme's screen
* Type a theme slug (ie `edin`, `twentysixteen`, etc...) for a theme that isn't already installed into the EditText
* Click the `Install WP.com theme on Jetpack site` button and check the log output to make sure the installation was successful
* Now try again with the same theme and make sure the log shows `The theme is already installed`

cc @aforcier, would you mind taking a look at this one?


fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
```diff
@@ -34,9 +34,14 @@ public FetchedCurrentThemePayload(@NonNull SiteModel site, @NonNull ThemeModel t
             this.site = site;
             this.theme = theme;
         }
+
+        @Override
+        public boolean isError() {
```
aforcier: We can drop this method, it's the same as the one it's overriding.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
```diff
@@ -21,7 +21,7 @@
 
 public class ThemeStore extends Store {
     // Payloads
-    public static class FetchedCurrentThemePayload extends Payload {
+    public static class FetchedCurrentThemePayload extends Payload<FetchThemesError> {
         public SiteModel site;
         public ThemeModel theme;
         public FetchThemesError error;
```
aforcier: With the typed `Payload`s, we no longer need to overwrite the `error` field in the subclass - this line (and the ones in `FetchedThemesPayload` and `ActivateThemePayload` below) can be removed.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
```diff
@@ -238,6 +254,21 @@ private void handleCurrentThemeFetched(FetchedCurrentThemePayload payload) {
         emitChange(event);
     }
 
+    private void installTheme(@NonNull ActivateThemePayload payload) {
+        if (payload.site.isJetpackConnected()) {
+            mThemeRestClient.installTheme(payload.site, payload.theme);
+        } else {
+            payload.error = new ActivateThemeError(ThemeErrorType.NOT_AVAILABLE.name(), null);
```
aforcier: This is pretty minor, but you could add a second constructor to `ActivateThemeError` that takes a `ThemeErrorType` instead of a `String`, so you don't have to convert to a string and back again (like is done [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/85609c69a1a057cd79a8583e9560d406f7f1c7f0/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java#L143-L151)).

tonyr59h:
Thanks for the review, @aforcier!

aforcier:
:shipit:

PR #570, opened 2017-09-27
ThemeStore: Add actions to delete a theme installed on a Jetpack site

tonyr59h:
Note: #569 should be merged first

To test:
* Go to the Jetpack site's wp-admin and check which themes are currently installed
* In the example app...sign into an account with the Jetpack site linked
* Go to the Theme's screen
* Type a theme slug (ie `edin`, `twentysixteen`, etc...) for a theme that isn't already installed into the EditText
* Click the `Delete theme on Jetpack site` button and check the log output to make sure there's an error (unknown theme)
* Now try again with an installed theme and make sure the log shows success

cc @aforcier, this one is not very urgent since WPAndroid doesn't expose this to the user yet.


fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
```diff
@@ -284,4 +291,19 @@ private void handleThemeActivated(@NonNull ActivateThemePayload payload) {
         event.error = payload.error;
         emitChange(event);
     }
+
+    private void deleteTheme(@NonNull ActivateThemePayload payload) {
+        if (payload.site.isJetpackConnected() && payload.site.isUsingWpComRestApi()) {
```
aforcier: This reminds me, it looks like the `isUsingWpComRestApi()` is missing for [this line](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/9903e64873ba868c00257adf840f6af7d1d1806c/fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java#L266).

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
```diff
@@ -284,4 +291,19 @@ private void handleThemeActivated(@NonNull ActivateThemePayload payload) {
...
+            mThemeRestClient.deleteTheme(payload.site, payload.theme);
+        } else {
+            payload.error = new ActivateThemeError("not_available", null);
```
aforcier: Should use the enum here instead of a String literal.

fluxc/src/main/tools/wp-com-endpoints.txt
```diff
@@ -55,6 +55,7 @@
 /sites/$site/themes
 /sites/$site/themes/mine
 /sites/$site/themes/$theme_ID#String/install
+/sites/$site/themes/$theme_ID#String/delete
```
aforcier: This should go above `install` 😱

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
tonyr59h: Wow, great memory you've got there!

tonyr59h: Fixed in 94f1d292e0abd907269e20583cc40380f8a39fda

aforcier:
This is looking good now - @tonyr59h I assume the `[PR] Not Ready for Review` label can be dropped?

aforcier:
Side note related to themes in general: I feel like saving all WordPress.com sites under a site ID `0` (when performing `FETCH_WP_COM_THEMES`) might come back to bite us one day (for example, if we decide to start using foreign key constraints more). This might be better in a separate table, without a site ID reference.

tonyr59h:
> I feel like saving all WordPress.com sites under a site ID 0 (when performing FETCH_WP_COM_THEMES) might come back to bite us one day

I definitely have my reservations about it, hadn't considered making a separate table for it but that's probably a smart idea.

aforcier:
> I definitely have my reservations about it, hadn't considered making a separate table for it but that's probably a smart idea.

Want to track it in an issue for now @tonyr59h?

aforcier:
:shipit:

PR #575, opened 2017-10-05
Feature/add social 2fa

theck13:
Add a method for WordPress.com social login two-factor authentication, `pushSocialAuth`.  The endpoint has a similar structure to `pushSocialLogin`, but requires different parameters and returns different values.  Because of that, I considered created new response, payload, and error classes, but I decided to update the existing classes instead to consolidate REST-like API calls.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
@@ -3,6 +3,7 @@ package org.wordpress.android.fluxc.network.rest.wpcom.account;
 import android.content.Context;
 import android.support.annotation.NonNull;
 
+import com.android.volley.DefaultRetryPolicy;
 import com.android.volley.RequestQueue;
 import com.android.volley.Response.Listener;
 import com.android.volley.VolleyError;
@@ -38,8 +39,12 @@ import java.util.Map;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import static com.android.volley.DefaultRetryPolicy.DEFAULT_BACKOFF_MULT;
+import static org.wordpress.android.fluxc.network.BaseRequest.DEFAULT_REQUEST_TIMEOUT;
+
 @Singleton
 public class AccountRestClient extends BaseWPComRestClient {
+    private static final String SOCIAL_AUTH_ENDPOINT_VERSION = "1";
     private static final String SOCIAL_LOGIN_ENDPOINT_VERSION = "1";
 
     private final AppSecrets mAppSecrets;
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -201,9 +218,58 @@ public void onErrorResponse(@NonNull BaseNetworkError error) {
...
+                },
+                new BaseErrorListener() {
+                    @Override
+                    public void onErrorResponse(@NonNull BaseNetworkError error) {
+                        AccountPushSocialResponsePayload payload = new AccountPushSocialResponsePayload(error);
+                        mDispatcher.dispatch(AccountActionBuilder.newPushedSocialAction(payload));
+                    }
+                }
+        );
+        request.setRetryPolicy(new DefaultRetryPolicy(DEFAULT_REQUEST_TIMEOUT, 0, DEFAULT_BACKOFF_MULT));
```
aforcier: I know I'm the one who recommended this, but I just realized there's a better way - you can just use [`request.disableRetries()`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/98e50a83f196b993b076c7de02967c8a02692d54/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java#L175-L177) here instead.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountSocialRequest.java
```diff
@@ -56,6 +56,12 @@ public String getBodyContentType() {
             JSONObject object = new JSONObject(responseBody);
             JSONObject data = object.getJSONObject("data");
             parsed.bearer_token = data.optString("bearer_token");
+            parsed.two_step_supported_auth_types = data.optJSONArray("two_step_supported_auth_types");
```
aforcier: It looks like we're storing this in the `AccountSocialResponse` but it's not actually used anywhere - shouldn't we float this up to the `AccountPushSocialResponsePayload` (and eventually the `OnSocialChanged` event), so the client knows what 2fa methods are supported for the user?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountSocialResponse.java
```diff
@@ -1,7 +1,15 @@
...
 import org.wordpress.android.fluxc.network.Response;
 
 public class AccountSocialResponse implements Response {
+    public JSONArray two_step_supported_auth_types;
     public String bearer_token;
+    public String two_step_nonce_authenticator;
+    public String two_step_nonce_backup;
+    public String two_step_nonce_sms;
+    public String two_step_notification_sent;
+    public String two_step_type;
```
aforcier: What is this field for? It doesn't seem to be populated at any point.

aforcier:
Gave social login -> 2fa needed response -> social auth -> error/success a full try, this seems to be working well 👍. Should be good to go aside from my couple of comments.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
theck13: I updated to `disableRetries` in ec4e474fb366a2de28c1cf2b380ea3d3c0f399ca.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountSocialRequest.java
theck13: I added the types to the response payload and changed method in 7f810b928354a8c29656cd4bb1ff903bfef9074d.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountSocialResponse.java
theck13: That should have been `two_step_supported_auth_types`.  I removed `two_step_type` in f6e47f63e670ad6841fd983f12e7a007bcb7add1.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -67,19 +69,33 @@ public AccountPushSocialResponsePayload(AccountSocialResponse response) {
...
+        public List<String> twoStepTypes;
         public String bearerToken;
         public String twoStepNonceAuthenticator;
         public String twoStepNonceBackup;
         public String twoStepNonceSms;
         public String twoStepNotificationSent;
-        public String twoStepType;
         public String userId;
+
+        private List<String> convertJsonArrayToStringList(JSONArray array) {
```
aforcier: This convinces me that we should refactor `AccountSocialResponse` to use Gson - but it's going to be a bit of an undertaking and I don't think it's worth bogging things down at this point.

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
```diff
@@ -591,7 +594,8 @@ private void handlePushSocialCompleted(AccountPushSocialResponsePayload payload)
         // Error; emit only social change.
         if (payload.isError()) {
             OnSocialChanged event = new OnSocialChanged();
-            event.error = new AccountSocialError(payload.error.type, payload.error.message, payload.error.nonce);
+            event.error = new AccountSocialError(payload.error.type, payload.error.message, payload.error.nonce,
+                    payload.error.twoStepTypes);
```
aforcier: It doesn't look like `payload.error.twoStepTypes` is ever assigned a value - did you mean for this to be passed as `event.twoStepTypes` instead?

theck13: That was my mistake.  The `twoStepTypes` is only passed in the successful response and not in the error response.  I removed it from the `AccountSocialError` class in 554e2a2e445d85dfca2391350ba3ba2e256fed48.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -67,19 +69,33 @@ public AccountPushSocialResponsePayload(AccountSocialResponse response) {
...
+            List<String> list = new ArrayList<>();
+
+            try {
+                for (int i = 0; i < array.length(); i++) {
```
aforcier: We'll get an NPE here if `array` is `null`.

theck13: Good <i>catch</i>.  I added handling for that in 9e29a5dc3a79890b211562bbf0848c3de9c38d8a.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -90,7 +90,7 @@ public AccountPushSocialResponsePayload(BaseNetworkError error) {
                 for (int i = 0; i < array.length(); i++) {
                     list.add(array.getString(i));
                 }
-            } catch (JSONException exception) {
+            } catch (NullPointerException | JSONException exception) {
```
aforcier: It's generally not recommended to catch runtime `Exception`s - it's better to guard against this earlier and just return the empty list.

theck13: I added the guard in 9646349dea0efc27b9a1c756a1e3126871dff890.

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
```diff
@@ -548,10 +582,16 @@ private void handlePushSettingsCompleted(AccountPushSettingsResponsePayload payl
     }
 
     private void handlePushSocialCompleted(AccountPushSocialResponsePayload payload) {
+        // Error; emit only social change.
```
aforcier: Out of context, this is probably one of my favorite comments in this repository 😀

aforcier:
:shipit:

PR #577, opened 2017-10-12
Feature/add social connect

theck13:
Add a method for social login to connect a social account with a WordPress.com account, `pushSocialConnect`.  It requires similar parameters to the `pushSocialAuth` and `pushSocialLogin` methods, but uses the `/me/social-login/connect/` authenticated endpoint.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
@@ -75,6 +75,8 @@ public class AccountRestClient extends BaseWPComRestClient {
         public AccountPushSocialResponsePayload(BaseNetworkError error) {
             this.error = new AccountSocialError(error.volleyError.networkResponse.data);
         }
+        public AccountPushSocialResponsePayload() {
+        }
         public List<String> twoStepTypes;
         public String bearerToken;
         public String twoStepNonceAuthenticator;
@@ -313,7 +315,8 @@ public class AccountRestClient extends BaseWPComRestClient {
                 new BaseErrorListener() {
                     @Override
                     public void onErrorResponse(@NonNull BaseNetworkError error) {
-                        AccountPushSocialResponsePayload payload = new AccountPushSocialResponsePayload(error);
+                        AccountPushSocialResponsePayload payload = new AccountPushSocialResponsePayload();
+                        payload.error = new AccountSocialError(((WPComGsonNetworkError) error).apiError, error.message);
                         mDispatcher.dispatch(AccountActionBuilder.newPushedSocialAction(payload));
                     }
                 }
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
@@ -281,6 +281,46 @@ public class AccountRestClient extends BaseWPComRestClient {
     }
 
     /**
+     * Performs an HTTP POST call to the v1.1 /me/social-login/connect/ endpoint.  Upon receiving a
+     * response (success or error) a {@link AccountAction#PUSHED_SOCIAL} action is dispatched with a
+     * payload of type {@link AccountPushSocialResponsePayload}.
+     *
+     * {@link AccountPushSocialResponsePayload#isError()} can be used to check the request result.
+     *
+     * No HTTP POST call is made if the given parameter map is null or contains no entries.
+     *
+     * @param idToken       OpenID Connect Token (JWT) from the service the user is using to
+     *                      authenticate their account.
+     * @param service       Slug representing the service for the given token (e.g. google).
+     */
+    public void pushSocialConnect(@NonNull String idToken, @NonNull String service) {
+        String url = WPCOMREST.me.social_login.connect.getUrlV1_1();
+
+        Map<String, Object> params = new HashMap<>();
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -281,6 +281,46 @@ public void onErrorResponse(@NonNull BaseNetworkError error) {
...
+                    @Override
+                    public void onResponse(AccountSocialResponse response) {
+                        AccountPushSocialResponsePayload payload = new AccountPushSocialResponsePayload(response);
+                        mDispatcher.dispatch(AccountActionBuilder.newPushedSocialAction(payload));
+                    }
+                },
+                new BaseErrorListener() {
+                    @Override
+                    public void onErrorResponse(@NonNull BaseNetworkError error) {
+                        AccountPushSocialResponsePayload payload = new AccountPushSocialResponsePayload(error);
```
aforcier: Since this social endpoint is able to use `WPComGsonRequest`, we can use the error that [`WPComGsonRequest.deliverBaseNetworkError()`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/2661ab61bcdd2aa6314825153961e16fe6d5ef01/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequest.java#L107-L108) has already parsed for us here.

Here's an example usage:

https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/2661ab61bcdd2aa6314825153961e16fe6d5ef01/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java#L79

(You'll also need to add a constructor to `AccountSocialError` like [this one](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/2661ab61bcdd2aa6314825153961e16fe6d5ef01/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java#L148-L151)).

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
```diff
@@ -274,11 +274,18 @@ public AccountSocialError(@NonNull byte[] response) {
...
                 String responseBody = new String(response, "UTF-8");
                 JSONObject object = new JSONObject(responseBody);
-                JSONObject data = object.getJSONObject("data");
-                this.nonce = data.optString("two_step_nonce");
-                JSONArray errors = data.getJSONArray("errors");
-                this.type = AccountSocialErrorType.fromString(errors.getJSONObject(0).getString("code"));
-                this.message = errors.getJSONObject(0).getString("message");
+                String error = object.optString("error");
+
+                if (!TextUtils.isEmpty(error)) {
```
aforcier: We can get rid of this block if we use the already parsed error `WPComGsonRequest` gives us in `AccountRestClient` (https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/577/files#r144525266).

aforcier:
@theck13 what's the expected successful response for this endpoint? I'm getting:

```js
{
  "redirect_to": ""
}
```

The result is an [emitted `OnSocialChanged` that is completely empty](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/2661ab61bcdd2aa6314825153961e16fe6d5ef01/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java#L592). Don't we need to somehow distinguish this response from a '2fa required' social signin?

(Also if we keep this in the same block the comment above that line, `"No error, but two-factor authentication is required; emit only social change."` needs to be updated.)

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
theck13: I added the new constructor in e3cf72e1d5932449764c8396f8fe296f6193944f and updated the error parsing in 77cf6370f11e4c38f77e2d46da674b5c1058ecec.

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
theck13: I removed that block in eb6296ed541e2912073cfc6b38b94de16145567e.

theck13:
Yep, that's the expected response.  It's not very useful since we aren't redirecting anywhere, but you made a good point; we should differentiate between two-factor authentication and social connect when receiving a successful response.  I added that in 7308089445dadfd212549a493186272ee7d2f4df.

fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
```diff
@@ -146,6 +147,7 @@ public OnSocialChanged(@NonNull AccountPushSocialResponsePayload payload) {
             this.nonceSms = payload.twoStepNonceSms;
             this.notificationSent = payload.twoStepNotificationSent;
             this.userId = payload.userId;
+            this.requiresTwoStepAuth = true;
```
aforcier: This is currently being set to `true` even for the case of a successful social connect request.

theck13: You're right.  I updated the logic with the `hasTwoStepTypes` method in 9be6303c91940142e92b7e6088bd90926bf51be6.

aforcier:
Works great! Tested social connect as well as social login/auth.

:shipit:

PR #578, opened 2017-10-13
Make Comment, Plugin, and Taxonomy stores Singletons

aforcier:
We missed the `@Singleton` annotation on a few stores, which made it possible for several instances of those stores to exist at once, with all of them receiving actions.

This manifested in the WPAndroid app as duplicate comments being created when replying to a post or comments - the number of duplicates depended on your navigation history through the app, which would create multiple `CommentStore`s.

In this PR I made all stores `@Singleton`s, as well as all network clients, and added Checkstyle rules enforcing both.

To test this PR:

1. Checkout `develop`
2. Run the FluxC example app
3. Hook up your network sniffer of choice
4. Open the 'Comments' view and fetch comments
5. Notice one fetch request sent in the network sniffer
6. Return to the main view and re-open the 'Comments' view a few times
7. Fetch comments again
8. Notice multiple (duplicate) requests are sent

This should no longer happen in this branch.

cc @maxme

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -57,10 +57,16 @@
         <property name="severity" value="error"/>
     </module>
 
+    <!-- Specific to FluxC -->
     <module name="RegexpSingleline">
         <property name="format" value="extends Payload[^&lt;]"/>
         <property name="message" value="Payload should not be used as a raw type"/>
         <property name="severity" value="error"/>
+    </module>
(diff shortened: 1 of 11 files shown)


maxme:
Looks good, I checked there were only 1 instance of our stores and network clients.
Also checked the new checkstyle regex was working.

:shipit:

PR #581, opened 2017-10-13
text hints are defined as non nullable string in ThreeEditTextDialog.…

maxme:
…newInstance (Kotlin) definition - results in a crash when it's called with a null argument from Java


aforcier:
🤦‍♂️ :shipit:

PR #580, opened 2017-10-13
Use kapt plugin

aforcier:
Updates to use the [kapt plugin](https://kotlinlang.org/docs/reference/kapt.html), allowing us drop the `generateStubs = true` config (and silence a `Original kapt is deprecated. Please add "apply plugin: 'kotlin-kapt'" to your build.gradle.` compiler warning).

Explicitly declaring a `kaptAndroidTest` dependency for the connected tests appears to be required as a result - the linked article above seems to say otherwise, but it doesn't seem to be the case.

cc @maxme

PR #586, opened 2017-10-19
Bump Dagger to 2.11

aforcier:
Bumps our dagger version to `2.11`, which is a prerequisite for eventually using [`dagger-android`](https://google.github.io/dagger/android.html) to simplify injections for activities/fragments/services, both in the example app and the real FluxC clients. (We have an issue open for this: https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/495.)

A follow-up PR will actually use `dagger-android` and overhaul injection in the example app.

I also took the opportunity to centralize our dagger and support library versions (https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/3c598e8f305b0346a4fb84e94e0c097f7fe28df5). We could do this for other dependencies too, but I just picked these two as we use them in multiple places and they should always be using the same version.

### Notes

1. ~#580 should be reviewed and merged first - it turns out using the newer `kapt` plugin form is required to play nice with `dagger-android`.~ There are issues with using the new `kapt` plugin, and it turns out not to be necessary to use `dagger-android`, so we're going to proceed without updating `kapt` for now.

2. Also, a PR in WPAndroid is required, bumping the `Dagger` version there to `2.11` and updating the FluxC hash (no other changes should be required in WPAndroid).

cc @maxme

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,8 +1,6 @@
 allprojects {
     apply plugin: 'checkstyle'
 
-    ext.kotlin_version = '1.1.51'
-
     repositories {
         jcenter()
     }
@@ -42,4 +40,11 @@ subprojects {
         classpath = configurations.ktlint
         args "-F", "src/**/*.kt"
     }
+}
+
+ext {
+    kotlinVersion = '1.1.51'
+
+    daggerVersion = '2.11'
+    supportLibraryVersion = '25.3.1'
 }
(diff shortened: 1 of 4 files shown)


aforcier:
Rebased out the previous reliance on the new kapt plugin (#580). Ready for review, @maxme!

PR #597, opened 2017-11-01
Convert ThemeFragment to Kotlin

tonyr59h:
Just looking to practice Kotlin a bit.

To test:
* Run the theme functions in the example app


example/src/main/java/org/wordpress/android/fluxc/example/ThemeFragment.kt
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    private fun prependToLog(s: String) = (activity as MainExampleActivity).prependToLog(s)
+
+    private fun getThemeIdFromInput(root: View?): String? {
+        val themeIdInput = if (root == null) null else root.findViewById(R.id.theme_id) as TextView
+        return themeIdInput?.text?.toString()
+    }
+
+    private val wpComSite: SiteModel?
```
aforcier: The converter automatically turns methods of type `getBla` into properties, but I'm not sure it's a good idea for this and the one below - we're pretending they're properties but calling them is actually performing a db call, which you wouldn't really expect. I think we should keep these two as `getWpComSite()` and `getJetpackConnectedSite()`.

example/src/main/java/org/wordpress/android/fluxc/example/ThemeFragment.kt
```diff
@@ -0,0 +1,246 @@
...
+        get() {
+            return siteStore.sites.firstOrNull { it != null && it.isWPCom }
```
aforcier: :+1:

example/src/main/java/org/wordpress/android/fluxc/example/ThemeFragment.kt
```diff
@@ -0,0 +1,246 @@
...
+                }
+            }
+        }
+
+        search_themes.setOnClickListener {
+            val term = getThemeIdFromInput(view)
+            if (TextUtils.isEmpty(term)) {
+                prependToLog("Please enter a search term")
+            } else {
+                dispatcher.dispatch(ThemeActionBuilder.newSearchThemesAction(ThemeStore.SearchThemesPayload(term!!)))
```
aforcier: I've been trying to avoid using the `!!` operator since it's kind of dangerous and often unnecessary. In this case we already know `term` is safe, but the `!!` will prevent you from realizing the danger if you refactored this method in the future and were no longer checking `term`'s nullness in this way.

I'm not sure what the best approach is in this case, though. Smart cast unfortunately isn't smart enough to figure out that `term` is safe, even if we used Kotlin's `String?.isNullOrEmpty()` function. We could add a (kind of redundant) `term == null ||` in front of `TextUtils.isEmpty(term)`, which would let us drop the `!!`. Or maybe `getThemeIdFromInput()` can return a `String` (replacing `null` with the empty string).

WDYT, @tonyr59h?

tonyr59h: I steer clear of `!!` as much as possible but with the `isEmpty()` check above this felt safe.

> Or maybe getThemeIdFromInput() can return a String (replacing null with the empty string).

I'm inclined to go this route if only for the opportunity to use the Elvis operator 😄

example/src/main/java/org/wordpress/android/fluxc/example/ThemeFragment.kt
tonyr59h: > I'm not sure it's a good idea for this and the one below - we're pretending they're properties but calling them is actually performing a db call,

Definitely agree, nice catch. Switched them over to methods in c0fe18a6891824b30c107e0aeeb4bd1fffc8a630

example/src/main/java/org/wordpress/android/fluxc/example/ThemeFragment.kt
aforcier: > I'm inclined to go this route if only for the opportunity to use the Elvis operator 😄

"Thank you, thank you very much"

aforcier:
Nice work, @tonyr59h! :shipit:

PR #599, opened 2017-11-04
Feature/add social sms

theck13:
Add a method for WordPress.com social login to send a two-factor authentication code via SMS, `pushSocialSms`.  The endpoint has a similar structure to `pushSocialAuth` and `pushSocialLogin`, but requires different parameters and returns different values.  Because of that, I decided to update the existing classes instead to consolidate REST-like API calls.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
@@ -13,6 +13,7 @@ import org.wordpress.android.fluxc.store.AccountStore.NewAccountPayload;
 import org.wordpress.android.fluxc.store.AccountStore.PushAccountSettingsPayload;
 import org.wordpress.android.fluxc.store.AccountStore.PushSocialAuthPayload;
 import org.wordpress.android.fluxc.store.AccountStore.PushSocialLoginPayload;
+import org.wordpress.android.fluxc.store.AccountStore.PushSocialSmsPayload;
 import org.wordpress.android.fluxc.store.AccountStore.UpdateTokenPayload;
 
 @ActionEnum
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
```diff
@@ -32,6 +32,8 @@
     PUSH_SOCIAL_CONNECT,    // request social connect remotely
     @Action(payloadType = PushSocialLoginPayload.class)
     PUSH_SOCIAL_LOGIN,      // request social login remotely
+    @Action(payloadType = PushSocialAuthPayload.class)
```
aforcier: I wonder if it would make sense to just use a new payload type for making the request. Two fields, `type` and `code`, are unused for the SMS request and have to be set as empty strings.

We could also just add a new constructor that just takes `userId` and `authenticatorNonce` - I'd be okay with that but IMO social login is convoluted enough that it's worth separating things out a bit more for clarity, and creating a new request payload type.

Similarly, I think `AccountPushSocialResponsePayload` is getting to be doing too much, but that one's less important since it's FluxC-internal.

aforcier:
Looks to be working well @theck13 👍 I just had [one comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/599#discussion_r149369205) about the payloads.

fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
theck13: I added a separate payload in 1da1b5fa7b18d85f3790e6df9d3cf06573444296.

aforcier:
:shipit:

PR #600, opened 2017-11-06
Add getPostByRemotePostId in PostStore

mzorz:
This PR adds the ability to obtain a `PostModel` from the `PostStore` given the `remote postId` is known as well as the `SiteModel` to which such a post belongs to.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/post/PostStoreUnitTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/post/PostStoreUnitTest.java
@@ -192,6 +192,17 @@ public class PostStoreUnitTest {
     }
 
     @Test
+    public void testGetPostByRemoteId() {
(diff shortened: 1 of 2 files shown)


aforcier:
@mzorz What's the use case for this? We've generally tried to avoid exposing remote IDs directly though FluxC methods if it could be helped (which maybe it can't in this case - just wondering what the background is).

aforcier:
Ah, I see the WPAndroid PR now - a workaround for losing local IDs when we update remote posts, got it.

aforcier:
Looks good!

This reminded me of the work @nbradbury did in #514 to prevent changing local IDs on fetch for media. I opened #601 to track applying that fix to other models (including posts), which might eventually allow us not to rely on the remote ID in WPAndroid anymore. But until we get to that this fix addresses the issue well 👍

:shipit:

PR #610, opened 2017-11-10
ThemeStore: Add actions for removing themes from local database

tonyr59h:
Simple patch since we don't need to deal with network calls and response action handling. This PR just adds `REMOVE_THEME` action for removing a single specific theme and `REMOVE_SITE_THEMES` action for removing all themes associated with a site (via local site ID).

I've tested this in WPAndroid and it's worked without a problem thus far. Tests coming in different PR.

cc @kwonye or @oguzkocer


tonyr59h:
@kwonye Tests added for the new actions. Unit tests weren't possible without exposing some `ThemeStore` methods but I've added end-to-end tests. Make sure all the tests in `ReleaseStack_ThemeTestJetpack` pass!

kwonye:
Thanks for writing the tests! Works great and tests all pass.

:shipit:

PR #612, opened 2017-11-14
ThemeStore: Resolve type mismatch in search response

tonyr59h:
@kwonye found a bug while reviewing the WPAndroid ThemeStore integration branch. Apparently search results return prices as `String` and other endpoints return an `Object`. So we have to create an entirely separate response class for the search results, that's what this PR does.

To test:
* Search `twenty` in the example app and expect 8 results
* Search `fifteen` and expect 2
* No crashes

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java
@@ -2,6 +2,7 @@ package org.wordpress.android.fluxc.network.rest.wpcom.theme;
 
 import android.content.Context;
 import android.support.annotation.NonNull;
+import android.text.TextUtils;
 
 import com.android.volley.RequestQueue;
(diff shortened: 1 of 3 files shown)


kwonye:
@tonyr59h would it make more sense to have the API team fix it on the server? Us managing two different types of themes doesn't sound ideal.

Or, can we have the ThemeModel be more flexible to handle both a string/object response?

tonyr59h:
> Or, can we have the ThemeModel be more flexible to handle both a string/object response?

With the way we deserialize network responses I don't think we're able to have one model handle two different types with the same name (`price` in this case). Best we could do is create a base response model with all fields except price then two subclasses with `price` of differing types. But I think that is more complicated than duplicating fields as we're doing here.

tonyr59h:
> would it make more sense to have the API team fix it on the server

Let me see what we can do about this.

tonyr59h:
So there's an undocumented `v1.2/themes` endpoint that we could use and it returns a `String price`. This would allow us to stick with just the one model. I'll open a separate PR to keep commit history down.

PR #613, opened 2017-11-15
ThemeStore: Use GET v1.2/themes endpoint to fetch themes

tonyr59h:
Related to #612. This PR uses the newer endpoint (v1.2) which fixes the type mismatch issue and allows us to simply update the type of `price` to a `String`.

cc @kwonye

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java
@@ -2,6 +2,7 @@ package org.wordpress.android.fluxc.network.rest.wpcom.theme;
 
 import android.content.Context;
 import android.support.annotation.NonNull;
+import android.text.TextUtils;
 
 import com.android.volley.RequestQueue;
 import com.android.volley.Response;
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java
```diff
@@ -123,16 +124,16 @@ public void onErrorResponse(@NonNull BaseNetworkError error) {
                 }));
     }
 
-    /** Endpoint: v1.1/themes */
+    /** [Undocumented!] Endpoint: v1.2/themes */
     public void fetchWpComThemes() {
-        String url = WPCOMREST.themes.getUrlV1_1();
-        add(WPComGsonRequest.buildGetRequest(url, null, MultipleWPComThemesResponse.class,
-                new Response.Listener<MultipleWPComThemesResponse>() {
+        String url = WPCOMREST.themes.getUrlV1_2() + "?number=500";
```
kwonye: Can we set that string as a constant please? I'm guessing this is the limit of themes we're requesting?

tonyr59h: > I'm guessing this is the limit of themes we're requesting?

Yeppers! Moved to constant in 9ca40cca77f6af03583c95ad86549cb005f5a07a

kwonye:
Tyty! I'll :shipit: once travis builds this.

Issue #618, opened 2017-11-16
Crash report 8.8-rc-2: NPE in AccountRestClient


maxme:
```
Fatal Exception: java.lang.NullPointerException: Attempt to read from field 'byte[] com.android.volley.NetworkResponse.data' on a null object reference
       at org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient$AccountPushSocialResponsePayload.<init>(AccountRestClient.java:77)
       at org.wordpress.android.fluxc.network.rest.wpcom.account.AccountRestClient$14.onErrorResponse(AccountRestClient.java:371)
       at org.wordpress.android.fluxc.network.BaseRequest.deliverError(BaseRequest.java:244)
       at com.android.volley.ExecutorDelivery$ResponseDeliveryRunnable.run(ExecutorDelivery.java:101)
       at android.os.Handler.handleCallback(Handler.java:751)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:154)
       at android.app.ActivityThread.main(ActivityThread.java:6776)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1496)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1386)
```

https://fabric.io/automattic/android/apps/org.wordpress.android/issues/5a0d417361b02d480d89d10f

aforcier:
Fixed in #620.

PR #620, opened 2017-11-16
Issue/618 crash google login

theck13:
### Fix
Add a check for a `null` network response while parsing the social response payload to avoid `NullPointerException` as described in https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/618.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
@@ -77,7 +77,11 @@ public class AccountRestClient extends BaseWPComRestClient {
             this.userId = response.user_id;
         }
         public AccountPushSocialResponsePayload(BaseNetworkError error) {
-            this.error = new AccountSocialError(error.volleyError.networkResponse.data);
+            if (error.volleyError.networkResponse == null || error.volleyError.networkResponse.data == null) {
+                this.error = new AccountSocialError("generic_error", "");
+            } else {
+                this.error = new AccountSocialError(error.volleyError.networkResponse.data);
+            }
         }
         public AccountPushSocialResponsePayload() {
         }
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -77,7 +77,11 @@ public AccountPushSocialResponsePayload(AccountSocialResponse response) {
             this.userId = response.user_id;
         }
         public AccountPushSocialResponsePayload(BaseNetworkError error) {
-            this.error = new AccountSocialError(error.volleyError.networkResponse.data);
+            if (error.volleyError.networkResponse == null || error.volleyError.networkResponse.data == null) {
+                this.error = new AccountSocialError("generic_error", "");
```
aforcier: I think we should avoid using a literal here when we already have a matching enum. I'd suggest adding a `AccountSocialError` constructor that takes an `AccountSocialErrorType` and a `String`, so we can pass `AccountSocialErrorType.GENERIC_ERROR` instead.

Actually, I think this might be a good opportunity to move [this logic](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/26a756c63c2cb534a2b2f268967f4a6c790526ae/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java#L292-L303) inside this payload constructor and out of the store, as it's pretty low-level logic (and this will make it more consistent with other situations where we have to parse `networkResponse`).

[`volleyErrorToAccountResponsePayload()` below does something similar to what I mean](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/26a756c63c2cb534a2b2f268967f4a6c790526ae/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java#L537-L548) - first it builds an error with `GENERIC_ERROR`, and then updates the type and message only if it can parse `networkResponse.data`.

theck13: I added the new constructor in 307309616c510f23ef800ef67c7b69f383f270c5.
I moved the parsing logic in 5846d60900a0455bd6b9c3cfbe6f7869c0d37eb9.

aforcier:
Looks good! :shipit:

PR #622, opened 2017-11-17
Package sample image and video in the test apk

maxme:
Package sample image and video in the test apk instead of defining a local property. It makes connected tests more portable. 

User just have to install `example-debug.apk` and `example-debug-androidTest.apk`, no other setup is needed on the host.


aforcier:
Nice, @maxme. I'm still trying this out, but I'm wondering - what do you think about keeping `TEST_LOCAL_IMAGE` and `TEST_LOCAL_VIDEO` and falling back to the packaged assets if the variables are not defined?

Mostly my concern is that the packaged assets are nice and large at 1.9MB and 5.6MB, which tends to make for more reliable upload tests (and so I think are a good choice for automated test running), but those can still slow down the tests a lot when running the entire media test suite. I think it's worth keeping the option to use local, smaller media for speedier repetitive testing. It's also on occasion been useful to run the tests with different file sizes for comparison to reproduce issues.

maxme:
> I think it's worth keeping the option to use local, smaller media for speedier repetitive testing. It's also on occasion been useful to run the tests with different file sizes for comparison to reproduce issues.

Good idea, I'll add `TEST_LOCAL_IMAGE` and `TEST_LOCAL_VIDEO` back.

example/tests.properties-example
```diff
@@ -186,11 +186,7 @@ TEST_WPORG_URL_JETPACK_SUBFOLDER_ENDPOINT = FIXME
...
-# These files should be at least 500KB, otherwise they may cause cancellation tests to fail on fast enough connections
-
-# Path (on device) to a test image for uploading, e.g. /sdcard/Pictures/flux-capacitor-schematics.png
-TEST_LOCAL_IMAGE = FIXME
-
-# Path (on device) to a test video for uploading, e.g. /sdcard/Video/1985-10-26-Einstein-temporal-experiment.mp4
-TEST_LOCAL_VIDEO = FIXME
+# Local files to override default samples. Keep the empty string to use default samples.
+# These files should be at least 500KB, otherwise they may cause cancellation tests to fail on fast enough connections.
+TEST_LOCAL_IMAGE =
```
aforcier: Could we also restore the comments with path examples? Those were added because there had been some confusion about whether to use a computer-local path or a device-local path.

maxme: Restored in 2a88ce62

example/tests.properties-example
```diff
@@ -183,10 +183,12 @@ TEST_WPORG_URL_JETPACK_SUBFOLDER = FIXME
 TEST_WPORG_URL_JETPACK_SUBFOLDER_ENDPOINT = FIXME
 
 # WP.com - Account with only one site: a Jetpack site with beta versions enabled: http://do.wpmt.co/jp-beta/
+# Path (on device) to a test image for uploading, e.g. /sdcard/Pictures/flux-capacitor-schematics.png
```
aforcier: Looks like these two comments were added to the wrong place, they need to be shifted down by one comment block.

maxme: 🤦‍♂️

Issue #529, opened 2017-08-03
Run unit tests on Travis (or perish in the attempt)


aforcier:
We used to run unit tests on Travis, then we changed the command we ran and accidentally stopped running them.

Attempts to turn them back on have resulted in process deaths and failed builds on Travis (https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/483, #528).

aforcier:
We did it! #623

PR #623, opened 2017-11-17
Run unit tests on Travis

aforcier:
A new attempt to get Travis to run unit tests. Last time failed with intermittent issues (see #529), but I haven't been able to get Travis to fail this branch so far - it's possibly something changed in their config.

Of course, we might just end up reverting this PR in a few days. 🤷‍♂️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ before_install:
 
 script:
   - ./gradlew assembleDebug assembleRelease
+  - ./gradlew test
   - ./gradlew lint || (grep -A20 -B2 'severity="Error"' */build/**/*.xml; exit 1)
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,7 +22,7 @@ before_install:
   - echo yes | android update sdk -a --filter platform-tools --no-ui --force
 
 script:
-  - ./gradlew assembleDebug assembleRelease
```
maxme: I would keep this line, I don't these task explicitly called when running the task `test` and I wonder if there are extra tasks after compilation and before `assembleDebug` not needed by unit tests.

maxme:
:shipit:

PR #627, opened 2017-11-21
Update Travis build to only unit test release branch

tonyr59h:
Builds were very slow earlier. Maybe limiting the unit testing to the release branch will help?

cc @aforcier, @maxme; any benefits of running the tests on variants other than the one we ship with?


aforcier:
:shipit:

PR #628, opened 2017-11-21
Move DebugOkHttpClient to core FluxC

aforcier:
Makes `DebugOkHttpClient` more generic and moves it to core FluxC (it was previously part of the example app). This allows other FluxC clients to use debug builds with interceptors without duplicating the code in `DebugOkHttpClient`.


maxme:
:shipit:

PR #629, opened 2017-11-21
Add Stetho to Instaflux

aforcier:
Adds Stetho to Instaflux for inspecting the database and network requests through Chrome inspection. I've found this useful while working on the WooCommerce plugin, since Instaflux is not importing that and I can test database config differences and so on without leaving the FluxC repo.

Relies on #628, which moves `DebugOkHttpClient` to core FluxC so it can be reused.

### DIFF
--- a/instaflux/build.gradle
+++ b/instaflux/build.gradle
@@ -71,4 +71,5 @@ dependencies {
 
     // Debug dependencies
     debugImplementation 'com.facebook.stetho:stetho:1.5.0'
+    debugImplementation 'com.facebook.stetho:stetho-okhttp3:1.5.0'
 }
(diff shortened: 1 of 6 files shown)


instaflux/build.gradle
```diff
@@ -63,4 +63,8 @@ dependencies {
     compile "com.google.dagger:dagger:$daggerVersion"
     kapt "com.google.dagger:dagger-compiler:$daggerVersion"
     provided 'org.glassfish:javax.annotation:10.0-b28'
+
+    // Debug dependencies
+    compile 'com.facebook.stetho:stetho:1.5.0'
+    compile 'com.facebook.stetho:stetho-okhttp3:1.5.0'
```
maxme: Should we use `debugCompile` for these 2 new dependencies?

aforcier: Oops, yep - rebased over `develop` and fixed this (it's now `debugImplementation`).

maxme:
:shipit:

PR #619, opened 2017-11-16
WooCommerce plugin - Initial setup

aforcier:
Sets up the basics for a new WooCommerce store. This is added in a separate module from core `fluxc`, allowing FluxC clients to choose to use the `WooCommerce` plugin or not.

This is targeting the`woocommerce-store` feature branch as there are several more pieces needed to complete the core setup for the plugin (see https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/615).

This PR adds:

* A new module, `:plugin:woocommerce`
* Two new fields from the WP.com API, `is_wpcom_store`, `woocommerce_is_active`, stored in the `SiteModel`, to allow us to identify Jetpack sites with WooCommerce installed
* A tweak to the endpoint class generator to support multiple modules calling `fluxc-processor`
* A demo on use of the plugin in the example app

A few things are missing to complete core setup of the plugin and will be added in subsequent PRs:
* A WooCommerce network client and a Dagger module that provides it
* Support for database table generation from models inside the plugin module ~(this will require some changes to wellsql, which are a WIP)~ changes complete, see #642 (which has been merged into this PR's branch)

To use the plugin in an eternal project, import it using Jitpack along with core FluxC:

```
compile('com.github.wordpress-mobile.WordPress-FluxC-Android:fluxc:[hash or snapshot]')
compile 'com.github.wordpress-mobile.WordPress-FluxC-Android:woocommerce:[hash or snapshot]'
```

PR #635, opened 2017-12-04
Add Jetpack WP-API tunnel support

aforcier:
Adds a new type of network request: `WPComJPTunnelGsonRequest`. This request takes calls to WP-API endpoints (intended for self-hosted sites), and tunnels them through the WordPress.com `/jetpack-blogs/$site/rest-api/` endpoint. This involves packing params intended for the WP-API endpoint in a specific way to be understood and proxied by the tunnel endpoint, and unwrapping the response from the WP.com API.

The `WPComJPTunnelGsonRequest` class has [a Javadoc explaining the details](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/37dfd5078d6ff5f9835dfaa714023387a3aac303/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/WPComJPTunnelGsonRequest.kt#L11-L87) of how requests are made using the REST API tunnel.

I added some unit tests verifying that the built requests have the right format for each type of method supported by WP-API (`GET`, `POST`, `PATCH`, `PUT`, `DELETE`), and also mocked tests of a full request/response for success and failure.

The first usage of this will be by the WooCommerce FluxC module, which will use the tunnel to call `wc/v2` (WooCommerce) endpoints on the Jetpack site. I will be adding some live connected tests at the same time.

This is really part of the setup work for WooCommerce (and https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/619 should be reviewed and merged first, and the base branch updated to `woocommerce-store`), but I'm separating this out into a standalone PR to keep PR sizes reasonable.

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -81,6 +81,8 @@ dependencies {
         exclude group: "com.mcxiaoke.volley";
     }
 
+    implementation 'com.google.code.gson:gson:2.8.0'
+
     // Dagger
     implementation "com.google.dagger:dagger:$daggerVersion"
     kapt "com.google.dagger:dagger-compiler:$daggerVersion"
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -91,6 +91,7 @@ dependencies {
     kapt "com.google.dagger:dagger-android-processor:$daggerVersion"
 
     testImplementation 'junit:junit:4.12'
+    testImplementation "org.jetbrains.kotlin:kotlin-test-junit:$kotlinVersion"
     testImplementation 'org.mockito:mockito-core:2.8.9'
     testImplementation 'org.robolectric:robolectric:3.6.1'
     testImplementation "com.nhaarman:mockito-kotlin-kt1.1:1.5.0"
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/WPComJPTunnelGsonRequest.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/WPComJPTunnelGsonRequest.kt
@@ -1,0 +1,86 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel
+
+import com.android.volley.Response
+import com.google.gson.reflect.TypeToken
+import org.wordpress.android.fluxc.generated.endpoint.WPCOMREST
+import org.wordpress.android.fluxc.network.BaseRequest.BaseErrorListener
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest
+import java.lang.reflect.Type
+import kotlin.reflect.KClass
+
+/**
+ * A request making a WP-API call on a Jetpack site via the WordPress.com <code>/jetpack-blogs/$site/rest-api</code>
+ * tunnel.
+ */
+object WPComJPTunnelGsonRequest {
+    /**
+     * Creates a new GET request to the given WP-API endpoint, calling it via the WP.com Jetpack WP-API tunnel.
+     *
+     * @param wpApiEndpoint the WP-API request endpoint (e.g. /wp/v2/posts/)
+     * @param siteId the WordPress.com site ID
+     * @param params the parameters to append to the request URL
+     * @param type the Type defining the expected response
+     * @param listener the success listener
+     * @param errorListener the error listener
+     *
+     * @param T the expected response object from the WP-API endpoint
+     */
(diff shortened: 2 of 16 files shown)


example/build.gradle
```diff
@@ -81,6 +81,8 @@ dependencies {
         exclude group: "com.mcxiaoke.volley";
     }
 
+    implementation 'com.google.code.gson:gson:2.8.0'
```
maxme: Do we need this here? `example` has `fluxc` as a dependency (Not a big deal, I'm just wondering)

aforcier: I think the reason that's there is that Gson is an `implementation` dependency in `fluxc` now, so it's not available to `example`. We could make it an `api` dependency in `fluxc`, but since this is only used for tests I wasn't sure I wanted to add it to FluxC's exported API. WDYT?

maxme: Right, and in most case an app that uses `fluxc` doesn't automatically needs `gson` at compile time (ie. we're not providing `gson` objects from `fluxc`), so it makes sens to keep it as a `implementation` dependency here.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/WPComJPTunnelGsonRequest.kt
```diff
@@ -0,0 +1,236 @@
...
+            }
+        }
+        return finalBody
+    }
+
+    private fun buildRestApiPath(path: String, params: Map<String, String>, method: String): String {
+        var result = path + "&_method=" + method
+        if (params.isNotEmpty()) {
+            for (param in params) {
+                result += "&" + param.key + "=" + param.value
```
maxme: Should we URI-encode param.key and param.value here?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/WPComJPTunnelGsonRequest.kt
```diff
@@ -0,0 +1,236 @@
...
+ * The WordPress.com endpoint will return the response it received from the WP-API endpoint, wrapped in a `data`
+ * object (see [JPTunnelWPComRestResponse]). The response is unwrapped, and the pure WP-API response is handed
+ * to the listeners.
+ *
+ * # Errors
+ *
+ * Any errors from WP-API are converted into usual WP.com API errors.
+ *
+ */
+object WPComJPTunnelGsonRequest {
```
maxme: Naming nitpick: I'm not a fan of "JP" for Jetpack, it reminds me of the nickname for "Jean Pierre" too much ;)

maxme:
Looks good, only 2 comments.

maxme:
For the record, to run only these specific tests from cli (as the full test suite is now quite big):

```
./gradlew :example:testDebugUnitTest --tests "*JetpackTunnelGsonRequestTest"
./gradlew :example:connectedAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.mocked.MockedStack_JetpackTunnelTest
```

aforcier:
> `./gradlew :example:connectedAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.mocked.MockedStack_JetpackTunnelTest`

Yes, I've been finding that one very useful since AS sometimes tries to run single Kotlin connected tests or test classes as unit tests and fails 😕. (You need to configure the run config manually, which is a PITA.)

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/WPComJPTunnelGsonRequest.kt
aforcier: Yeah we probably should - done in b68c9e7.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/WPComJPTunnelGsonRequest.kt
aforcier: I agonized about the naming too - I renamed this to `JetpackTunnelGsonRequest` in 
cfdde56, I think the 'Jetpack' part implies that WP.com is involved enough that we don't need the 'WPCom' naming.

aforcier:
Ready for another pass @maxme!

maxme:
Looks good - :shipit:

PR #639, opened 2017-12-05
Feature/add social signup

theck13:
Add a method for social signup, `pushSocialSignup`, to sign up a new user with a WordPress.com account via their social account.  It requires similar parameters to the `pushSocialAuth`, `pushSocialConnect`, `pushSocialLogin`, and `pushSocialSms` methods, but uses the `/users/social/new/` authenticated endpoint.


aforcier:
Looks good, nice work on this - I think the end solution you arrived at with the extra `OnAuthenticationChanged` param keeps things pretty clear and understandable 👍.

:shipit:

PR #640, opened 2017-12-06
Add support for WP.com V2 endpoints to code generator

aforcier:
This adds a new endpoint text file, `wp-com-v2-endpoints.txt`, which is used to generate an endpoint class for WordPress.com V2 endpoints.

These are generally used just like any other WP.com (V1) REST endpoints (as part of a `WPComGsonRequest`, and within a network client that extends `BaseWPComRestClient`). The only difference is to use the generated `WPCOMV2` class instead of `WPCOMREST` to build the endpoint URL.

Note: The V2 endpoints don't seem to currently have a versioning structure, so all FluxC offers is the one kind of URL (https://public-api.wordpress.com/wpcom/v2/...). We can add new URL methods to `WPComV2Endpoint.java` in the future if needed (if `/wpcom/v2.1/` and such are used).

To test, rebuild the project on this branch and run the unit tests (`./gradlew testDebug`).

cc @aerych

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/WPComV2EndpointTest.java
+++ b/example/src/test/java/org/wordpress/android/fluxc/WPComV2EndpointTest.java
@@ -1,0 +1,23 @@
+package org.wordpress.android.fluxc;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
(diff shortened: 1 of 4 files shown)


aerych:
Thanks for adding this @aforcier :) 

I've spent some time refreshing my memory about fluxc, then took a peek this branch.  Everything looks good afaict. ;)   Tests pass.   I also tested the example app again for paranoia.
:shipit:

PR #642, opened 2017-12-11
WooCommerce plugin - WellSql config

aforcier:
Adds an example model to the WooCommerce plugin, along with configuration necessary to only add the corresponding DB table if the host app is using the WC plugin.

This is a WIP PR - ~for one, https://github.com/wordpress-mobile/wellsql/pull/11 (which is where most of the interesting stuff is happening) needs to be reviewed and merged, and this PR updated back to a release version of wellsql and not a Jitpack hash~ (done in 6e61243).

Note: One thing not implemented yet is the case of enabling the WC plugin after the database has already been set up, or removing the plugin, going through some FluxC upgrades and re-enabling the plugin. The assumption here is that the only client for the WooCommerce plugin module is the WC app, and that will have the plugin module added from the very beginning - I will add some checks/handling for this in the near future just in case though.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,6 @@ allprojects {
 
     repositories {
         jcenter()
-        maven { url "https://jitpack.io" }
     }
 
     task checkstyle(type: Checkstyle) {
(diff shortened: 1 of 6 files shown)


maxme:
:shipit:

PR #659, opened 2017-12-22
Added Portfolio content type support

michalkarmelita:
This PR adds support for multiple content types for posts. 

Currently `PostModel` can be either `Post` or `Page` which is determined by boolean flag saved into database. In order to support more than two I've added `ContentType` enum which allows to support multiple types. All usages of boolean flag has been refactored and I've added support for portfolio content. 

To make it consistent I've also changed DB model. Boolean `IS_PAGE` column has been replaced by text `TYPE`.


fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -395,10 +433,13 @@ public void onAction(Action action) {
...
             case SEARCH_POSTS:
-                searchPosts((SearchPostsPayload) action.getPayload(), false);
+                searchPosts((SearchPostsPayload) action.getPayload(), POST);
                 break;
             case SEARCH_PAGES:
-                searchPosts((SearchPostsPayload) action.getPayload(), true);
+                searchPosts((SearchPostsPayload) action.getPayload(), PAGE);
+                break;
+            case SEARCH_PORTFOLIOS:
+                searchPosts((SearchPostsPayload) action.getPayload(), PAGE);
```
aforcier: This should be `PORTFOLIO`.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -216,7 +221,8 @@ public static PostErrorType fromString(String string) {
     // Ensures that the UploadStore is initialized whenever the PostStore is,
     // to ensure actions are shadowed and repeated by the UploadStore
     @SuppressWarnings("unused")
-    @Inject UploadStore mUploadStore;
+    @Inject
```
aforcier: We don't have an automated style check for this yet, but we inline `@Inject` for FluxC components, it looks like this was changed by your auto-formatter.

example/src/main/res/layout/fragment_posts.xml
```diff
@@ -1,38 +1,55 @@
...
-        android:text="Search" />
+        <Button
+            android:id="@+id/fetch_first_site_portfolios"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:text="Fetch portfolios from first site"/>
 
-</LinearLayout>
+    </LinearLayout>
+</ScrollView>
```
aforcier: It looks like the newline that was previously at the end of the file here was accidentally removed.

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -410,7 +435,7 @@ public boolean equals(Object other) {
 
     public boolean supportsLocation() {
         // Right now, we only disable for pages.
-        return !isPage();
+        return getContentType() != PAGE;
```
aforcier: Do portfolios support location?

fluxc/src/main/java/org/wordpress/android/fluxc/model/post/ContentType.java
```diff
@@ -0,0 +1,25 @@
+package org.wordpress.android.fluxc.model.post;
+
+public enum ContentType {
```
aforcier: I think a more natural name for this would be `PostType` - what do you think @michalkarmelita?

example/src/main/java/org/wordpress/android/fluxc/example/PostsFragment.kt
```diff
@@ -90,14 +102,24 @@ class PostsFragment : Fragment() {
 
         val firstSite = getFirstSite()
         if (!postStore.getPostsForSite(firstSite).isEmpty()) {
```
aforcier: This is pretty minor, but you won't get logging for posts or portfolio fetches if there aren't any stored posts, since this line is only checking for 'post'-type posts specifically.

example/src/main/java/org/wordpress/android/fluxc/example/PostsFragment.kt
```diff
@@ -90,14 +102,24 @@ class PostsFragment : Fragment() {
...
-            if (event.causeOfChange == PostAction.FETCH_POSTS || event.causeOfChange == PostAction.FETCH_PAGES) {
-                prependToLog("Fetched " + event.rowsAffected + " posts from: " + firstSite.name)
+            if (event.causeOfChange in Arrays.asList(PostAction.FETCH_POSTS, PostAction.FETCH_PAGES, PostAction.FETCH_PORTFOLIOS)) {
```
aforcier: I'm not sure why this isn't triggering our style checker, but this line is over our `120` character limit.

aforcier: We could use Kotlin's `listOf()` here instead of `Arrays.asList()`.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
```diff
@@ -173,6 +173,15 @@ public void onUpgrade(SQLiteDatabase db, WellTableManager helper, int oldVersion
                 AppLog.d(T.DB, "Migrating to version " + (oldVersion + 1));
                 db.execSQL("alter table TermModel add POST_COUNT integer;");
                 oldVersion++;
+            case 20:
+                AppLog.d(T.DB, "Migrating to version " + (oldVersion + 1));
+                db.execSQL("alter table PostModel rename to PostModelOld");
+                helper.createTable(PostModel.class);
```
aforcier: Unfortunately we can't rely on `helper.createTable` for migrations, since this will use the latest version of the `PostModel`, not the one at case `20`. This means that when we make future changes to the `PostModel`, they'll also be applied here instead of at the later migration step they were supposed to be, and this could lead to a crash on migration.

aforcier:
This changeset is breaking a few of the [connected tests](https://github.com/wordpress-mobile/WordPress-FluxC-Android#building-and-running-tests-and-the-example-app) (`example/src/androidTest`), as they also need to be refactored to reflect the post type change.

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
michalkarmelita: How can I check it? If web editor is our source of truth then: no, portfolios doesn't support location. I can see location in settings for `POST` only.

fluxc/src/main/java/org/wordpress/android/fluxc/model/post/ContentType.java
michalkarmelita: It was super confusing for me at the beginning. Because all we're playing with here is a post and then it can be normal `post`, `page` or `portfolio`. But you know the project better and if you feel that `PostType` is more natural I'm fine with that.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
michalkarmelita: Yeah, looks like my auto-formatter broke this.

michalkarmelita:
Sorry, my bad. I fixed Unit tests, but missed regression tests. I'll fix it.

PR #667, opened 2017-12-28
Checkstyle/ktlint tweaks

aforcier:
This makes a few small adjustments to our lint rules:

1. The `./gradlew checkstyle` task now also checks `xml` files (the Android Studio plugin was already doing this)
2. I noticed checkstyle wasn't enforcing line length on Kotlin files, so I enabled a `max_line_length` check for ktlint

Also bumped Kotlin and ktlint versions, and fixed some new `ktlint` errors.

cc @maxme

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[*.{kt,kts}]
+max_line_length=120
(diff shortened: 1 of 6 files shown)


maxme:
:shipit: - I'll have to update Aztec config accordingly

PR #685, opened 2018-01-17
Feature/add username suggestions

theck13:
Add a method for fetching username suggestions for a WordPress.com account, `fetchUsernameSuggestions`.  It requires the `name` parameter using the `/wpcom/v2/users/username/suggestions` unauthenticated endpoint.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
@@ -17,6 +17,7 @@ import org.wordpress.android.fluxc.Payload;
 import org.wordpress.android.fluxc.action.AccountAction;
 import org.wordpress.android.fluxc.generated.AccountActionBuilder;
 import org.wordpress.android.fluxc.generated.endpoint.WPCOMREST;
+import org.wordpress.android.fluxc.generated.endpoint.WPCOMV2;
 import org.wordpress.android.fluxc.model.AccountModel;
 import org.wordpress.android.fluxc.network.BaseRequest.BaseErrorListener;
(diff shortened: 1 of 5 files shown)


nbradbury:
@theck13 This looks good, but do you think we should add a test to fetch suggestions for a non-empty name? Right now the only test is fetching them for an empty name.

theck13:
> do you think we should add a test to fetch suggestions for a non-empty name?

That's good idea.  I added a successful test in 4009e3ebdcbf33854d2ab2cd55f3f7f56c510415.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -218,6 +233,43 @@ public void onErrorResponse(@NonNull BaseNetworkError error) {
...
+     *
+     * @param name  Text (e.g. display name) from which to create username suggestions
+     */
+    public void fetchUsernameSuggestions(@NonNull String name) {
+        String url = WPCOMV2.users.username.suggestions.getUrl();
+
+        Map<String, String> params = new HashMap<>();
+        params.put("name", name);
+
+        add(WPComGsonRequest.buildGetRequest(url, params, UsernameSuggestionsResponse.class,
```
aforcier: Does this need to be an authed request @theck13? Or can we use `addUnauthedRequest`?

theck13: It _can_ in fact be an unauthenticated request.  I updated the method in ec186b035a9719daddb2e036e7f91cf80c804df3.

nbradbury:
:shipit:

PR #689, opened 2018-01-18
issue/688-jetpack-wordpress-version

### LINKED ISSUE #688: Store WordPress version for Jetpack sites
The "software version" for Jetpack sites is currently null. From Slack:

> looks like the reason is that we’re only populating the `mSoftwareVersion` field for XML-RPC site fetches, and that looks like a Jetpack site fetched over the rest API
if we wanted to store the WP version in that case, it looks like there’s a `options.software_version` field the API returns that we can use, we’d just need to add it to `SiteWPComRestResponse`:
https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/3435314117ce7632070ecabed71d635ac061b328/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteWPComRestResponse.java#L12
(and store it to `mSoftwareVersion` in the `SiteRestClient`)

nbradbury:
Fixes #688 - `software_version` is now stored for Jetpack sites.


aforcier:
:shipit:

PR #699, opened 2018-01-24
Add checkstyle rules for @Inject annotation

aforcier:
Adds a few new checkstyle rules to standardize some syntax around `@Inject` that have come up several times.

#### Allowed:

```java
@Inject AccountStore mAccountStore;
```

```java
@Inject protected AccountStore mAccountStore;
```

```kotlin
@Inject internal lateinit var accountStore: AccountStore
```

(Note: There's a style rule for Kotlin that will flag anything like `internal @Inject var ...` - we could make this an error, but it wouldn't cover java files anyway, and is redundant with the checkstyle rule.)

#### Not allowed:

```java
protected @Inject AccountStore mAccountStore;
```

The checkstyle for this excludes any lines including `class`, since this is valid Kotlin:

```kotlin
class Something @Inject constructor(...) {

}
```

#### Allowed:

```java
@Inject AccountStore mAccountStore;
```

#### Not allowed:

```java
@Inject
AccountStore mAccountStore;
```

This one flagged one false positive in `CommentStore` where the type was separated from the name of a constructor field - we should probably not allow this anyway, but since this was the only false positive I just fixed it manually.

I tested these rules in WPAndroid and WCAndroid, and they seem to catch all the problematic cases with no false positives.

cc @maxme (more work for project easter bunny, I guess!)

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -57,6 +57,11 @@
         <property name="severity" value="error"/>
     </module>
 
+    <module name="RegexpSingleline">
+        <property name="format" value="^(?!.*\bclass\b).*\S.*@Inject"/>
+        <property name="message" value="In-line inject annotation should precede the rest of the declaration"/>
+    </module>
+
     <!-- Specific to FluxC -->
     <module name="RegexpSingleline">
         <property name="format" value="extends Payload[^&lt;]"/>
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -58,7 +58,7 @@
     </module>
 
     <module name="RegexpMultiline">
-        <property name="format" value="@Inject\n.*[^{,(]$"/>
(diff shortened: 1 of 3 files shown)


aforcier:
Also updated the shared IDEA configs to prevent the IDE from auto-line wrapping annotations.

config/checkstyle.xml
```diff
@@ -57,6 +57,16 @@
         <property name="severity" value="error"/>
     </module>
 
+    <module name="RegexpMultiline">
+        <property name="format" value="@Inject(\n|\r\n).*[^{,(]$"/>
```
maxme: +1 for `\n|\r\n`, regex looks good since we'll never have inject fields that require affectation (and thus `()` or `{}`)

maxme:
:shipit:

PR #703, opened 2018-01-29
AccountStore connected test tweaks

aforcier:
Some improvements to the connected `AccountStore` tests:

* Ensure we start with consistent state (e.g., logged out if we're testing logging in)
* Dispatches are no longer sent before `CountDownLatch`es were set - this was a race condition risk
* The one test that makes changes to the account (`testWPComPost()`) now makes sure we're logged into the test account first

Overall, we could make these 100% atomic by always making sure to log out and back in at the start of tests - I'm hesitant to do that since many of the tests don't rely on being logged into any particular account, and all that extra network activity will slow down the tests. We can do that if we're running into problems otherwise.

cc @maxme - not sure whether this addresses the issues you were seeing, might need more info from you

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_AccountTest.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_AccountTest.java
@@ -66,6 +66,10 @@ public class ReleaseStack_AccountTest extends ReleaseStack_Base {
     }
 
     public void testWPComAuthenticationOK() throws InterruptedException {
+        if (mAccountStore.hasAccessToken()) {
+            signOut();
+        }
+
         mNextEvent = TestEvents.AUTHENTICATE;
         authenticate(BuildConfig.TEST_WPCOM_USERNAME_TEST1, BuildConfig.TEST_WPCOM_PASSWORD_TEST1);
     }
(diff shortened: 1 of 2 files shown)


maxme:
> cc @maxme - not sure whether this addresses the issues you were seeing, might need more info from you

No, that doesn't solve the problem I get, I'll investigate on my side.

aforcier:
@maxme I think I found the culprit, I had `testChangeWPComUsernameInvalidInputError` fail when running `cAT` on a different branch. This looks suspicious:

screen shot 2018-02-12 at 6 57 50 pm

Fixed in 8c4cb65 and 7b093de by making sure the mock successful account login test clears out the account store after it's done.

aforcier:
Updated with the JUnit changes from #721.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_AccountTest.java
```diff
@@ -106,15 +111,22 @@ public void testWPComPost() throws InterruptedException {
...
+
         mNextEvent = TestEvents.POSTED;
         PushAccountSettingsPayload payload = new PushAccountSettingsPayload();
         String newValue = String.valueOf(System.currentTimeMillis());
         mExpectAccountInfosChanged = true;
         payload.params = new HashMap<>();
         payload.params.put("description", newValue);
-        mDispatcher.dispatch(AccountActionBuilder.newPushSettingsAction(payload));
         mCountDownLatch = new CountDownLatch(1);
+        mDispatcher.dispatch(AccountActionBuilder.newPushSettingsAction(payload));
```
maxme: That wasn't the issue, but +1

maxme:
```
$ ./gradlew :example:cAT -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.release.ReleaseStack_AccountTest
Starting 19 tests on [...]
BUILD SUCCESSFUL in 36s
```

Looks good :shipit: - I'll try to run the test suite in Firebase.

PR #721, opened 2018-02-23
Replace InstrumentationTestCase by JUnit tests and use the test orchestrator 

maxme:
Purpose of the PR is to replace InstrumentationTestCase by JUnit tests. Tests will run independently and if one crashes, the test suite will continue.

This PR replaces the previous #702

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -80,7 +80,8 @@ dependencies {
     // WordPress libs
     implementation ('org.wordpress:utils:1.16.0') {
         // Using official volley package
-        exclude group: "com.mcxiaoke.volley";
+        exclude group: "com.mcxiaoke.volley"
+        exclude group: "com.android.support"
     }
 
     // Dagger
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -100,12 +100,11 @@ dependencies {
     androidTestImplementation 'org.mockito:mockito-core:1.10.19'
     kaptAndroidTest "com.google.dagger:dagger-compiler:$daggerVersion"
     androidTestCompileOnly 'org.glassfish:javax.annotation:10.0-b28'
+    // Test orchestrator
+    androidTestImplementation 'com.android.support.test:runner:1.0.1'
+    androidTestUtil 'com.android.support.test:orchestrator:1.0.1'
 
     // Debug dependencies
     debugImplementation 'com.facebook.stetho:stetho:1.5.0'
     debugImplementation 'com.facebook.stetho:stetho-okhttp3:1.5.0'
-
-    // Test orchestrator
-    androidTestImplementation 'com.android.support.test:runner:1.0.1'
-    androidTestUtil 'com.android.support.test:orchestrator:1.0.1'
 }
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -41,7 +41,10 @@ android {
     }
 
     testOptions {
-        execution 'ANDROID_TEST_ORCHESTRATOR'
+        // Don't use the test orchestrator yet, some of our connected testsare sharing state to reduce network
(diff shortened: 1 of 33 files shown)


example/build.gradle
```diff
@@ -100,4 +104,8 @@ dependencies {
     // Debug dependencies
     debugImplementation 'com.facebook.stetho:stetho:1.5.0'
     debugImplementation 'com.facebook.stetho:stetho-okhttp3:1.5.0'
+
+    // Test orchestrator
+    androidTestImplementation 'com.android.support.test:runner:1.0.1'
```
aforcier: Should we group these with the rest of the `androidTestImplementation`s a bit further up?

aforcier:
Still getting through reviewing the functionality, but I noticed 6 tests missing annotations: one in `ReleaseStack_SiteTestWPCom.java`, one in `ReleaseStack_PluginTestJetpack.java`, and four in `ReleaseStack_WPOrgPluginTest.java`.

Also there's a checkstyle violation 😀

aforcier:
@maxme It looks like the switch to the orchestrator has made a pretty big change: the `sSite` static field in `ReleaseStack_WPComBase` now seems to be reset after every test method, which means we're doing a full auth/account fetch/site fetch for each test (because [this](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/7f8ffac5285b62b0c48e6a8121bf25ae7eca795b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WPComBase.java#L48) is always false), which is slowing down the tests (and probably opening us up to rate limiting).

(I was wondering why `testChangeWPComUsernameInvalidInputError` never failed in this branch without the fix from #703, this is why.)

maxme:
> @maxme It looks like the switch to the orchestrator has made a pretty big change: the sSite static field in ReleaseStack_WPComBase now seems to be reset after every test method, which means we're doing a full auth/account fetch/site fetch for each test (because this is always false), which is slowing down the tests (and probably opening us up to rate limiting).

Ok, I'll have a look at this, hopefully we can just load the account in the init method.

maxme:
> Still getting through reviewing the functionality, but I noticed 6 tests missing annotations: one in ReleaseStack_SiteTestWPCom.java, one in ReleaseStack_PluginTestJetpack.java, and four in ReleaseStack_WPOrgPluginTest.java.

Fixed in e01036d - and double checked with:

```shell
ag -o --nofilename '@Test[^a]+void test.*\(' | xargs -n 4
```

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_TaxonomyTestWPCom.java
```diff
@@ -382,7 +382,7 @@ private void deleteTerm(TermModel term) throws InterruptedException {
         assertTrue(mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
     }
 
-    private void testUpdateExistingTerm(TermModel term) throws InterruptedException {
+    private void checkUpdateExistingTerm(TermModel term) throws InterruptedException {
```
aforcier: The naming of this one threw me off too, good call 👍

aforcier:
Per our discussion, it makes sense to merge this and bring in the JUnit changes even with the orchestrator disabled. When we find a way to 'hide' the `SiteModel` from the orchestrator so we can reuse it on different tests, we can turn it back on.

(Note: This takes us back to tests failing in `ReleaseStack_AccountTest` when running `gradlew cAT`, but I confirmed that the work in #703 still works as a fix combined with this branch 👍.)

:shipit:

PR #731, opened 2018-03-01
Have /domains/suggestions/ call be authenticated

### LINKED ISSUE #730: SiteRestClient.suggestDomains call need to be authenticated
Even though the `/domains/suggestions/` call can be performed unauthenticated as well, the authenticated can return different results. Related to https://github.com/wordpress-mobile/WordPress-Android/issues/7336 in particular, the call needs to be authenticated so it returns the exact match for user's username if that's available (e.g. for a site-less user).

hypest:
Fixes #730 

* Have `suggestDomains()` do its call authenticated
* Move the `suggestDomains()` call up in the module to be closer to the rest of the authenticated calls source code

### DIFF
--- a/example/src/main/res/layout/fragment_signed_out_actions.xml
+++ b/example/src/main/res/layout/fragment_signed_out_actions.xml
@@ -39,7 +39,7 @@
         android:id="@+id/domain_suggestions"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:text="Domain suggestions" />
+        android:text="Domain suggestions (results differ if authenticated)" />
 
     <Button
         android:id="@+id/connect_site_info"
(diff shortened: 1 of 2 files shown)


maxme:
The patch looks good, but I don't see any difference in the results (even when requesting 36 names). I do see the token in the request, I wonder if another parameter was needed in the request or if the API don't care about my test username?

Can we update the label on the Example app button? something like `Domain suggestions (different results when authenticated)`?

hypest:
I should be more analytical in what's the difference in response from the API. So, to see the diff you need to login with an account that has no sites yet. Because the "username.wordpress.com" domain is reserved for usage only for that account, it won't be offered as result if the call is un-authenticated. Instead, the API will indeed offer it if the call bears an auth token for that particular account. Does that make sense?

hypest:
> Can we update the label on the Example app button? something like Domain suggestions (different results when authenticated)?

Makes sense, will do that.

hypest:
Updated the button label with 41b5c43.

maxme:
Ah ok, I was expecting other names related to my username (like `username-catblog.wordpress.com`). :shipit:

PR #736, opened 2018-03-07
Import checkstyle and code style config changes from WPAndroid

aforcier:
Brings in the latest changes from WPAndroid, consolidating new changes here with those we made for the refactor on that side.

Once this is done, we can update https://github.com/Automattic/style-config-android, and import the fetchstyle gradle plugin in both projects 🎉

cc @maxme

### DIFF
--- a/.idea/checkstyle-idea.xml
+++ b/.idea/checkstyle-idea.xml
@@ -3,12 +3,12 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:FluxC Style" />
+        <entry key="active-configuration" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:a8c Style" />
         <entry key="checkstyle-version" value="8.2" />
         <entry key="copy-libs" value="false" />
         <entry key="location-0" value="BUNDLED:(bundled):Sun Checks" />
         <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
-        <entry key="location-2" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:FluxC Style" />
+        <entry key="location-2" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:a8c Style" />
         <entry key="scan-before-checkin" value="false" />
         <entry key="scanscope" value="AllSourcesWithTests" />
         <entry key="suppress-errors" value="false" />
--- a/.idea/checkstyle-idea.xml
+++ b/.idea/checkstyle-idea.xml
@@ -10,7 +10,7 @@
         <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
         <entry key="location-2" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:a8c Style" />
         <entry key="scan-before-checkin" value="false" />
-        <entry key="scanscope" value="JavaOnlyWithTests" />
+        <entry key="scanscope" value="AllSourcesWithTests" />
         <entry key="suppress-errors" value="false" />
       </map>
     </option>
--- a/.idea/checkstyle-idea.xml
+++ b/.idea/checkstyle-idea.xml
@@ -10,7 +10,7 @@
         <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
         <entry key="location-2" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:a8c Style" />
         <entry key="scan-before-checkin" value="false" />
-        <entry key="scanscope" value="AllSourcesWithTests" />
+        <entry key="scanscope" value="JavaOnlyWithTests" />
         <entry key="suppress-errors" value="false" />
       </map>
     </option>
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -3,6 +3,7 @@
   <component name="ProjectCodeStyleSettingsManager">
     <option name="PER_PROJECT_SETTINGS">
       <value>
+        <option name="FIELD_NAME_PREFIX" value="m" />
         <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
         <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
         <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -30,10 +30,14 @@
             <emptyLine />
           </value>
         </option>
+        <option name="ENABLE_JAVADOC_FORMATTING" value="false" />
         <option name="RIGHT_MARGIN" value="120" />
         <AndroidXmlCodeStyleSettings>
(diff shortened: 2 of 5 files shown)


aforcier:
Downgraded to Java-only for the checkstyle plugin since it's causing problems in WPAndroid with `.js` and `.sql` files and we want to keep things consistent. Hopefully we can change this in the future and target `.java`, `.kt`, and `.xml` files specifically. In the meantime, the `checkstyle` task is still covering those files.

maxme:
I updated the checkstyle config to allow _//noinspection_ comments - see https://github.com/wordpress-mobile/WordPress-Android/pull/7430

.idea/checkstyle-idea.xml
```diff
@@ -3,14 +3,14 @@
...
+        <entry key="active-configuration" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:a8c Style" />
         <entry key="checkstyle-version" value="8.2" />
         <entry key="copy-libs" value="false" />
         <entry key="location-0" value="BUNDLED:(bundled):Sun Checks" />
         <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
-        <entry key="location-2" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:FluxC Style" />
+        <entry key="location-2" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:a8c Style" />
         <entry key="scan-before-checkin" value="false" />
-        <entry key="scanscope" value="AllSourcesWithTests" />
+        <entry key="scanscope" value="JavaOnlyWithTests" />
```
maxme: Sad but necessary.

Ref. https://github.com/jshiell/checkstyle-idea/issues/386

.idea/codeStyleSettings.xml
```diff
@@ -3,6 +3,7 @@
   <component name="ProjectCodeStyleSettingsManager">
     <option name="PER_PROJECT_SETTINGS">
       <value>
+        <option name="FIELD_NAME_PREFIX" value="m" />
```
maxme: Should we add `s` for static field names?

.idea/codeStyleSettings.xml
```diff
@@ -30,10 +31,14 @@
             <emptyLine />
           </value>
         </option>
+        <option name="ENABLE_JAVADOC_FORMATTING" value="false" />
```
maxme: Note I changed that to run `Restyle code` without changing every javadoc, but we could re-enable it.

config/checkstyle.xml
```diff
@@ -52,7 +52,7 @@
     </module>
 
     <module name="RegexpSingleline">
-        <property name="format" value="@[\S]*(\([{]?\&quot;.*\&quot;[}]?\)++|\([^\&quot;]*\)++)[^ \t]"/>
+        <property name="format" value="[^\S\r\n]@[\S]*(\([{]?\&quot;.*\&quot;[}]?\)++|\([^\&quot;]*\)++)[^ \t]"/>
```
maxme: 👍 This change was made to avoid matching `@` in strings, like in `emailAddress.split("@")`.

config/checkstyle.xml
```diff
@@ -74,25 +74,27 @@
...
         <property name="message" value="Network clients must be annotated with @Singleton"/>
     </module>
 
-    <!-- Specific to FluxC / WordPress -->
+    <!-- Specific to WordPress -->
     <module name="RegexpSingleline">
       <property name="format" value=".*dotcom(?!.*checkstyle ignore)"/>
       <property name="ignoreCase" value="true"/>
-      <property name="message" value="Kim is not happy, you should not use _dotcom_, use _wpcom_ instead"/>
+      <property name="message" value="You should not use _dotcom_, use _wpcom_ instead"/>
```
maxme: 😭

config/checkstyle.xml
```diff
@@ -74,25 +74,27 @@
...
 
+    <!-- Specific to FluxC -->
     <module name="RegexpMultiline">
         <property name="format" value="(?&lt;!^@Singleton)\n.*extends Store[\s{]"/>
         <property name="message" value="Stores must be annotated with @Singleton"/>
     </module>
 
+    <!-- Specific to FluxC -->
     <module name="RegexpMultiline">
         <property name="format" value="(?&lt;!^@Singleton)\n.*extends \w.*Client[\s{]"/>
```
maxme: In utils, this fails for the class `WPWebChromeClient`  in `libs/utils/WordPressUtils/src/main/java/org/wordpress/android/util/helpers/WPWebChromeClient.java`. Maybe we should specify this test for class`BaseXMLRPCClient` and `BaseWPComRestClient` (and eventually change this when we add new base classes)

maxme: Actually there are many other cases of classes extending `WebViewClient` and `WebChromeClient` in wpandroid.

.idea/codeStyleSettings.xml
```diff
@@ -72,11 +74,35 @@
...
         </Objective-C-extensions>
         <XML>
-          <option name="XML_KEEP_LINE_BREAKS" value="false" />
-          <option name="XML_ALIGN_ATTRIBUTES" value="false" />
-          <option name="XML_SPACE_INSIDE_EMPTY_TAG" value="true" />
+          <option name="XML_LEGACY_SETTINGS_IMPORTED" value="true" />
         </XML>
         <codeStyleSettings language="JAVA">
+          <option name="ALIGN_MULTILINE_CHAINED_METHODS" value="true" />
+          <option name="ALIGN_MULTILINE_PARAMETERS_IN_CALLS" value="true" />
```
maxme: We should remove that line

Ref. https://github.com/wordpress-mobile/WordPress-Android/pull/7388#issuecomment-371436745
cc @hypest

maxme:
- [ ] Enable javadoc formatting?
- [ ] Static field names prefixed by `s`?
- [ ] Remove `ALIGN_MULTILINE_PARAMETERS_IN_CALLS` from AS style config

.idea/codeStyleSettings.xml
aforcier: Sure, done in a20de11.

.idea/codeStyleSettings.xml
aforcier: It's back: 8944fd5.

config/checkstyle.xml
aforcier: Done in 59bc97f...kinda wishing for separate configs more now, pretty easy to add a new client and never catch this 😞.

Think I'll look at that as a possibility after this PR.

.idea/codeStyleSettings.xml
aforcier: Done in 31ea5e1.

aforcier:
* [ ] Use `fileExtensions` and `BeforeExecutionExclusionFileFilter` settings in `checkstyle.xml` to restrict to `java`/`kt`/`xml` files and exclude `gen` folders
* [ ] Restore 'all sources' checkstyle plugin config
* [ ] Drop the (now redundant) include/excludes from the gradle `checkstyle` task config

config/checkstyle.xml
maxme: Right, so maybe `Base.*Client` would be less risky?

aforcier: Yeah that's safer, updated in 223e32d.

aforcier:
Ready for review!

aforcier:
Added an inspection rule that was missing from FluxC in 0385b6e, allows us to use backticks in method names in kotlin tests.

maxme:
Looks good - :shipit:

PR #740, opened 2018-03-08
Localize account creation

hypest:
Addresses the account related part of #609 by setting the `locale` body param of the ~`/users/new` and~ `auth/send-signup-email`.

To test:
1. Put the device on the French language/locale
2. Run the example app and head over to the "Signed out actions" section
3. Tap on "Send magic link sign e-mail" and input an email you have access to but *not* already registered with WPCOM and hit "Next"
4. Access that email account and verify that a new email has arrived and it's in French

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
@@ -18,8 +18,9 @@ import org.wordpress.android.util.LanguageUtils;
 
 public abstract class BaseWPComRestClient {
     private AccessToken mAccessToken;
-    private final Context mAppContext;
     private final RequestQueue mRequestQueue;
+
+    protected final Context mAppContext;
     protected final Dispatcher mDispatcher;
     protected UserAgent mUserAgent;
 
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
@@ -18,9 +18,9 @@ import org.wordpress.android.util.LanguageUtils;
 
 public abstract class BaseWPComRestClient {
     private AccessToken mAccessToken;
+    private final Context mAppContext;
     private final RequestQueue mRequestQueue;
 
-    protected final Context mAppContext;
     protected final Dispatcher mDispatcher;
     protected UserAgent mUserAgent;
 
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedNetworkModule.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedNetworkModule.java
@@ -68,8 +68,9 @@ public class MockedNetworkModule {
 
     @Singleton
     @Provides
-    public Authenticator provideAuthenticator(Dispatcher dispatcher, AppSecrets appSecrets, RequestQueue requestQueue) {
-        Authenticator authenticator = new Authenticator(dispatcher, requestQueue, appSecrets);
+    public Authenticator provideAuthenticator(Context appContext, Dispatcher dispatcher, AppSecrets appSecrets,
+                                              RequestQueue requestQueue) {
+        Authenticator authenticator = new Authenticator(appContext, dispatcher, requestQueue, appSecrets);
         Authenticator spy = spy(authenticator);
 
         // Mock Authenticator with correct user: test/test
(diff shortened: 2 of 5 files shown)


aforcier:
@hypest tested this in French and Japanese. It works well for account creation, but doesn't seem to be for magic link signup e-mail. The request looks right:

```json
{
	"client_secret": "secret",
	"locale": "ja",
	"email": "email-address",
	"client_id": "id"
}
```

But the e-mail always arrives in English. Is it possible something changed server-side?

hypest:
> Is it possible something changed server-side?

Oops, my bad. The server side changes for the signup-email are not deployed to production yet! (For reference, the Phabricator diff is: D10746).

I will raise the `Not Ready for Review` label until it gets deployed and we can take it from there. Thanks for the review so far @aforcier and sorry for the trouble :(

yurynix:
Hi :wave:,
Most WPCOM REST endpoints accept `locale`, but some, such as `/users/new` are an outlier for some reason.

However, `/auth/send-signup-email`, IMO, should follow whatever the other endpoints are doing and get the locale via query arg.

I've tried to push a commit to change that in this PR, but couldn't 😭 , here's the patch:

```diff
diff --git a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/auth/Authenticator.java b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/auth/Authenticator.java
index 73877815..40147dc5 100644
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/auth/Authenticator.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/auth/Authenticator.java
@@ -3,6 +3,7 @@ package org.wordpress.android.fluxc.network.rest.wpcom.auth;
 import android.content.Context;
 import android.support.annotation.NonNull;
 import android.text.TextUtils;
+import android.net.Uri;
 
 import com.android.volley.NetworkResponse;
 import com.android.volley.ParseError;
@@ -210,12 +211,15 @@ public class Authenticator {
         String url = payload.isSignup ? WPCOMREST.auth.send_signup_email.getUrlV1_1()
                 : WPCOMREST.auth.send_login_email.getUrlV1_3();
 
+        url = Uri.parse(url).buildUpon()
+                .appendQueryParameter("locale", LanguageUtils.patchDeviceLanguageCode(LanguageUtils.getCurrentDeviceLanguage(mAppContext).getLanguage()))
+                .build()
+                .toString();
+
         Map<String, Object> params = new HashMap<>();
         params.put("email", payload.emailOrUsername);
         params.put("client_id", mAppSecrets.getAppId());
         params.put("client_secret", mAppSecrets.getAppSecret());
-        params.put("locale", LanguageUtils.patchDeviceLanguageCode(
-                LanguageUtils.getCurrentDeviceLanguage(mAppContext).getLanguage()));
 
         WPComGsonRequest request = WPComGsonRequest.buildPostRequest(url, params, AuthEmailWPComRestResponse.class,
                 new Response.Listener<AuthEmailWPComRestResponse>() {
```

yurynix:
D10746 deployed in prod today

hypest:
Changes since last time:
* Backend has changed the implementation and now accepts the `locale` param passed as a query parameter like the rest of the API endpoints
* The `lang-Country` format is now accepted as normal
* All those changes are deployed to production so, no need to test against a sandbox
* Only the `Authenticator` needs change after all, to include the `locale` param in the query

I've removed the Not Ready For Review flag @aforcier , ready for another pass, thanks!

aforcier:
@hypest: Gave this another try, but both new user signup and magic link signup e-mails are sending e-mails in English, even though I have the device set to French and `?locale=fr_CA` is appended to the request URL.

aforcier:
Hmm looks like the issue now is that `fr_CA` is unsupported, maybe that's expected behavior?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
```diff
@@ -18,8 +18,9 @@
 
 public abstract class BaseWPComRestClient {
     private AccessToken mAccessToken;
-    private final Context mAppContext;
     private final RequestQueue mRequestQueue;
+
+    protected final Context mAppContext;
```
aforcier: Looks like we can return this one back to `private` now.

aforcier:
To summarize our conversation, it looks like this is working for magic link signup emails, except some regions (like `fr_CA`) aren't supported - but this looks to be a backend issue and not something we'll need to address client-side.

`/users/new/` is not localizing at all (and seems to still required a `locale=xy` parameter in the body to work), but we'll bring this up too, and anyway this is the already existing behavior of FluxC. So, we'll consider this PR to only update magic link signup (which is all the WordPress app is using now anyway).

aforcier:
Just one tiny comment and then good to go @hypest!

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
hypest: Defo! Done with 36232c7.

aforcier:
:shipit:

PR #746, opened 2018-03-09
Add checks for disk space quota before media uploads.

SergioEstevao:
This PR adds an extra check before upload that check if the file is larger than the maximum space available on the disk quota for the site.


aforcier:
Works great! Nice enhancement @SergioEstevao :shipit:

PR #747, opened 2018-03-12
Update to ktlint 0.21.0

aforcier:
Updates `ktlint` to `0.21.0` and fixes new style violations introduced.

### Main changes:

1. Functions that can't fit on a single line must have each of their parameters on a separate line, indented with normal (4 space) indentation:

```
fun newInstance(
    onClickListener: Listener,
    text1Hint: String,
    text2Hint: String,
    text3Hint: String
): ThreeEditTextDialog {
    ...
}
```

This one is also enforced in the AS formatter via changes to the shared IDEA config file.

2. Operators need to appear before a wrapped line, not after:

```
prependToLog("error: " + event.error.type + "Some more text here: " +
        event.error.message)
```

This is, sadly, the opposite of our rule for Java (which is a checkstyle default), which forces the operators to appear at the beginning of the new line. There's some hope that this may be addressed in the future, but for now we'll have to live with the inconsistency.

cc @maxme

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -47,6 +47,7 @@
           </option>
           <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="2147483647" />
           <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="2147483647" />
+          <option name="CONTINUATION_INDENT_IN_PARAMETER_LISTS" value="false" />
         </JetCodeStyleSettings>
         <Objective-C-extensions>
           <file>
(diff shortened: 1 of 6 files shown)


aforcier:
For a more substantial demo of the changes, see https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/eabb35b33c805d422b0bdb0414dd3fc69ab9571e, which applies the new rules to the `woocommerce-store` branch.

shyiko:
Hey @aforcier. 
Just in case you missed https://github.com/shyiko/ktlint/issues/163:  
+/- is wrapping is unlikely to change due to the following - https://github.com/shyiko/ktlint/issues/163#issuecomment-369418775. 
I know it's different from checkstyle's OperatorWrap default but there is little we can do (as unfortunate as it is).

aforcier:
Thanks @shyiko! Yes I read that, I was hopeful that the Android style guide might end up deciding to keep at least `&&`, `||` on the break-before-operator behavior, and make `+`/`-` exceptions since there's the unary operator problem. (In which case ktlint would probably be updated to reflect that, per your comment.) I probably should have used `&&` in my example above since `+` probably won't change 😀

aforcier:
With #754 merged, rebased this branch and fixed new ktlint violations from code in that branch.

maxme:
- `./gradlew ktlint`: ✅ 
- Sad 🐼 for the trailing `+`: ✅

maxme:
:shipit:

PR #752, opened 2018-03-14
Reset the site object for each WP.com connected test class

aforcier:
Prevents an issue when running all connected tests, where the Jetpack media tests log into a different account, but then the WP.com media tests attempt to use the same account and the existing `sSite`, which was set from the last WP.com tests, causing an auth error. This is a clean way to ensure each test class starts with a fresh WP.com session (but doesn't log out and back in for each individual test).

cc @maxme


maxme:
Looks good - :shipit:

PR #753, opened 2018-03-14
Limit theme activation tests to _Twenty ..._ themes

maxme:
This prevents the test  `ReleaseStack_SiteTestWPCom > testFetchPostFormats` from failing when the theme is switched to a non-post formats supported theme.


aforcier:
Smooth fix. Combined with #752, `./gradlew cAT` is consistently passing for me. 🎉

:shipit:

PR #754, opened 2018-03-15
Add WooCommerce plugin and basic store

aforcier:
Adds the WooCommerce plugin with basic setup, along with the Jetpack tunnel network request.

All code in this PR has already been reviewed - the PRs were #619, #635, and #642. The only change has been to update the tests to use the new JUnit setup (d09bdeb).

### Summary of changes:
* A new module, `:plugin:woocommerce`
* Two new fields from the WP.com API, is_wpcom_store, woocommerce_is_active, stored in the SiteModel, to allow us to identify Jetpack sites with WooCommerce installed
* A tweak to the endpoint class generator to support multiple modules calling fluxc-processor
* A demo on use of the plugin in the example app
* A new type of network request: `WPComJPTunnelGsonRequest`. This request takes calls to WP-API endpoints (intended for self-hosted sites), and tunnels them through the WordPress.com /jetpack-blogs/$site/rest-api/ endpoint.
* Brings in addon support for Wellsql (https://github.com/wordpress-mobile/wellsql/pull/11), and sets up the first WooCommerce table using this system. See the description of #642 for some more background and caveats.

This makes some changes to `WellSqlConfig` I thought I should highlight here.

One, [we don't need to add new Models to a list in `WellSqlConfig` anymore](https://github.com/wordpress-mobile/WordPress-FluxC-Android/compare/woocommerce-store?expand=1#diff-7b4c47eb749776091657a82024f162ceL42) - models annotated with `@Table` are automatically picked up, according to the active addons.

Two, migrations for addon tables a bit special. The db version number should increase, and a new entry should be added to `onUpgrade()`, but it should call `migrateAddOn()` instead of doing the migration in place. `migrateAddOn()` does its own database migration check, and conditionally upgrades the database if the addon corresponding to that migration is currently enabled (this is to avoid attempting to modify tables that don't exist). Apps not using that addon (e.g., the WordPress app), will just process this as a database version bump, with no actual changes.

cc @maxme


aforcier:
Note: Conflicts with https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/747 - when one is merged, the other should be updated with lint fixes to prevent breaking `ktlint` in `develop`.

maxme:
Code has been reviewed already and tests are OK:
```
$ ./gradlew :example:cAT -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.mocked.MockedStack_JetpackTunnelTest
Starting 3 tests on [...]
BUILD SUCCESSFUL in 12s
```

:shipit:

PR #756, opened 2018-03-16
Activity Log store - Part 1 - API + models

planarvoid:
First part of Activity store Data layer - written in Kotlin. It is based on normal implementation of Model, RestClient and Store with some tweaks. The PR is little bit too big but it should be fairly easy to review. Let's take this PR as a discussion point about some fined details of Kotlin standards. My standards are based on many discussions we had when we were moving to Kotlin in the past. 

I'll try to describe some questionable parts with focus on Kotlin. If there is anything that seems out of order, let me know.  

**Model**
I'm using Kotlin `data classes` for the model. You can read more about the advantages here - https://antonioleiva.com/data-classes-kotlin/ They are very concise and help us keep our objects immutable. From my experience immutability is extremely important, especially with parallel environment and with regards to going functional one day. 
WellSql does not support immutable objects (and fields) so I'm proposing a concept of mutable builders (for the sake of WellSql and easier object creation). Builder is an internal class that contains all the fields that are not logically part of the object (in this case for example `siteId` for an `ActivityModel`). 
**RestClient**
This class should be pretty straightforward. I don't think that inheritance helps this class and I would prefer using a composition pattern. Currently it's not really unit testable. However, my implementation just follows our current standards. The `ActivityRestClient` is currently missing the storage logic which will come in another PR. 
**Testing**
Unit testing should be fairly simple. Integration tests are currently happening in `ReleaseStack_ActivityTestWPCom` which is still Java. I'm having an issue with Kotlin integration tests where I'm not able to run just one test/file directly from IDE (it considers the test as Unit test and fails with `Empty test suite`). However, it is possible to run the whole package of tests. If anyone knows a solution, let me know.


aforcier:
> I'm using Kotlin data classes for the model. You can read more about the advantages here - https://antonioleiva.com/data-classes-kotlin/ 

Yeah, data classes were one of the biggest reasons we were excited for Kotlin in FluxC, so much less boilerplate.

> WellSql does not support immutable objects (and fields) so I'm proposing a concept of mutable builders (for the sake of WellSql and easier object creation). Builder is an internal class that contains all the fields that are not logically part of the object (in this case for example siteId for an ActivityModel).

I like this approach, though the resulting syntax from working around WellSql seems pretty troublesome (SQL queries return `*Model.Builder` instead of `*Model`, for example)...just one more reason to think about a move to Room at some point.

It's difficult to evaluate this and look for problems since the models are not being used as database tables at this stage. I'll have to experiment with this a little (or wait for a later PR of yours that will demonstrate it).

> I'm having an issue with Kotlin integration tests where I'm not able to run just one test/file directly from IDE (it considers the test as Unit test and fails with Empty test suite).

Yes, I've run into this too, it appears to be an AS bug. You can get it to work by manually setting up a run configuration for that test in AS as an instrumented test. You can also just run single tests from the commandline:

```
./gradlew cAT -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.release.ReleaseStack_ActivityTestWPCom
```

We already [have a connected test in Kotlin](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/80c252666e397f638b1228261f276b702d95d9e7/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_JetpackTunnelTest.kt), with some more on the way, so don't feel like you need to write them in Java if you don't want to.

fluxc-annotations/src/main/java/org/wordpress/android/fluxc/annotations/endpoint/WPComEndpoint.java
```diff
@@ -7,6 +7,7 @@
     private static final String WPCOM_PREFIX_V1_2 = WPCOM_REST_PREFIX + "/rest/v1.2";
     private static final String WPCOM_PREFIX_V1_3 = WPCOM_REST_PREFIX + "/rest/v1.3";
     private static final String WPCOM_PREFIX_V0 = WPCOM_REST_PREFIX;
+    private static final String WPCOM_PREFIX_V2 = WPCOM_REST_PREFIX + "/wpcom/v2";
```
aforcier: The "V2" WP.com endpoints actually live in `WPComV2Endpoint`, which corresponds to a generated `WPCOMV2` class. This change to `WPComEndpoint` can be dropped and the `/activity-log/$site/rewind/to/$rewind_ID#String` endpoint moved to `wp-com-v2-endpoints.txt`.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityStore.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.wordpress.android.fluxc.model.activity.ActivityModel
+import org.wordpress.android.fluxc.model.activity.RewindStatusModel
+import org.wordpress.android.fluxc.network.BaseRequest
+import org.wordpress.android.fluxc.network.rest.wpcom.activity.ActivityRestClient
+import org.wordpress.android.util.AppLog
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class ActivityStore
```
aforcier: I think the `Activity*` naming scheme can be a bit misleading, with `Activity` having significance on its own in Android and being pretty generic in general. How about `ActivityLogStore`, and so on?

planarvoid: I wasn't sure about this but I agree with you. Would you still call the objects `ActivityModel`? Or would you add the `Log` all over the place?

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityStore.kt
```diff
@@ -0,0 +1,104 @@
...
+@Inject constructor(private val mActivityRestClient: ActivityRestClient,
+                    dispatcher: Dispatcher) : Store(dispatcher) {
+    @Subscribe(threadMode = ThreadMode.ASYNC)
+    override fun onAction(action: Action<*>) {
+        val payload = action.payload ?: return
```
aforcier: This works for now, but it's possible to have payload-less actions. Checking the payload type instead of the action also seems error prone, as it's possible to have different actions that use the same payload type. Our pattern is to filter down to only actions relevant to the current store, and then switch through the action type (see [PostStore](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/80c252666e397f638b1228261f276b702d95d9e7/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java#L355-L360), for example).

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityStore.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.wordpress.android.fluxc.model.activity.RewindStatusModel
+import org.wordpress.android.fluxc.network.BaseRequest
+import org.wordpress.android.fluxc.network.rest.wpcom.activity.ActivityRestClient
+import org.wordpress.android.util.AppLog
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class ActivityStore
+@Inject constructor(private val mActivityRestClient: ActivityRestClient,
```
aforcier: Our convention is not to use `m`/`s` prefixes when defining Kotlin properties - this should just be `activityRestClient: ActivityRestClient`.

aforcier: (There are some other uses that need to be updated too elsewhere in the changeset.)

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityStore.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    private fun fetchActivitiesRewind(fetchActivitiesRewindPayload: FetchRewindStatePayload) {
+        mActivityRestClient.fetchActivityRewind(fetchActivitiesRewindPayload.site,
+                fetchActivitiesRewindPayload.number,
+                fetchActivitiesRewindPayload.offset)
+    }
+}
+
+// Payloads
+data class FetchActivitiesPayload(val site: SiteModel,
```
aforcier: Our convention is to nest these inside the `Store` class they belong to, since semantically they don't really belong to `org.wordpress.android.fluxc.store`, and they have the potential to cause import conflicts. Could you move these to be inner classes?

aforcier: Also, what is the reasoning for these being `data` classes? It seems like we could avoid all the auto-generated functions that come with data classes in the case of these types of objects.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityStore.kt
planarvoid: I was going full Kotlin and got carried away, good point 👍

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityStore.kt
planarvoid: yay 🎉

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityStore.kt
aforcier: I would go for `ActivityLogModel`, just for maximum clarity.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityStore.kt
planarvoid: I'm used to using `data` class as default for model. In this case I don't think it gives us much. What would you default to? `sealed` class or just default class?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityRestClient.kt
```diff
@@ -0,0 +1,225 @@
...
+@Inject constructor(appContext: Context, dispatcher: Dispatcher, requestQueue: RequestQueue,
+                    accessToken: AccessToken, userAgent: UserAgent)
+    : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
+    fun fetchActivity(site: SiteModel, number: Int, offset: Int) {
+        val url = WPCOMREST.sites.site(site.siteId).activity.urlV2
+        val pageNumber = offset / number + 1
+        val params = mapOf("page" to pageNumber.toString(), "number" to number.toString())
+        val request = buildGetRequest(
+                url, params, ActivitiesResponse::class.java,
+                {
```
aforcier: This is pretty minor, but I think readability here would be improved a bit by naming the function parameter (and not using `it`), so it's immediately clear that this is the successful response block and the below is the error block. Especially since this is still early days for Kotlin in FluxC and it would be best to keep things readable. Maybe `{ response ->`, and below `networkError`.

planarvoid: This is a good point! And it might also help people who are not used to Kotlin.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestWPCom.kt
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+import javax.inject.Inject
+
+import junit.framework.Assert.assertNotNull
+import junit.framework.Assert.assertTrue
+
+class ReleaseStack_ActivityLogTestWPCom : ReleaseStack_WPComBase() {
```
aforcier: These tests fail if the first site in the WordPress.com account specified in `TEST_WPCOM_USERNAME_TEST1`/`TEST_WPCOM_PASSWORD_TEST1` is not a Jetpack site, but a pure WordPress.com site. `ReleaseStack_WPComBase` is meant to be used for 'agnostic' tests that should work for both site types.

We normally extend `ReleaseStack_Base` instead, and handling auth yourself in this class, using the `TEST_WPCOM_USERNAME_SINGLE_JETPACK_ONLY`/`TEST_WPCOM_PASSWORD_SINGLE_JETPACK_ONLY` test credentials (here's [an example](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/d17252bb095762dc47c911023baf3561e3f09849/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java#L62-L81)).

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestWPCom.kt
```diff
@@ -0,0 +1,76 @@
...
+    private val incomingActions: MutableList<Action<*>> = mutableListOf()
+    @Inject lateinit var activityLogStore: ActivityLogStore
+    private var mBackup: CountDownLatch? = null
```
aforcier: What's the purpose of `mBackup`?

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.wordpress.android.fluxc.generated.ActivityActionBuilder
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.rest.wpcom.activity.ActivityLogRestClient
+
+@RunWith(MockitoJUnitRunner::class)
+class ActivityLogStoreTest {
+    @Mock private lateinit var activityLogRestClient: ActivityLogRestClient
+    @Mock private lateinit var dispatcher: Dispatcher
+    @Mock private lateinit var siteModel: SiteModel
+    private lateinit var mActivityLogStore: ActivityLogStore
```
aforcier: This should drop the `m` prefix.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -0,0 +1,225 @@
...
+import org.wordpress.android.fluxc.store.ActivityLogStore.FetchedActivitiesPayload
+import org.wordpress.android.fluxc.store.ActivityLogStore.FetchedRewindStatePayload
+import org.wordpress.android.fluxc.store.ActivityLogStore.RewindStatusError
+import org.wordpress.android.fluxc.store.ActivityLogStore.RewindStatusErrorType
+import java.util.Date
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+open class ActivityLogRestClient
```
aforcier: Does this need to be `open`?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -0,0 +1,225 @@
...
+                restoreId = response.restoreResponse?.rewind_id,
+                restoreErrorCode = response.restoreResponse?.error_code,
+                restoreState = response.restoreResponse?.status,
+                restoreProgress = response.restoreResponse?.progress,
+                restoreMessage = response.restoreResponse?.message,
+                restoreFailureReason = response.restoreResponse?.reason)
+        return FetchedRewindStatePayload(rewindStatusBuilder.build(), site, number, offset)
+    }
+
+    private fun error(site: SiteModel, number: Int, offset: Int, errorType: RewindStatusErrorType) =
```
aforcier: This function could probably use a more descriptive name.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -0,0 +1,225 @@
...
+        }
+        if (error is WPComGsonRequest.WPComGsonNetworkError) {
+            if ("unauthorized" == error.apiError) {
+                errorType = authorizationRequired
+            }
+        }
+        return errorType
+    }
+
+    private data class ActivitiesResponse(val totalItems: Int?,
```
aforcier: Similarly to [this comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/756#discussion_r175779590), it doesn't seem like we need the extra overhead of `data` classes here (and for the inner classes as well), since these are only generated by Gson and don't require getters/setters/etc.

aforcier:
Okay @planarvoid, second pass done! As we discussed, since the builder pattern is changed in #758 (in a way that looks to address my main concern) and this branch is now targeting a feature branch, I'm good with checking up on that aspect in that PR, where the database implications can be checked out properly, and merging this one when the rest of the feedback has been addressed.

fluxc/src/main/java/org/wordpress/android/fluxc/action/ActivityAction.java
```diff
@@ -0,0 +1,21 @@
+package org.wordpress.android.fluxc.action;
+
+import org.wordpress.android.fluxc.annotations.Action;
+import org.wordpress.android.fluxc.annotations.ActionEnum;
+import org.wordpress.android.fluxc.annotations.action.IAction;
+import org.wordpress.android.fluxc.store.ActivityLogStore;
+
+@ActionEnum
+public enum ActivityAction implements IAction {
```
aforcier: We should rename this one to `ActivityLogAction` too.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestWPCom.kt
planarvoid: I haven't noticed that! thanks 👍

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestWPCom.kt
planarvoid: At one point I had some issues with improper `CountDownLatch` teardown, I'll remove it

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
planarvoid: It seems that it doesn't! I remember that when a class wasn't `open`, it was not possible to mock it but it seems like it's not the case anymore.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
planarvoid: I'm renaming it to `buildErrorPayload`

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
planarvoid: I used data class for the `equals` method. I'll check if all the test are still passing after this change but comparing the objects manually gets boilerplatey.

fluxc/src/main/java/org/wordpress/android/fluxc/action/ActivityAction.java
planarvoid: I went over the whole change and renamed everything to `ActivityLog*`

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestWPCom.kt
aforcier: I agree - I worked on something like this for the WooCommerce tests, should be opening a PR for that soon. After that one we can probably extend it for Jetpack and use it for tests like this.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
aforcier: Ah yes - there's a workaround for this by using the `MockMaker` extension for Mockito, it was added for FluxC unit tests in #709.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
aforcier: Agreed - if it turns out `equals` is needed, it's worth keeping these as `data` classes 👍

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Tests with real credentials on real servers using the full release stack (no mock)
+ */
+class ReleaseStack_ActivityLogTestJetpack : ReleaseStack_Base() {
+    private val incomingActions: MutableList<Action<*>> = mutableListOf()
+    @Inject lateinit var activityLogStore: ActivityLogStore
+    @Inject internal lateinit var siteStore: SiteStore
+    @Inject internal lateinit var accountStore: AccountStore
+
+    private var mNextEvent: TestEvents? = null
```
aforcier: Can drop the `m` prefix 😉

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
```diff
@@ -0,0 +1,170 @@
...
+            throw AssertionError("Unexpected error occurred with type: " + event.error.type)
+        }
+        assertTrue(siteStore.hasSite())
+        assertEquals(TestEvents.SITE_CHANGED, mNextEvent)
+        mCountDownLatch.countDown()
+    }
+
+    @Subscribe
+    fun onSiteRemoved(event: OnSiteRemoved) {
+        AppLog.e(T.TESTS, "site count " + siteStore.sitesCount)
```
aforcier: Should be an `i` or `d`-level log.

aforcier: Actually, we shouldn't need to subscribe to `OnSiteRemoved` events here at all, unless I missed something?

aforcier:
Looks good @planarvoid, just a couple of really minor comments

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
planarvoid: copy & paste 😉

PR #757, opened 2018-03-20
Activity Log store - part 2 - Cleanup and test ActivityRestClient

planarvoid:
- wrap static methods and parent class and inject them instead in order to make ActivityRestClient testable. Using parent WPComRestClient made it hard to unit test the class so I made it not abstract and injected it instead. 
- unit test ActivityRestClient
- remove `page` and `offset` params from RewindStatus because they were not used


aforcier:
> wrap static methods and parent class and inject them instead in order to make ActivityRestClient testable. Using parent WPComRestClient made it hard to unit test the class so I made it not abstract and injected it instead.

I'm a bit torn on this one. I'd love for us to have more unit testability in FluxC, but on the other hand part of the reason `BaseWPComRestClient` is abstract (and its internals `protected`) and not injected is to discourage using it directly outside of FluxC - so it's not possible from the WordPress app, for example, to create your own network request and `add()` it to `BaseWPComRestClient`. I'm concerned about weakening the behavior of the production library (and even internally, making all the `BaseWPComRestClient` internals `public`) for tests. This also feels like a somewhat expensive approach: each client using this approach now also requires an instance of a `BaseWPComRestClient` and of a `WPComGsonRequestBuilder` to be created, which wasn't necessary before.

I don't have a complete suggestion for an alternative or improvement in mind right now. I do wonder if we could use mocked connected tests (we have a few of these in [`/mocked`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/tree/80c252666e397f638b1228261f276b702d95d9e7/example/src/androidTest/java/org/wordpress/android/fluxc/mocked), though I don't think something exactly like this is in there yet) combined with unit tests specifically for network response -> model conversions for an acceptable alternative with similar coverage. Any mocked tests that were written at this point feel like they would be replicating much of the checks these unit tests are doing.

I'd like to think about this one some more and maybe let others weigh in as well, but wanted to share my first impressions of this approach and see what you thought @planarvoid.

planarvoid:
> ...so it's not possible from the WordPress app, for example, to create your own network request and `add()` it to `BaseWPComRestClient`

It is still possible now to override the `BaseWPComRestClient` in WPAndroid to create your own `FeatureRestClient` and override it to get the same behaviour. I don't think it's much safer than injecting it. If it would make you feel safer, I can create a wrapper and inject that instead of changing the `BaseWPComRestClient`. I can see that enabling (for example) `CommentRestClient.add()` on an existing `RestClient` could be bad (and confusing). Let me know what you think after I implement the change.

> This also feels like a somewhat expensive approach: each client using this approach now also requires an instance of a `BaseWPComRestClient`

There shouldn't be a big difference in this case since every RestClient extends from BaseClient anyway, there will be the same amount of instances of everything inside the `BaseWPComRestClient` (there will just be 2 classes instead of 1 per feature). And this is a safe approach, in future I can see the `BaseWPComRestClient` as a singleton.

> and of a WPComGsonRequestBuilder to be created, which wasn't necessary before

True, however, this class is super lightweight. I will make it a singleton.

Overall, I feel like adding some classes for the sake of unit testability and separation of concerns is a win. I'll try to address your concerns and make it safer @aforcier

aforcier:
Summarizing our discussion for posterity: we agreed that the request builder was a great improvement on its own, and dropping the `WPComRestClient` was not a big hit to testability. This solves the issue of exposing too much of the network internals outside the library (though it would be great to improve this and lock it down completely, but that's way outside the scope here).

:shipit:, nice work @planarvoid!

PR #758, opened 2018-03-21
Activity Log store - Part 3 - SQL Utils + tests

planarvoid:
- `ActivityLogSqlUtils` is now an injected class (it makes it easier to unit test the store)
- first two methods to store and retrieve activities per site
- I moved the DB object to the `ActivityLogSqlUtils` class. I think it belongs to the DB layer. This way it could be used only to store/retrieve objects from DB and we can directly create the domain object (`ActivityLogModel`) and return it. The domain object is immutable and doesn't contain unnecessary fields. Let me know if this makes sense @aforcier
- I don't really see the advantage of 2 events being passed around - the first one being dispatched when we receive an API response (`FetchedActivities`) and the `OnChange` event once we insert the data into the DB. I would merge these two actions and only send the latter. However I was trying to follow the patterns and not make unnecessary changes.


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
```diff
@@ -56,6 +56,7 @@ public WellSqlConfig(Context context) {
         add(TermModel.class);
         add(ThemeModel.class);
         add(WPOrgPluginModel.class);
+        add(ActivityLogSqlUtils.Builder.class);
```
aforcier: Make sure to merge latest `develop` in - the behavior here has changed, adding to a `TABLES` variable is no longer needed.

aforcier:
Note: You'll need to add a db migration step to `WellSqlConfig` for the two new tables - [here's an example](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/8f5bd515783e48f1cf129ef7f181806cddffe96d/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java#L133-L140) (you can just write one migration step for this whole feature branch, no need to add steps for each change you make to the db before this has shipped to production).

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
```diff
@@ -224,6 +224,15 @@ public void onUpgrade(SQLiteDatabase db, WellTableManager helper, int oldVersion
                 db.execSQL("ALTER TABLE SiteModel ADD IS_WP_COM_STORE INTEGER");
                 db.execSQL("ALTER TABLE SiteModel ADD HAS_WOO_COMMERCE INTEGER");
                 oldVersion++;
+            case 27:
```
aforcier: Forgot to mention this - you'll also need to bump the db version [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/e7cd900ee9bdea8fc46c58a2a48f6010a3ba4c97/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java#L44) to `28`.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
```diff
@@ -41,10 +51,29 @@ class ActivityLogStore
...
+            emitChange(OnActivityLogFetched(payload.error, action))
+        }
+    }
+
     private fun fetchActivitiesRewind(fetchActivitiesRewindPayload: FetchRewindStatePayload) {
         activityLogRestClient.fetchActivityRewind(fetchActivitiesRewindPayload.site)
     }
 
+    // Actions
+    data class OnActivityLogFetched(val rowsAffected: Int,
```
aforcier: This one can also become a plain non-`data` class.

planarvoid: There is an equality check in the unit test which wouldn't work if this wasn't a data class

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
planarvoid: I thought that if the version is 27, it would go through the `case 27`

aforcier: Yeah it's a bit confusing - the number being switched here is the `oldVersion` of the database, so this case is declaring what to do when upgrading the database from `27` to whatever the current number is (at least `28` for this to run).

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
aforcier: Missed that, fair enough 👍

aforcier:
:shipit:!

PR #770, opened 2018-03-29
Activity Log Store - Part 5 - Rewind

planarvoid:
- implement rewind operation
- rewind is still on V1 rest endpoint - https://betterjetpackxp.wordpress.com/2017/10/05/activity-log-on-the-mobile-apps/
- it is not possible to reliably test rewind with real API so I'm just testing the failure (and that the endpoint is there) and unit testing the rest

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
@@ -85,10 +85,10 @@ class ActivityLogStore
     }
 
     private fun emitRewindResult(payload: RewindResultPayload, action: ActivityLogAction) {
-        if (payload.restoreId != null) {
-            emitChange(OnRewind(restoreId = payload.restoreId, causeOfChange = action))
-        } else if (payload.error != null) {
+        if (payload.error != null) {
             emitChange(OnRewind(payload.error, action))
+        } else {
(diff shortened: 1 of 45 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
```diff
@@ -78,6 +84,14 @@ class ActivityLogStore
         }
     }
 
+    private fun emitRewindResult(payload: RewindResultPayload, action: ActivityLogAction) {
+        if (payload.restoreId != null) {
+            emitChange(OnRewind(restoreId = payload.restoreId, causeOfChange = action))
+        } else if (payload.error != null) {
+            emitChange(OnRewind(payload.error, action))
+        }
```
aforcier: Perhaps a null `restoreId` should be caught in the rest client and floated up here as a new error, `MISSING_RESTORE_ID` or something like that?

aforcier:
Looks good to me overall, just had one comment/question - will let @0nko verify the functionality.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
planarvoid: good point. I should revert this logic as I did in AL

0nko:
:shipit:

PR #774, opened 2018-04-06
Issue/add notification subscriptions

theck13:
Update the `SubscriptionResponsePayload` class to include the new `SubscriptionType` in order to differentiate between subscriptions in the `onSubscriptionUpdated` callback used in https://github.com/wordpress-mobile/WordPress-Android/issues/7540.


nbradbury:
The first time I ran this build the list of followed sites in settings was empty. I went to the reader and noticed it showed all my followed sites, but when I returned to settings they still weren't showing (but they did eventually show up). 

This made me wonder if we're storing followed sites in two places - once in the ReaderBlogTable and again in FluxC - and I **think** this is the case. If so, we'll want to drop the old method of fetching followed sites and convert to the FluxC version (but not as part of this PR, especially since @mzorz is currently updating the app's reader services).

PR #776, opened 2018-04-10
WooCommerce: Store order details

aforcier:
Adds more fields to the `WCOrderModel`, mainly customer details, payment information, and the product list. Also adds some some unit and mocked network tests.

cc @AmandaRiu


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/OrderAddress.kt
```diff
@@ -0,0 +1,46 @@
...
+                address1 = orderModel.shippingAddress1
+                address2 = orderModel.shippingAddress2
+                city = orderModel.shippingCity
+                state = orderModel.shippingState
+                postcode = orderModel.shippingPostcode
+                country = orderModel.shippingCountry
+            }
+        }
+    }
+}
```
AmandaRiu: 🥇 What a clever way to handle this!

PR #778, opened 2018-04-11
GDPR - Add Tracks Opt Out field to Account Settings

daniloercoli:
As part of the GDPR efforts, WordPress.com recently introduced a new setting to opt-out. We want to sync the opt-out in the apps with that value whenever the user is signed in.

The setting is stored as tracks_opt_out in me/settings.

Once this PR is merged I will take care of implementing the opt-out in wp-android as soon as possible.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_AccountTest.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_AccountTest.java
@@ -187,6 +187,66 @@ public class ReleaseStack_AccountTest extends ReleaseStack_Base {
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


nbradbury:
@daniloercoli The code looks good, but given the importance of this setting for GDPR, what do you think about adding a simple test? I'm thinking we'd benefit from one that changes the opt-out setting for a site and ensures the change is reflected when the site is retrieved.

daniloercoli:
Right, simple tests were already there, and i've added connected tests as you've suggested.
- The first test just pushes `tracks_opt_out` without changing it, and verifies that's correct after the push.
- The second test changes `tracks_opt_out`, and verifies that the new value is on the server.

nbradbury:
:shipit:

PR #779, opened 2018-04-11
WooCommerce: Add customer_note mapping to WCOrderModel

AmandaRiu:
Adds the new `WCOrderModel.customerNote` field mapped from the order api response element `customer_note`.

Tests have also been updated where applicable.

cc: @aforcier


aforcier:
:shipit: 🎉

PR #780, opened 2018-04-12
Activity store - part 6 - move pagination to store

planarvoid:
- change logic so the client doesn't need to know what page it's on. It requests either the first or a next page. On first the store cleans up the DB, on next page it reads size of activity log in DB and loads the next page


0nko:
:shipit:

PR #786, opened 2018-04-19
Issue/785 changed posts count

### LINKED ISSUE #785: Return total number of local changes
In order to address [this WPAndroid issue](https://github.com/wordpress-mobile/WordPress-Android/issues/7655), we need to add a method which returns the total number of local changes.

nbradbury:
Resolves #785 - returns a count of the total number of posts with local changes across all sites.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java
@@ -161,4 +161,14 @@ public class PostSqlUtils {
                 .equals(PostModelTable.IS_LOCALLY_CHANGED, true)
                 .endGroup().endGroup().endWhere().getAsCursor().getCount() > 0;
     }
+
+    public static int getNumLocalChanges() {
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java
```diff
@@ -161,4 +161,14 @@ public static boolean getSiteHasLocalChanges(SiteModel site) {
                 .equals(PostModelTable.IS_LOCALLY_CHANGED, true)
                 .endGroup().endGroup().endWhere().getAsCursor().getCount() > 0;
     }
+
+    public static int getNumLocalChanges() {
```
aforcier: May I suggest a wrapper method for this in `PostStore` for consistency? We've generally tried to avoid making calls to `*SqlUtils` from outside FluxC, and only expose database reads via the corresponding `*Store` (ideally we'll one day refactor things so the `SqlUtils` are not exposed outside of FluxC at all).

aforcier:
:shipit:

PR #795, opened 2018-05-02
Update connect/site-info API call

### LINKED ISSUE #796: Update connect/site-info API call
The `connect/site-info` endpoint has been updated to accept the URL as a query
parameter rather than parts of the path.  We should consider updating the call as it will simplify the logic.
See https://github.com/wordpress-mobile/WordPress-iOS/pull/9245 for additional context.  (props @sirreal)

The test plan can be the same as the iOS app.  Additionally, I was able to perform a practical test with my self-hosted site by editing the xmlrpc.php file and introducing a fatal syntax error.

sirreal:
A new site-info endpoint has been added that receives the URL as a query
parameter rather than parts of the path (D11731-code)

This simplifies the API call and is less error-prone.

What was in the path, with `/` -> `::` (slug-style) encoding:

`https://public-api.wordpress.com/rest/v1.1/connect/site-info/https/example.com::site::path`

Is now a query param (no need to think about encoding):

`https://public-api.wordpress.com/rest/v1.1/connect/site-info?url=https%3A//example.com/site/path`

Aligns with https://github.com/Automattic/wp-calypso/pull/24034
Related iOS update: https://github.com/wordpress-mobile/WordPress-iOS/pull/9245

The API logic and response are unchanged.

Fixes #796

cc: @seear

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
@@ -378,13 +378,12 @@ public class SiteRestClient extends BaseWPComRestClient {
             return;
         }
 
-        // Sanitize and encode the Url for the API call.
-        String sanitizedURL = UrlUtils.removeScheme(siteUrl);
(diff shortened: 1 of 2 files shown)


aforcier:
Works well @sirreal, thanks for the patch!

:shipit:

PR #797, opened 2018-05-04
API: Add urlAfterRedirects from connect/site-info response

sirreal:
The API returns a `urlAfterRedirects`, which returns the URL after following HTTP redirections.

Add this data to the store.

See D11529-code and D12573-code

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/ConnectSiteInfoResponse.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/ConnectSiteInfoResponse.java
@@ -9,4 +9,5 @@ public class ConnectSiteInfoResponse implements Response {
     public boolean isJetpackActive;
     public boolean isJetpackConnected;
     public boolean isWordPressDotCom; // CHECKSTYLE IGNORE
+    public String urlAfterRedirects;
 }


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -703,6 +703,7 @@ private ConnectSiteInfoPayload connectSiteInfoFromResponse(String url, ConnectSi
         info.isJetpackConnected = response.isJetpackConnected;
         info.isWordPress = response.isWordPress;
         info.isWPCom = response.isWordPressDotCom; // CHECKSTYLE IGNORE
+        info.urlAfterRedirects = info.urlAfterRedirects;
```
aforcier: This needs to be `... = response.urlAfterRedirects`, and we need to add `urlAfterRedirects` to the list of expected params from the API in [`ConnectSiteInfoResponse`](https://github.com/sirreal/WordPress-FluxC-Android/blob/27ad19c55e3fe632753ed13a673eb0d8018139c5/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/ConnectSiteInfoResponse.java).

sirreal: Thanks @aforcier, I've addressed those points.

PR #800, opened 2018-05-08
Support turning order models into id sets and back

aforcier:
Adds `OrderIdentifier`, which is a composite id for orders made up their local id, remote id, and local site id (really a `typealias`'d `String`). This is meant to be a safe way to serialize and deserialize a partial `WCOrderModel` (e.g. for passing it in a `Bundle`), without the risk of accidentally storing and restoring the wrong id (e.g. storing the local site id instead of the local order id, and then attempting to load the order by its local order id).

I removed `getOrderByLocalOrderId()` entirely - it's possible it might have some other use in the future, but for now it's not needed for the implementation and it seems best to remove temptation 😛 

cc @AmandaRiu


aforcier:
While reviewing this I noticed that the notes created are always "system" notes on the Woo site, and not normal, site user-created private notes as I would expect. This seems to be a limitation of the API that has already been addressed in development, and we should be able to make use of it once that's rolled out to core. I opened a ticket Woo app-side to track this.

PR #801, opened 2018-05-09
Fix wrong mapping of actor type in ActivityLogModel

planarvoid:
Passing ActivityLog type instead of Actor type while reading data from database


aforcier:
:shipit:

PR #805, opened 2018-05-15
Fix wrong table name in migration in activity log

planarvoid:
- there was a wrong (old) table name - `ActivityLogModel` - while creating the table through migration which was crashing the app on `no such table: ActivityLog`
- I'm dropping the `ActivityLogModel` table if it exists and creating `ActivityLog` table if it doesn't. Since the `CREATE ActivityLog` script runs automatically on install, there is a chance some users have both tables.


0nko:
🎉

PR #808, opened 2018-05-17
Implement rewind status model changes

planarvoid:
I found out that the `RewindStatus` coming from the API is completely different than what we expect. This PR implements the current state & tests everything from https://activitylogrewind.wordpress.com/2017/12/21/jetpack-rewind-state-machine/.


planarvoid:
@0nko yeah, I thought we don't need it at the moment and it would clutter the model. I was considering not implementing the `credentials` either

PR #809, opened 2018-05-18
WooCommerce: Add WCOrderStatsModel

aforcier:
Adds the `WCOrderStatsModel` used for storing order stats.

The `WCOrderStatsModel` is used in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/804 (and this PR should be reviewed along with that one for context), but this PR targets `develop` rather than the `wc-stats` feature branch. The reason is so that the Woo beta app can target the `wc-stats` feature branch without causing future migration issues, by having the database table already present in FluxC's `develop`.

I also spotted and fixed an issue (https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/809/commits/7c67c7f72a44281e93f73d37cd5b63396df0ac11) with the version used for the earlier order table migration in #803 - I think the number was missed when the PR was updated after a conflicting migration was merged into `develop`. This only affected folks upgrading from an older version to the current latest Woo app (`0.8`) - once this fix is baked into the Woo app, upgrading from `< 0.8` -> `0.9` will work properly (anyone using `0.8` will have had to clear data by this point anyway).


aforcier:
Ready for review, @AmandaRiu!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderStatsModel.kt
```diff
@@ -15,6 +15,18 @@ data class WCOrderStatsModel(@PrimaryKey @Column private var id: Int = 0) : Iden
...
+        val responseType = object : TypeToken<List<String>>() {}.type
+        gson.fromJson(fields, responseType) as? List<String> ?: emptyList()
+    }
+
+    // The data JSON deserialized into a list of lists of arbitrary type
+    val dataList by lazy {
+        val responseType = object : TypeToken<List<List<Any>>>() {}.type
+        gson.fromJson(data, responseType) as? List<List<Any>> ?: emptyList()
+    }
+
```
AmandaRiu: Really slick improvement! ⚙️

PR #810, opened 2018-05-21
Fix DB migration for RewindStatusCredentials

planarvoid:
- recreate RewindStatusCredentials table from a new script


0nko:
:shipit:

PR #811, opened 2018-05-21
Fix changed Rewind model according to response

planarvoid:
- Rewind response doesn't match description so this PR is changing the model to match the response


planarvoid:
@0nko fixed 👍

0nko:
🚢

PR #815, opened 2018-05-28
Issue/crash subscription error

theck13:
### Fix
Update the [`handleUpdatedSubscription` method ](https://github.com/wordpress-mobile/WordPress-FluxC-Android/compare/issue/crash-subscription-error?expand=1#diff-21a42e5cb9fc1d9531f135b868bc3e3aR1191) to assign the `OnSubscriptionUpdated` event error from the `SubscriptionResponsePayload` payload rather than creating a new `SubscriptionError`.   This solves a `NullPointerException` crash when `event.error` is `null` as described in https://github.com/wordpress-mobile/WordPress-Android/issues/7796.


aforcier:
:shipit:

PR #819, opened 2018-05-31
Fix error in rewind status mapping

planarvoid:
- there was a missing restore ID when mapping `Rewind` from DB and this made it null in the result
- progress is now nullable in `Rewind`
- I've noticed that the `RewindResponse` contains extra fields with an `error` so I've included them and created a new `API_ERROR` type


0nko:
:shipit:

PR #831, opened 2018-06-19
Delete scheduled posts in connected tests

aforcier:
One of our connected tests creates a post and then makes it scheduled. This isn't a problem for the post tests, but it causes an issue for the comment tests (WP.com specifically, at the moment). This is because the comment tests rely on a published to comment on. If there are 20 scheduled posts, all the posts we fetch are scheduled, and there's no published post for the comment tests to use.

(This was causing the tests to fail consistently on Firebase, since the test site reached 20 scheduled posts.)

So, this PR just deletes the scheduled post at the end of the test.

There are other failing tests that are flaky, we need to lock those down too but this one is highest priority since it's guaranteed test failure.

cc @oguzkocer


aforcier:
> Should we introduce helpers for both classes for deleting a remote post

Makes sense! Added in d9c738e.

PR #832, opened 2018-06-21
Delete Activity Log data only when new data is succesfully fetched

planarvoid:
- it makes sense to delete the activity log only when we have successfully fetched the new data. Checking the offset makes sure we only delete the previous data when the new data starts at 0 (it is a new fetch of the first page)


0nko:
I've tested this PR and it seems the rewind progress state is still not being updated:

1. Try starting a rewind (from the web, for example)
2. Add a breakpoint in the rewind service to inspect the rewind state LiveData
3. Go to the Activity log in the app
4. Inspect the rewind state LiveData and notice the value is "FINISHED"
5. Inspect the rewind progress LiveData and notice the value is null

0nko:
:shipit:

planarvoid:
@0nko I've fixed this in WPAndroid so we should be fine merging this PR

PR #833, opened 2018-06-26
Move OAuth token storage location

aforcier:
Moves the WordPress.com OAuth token storage location away from the default shared preferences location for the host app and into a custom location.

The backstory here is that we don't want the token to be stored as part of Android's auto backup feature, but we generally do want to include the default shared preferences file in the backup. With this change, the WP.com OAuth token won't be lumped in by default.

**This PR shouldn't be merged until the WPAndroid PR has also been tested and we're sure there are no migration issues.**

**To test:**
There will be a PR in WPAndroid making use of this, but this can also be tested with the FluxC example app:

1. Build `develop`, run the example app, log in to WordPress.com
2. Build this branch and run the example app
3. The token should be migrated to a new location, and you should stay logged in

Using Stetho, under Resources > Local Storage, you should see:
`org.wordpress.android.fluxc.example_preferences`, which is the default shared preferences location, and should not contain the token
and `org.wordpress.android.fluxc.example_fluxc-preferences`, which will contain the token

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WPComBase.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WPComBase.java
@@ -1,7 +1,6 @@
 package org.wordpress.android.fluxc.release;
 
 import android.content.Context;
-import android.preference.PreferenceManager;
 
 import org.greenrobot.eventbus.Subscribe;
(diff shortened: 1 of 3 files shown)


mzorz:
Followed the steps as indicated in https://github.com/wordpress-mobile/WordPress-Android/pull/7949 and confirmed it works as expected. 
:shipit:

PR #836, opened 2018-07-05
Use correct year format for week stats requests

aforcier:
Fixes a bug where I was accidentally using the [week-year](https://docs.oracle.com/javase/7/docs/api/java/util/GregorianCalendar.html) instead of the year for weeks requests to the order stats endpoint.

(This doesn't generally cause any obvious problems for now, but week years are different from normal years and may have caused problems later. Also it's only supported on API 24+, so it would have broken things for older Android versions.)

I also made the stats quantity constants public so they can be referenced from the host app if need be.


nbradbury:
:shipit:

PR #846, opened 2018-07-13
Support retrying Jetpack tunnel GET requests on timeout

aforcier:
Adds a mechanism for silently retrying GET requests through the Jetpack tunnel when we get a timeout error.

The Jetpack timeout error basically happens when a request through to the remote site ends up taking longer than 5 seconds. The response we get through the Jetpack tunnel looks like this:

```js
{
  "error": "http_request_failed",
  "message": "cURL error 28: Operation timed out after 5001 milliseconds with x bytes received"
}
```

(With HTTP status `400`.)

Retrying the request manually immediately or after a few seconds has tended to work in my experience, so the hope is that we can avoid too much user friction and just have a slight delay in completing the network action most of the time.

In the longer term I'd like us to expand this solution in a few ways:
1. Support POST/PUT/PATCH retries

This is trickier since in general these methods are not safe, and we can't call them multiple times. We would probably want to do something more complex than just retry in those cases, and maybe somehow verify if the resource was created on the server, and only retry in that case.

2. Track retry events

It might be nice to have some data on all this, patching through to analytics when a timeout occurs, for what endpoint, and whether a retry was successful.

### To test

This one's fun 😀 It's not easy to reliably trigger a legitimate Jetpack timeout error. It's pretty rare for fetching orders and order notes, and seems to appear a bit more often for `product` WooCommerce endpoints, which we don't use in FluxC yet. It's probably possible to set it up by modifying a site and adding some delays to PHP calls, but we can simulate the whole thing using Charles.

One way is to set a breakpoint on a request, like the one made when fetching orders through the example app, and modify the raw response to return the error message above and a HTTP 400.

A better way is to use Charles' [Rewrite function](https://www.charlesproxy.com/documentation/tools/rewrite/). This can be setup to always modify requests of a certain type according to your specifications. Mine looks like this:

charles-rewrite

(Replace the `11111111` in the URL with the ID of the site you're testing with.)

This will allow you to test the retry behavior. You should see 3 total failed requests come through to Charles with this active - the first failed attempt, and then the two retries. You should only get an error message at the UI level after the last failed attempt.

To test success after a retry, you can turn off Rewrite after the first retry is done - there's a 5 second delay between the first and second retry. You can also add more time by manually increasing [this value](https://github.com/wordpress-mobile/WordPress-FluxC-Android/compare/feature/wc-5-second-timeout-retry?expand=1#diff-42f3c5eee34b95ebe1f502c77163420eR36).

cc @nbradbury

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_JetpackTunnelTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_JetpackTunnelTest.kt
@@ -16,6 +16,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComErrorListener
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTimeoutRequestHandler
 import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequest
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -62,7 +63,8 @@ class OrderRestClient(
                     val orderError = networkErrorToOrderError(networkError)
                     val payload = FetchOrdersResponsePayload(orderError, site)
                     mDispatcher.dispatch(WCOrderActionBuilder.newFetchedOrdersAction(payload))
-                })
+                },
+                { request: WPComGsonRequest<*> -> add(request) })
```
nbradbury: Being so new to Kotlin I'm having trouble grokking what this line does. Any chance you can explain it?

aforcier: Sure thing!

So [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/846/files#diff-b161c6048faf9b78ed1d48967aa63480R112) I added a new parameter to `buildGetRequest()`:

```kotlin
...,
jpTimeoutListener: ((WPComGsonRequest<*>) -> Unit)?
) ...
```

`jpTimeoutListener` is not a normal object but a function - one that takes a single argument of a `WPComGsonRequest`, and doesn't return anything (we don't care what the function actually does, that's up to the whoever is calling `buildGetRequest()`). (It's also nullable but that doesn't really mean anything different than usual.)

In this line here we're creating a function that matches what `buildGetRequest()` expects so we can satisfy that argument. The curly brackets indicate a function (specifically a [lambda expression](https://kotlinlang.org/docs/reference/lambdas.html#lambda-expressions-and-anonymous-functions)), and we're declaring a function that takes one argument, of type `WPComGsonRequest` (which we're calling `request`). The body of the function is up to us, and we're using it to pass the `request` to our `add()` function.

So basically it's like defining a listener that implements an interface and putting `add(request)` inside a method of that interface as we would do in Java, but instead of having to call something like `listener.onJpTimeout(request)` [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/846/files#diff-42f3c5eee34b95ebe1f502c77163420eR62), we're basically just calling the callback function directly and there's no wrapped listener object.

You can also look at it this way - we could have declared the function explicitly further up, before we started defining the request:

```kotlin
val jpTimeoutListener = fun(request: WPComGsonRequest<*>) {
    add(request)
}
```

Then, on this line, instead of

```kotlin
{ request: WPComGsonRequest<*> -> add(request) }
```

we would just have

```kotlin
jpTimeoutListener
```

This also helps read the call to `jpTimeoutListener` [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/846/files#diff-42f3c5eee34b95ebe1f502c77163420eR62), since you can see that we're literally calling the jpTimeoutListener function as it is above, with the argument `request`. The only different is that we're doing it inline, so we don't have to give the function a name.

Hope this was helpful - I'm happy to answer anything else that might come up, I know this is a tricky part of Kotlin to grok (and the networking code isn't the simplest place to deal with it in 😬).

nbradbury: > You can also look at it this way - we could have declared the function explicitly further up, before we started defining the request:

Aha, that made it clear to me - thanks for taking the time to explain it!

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_JetpackTunnelTest.kt
```diff
@@ -108,6 +111,78 @@ class MockedStack_JetpackTunnelTest : MockedStack_Base() {
...
 
+    @Test
+    fun testJetpackTimeoutErrorResponseContinuous() {
+        // Simulate continuously receiving Jetpack timeout error from the server
+        // Expect to make the maximum number of retry attempts, and then finally receive the timeout as a normal error
+        val countDownLatch = CountDownLatch(JetpackTimeoutRequestHandler.DEFAULT_MAX_RETRIES + 1)
+        val url = "/"
+        var retriesAttempted = 0
+
+        val request = JetpackTunnelGsonRequest.buildGetRequest(url, 567, mapOf(),
```
nbradbury: Perhaps this is picky, but I'm wondering why we've hardcoded siteId 567 throughout? Seems like that should be a constant.

aforcier: It could be - I guess my thinking was that the number is totally arbitrary so having a constant almost suggests that the value is significant (e.g. is checked somewhere) when it's not.

...but we can solve that by just naming the constant appropriately. Updated in 6ed6df1 - also did the same for the unit test counterpart class.

nbradbury: > I guess my thinking was that the number is totally arbitrary so having a constant almost suggests that the value is significant

Oh, I was assuming that was a real site ID - I didn't know it was arbitrary.

nbradbury:
Really nice work @aforcier - and I learned a bit about Kotlin by reviewing this. :shipit:

PR #854, opened 2018-07-29
Convert OrderStatus to an enum for improved flexibility

AmandaRiu:
Converted the OrderStatus class from a static object to an enum for more flexibility in situations where the values must be iterated over. The enum also allows the core label to be applied in a single place. 

Also added some unit test suites for the orders and stats feature, and then an overall woo test suite to run all the woo unit tests. This just makes it easier to run tests when inside Android Studio. 

**To Test**

- Run Unit and Instrumented WC tests
- In  the Woo page of the Example app, test using the two order status options


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderStatus.kt
```diff
@@ -1,12 +1,29 @@
...
-    const val ON_HOLD = "on-hold"
-    const val PROCESSING = "processing"
-    const val COMPLETED = "completed"
-    const val CANCELLED = "cancelled"
-    const val REFUNDED = "refunded"
-    const val FAILED = "failed"
+/**
+ * Standard Core WooCommerce order statuses
+ */
+enum class OrderStatus(val label: String, val value: String) {
```
aforcier: @AmandaRiu I'm not completely sure about this. Enums should be exhaustive, and in this case it's not, which is a bit misleading. It would be pretty easy in the future for a developer to loop through `OrderStatus` and miss out on covering custom statuses, or define functions that require an `OrderStatus` instead of a `String`.

Also, there are a couple of statuses we'll likely support natively in the app because they're defined by popular extensions: [`pre-ordered`](https://docs.woocommerce.com/document/pre-orders/) and [`partially-paid`](https://docs.woocommerce.com/document/woocommerce-deposits/) (those might not be the exact slug names). But we won't want to lump those into the enum, since we'll only show them as options if the store has those extensions enabled.

Maybe we can instead assemble a list of supported statuses dynamically client-side? Alternatively, maybe calling this `CoreOrderStatus` adds enough clarity, though we'll want a plan for storing the `pre-ordered` and `partially-paid` statuses outside it. WDYT?

AmandaRiu: @aforcier I agree this is not exhaustive, it's an interim until we get to the point where we are using the API to poll for the available order status values and hopefully store them in a database where things are more dynamic. The purpose of this was just to convert the pre-existing OrderStatus object with a static set of constants for each core status and then convert it into something that can be used with adapters, as well as to keep the label and value in a single place. I could rename it to CoreOrderStatus if you really think it will help, but I don't believe this class should be around long enough. The moment we implement filtering by custom order statuses this one would already be defunct. WDYT?

AmandaRiu: Spoke with @aforcier and we decided to rename this class to `CoreOrderStatus` and in the future we can just create another Enum for any extension-related order statuses such as `partially-paid` and `pre-ordered`. Ideally this would all be fed from the API but currently we attach colors to order status tags and those colors are not provided by the api so it seems this class will be around much longer if we don't get something added to the api.

AmandaRiu:
@aforcier - name updated in 6cacd35fde3cf39305f6a2c5afb9a5380e0c3bcf . Ready for review :)

example/src/test/java/org/wordpress/android/fluxc/wc/WooTestSuite.kt
```diff
@@ -0,0 +1,14 @@
+package org.wordpress.android.fluxc.wc
+
+import org.junit.runner.RunWith
+import org.junit.runners.Suite
+import org.junit.runners.Suite.SuiteClasses
+import org.wordpress.android.fluxc.wc.order.OrdersTestSuite
+import org.wordpress.android.fluxc.wc.stats.StatsTestSuite
+
+@RunWith(Suite::class)
+@SuiteClasses(
```
aforcier: To be honest, I'm not sure I see the value in these suites. For connected tests that take a lot of time, sure - but unit tests are quite speedy (even running the entire suite), and it seems preferable to encourage running the entire set of unit tests instead.

Also, when running the entire test set this ends up running all these tests twice in AS, and it looks like 3 times over the command-line (`301` (no suites) vs `326` (AS with suites)  vs `352` (terminal with suites)).

aforcier:
@AmandaRiu this is ready to go apart from my comment above about the test suites

example/src/test/java/org/wordpress/android/fluxc/wc/WooTestSuite.kt
AmandaRiu: I had added them mostly because of all the issues with running tests directly in Android Studio. Many times it just cannot find the tests, but with a test suite, it has no issues at all. I can remove them tho.

> Also, when running the entire test set this ends up running all these tests twice in AS, and it looks like 3 times over the command-line (301 (no suites) vs 326 (AS with suites) vs 352 (terminal with suites)).
That's interesting.

aforcier:
This is good to go @AmandaRiu  :shipit: 

Won't actually merge until the Woo Android counterpart PR has been reviewed and is ready, since this is a breaking change, but feel free to do it yourself (or @nbradbury can) when that happens if I'm not around 🙂

nbradbury:
:shipit:

PR #855, opened 2018-07-31
Pass isPages parameter to error on FETCH_PAGES

planarvoid:
We were not passing the `isPages` parameter on error. This causes the `causeOfChange` field to be `FETCH_POSTS` instead of `FETCH_PAGES` when we are fetching pages.


aforcier:
> LGTM, but there is something funky going on with Travis :/.

The issue is that the `PostXMLRPCClient` also uses `FetchPostsResponsePayload` and needs to be updated to the new usage as well.

planarvoid:
@aforcier I've fixed the issue 👍

0nko:
:shipit:

PR #856, opened 2018-08-02
Add ability to fetch order counts with optional filter

AmandaRiu:
Add logic to process request for a count of orders optionally filtered by order status. This is part of the project to add a "Orders to fulfill" card to the dashboard of the WooCommerce-Android app.

### API
No API endpoint for requesting just the total count of orders so while I've added a new `FETCH[ED]_ORDERS_COUNT` action, I'm reusing the fetchOrders() api endpoint to fetch the list of orders and just counting the results. Since it's done this way, I've continued to use the default value for the `per_page` parameter. So the maximum count returned will be this value. I plan on using the `canLoadMore` value to determine how the count should be displayed. 

**Using this view as an example:**

orders

_If `canLoadMore=true` and `count=50`_, then the title would be:

> **You have 50+ orders to fulfill**

### DB
The results of using this new feature are **not saved to the database**, even though we are fetching full orders to build the result from. This is to keep the orders list view in proper sync since currently we drop all orders from the table and repopulate it after after successful orders fetch. This method will definitely be used by the dashboard to get and display a count, but we have no idea what the current order list view is filtered by or actively displaying. If the user clicks the **VIEW ORDERS** button, then the app will navigate to the orders view and a request to officially fetch all `processing` orders would be executed.

### Dispatching and handling the response
Since the database is not at all involved in this feature, the OnOrderChanged event contains all the information needed to fulfill the request. I've reused the `rowsAffected` property and populate it with the total order count and the `canLoadMore` property controls whether or not a `+` symbol is appended to the total count in the UI.  

**Note:** _I actually struggled with this because the `rowsAffected` field is always used to represent total records in the database, so my use is actually misleading. Would love to get some other peoples opinions on this and will happily add this field if anyone else also thinks it's a good idea._ 🤓

## Testing

1. Added new tests to the release and mocked WC orders tests.
2. Added a new button to the Example app:

screenshot_1533188655

Clicking the `FETCH ORDERS COUNT` button will open a dialog window: 

screenshot_1533188691

**Options:**
- Leave blank: fetch all orders (no filter)
screen shot no filter

- Enter a single order status: will fetch a count of orders by that order status.

screen shot 2018-08-02 at 12 47 03 am

 cc: @nbradbury @aforcier

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/WooCommerceFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/WooCommerceFragment.kt
@@ -198,6 +198,9 @@ class WooCommerceFragment : Fragment() {
 
         getFirstWCSite()?.let { site ->
             wcOrderStore.getOrdersForSite(site).let { orderList ->
+                // We check if the rowsAffected value is zero because not all events will causes data to be
+                // saved to the orders table (such as the FETCH-ORDERS-COUNT...so the orderList would always
+                // be empty even if there were orders available.
                 if (orderList.isEmpty() && event.rowsAffected == 0) {
                     prependToLog("No orders were stored for site " + site.name + " =(")
                     return
--- a/example/src/main/java/org/wordpress/android/fluxc/example/WooCommerceFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/WooCommerceFragment.kt
@@ -12,6 +12,7 @@ import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.action.WCOrderAction.FETCH_ORDERS
+import org.wordpress.android.fluxc.action.WCOrderAction.FETCH_ORDERS_COUNT
 import org.wordpress.android.fluxc.action.WCOrderAction.FETCH_ORDER_NOTES
 import org.wordpress.android.fluxc.action.WCOrderAction.POST_ORDER_NOTE
 import org.wordpress.android.fluxc.action.WCOrderAction.UPDATE_ORDER_STATUS
@@ -25,6 +26,7 @@ import org.wordpress.android.fluxc.model.WCOrderNoteModel
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.CoreOrderStatus
 import org.wordpress.android.fluxc.store.WCOrderStore
 import org.wordpress.android.fluxc.store.WCOrderStore.FetchOrderNotesPayload
+import org.wordpress.android.fluxc.store.WCOrderStore.FetchOrdersCountPayload
 import org.wordpress.android.fluxc.store.WCOrderStore.FetchOrdersPayload
 import org.wordpress.android.fluxc.store.WCOrderStore.OnOrderChanged
 import org.wordpress.android.fluxc.store.WCOrderStore.PostOrderNotePayload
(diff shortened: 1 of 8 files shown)


AmandaRiu:
@nbradbury Ready for review, but can't be merged until #854 is merged.

nbradbury:
It's a shame there's no endpoint for requesting a count since getting the order list can be a heavy response, but I just perused the Woo API docs and I also see no other way.

nbradbury:
> I actually struggled with this because the rowsAffected field is always used to represent total records in the database, so my use is actually misleading. Would love to get some other peoples opinions on this and will happily add this field if anyone else also thinks it's a good idea.

I'm torn on this as well but it seems like rather than rely on `OnOrderChanged` we should have something specific to getting just the count. Is there some reason we _have_ to use `OnOrderChanged`?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -155,6 +182,10 @@ class WCOrderStore @Inject constructor(dispatcher: Dispatcher, private val wcOrd
         wcOrderRestClient.fetchOrders(payload.site, offset, payload.statusFilter)
     }
 
+    private fun fetchOrdersCount(payload: FetchOrdersCountPayload) {
+        with(payload) { wcOrderRestClient.fetchOrders(site, 0, statusFilter, countOnly = true) }
```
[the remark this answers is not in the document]
AmandaRiu: @nbradbury I could do that, but then I'd have to also label the `statusFilter` param as well since in Kotlin once you declare the argument label you have to do it for every argument following.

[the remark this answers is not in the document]
nbradbury: Bah, forgot about that. Ignore me :)

example/src/main/java/org/wordpress/android/fluxc/example/WooCommerceFragment.kt
```diff
@@ -178,7 +198,7 @@ class WooCommerceFragment : Fragment() {
 
         getFirstWCSite()?.let { site ->
             wcOrderStore.getOrdersForSite(site).let { orderList ->
-                if (orderList.isEmpty()) {
+                if (orderList.isEmpty() && event.rowsAffected == 0) {
```
nbradbury: I think this could use a comment explaining why we're now checking `event.rowsAffected`.

example/src/main/java/org/wordpress/android/fluxc/example/WooCommerceFragment.kt
```diff
@@ -204,6 +224,12 @@ class WooCommerceFragment : Fragment() {
                             prependToLog("Fetched ${event.rowsAffected} orders from: ${site.name}")
                         }
                     }
+                    FETCH_ORDERS_COUNT -> {
+                        val append = if (event.canLoadMore) "+" else ""
+                        event.statusFilter?.let {
+                            prependToLog("Count of $it orders: ${event.rowsAffected}$append")
```
nbradbury: TIL `$it`!

example/src/main/java/org/wordpress/android/fluxc/example/WooCommerceFragment.kt
AmandaRiu: Good idea...added in #a6e2306d75d7afa99e40fa3c832815e504ee13b1

AmandaRiu:
@nbradbury,

Thanks for looking this over!

> It's a shame there's no endpoint for requesting a count since getting the order list can be a heavy response, but I just perused the Woo API docs and I also see no other way.

Yeah I know...[I've submitted a ticket](https://github.com/woocommerce/wc-api-dev/issues/116) tho!

> I'm torn on this as well but it seems like rather than rely on OnOrderChanged we should have something specific to getting just the count. Is there some reason we have to use OnOrderChanged?

I'm using this because we are hitting the orders api, and eventually, that information will be included in the metadata of that fetch orders response. 

I added a little code comment as suggested. This is ready for a 2nd pass. Thank you!

nbradbury:
:shipit:

PR #857, opened 2018-08-02
Included API to check domain name availability

usamahamid:
This in an API required for implementing [custom domain association feature](https://github.com/wordpress-mobile/WordPress-Android/issues/7923).

When User associates a custom domain, mobile client has to validate if the domain is available, mappable and supports privacy protection. This information will be utilized to make a decision; on whether to let user select a domain and whether to show the user with privacy protection option, during domain registration.


usamahamid:
@oguzkocer and @malinajirka. 
Included API to check availability of a domain name.
Please do take a look, when you have a free moment. :+1:

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -413,6 +438,19 @@ public OnSuggestedDomains(@NonNull String query, @NonNull List<DomainSuggestionR
         }
     }
 
+    public static class OnDomainAvailabilityChecked extends OnChanged<DomainAvailabilityError> {
+        public @Nullable DomainAvailabilityModel model;
```
oguzkocer: Isn't `model` too generic of a name? I wonder if we even need a model for this response. Maybe we should just remove `DomainAvailabilityModel ` and add its fields to `OnDomainAvailabilityChecked` directly. What do you think?

P.S: If we do keep it, I'd love to see the name change to be applied to the constructor parameter as well. All our models end with the `model` suffix, so I don't really like to use it as a parameter/property/variable name.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -413,6 +438,19 @@ public OnSuggestedDomains(@NonNull String query, @NonNull List<DomainSuggestionR
...
+        public OnDomainAvailabilityChecked(@Nullable DomainAvailabilityModel model,
+                                           @Nullable DomainAvailabilityError error) {
+            this.model = model;
+            this.error = error;
+        }
+        public OnDomainAvailabilityChecked(@NonNull DomainAvailabilityError error) {
```
oguzkocer: This constructor is not used.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -609,6 +647,11 @@ public static AutomatedTransferErrorType fromString(String type) {
         }
     }
 
+    public enum DomainAvailabilityErrorType {
+        INVALID_DOMAIN_NAME,
+        GENERIC_ERROR;
```
oguzkocer: Nitpick: No need for the `;` here. Let's make AS happy :)

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -1317,6 +1366,20 @@ private void handleFetchedPlans(FetchedPlansPayload payload) {
         emitChange(new OnPlansFetched(payload.site, payload.plans, payload.error));
     }
 
+    private void checkDomainAvailability(String domainName) {
+        if (domainName == null || domainName.isEmpty()) {
```
oguzkocer: We can use `TextUtils.isEmpty(domainName)` instead.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -219,6 +220,17 @@ public AutomatedTransferStatusResponsePayload(@NonNull SiteModel site, Automated
         }
     }
 
+    public static class DomainAvailabilityPayload extends Payload<DomainAvailabilityError> {
```
oguzkocer: I'd consider renaming this to `DomainAvailabilityResponsePayload`. I think it'll depend on what we do about the `DomainAvailabilityModel`. If we keep it, the names might be too long, so I am OK with not changing the name in that case. Otherwise, I really like to keep consistent with most of our code base and use the `Response` suffix in the response payloads.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
usamahamid: I would _prefer_ to keep the `DomainAvailabilityModel`, at the least for the purpose of holding `AvailabilityStatus` and `Mappability` Enums (Introduced in 51b9bf3). What do you think?

usamahamid:
@oguzkocer and @malinajirka, 
I have updated the PR as per your comments :slightly_smiling_face: and it is ready for another pass.
Please do check it out. when you have a free moment.

But Travis seems to fail in a unrelated test case. :thinking: 
Is there anything I can do to fix this?

malinajirka:
Thanks @usamahamid! It looks good overall. Great job! I've left one inline comment. However, since @oguzkocer started this review, let's let him take a look as well. Thanks!;)

>But Travis seems to fail in an unrelated test case. 🤔
Is there anything I can do to fix this?

Integration tests may sometime fail from various reasons - for example when the internet connection is unstable. First thing I usually try is simply restarting the build.

Click on the failing build on Github -> click on the "Restart Build" button in the top right corner.
454984a5e1a3e1c4b2987c2ffd183181

I'm wasn't sure if you have permissions to do that, so I restarted the build and it passed 👍.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/DomainAvailabilityResponse.kt
```diff
@@ -0,0 +1,11 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.site
+
+class DomainAvailabilityResponse(
```
malinajirka: I'm a bit concerned that it might not be safe to have NonNull fields in the response object. Unfortunately, AFAIK we don't have strict specification of the API, so we must just assume that all the attributes are nullable. You can emit a payload with a `DomainAvailabilityError` if the response doesn't contain some of the required attributes.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
oguzkocer: I think we can add the enums to `SiteStore` for now. I am even considering if we want a `DomainStore`, not as part of this PR, but maybe something to add in the future to make domain related stuff there and cleanup the `SiteStore` and `SiteRestClient` a bit.

I think having the fields in the `OnDomainAvailabilityChecked` is a bit cleaner as we have been using `**Model`s mostly for things we save to DB, although there are cases this is not true.

If you prefer having a model, I am cool with that, so no need to change!

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/DomainAvailabilityResponse.kt
oguzkocer: I don't think we can tell which fields are required on the FluxC side @malinajirka. I was a little surprised by having the status and errors etc in the success response, but the API is designed that way and if we tried to manually throw error in certain cases, I think we might introduce some inconsistencies. This is something I considered during the original PR review as well.

About the nullability though, you're right, we should have everything nullable in the response object and most likely for the response model `DomainAvailabilityModel` as well. cc @usamahamid

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -730,4 +772,16 @@ private ConnectSiteInfoPayload connectSiteInfoFromResponse(String url, ConnectSi
         info.urlAfterRedirects = response.urlAfterRedirects;
         return info;
     }
+
+    private DomainAvailabilityModel responseToDomainAvailability(DomainAvailabilityResponse response) {
+        Integer productId = response.getProduct_id();
+        String productSlug = response.getProduct_slug();
+        String domainName = response.getDomain_name();
+        // Utilize Locale.ROOT to fix ktlint
+        AvailabilityStatus status = AvailabilityStatus.valueOf(response.getStatus().toUpperCase(Locale.ROOT));
```
oguzkocer: I believe both of these `.valueOf` statements will crash if we get a `status` or `mappable` that's not in the enums. I think we should not allow such crashes in the app, even if it's unlikely. A better approach would be to follow the same pattern we use for error types, here is an example (one you wrote ;): https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java#L514-L531

Furthermore, I really, really don't like that we are using a locale here. I understand that it was done to fix a `ktlint` issue, but we should either do it without the locale in a different way to make `ktlint` happy (which the above pattern will do) or we should disable to `ktlint` check for this. Lints are supposed to increase the code quality which I don't think is the case here. And even if we used locale for some reason, I think `ROOT` might not be correct as, if I understand correct, that might end up changing the string (especially symbols).

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/DomainAvailabilityResponse.kt
usamahamid: Understandable. While I did test out the API extensively, maybe there are unique scenarios where API might fail to return those attributes. :+1: 
I have updated the attributes be `nullable` (in 95f70b4). Also, sincein this API we are interested mainly in the `AvailabilityStatus` and `Mappability`, I have introduced `UNKNOWN_STATUS` in both to be default/fallback, in case they are `null` or unidentified.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
usamahamid: Of course, :+1:. 
While I did not encounter any other statuses, it is always best to plan for a fallback.  
I have introduced `UNKNOWN_STATUS` as a fallback and have handled it same as in the example. Also as mentioned this side-stepped the `ktlint` issue. :100:

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
usamahamid: If `**Model` is generally utilized to define DB entities, it might be best to avoid using the same pattern here. 
As such I have removed the `DomainAvailabilityModel` and utilized `OnDomainAvailabilityChecked` event to hold the required attributes.

Maybe we should take some time , when we are free, to iron out/refactor such scenarios into common pattern :thinking:. What do you think? (cc @malinajirka)

usamahamid: @malinajirka.
Additionally, I have also been thinking about refactoring all `Domain` related API to `DomainStore`, `DomainRestClient`, `DomainActions`... (since Custom domain association, requires a couple more API related to `Domain`). 
I will probably make this refactor into a separate PR, for ease. What do you think?

malinajirka:
Thanks @usamahamid for making all the changes! It looks good, but I have one more comment. 

Moving both `AvailabilityStatus` and `Mappability` from `DomainAvailabilityModel` to `SiteStore` made their names a bit confusing. SiteStore.AvailabilityStatus -> I'd expect to get a status for a site. I think we might consider renaming them. WDYT?

Aside from this, I think it's ready for merge.

usamahamid:
@malinajirka, Of course. You are right. :100: 
I have renamed the `AvailabilityStatus`  to `DomainAvailabilityStatus` and `Mappability` to `DomainMappabilityStatus` to be more clear/concise (in b09d588) and the PR is ready for another pass. 
Please do check it out, when you have a free moment.

malinajirka:
I'm merging this PR even though it's in 'Changes requested' state as all the issue has been resolved.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
malinajirka: Thanks for the suggestions! It definitely sounds like a good idea to me. However, since there are still many tasks we need to finish before this feature is shippable, I'd postpone the refactoring. We might also consider discussing it with other devs, if they agree. Wdyt?

usamahamid: Great! 
We can put this on ice, for now. :+1:

PR #861, opened 2018-08-08
Feature/trashed pages

0nko:
This PR adds an optional parameter that specifies the post status types to be loaded.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/PageStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/PageStore.kt
@@ -69,10 +69,10 @@ class PageStore @Inject constructor(private val postStore: PostStore, private va
     suspend fun requestPagesFromServer(site: SiteModel): OnPostChanged = suspendCoroutine { cont ->
         this.site = site
         postLoadContinuation = cont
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/PageStore.kt
```diff
@@ -69,11 +70,12 @@ class PageStore @Inject constructor(private val postStore: PostStore, private va
...
         postLoadContinuation = cont
-        requestMore(site, false)
+        fetchPages(site, false)
     }
 
-    private fun requestMore(site: SiteModel, loadMore: Boolean) {
-        val payload = FetchPostsPayload(site, loadMore)
+    private fun fetchPages(site: SiteModel, loadMore: Boolean) {
+        val payload = FetchPostsPayload(site, loadMore,
+                listOf(PostStatus.DRAFT, PostStatus.PUBLISHED, PostStatus.SCHEDULED, PostStatus.TRASHED))
```
planarvoid: could we add a test for this parameter?

0nko: I added a unit test that verifies the correct payload.

PR #864, opened 2018-08-09
Issue/863 wc top earners

### LINKED ISSUE #863: Add support for Woo "top earners"
In order to show top earners in the WooCommerce app, we need to add the `sites/$siteId/stats/top-earners` endpoint. Note this is a wp.com endpoint rather than a Woo one. Ex:

https://public-api.wordpress.com/wpcom/v2/sites/$siteId/stats/top-earners?unit=day&date=2018-08-08&limit=10

nbradbury:
Resolves #863 - adds support & related tests for the Woo "top earners" endpoint. Note that the top earners data is **not** stored by SQLite since I'm assuming we don't need to cache it. If I'm incorrect about this, I can add the SQLite storage in a subsequent PR.

cc: @AmandaRiu @aforcier

### DIFF
--- a/.idea/checkstyle-idea.xml
+++ b/.idea/checkstyle-idea.xml
@@ -3,12 +3,12 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:a8c Style" />
+        <entry key="active-configuration" value="LOCAL_FILE:$PROJECT_DIR$/config/checkstyle.xml:a8c Style" />
         <entry key="checkstyle-version" value="8.2" />
         <entry key="copy-libs" value="false" />
         <entry key="location-0" value="BUNDLED:(bundled):Sun Checks" />
         <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
-        <entry key="location-2" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:a8c Style" />
+        <entry key="location-2" value="LOCAL_FILE:$PROJECT_DIR$/config/checkstyle.xml:a8c Style" />
         <entry key="scan-before-checkin" value="false" />
         <entry key="scanscope" value="AllSourcesWithTests" />
         <entry key="suppress-errors" value="false" />
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
@@ -1,11 +1,11 @@
 package org.wordpress.android.fluxc.model
 
-public class WCTopEarnerModel {
+class WCTopEarnerModel {
     var id: Int = 0
-    var currency: String? = ""
-    var image: String? = ""
-    var name: String? = ""
-    var price: Int = 0
+    var currency: String = ""
+    var image: String = ""
+    var name: String = ""
+    var price: Float = 0.0f
     var quantity: Int = 0
-    var total: Int = 0
+    var total: Float = 0.0f
 }
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
@@ -1,0 +1,11 @@
+package org.wordpress.android.fluxc.model
+
+public class WCTopEarnerModel {
+    var localSiteId = 0
+    var currency = ""
+    var image = ""
+    var name = ""
+    var price = ""
+    var quantity = ""
+    var total = ""
+}
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
@@ -1,11 +1,11 @@
 package org.wordpress.android.fluxc.model
 
 public class WCTopEarnerModel {
-    var localSiteId = 0
-    var currency = ""
-    var image = ""
-    var name = ""
-    var price = ""
-    var quantity = ""
-    var total = ""
+    var id: Int = 0
+    var currency: String? = ""
+    var image: String? = ""
+    var name: String? = ""
+    var price: Int = 0
+    var quantity: Int = 0
+    var total: Int = 0
 }
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
@@ -5,7 +5,7 @@ class WCTopEarnerModel {
     var currency: String = ""
     var image: String = ""
     var name: String = ""
-    var price: Float = 0.0f
+    var price: Double = 0.0
     var quantity: Int = 0
-    var total: Float = 0.0f
+    var total: Double = 0.0
 }
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
@@ -7,19 +7,19 @@ import org.wordpress.android.fluxc.generated.WCStatsActionBuilder
 import org.wordpress.android.fluxc.generated.endpoint.WPCOMV2
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.WCOrderStatsModel
+import org.wordpress.android.fluxc.model.WCTopEarnerModel
 import org.wordpress.android.fluxc.network.BaseRequest
 import org.wordpress.android.fluxc.network.UserAgent
 import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComGsonNetworkError
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.orderstats.TopEarnersStatsApiResponse.TopEarner
 import org.wordpress.android.fluxc.store.WCStatsStore.FetchOrderStatsResponsePayload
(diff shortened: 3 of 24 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/TopEarnersStatsApiResponse.kt
```diff
@@ -0,0 +1,48 @@
...
+
+private class TopEarnersDeserializer : JsonDeserializer<TopEarnersStatsApiResponse> {
+    @Throws(JsonParseException::class)
+    override fun deserialize(
+        json: JsonElement,
+        typeOfT: Type,
+        context: JsonDeserializationContext
+    ): TopEarnersStatsApiResponse {
+        val jsonObject = json.asJsonObject
+        val jsonBody = jsonObject.getJsonObject("body")
```
aforcier: The response as currently setup in FluxC doesn't have a `body` element - `date`, `unit`, and `data` are top-level elements. (This also applies to the sample data file for the tests.)

WordPress.com API responses are only wrapped in a `body` tag if the `_envelope=1` param is added to the request, which is something Calypso needs to do (I believe) but not something we need.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/TopEarnersStatsApiResponse.kt
```diff
@@ -0,0 +1,48 @@
...
+import java.lang.reflect.Type
+
+@JsonAdapter(TopEarnersDeserializer::class)
+class TopEarnersStatsApiResponse(
+    val date: String,
+    val unit: String,
+    val topEarners: List<WCTopEarnerModel>
+)
+
+private class TopEarnersDeserializer : JsonDeserializer<TopEarnersStatsApiResponse> {
```
aforcier: I'm not sure I follow why we need a deserializer in this case. We should be able to do something like:

```
class TopEarnersStatsApiResponse : Response {
    val date: String? = null
    val unit: String? = null
    val data: List<TopEarner>? = null

    class TopEarner {
        @SerializedName("ID")
        val id: Int? = null
        ....
    }
}
```

And handle converting the `TopEarner`s into `WCTopEarnerModels` in the `OrderStatsRestClient` as usual.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
```diff
@@ -0,0 +1,11 @@
+package org.wordpress.android.fluxc.model
+
+public class WCTopEarnerModel {
+    var id: Int = 0
+    var currency: String? = ""
```
aforcier: These shouldn't need to be nullable - we can set the value to the empty string if the field is missing in the API response.

.idea/checkstyle-idea.xml
```diff
@@ -3,12 +3,12 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration" value="LOCAL_FILE:$PRJ_DIR$/config/checkstyle.xml:a8c Style" />
+        <entry key="active-configuration" value="LOCAL_FILE:$PROJECT_DIR$/config/checkstyle.xml:a8c Style" />
```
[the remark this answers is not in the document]
nbradbury: Done!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt
```diff
@@ -0,0 +1,11 @@
...
+    var image: String? = ""
+    var name: String? = ""
+    var price: Int = 0
```
aforcier: I believe the `price` and `total` values from the API can have decimal points.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -87,6 +90,31 @@ class OrderStatsRestClient(
         add(request)
     }
 
+    fun fetchTopEarnersStats(site: SiteModel, unit: OrderStatsApiUnit, quantity: Int, force: Boolean = false) {
+        val date = SimpleDateFormat("yyyy-MM-dd").format(Date())
+        val url = WPCOMV2.sites.site(site.siteId).stats.top_earners.url
+        val params = mapOf(
+                "unit" to unit.toString(),
+                "date" to date,
+                "quantity" to quantity.toString())
```
aforcier: The API seems to call this field `limit`, not `quantity`.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -87,6 +90,31 @@ class OrderStatsRestClient(
         add(request)
     }
 
+    fun fetchTopEarnersStats(site: SiteModel, unit: OrderStatsApiUnit, quantity: Int, force: Boolean = false) {
+        val date = SimpleDateFormat("yyyy-MM-dd").format(Date())
```
aforcier: This should use the current time according to the site, not the device's date - [here's](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/4bffaba435d3b523cc5da457f7d55486a31d12b1/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt#L199) how it's handled for order stats.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt
```diff
@@ -194,6 +228,16 @@ class WCStatsStore @Inject constructor(
         emitChange(onStatsChanged)
     }
 
+    private fun handleFetchTopEarrnersStatsCompleted(payload: FetchTopEarnersStatsResponsePayload) {
```
aforcier: Typo in the function name, `s/Earrners/Earners/` 😉

aforcier:
@nbradbury did a quick first pass, left a few comments.

May I suggest adding a new button to the `WooCommerceFragment` for fetching top earner stats, similar to the order stats one (maybe a normal + forced pair of buttons)?

nbradbury:
@aforcier Thanks for the thorough review!. I've addressed everything you uncovered so ready for another round.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/TopEarnersStatsApiResponse.kt
```diff
@@ -1,48 +1,21 @@
...
-        model.currency = json.getString("currency")
-        model.image = json.getString("image")
-        model.name = json.getString("name", unescapeHtml4 = true)
-        model.price = json.getInt("price")
-        model.quantity = json.getInt("quantity")
-        model.total = json.getInt("total")
-        return model
+    class TopEarner {
+        @SerializedName("ID")
+        val id: Int = 0
```
aforcier: Unfortunately we can't trust that these won't be empty/null from the API, so we need to keep all of `TopEarner`'s properties nullable. When converting them to  a `WCTopEarnerModel` in the client, we'll need to do something like:

```kotlin
this.id = topEarner.id ?: 0
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -90,17 +90,30 @@ class OrderStatsRestClient(
...
                 "unit" to unit.toString(),
                 "date" to date,
-                "quantity" to quantity.toString())
+                "limit" to limit.toString())
 
         val request = WPComGsonRequest.buildGetRequest(url, params, TopEarnersStatsApiResponse::class.java,
                 { response: TopEarnersStatsApiResponse ->
-                    val payload = FetchTopEarnersStatsResponsePayload(site, unit, response.topEarners)
+                    val wcTopEarners = ArrayList<WCTopEarnerModel>()
+                    for (topEarner: TopEarner in response.data!!) {
```
aforcier: We can avoid a force unwrap via `!!` (which could crash if `data` is null) and do this in a slightly more idiomatic way:

```kotlin
val topEarners = response.data?.map {
    WCTopEarnerModel().apply {
        id = it.id ?: 0
        currency = it.currency ?: ""
        image = it.image ?: ""
        name = it.name ?: ""
        price = it.price ?: 0F
        this.quantity = it.quantity ?: 0
        total = it.total ?: 0F
    }
} ?: emptyList()
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/TopEarnersStatsApiResponse.kt
nbradbury: Of course 🤦‍♂️ Fixed in 182e4f2.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
nbradbury: So **that's** what `map` is for - nice! Addressed in 3b85106.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -100,15 +100,15 @@ class OrderStatsRestClient(
...
-                    for (topEarner: TopEarner in response.data!!) {
-                        WCTopEarnerModel().apply {
-                            this.id = topEarner.id ?: 0
-                            this.currency = topEarner.currency ?: ""
-                            this.image = topEarner.image ?: ""
-                            this.name = topEarner.name ?: ""
-                            this.price = topEarner.price ?: 0.0f
-                            this.quantity = topEarner.quantity ?: 0
-                            this.total = topEarner.total ?: 0.0f
+                    response.data?.map {
```
aforcier: Note that you can directly set `val topEarners = response.data?.map {...` here, and avoid the need to define it on the line above, as well as dropping the final `wcTopEarners.add(this)` call. `map {}` will understand that you're trying to return a `WCTopEarnerModel` for each entry in `response`, and will create the list implicitly.

PR #866, opened 2018-08-10
NotificationStore with device registration/unregistration

aforcier:
Adds a basic `NotificationStore`, with support for registering and unregistering the device for push notifications with WordPress.com (using a GCM/FCM token).

Also adds new support for a site ID parameter during registration, which configures the server to only send notifications for that site (and not for any site on the account). This is intended for use with the WooCommerce app.

This replicates much of the registration/deregistration work done in the WordPress app - we could migrate that over to use the `NotificationStore` at some point too.

**Not ready for review**

### TODO:
* [x] Move device registration parameter config to FluxC
* [x] Move UUID creation and storage logic to FluxC
* [x] Define preset values for app key
* [x] Add tests

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_NotificationTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_NotificationTest.kt
@@ -1,0 +1,104 @@
+package org.wordpress.android.fluxc.mocked
+
+import com.google.gson.JsonObject
+import org.greenrobot.eventbus.Subscribe
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.TestUtils
+import org.wordpress.android.fluxc.action.NotificationAction
+import org.wordpress.android.fluxc.annotations.action.Action
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.module.ResponseMockingInterceptor
+import org.wordpress.android.fluxc.network.rest.wpcom.notifications.NotificationRestClient
+import org.wordpress.android.fluxc.store.NotificationStore.NotificationAppKey
+import org.wordpress.android.fluxc.store.NotificationStore.RegisterDeviceResponsePayload
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+import kotlin.properties.Delegates
+
+/**
+ * Tests using a Mocked Network app component. Test the network client itself and not the underlying
+ * network component(s).
+ */
+class MockedStack_NotificationTest : MockedStack_Base() {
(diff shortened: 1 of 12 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_NotificationTest.kt
```diff
@@ -0,0 +1,104 @@
...
+        // No site was specified, so the site selection param should be omitted
+        assertNull(requestBodyMap["selected_blog_id"])
+
+        val payload = lastAction!!.payload as RegisterDeviceResponsePayload
+
+        assertEquals(responseJson.get("ID").asString, payload.deviceId)
+    }
+
+    @Test
+    fun testRegistrationWooCommerce() {
```
nbradbury: Should there also be a test for unregistering the device from push notifs?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
```diff
@@ -0,0 +1,95 @@
...
+                "device_token" to gcmToken,
+                "device_family" to "android",
+                "app_secret_key" to appKey.value,
+                "device_name" to deviceName,
+                "device_model" to "${Build.MANUFACTURER} ${Build.MODEL}",
+                "app_version" to PackageUtils.getVersionName(appContext),
+                "version_code" to PackageUtils.getVersionCode(appContext).toString(),
+                "os_version" to Build.VERSION.RELEASE,
+                "device_uuid" to uuid,
+                "selected_blog_id" to site?.siteId.toString().takeIf { site != null }
```
nbradbury: I'm confused by this line. If the passed site _is_ null, then we end up setting `selected_blog_id` to "null."

screen shot 2018-09-22 at 10 57 59 am

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_NotificationTest.kt
aforcier: Sure! Added one in 4470a01.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
aforcier: Ah, my mistake - I had a few versions of how to do this and I must've garbled the end result that I actually committed.

Fixed in fe77af6 - that line is supposed to return a null value overall if `site == null`, but then the whole thing needs to be wrapped in a `listOfNotNull()` to filter out any `null` values before creating a map (it's a bit verbose unfortunately).

Ultimately it didn't matter because null body params are filtered out by Volley before being sent, but we may as well do it properly here.

I also updated the corresponding test line to more accurately check that the property is undefined rather than null - this wouldn't have failed the test before anyway though, since the test checks the params actually sent to the server, by which time the `null` was stripped out.

nbradbury:
Looks good! :shipit:

PR #868, opened 2018-08-13
Introduced Fetch Supported Countries API

usamahamid:
This is an API required for implementing [custom domain association feature](https://github.com/wordpress-mobile/WordPress-Android/issues/7923).

When User associates a custom domain, they have to input and confirm basic information about themselves.
In this screen, they will be required to select one of the supported countries.

This API provides a list of supported countries from which User can select.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SupportedCountriesResponse.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SupportedCountriesResponse.kt
@@ -1,0 +1,6 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.site
+
+class SupportedCountriesResponse(
+    val code: String?,
+    val name: String?
+)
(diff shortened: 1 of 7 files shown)


usamahamid:
@malinajirka , This PR introduces a new API to fetch supported countries to be listed in the domain registration page.
Please do check it out, when you have a free moment.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SupportedCountriesResponse.kt
```diff
@@ -0,0 +1,6 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.site
+
+class SupportedCountriesResponse(
```
malinajirka: Just wondering whether we shouldn't rename this class to `SupportedCountryResponse` to be consistent with `DomainSuggestionResponse`. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -237,6 +238,19 @@ public DomainAvailabilityResponsePayload(@NonNull DomainAvailabilityError error)
         }
     }
 
+    public static class SupportedCountriesResponsePayload extends Payload<SupportedCountriesError> {
+        public List<SupportedCountriesResponse> supportedCountriesResponses;
```
malinajirka: This is just a nitpick, but we might consider renaming this to `supportedCountries` to be consistent with other Payload classes. Wdyt?

malinajirka:
Thanks @usamahamid, it looks good overall. I've left few minor inline comments. 

I'm also wondering whether the name 'supportedCountries' in SiteStore isn't a bit confusing. AFAIU it's a list of supported countries for the specific domain, right?

malinajirka:
Thanks for the changes! Looks good! ;)

PR #870, opened 2018-08-13
Remote jetpack install

planarvoid:
- add rest client and store for Jetpack to enable remote install
- add interface in both standard event driven arch and coroutines
- bump kotlin
- bump coroutines

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/store/JetpackStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/JetpackStoreTest.kt
@@ -1,0 +1,87 @@
+package org.wordpress.android.fluxc.store
+
+import com.nhaarman.mockito_kotlin.eq
+import com.nhaarman.mockito_kotlin.verify
+import com.nhaarman.mockito_kotlin.whenever
+import kotlinx.coroutines.experimental.delay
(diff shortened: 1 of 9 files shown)


example/src/test/java/org/wordpress/android/fluxc/store/JetpackStoreTest.kt
```diff
@@ -0,0 +1,87 @@
...
+    @Mock private lateinit var site: SiteModel
+    private lateinit var jetpackStore: JetpackStore
+
+    @Before
+    fun setUp() {
+        jetpackStore = JetpackStore(jetpackRestClient, dispatcher)
+    }
+
+    @Test
+    fun `on install triggers rest client and returns success`() = runBlocking {
```
0nko: Cool method names :D

0nko:
We can :shipit: once the checks are green.

example/src/test/java/org/wordpress/android/fluxc/store/JetpackStoreTest.kt
planarvoid: not sure we want to write them like that but I think it's more readable for tests 😄

0nko:
:shipit:

PR #871, opened 2018-08-13
Feature/introduce portfolio

glung:
/!\ Will be based on https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/869
**So start the review on commit 3** (https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/871/commits/1c929e7f3df3112b8f054c56fe1208c23107c3c2)

I putting on an early PR to get feedback. 

# Scope

After talking to @planarvoid, it seemed to be a good idea to introduce the `Portfolio` post type, the way the `Page` are being refactored - or let's say, there is no apparent good reason to not do it.

# Details 
- Introduce PostType.PortofolioType
- Introduce a `PortofolioModel` and `PortofolioStore` - basically the same as `PageStore`. 
- Update the clients to use `PortofolioType`

# Note

According to this chart, Portofolio can share the same properties in PostModel (without Geolocation). But it seems we reuse `category_id` and `tags` for `project category` and `project tags` which seems reasonable to me. 

This chart was build out WordPress.com editor. 

screen shot 2018-08-13 at 7 36 36 pm

cc @kwonye @0nko


glung:
Closing since the interesting bits were cherry-picked in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/869

PR #872, opened 2018-08-13
Device Connection Monitoring

AmandaRiu:
Added a new broadcast receiver to listen for changes in device connectivity. The logic has been borrowed from the current WordPress-Android app and has been consolidated to a single class in FluxC named `ConnectionChangeReceiver`. The App using this functionality would need to register this receiver to get connection updates. This can be done globally in the `Application` class, or in a single `Activity`. I've added it globally to the Example App for example implementation and testing purposes.

When a change in network status that effects the previously known network availability, a new `OnNetworkStatusChanged` event will be dispatched. 

### Design notes
I added this feature to FluxC instead of the Woo App for two reasons:
1. FluxC is responsible for the network and database layer of the Woo App already. 
2. To make use of the existing `Dispatcher` pattern and it's implementation of the `EventBus`. Since this feature does not make use of a database table or take incoming actions, I did not add it to FluxC as a store. 

I added the feature to core so it can be used across any app utilizing fluxC. The WordPress-Android app has a similar implementation already, but that would not be affected since the class being registered is different. They could eventually cut over to using this one however. 

### To Test
1. Make sure the device is connected to the internet.
2. Open the Example App - the current connectivity status will appear in the log section.
3. Put the device in Airplane mode - the updated connectivity status will appear in the log section

screen shot 2018-08-13 at 5 51 12 pm

cc: @aforcier @nbradbury

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ExampleApp.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ExampleApp.kt
@@ -2,17 +2,29 @@ package org.wordpress.android.fluxc.example
 
 import android.app.Activity
 import android.app.Application
+import android.content.IntentFilter
+import android.net.ConnectivityManager
+import android.os.Bundle
 import com.yarolegovich.wellsql.WellSql
 import dagger.android.AndroidInjector
 import dagger.android.DispatchingAndroidInjector
 import dagger.android.HasActivityInjector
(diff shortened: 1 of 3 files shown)


AmandaRiu:
After discussing this with @nbradbury and @aforcier, it has been decided this functionality should stay in the main app since it's really a device-level feature. Closing.

PR #873, opened 2018-08-13
Merge code that went into wrong branch

AmandaRiu:
PR #856 was erroneously merged into this branch instead of develop. Verified the changes are the same (this 232 branch was the one that was merged right before it so no changes there). This just makes sure those changes get into develop.


nbradbury:
:shipit:

PR #878, opened 2018-08-16
Quick Start api

khaykov:
This PR adds a mobile quick start API that is used to notify the server that mobile quick start is completed for a specific site, and that completion notification should be sent to the client.

There is no special errors and responses returns either `{success:true}` or `{success:false}`.


khaykov:
Thanks! I'll restart the build and see maybe it'll help.
About the MobileQuickStart thing - I thought that we might get a quick start on other platforms at some point, and do something with it from mobile, but it will probably not work like this, and mobile realistically should not care about any other "quick starts".

PR #879, opened 2018-08-17
Add page reload to jetpack install

planarvoid:
- The Jetpack store installs Jetpack to self-hosted sites remotely. This PR adds a step to the flow that refetches the SiteModel after the install call. It's necessary in order to let the app know that the field `isJetpackInstalled` is now updated. 
- I'm using advantages of coroutines - merging these two operations together and returning the result once both are done instead of sending events. 
- To test: there is nothing to test here until the related WPAndroid PR is done

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/JetpackStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/JetpackStore.kt
@@ -61,7 +61,7 @@ class JetpackStore
     }
 
     private suspend fun reloadSite(site: SiteModel) = suspendCoroutine<Unit> { cont ->
-        launch {
+        launch(CommonPool) {
             delay(5000)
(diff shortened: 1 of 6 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/JetpackStore.kt
```diff
@@ -45,6 +65,18 @@ class JetpackStore
         }
     }
 
+    private suspend fun reloadSite(site: SiteModel, context: CoroutineContext) = suspendCoroutine<Unit> { cont ->
+        siteStore.onAction(SiteActionBuilder.newFetchSiteAction(site))
+        siteContinuation = cont
+        launch(CommonPool) {
```
0nko: You're not using the supplied coroutine context here. Is that intentional?

0nko:
:shipit:

PR #880, opened 2018-08-17
Support scheme param for magic link e-mail requests

### LINKED ISSUE #834: Support setting the scheme for magic links
We currently don't specify a scheme when requesting a magic link. This means that the server defaults to creating the magic link with a `wordpress://` scheme and e-mailing it.

This isn't currently an issue, but with the WooCommerce app also using magic links, we should allow the scheme to be specified (the `woocommerce://` scheme has already been added to the backend).

Currently, opening any magic link while having both apps installed will open a system pop-up dialog allowing you to choose which app to open the magic link with. The issue is that if a user selects 'always', future magic links will always open the app they selected, and they won't be able to get it to work with the other app (not without tweaking their settings).

aforcier:
Closes #834. Adds support for a `scheme` parameter when generating magic link requests, allowing magic links to launch the app that requested them (unlike the current situation, where magic links requested from the WordPress or WooCommerce app will offer to open both apps).

To test:
This should have no effect on the WordPress app (this should be verified) - magic links created from it should open the WordPress app (or offer a choice between release and dev versions if both are installed).

A PR will be opened in the WooCommerce app making use of the `woocommerce` scheme, with more testing information.


nbradbury:
Looks good! :shipit:

PR #883, opened 2018-08-22
Issue/882 detect woo empty store

### LINKED ISSUE #882: Woo: detect empty store
We need to detect when the user is viewing a new store, and the simplest way to do that is to check whether the store has **ever** had any orders.

nbradbury:
Resolves #882 - uses `GET /wc/v2/orders` to request a single order of a specific type (or any type) to determine whether any orders exist.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
@@ -18,6 +18,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComGson
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
 import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequest
 import org.wordpress.android.fluxc.store.WCOrderStore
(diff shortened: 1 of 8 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -89,6 +90,42 @@ class OrderRestClient(
...
+        val params = mapOf(
+                "per_page" to "1",
+                "offset" to "0",
+                "status" to statusFilter)
+        val request = JetpackTunnelGsonRequest.buildGetRequest(url, site.siteId, params, responseType,
+                { response: List<OrderApiResponse>? ->
+                    val orderModels = response?.map {
+                        orderResponseToOrderModel(it).apply { localSiteId = site.id }
+                    }.orEmpty()
+                    val hasOrders = orderModels.size > 0
```
AmandaRiu: lint warning: This can be changed to `orderModels.isNotEmpty()`

nbradbury:
> This all looks good, but I'm curious why we didn't just use the existing FetchOrdersCountPayload? We can pass it any statusFilter and it'd give you the same information. This new one may be very slightly faster since it's limiting the result to 1,

That's exactly the reason I didn't reuse the existing `FetchOrdersCountPayload`. Limiting the response to a single order saves a *lot* on bandwidth, and I wasn't sure if the backend would support returning only a count in time for our open beta. I should've clarfied this in the PR description, but I look at this as a stopgap until the backend provides a more efficient solution.

AmandaRiu:
Thanks for the clarification 👍 

Looks good! :shipit:

PR #886, opened 2018-08-30
Activity log coroutines

planarvoid:
Converting Activity log to coroutines consists of two parts. First one is using new methods that instead of creating a Request and triggering it asynchronously encapsulate it to a suspend function. The result gets converted to a Response object. This makes it much easier to unit test the `RestClient`.
The second part is converting the `ActivityLogStore`. I'm still keeping original `onAction` which should behave exactly the same as before. I'm also making the suspend function public in order of using these in WPAndroid.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequestBuilder.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequestBuilder.kt
@@ -40,7 +40,7 @@ class WPComGsonRequestBuilder
         url: String,
         params: Map<String, String>,
         clazz: Class<T>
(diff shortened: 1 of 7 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequestBuilder.kt
```diff
@@ -28,6 +28,26 @@ class WPComGsonRequestBuilder
...
     }
 
+    /**
+     * Creates a new GET request.
+     * @param restClient rest client that handles the request
+     * @param url the request URL
+     * @param params the parameters to append to the request URL
+     * @param clazz the class defining the expected response
+     */
+    suspend fun <T> syncGetRequest(
```
0nko: Nice refactoring 👍.

PR #889, opened 2018-09-04
Feature/formattable content

### LINKED ISSUE #888: Extract Notifications - Introduce FormattableContent
The structure in the API response for Notifications and Activity Log is the same. We decided to refactor Notifications so we can reuse the same logic. 
We currently use raw json instead of type safe objects. 

Goal of this task is to 
- introcuce `FormattableContent` which represents content of both Notifications and Activities (from Activity Log)
- create a mapping from the raw json into a `FormattableContent`.

planarvoid:
Fixes #888 
- The result of this PR should be rich content inside ActivityLog items. Currently we're just using the `text` field from the `Content` object. In this PR I'm introducing a `FormattableContent` matching the object coming from the backend for `ActivityLog` and `Notifications`. It is the same structure used in both features and it allows us to have clickable `ActivityLog` items (for example - an item mentioning a site being created can link to the site itself). As part of this PR I'm using the new model in ActivityLog. I'm storing the whole JSON Content inside the database because the `FormattableContent` is too rich and prone to changes. It should be OK since we wipe the DB on every change (and use it basically as a cache). All the fields on `FormattableContent` are optional. There is a public mapper to be used in `Notifications`.

This PR has to be merged at the same time as the WPAndroid changes to FluxC.
There is nothing to test without the WPAndroid changes.


example/src/test/java/org/wordpress/android/fluxc/tools/FormattableContentMapperTest.kt
```diff
@@ -0,0 +1,141 @@
...
+            "              \"context\": \"single\",\n" +
+            "              \"id\": 17,\n" +
+            "              \"root_id\": 68\n" +
+            "            }\n" +
+            "          ]\n" +
+            "        }"
+
+    @Before
+    fun setUp() {
+        val gson = Gson()
```
malinajirka: Can't we use an injected instance?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
```diff
@@ -335,6 +335,15 @@ public void onUpgrade(SQLiteDatabase db, WellTableManager helper, int oldVersion
                 db.execSQL("CREATE TABLE IF NOT EXISTS QuickStartStatusModel (_id INTEGER PRIMARY KEY AUTOINCREMENT,"
                            + "SITE_ID INTEGER,IS_COMPLETED INTEGER,IS_NOTIFICATION_RECEIVED INTEGER)");
                 oldVersion++;
+            case 40:
```
malinajirka: Shouldn't we increment the version number in `getDbVersion`? 

Tbh the current state is quite confusing, since the version is `40`, but there wasn't any case for `40` and some users might have already upgraded the database before the case was added.

aforcier: Drive-by comment: The version number in `getDbVersion` should definitely be incremented.

The switch in this method is not over `newVersion`, but over `oldVersion`. So what this case is saying is "if the old version is 40, perform these steps". Once you run out of cases (and you've incremented the `oldVersion` at each step), you should hit the number defined in `getDbVersion`, which should be `== newVersion`.

Incidentally @planarvoid we should add an `oldVersion++;` at the end of the case as with all other cases - omitting it won't break anything right away, but it's convention and helps prevent the next person from writing a migration that won't actually be executed.

planarvoid: Ah I see, I missed the old version increase 👍. It's not on purpose

malinajirka:
LGTM

PR #895, opened 2018-09-11
Issue/revisions endpoints

khaykov:
This PR adds API calls for Revisions endpoints for both Pages and Posts.

`RevisionModel` contains all the info from the response in a more flatter and easier to use format.

When something is wrong, endpoints return generic network errors or empty revisions array.

This is probably not a final version of the client, as our requirements might update when we start developing client side.


fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -276,8 +322,8 @@ public int getUploadedPagesCountForSite(SiteModel site) {
      */
     public PostModel getPostByLocalPostId(int localId) {
         List<PostModel> result = WellSql.select(PostModel.class)
-                .where().equals(PostModelTable.ID, localId).endWhere()
-                .getAsModel();
+                                        .where().equals(PostModelTable.ID, localId).endWhere()
+                                        .getAsModel();
```
theck13: Can we revert these deep indents for better readability?

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -291,9 +337,9 @@ public PostModel getPostByLocalPostId(int localId) {
      */
     public PostModel getPostByRemotePostId(long remoteId, SiteModel site) {
         List<PostModel> result = WellSql.select(PostModel.class)
-                .where().equals(PostModelTable.REMOTE_POST_ID, remoteId)
-                .equals(PostModelTable.LOCAL_SITE_ID, site.getId()).endWhere()
-                .getAsModel();
+                                        .where().equals(PostModelTable.REMOTE_POST_ID, remoteId)
+                                        .equals(PostModelTable.LOCAL_SITE_ID, site.getId()).endWhere()
+                                        .getAsModel();
```
theck13: Can we revert these deep indents for better readability?

khaykov:
Thanks! Fixed indents. Interesting fact is that I was using the code style settings from the project, and they produce this type of indentation. I switched back to default AS style, and it works as expected 🤷‍♂️

PR #896, opened 2018-09-12
Replace runBlocking in unit tests with a new function called test

planarvoid:
When you run a test that's testing a suspended function, coroutines (in the current version) require you to wrap the test in `runBlocking`. Since `runBlocking` might return a result, the test only works when it doesn't return any value. Otherwise you have to use `runBlocking<Unit>`. I felt like this is unnecessary so I'm creating here a helper function that replaces both of these calls with a simpler `test` one. This way you don't have to worry about any return value and the meaning is clear. The method is introduced only in `CoroutineTestUtils` in tests so it's not visible from production code.

Let me know if the method name works for you.


0nko:
Looks good :shipit:

PR #901, opened 2018-09-14
Add fixed list of stats

planarvoid:
- this PR is introducing a `StatsStore` which now returns a fixed list of Insights types for the insights screen in Stats and a list of block types for the Day/Week/Month/Year screen. 
- As a next step we'll replace this with a user selected list of types in a custom order.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/StatsStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/StatsStore.kt
@@ -6,7 +6,7 @@ import org.wordpress.android.fluxc.store.StatsStore.InsightsTypes.COMMENTS
 import org.wordpress.android.fluxc.store.StatsStore.InsightsTypes.FOLLOWERS
 import org.wordpress.android.fluxc.store.StatsStore.InsightsTypes.LATEST_POST_SUMMARY
 import org.wordpress.android.fluxc.store.StatsStore.InsightsTypes.MOST_POPULAR_DAY_AND_HOUR
-import org.wordpress.android.fluxc.store.StatsStore.InsightsTypes.PUBLICISE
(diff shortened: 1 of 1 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/StatsStore.kt
```diff
@@ -0,0 +1,78 @@
...
+    suspend fun getInsights(): List<InsightsTypes> = withContext(coroutineContext) {
+        return@withContext listOf(
+                LATEST_POST_SUMMARY,
+                TODAY_STATS,
+                ALL_TIME_STATS,
+                MOST_POPULAR_DAY_AND_HOUR,
+                COMMENTS,
+                TAGS_AND_CATEGORIES,
+                FOLLOWERS,
+                PUBLICISE
```
0nko: Spelling: should be PUBLICIZE.

fluxc/src/main/java/org/wordpress/android/fluxc/store/StatsStore.kt
```diff
@@ -0,0 +1,78 @@
...
+        MOST_POPULAR_DAY_AND_HOUR,
+        ALL_TIME_STATS,
+        FOLLOWER_TOTALS,
+        TAGS_AND_CATEGORIES,
+        ANNUAL_SITE_STATS,
+        COMMENTS,
+        FOLLOWERS,
+        TODAY_STATS,
+        POSTING_ACTIVITY,
+        PUBLICISE
```
0nko: Same, spelling.

fluxc/src/main/java/org/wordpress/android/fluxc/store/StatsStore.kt
planarvoid: I was following the designs, British vs American English. I'll change it 👍

PR #903, opened 2018-09-18
Issue/support for local revisions

khaykov:
This PR adds functionality for storing and retrieving of local revisions.

The response we are getting from revisions API is not the easiest one to work with (two arrays of objects loosely linked together) so the majority of code is dealing with converting models to/from format convenient for the client to/from a format acceptable for storage in DB. We have to store revision info two separate tables and glue them together (technically, we can store them in one table, but this option is a bit more confusing in my opinion).

Revisions are part of a PostStore. I was thinking about separating them, but this adds even more boilerplate code for very little functionality. Revisions do not exist outside of the post/page context, so this made the most sense to me.

Again, this PR is based on loose requirements of a client, so if you think something must be changed do let me know.


example/src/test/java/org/wordpress/android/fluxc/post/PostRevisionModelTest.kt
```diff
@@ -0,0 +1,110 @@
...
+        assertEquals(DiffOperations.DELETE, contentDiffs[2].operation)
+        assertEquals("del content", contentDiffs[2].value)
+    }
+
+    @Test
+    fun testRevisionModelEquals() {
+        val sampleRevision1 = PostTestUtils.generateSamplePostRevision()
+        val sampleRevision2 = PostTestUtils.generateSamplePostRevision()
+
+        assertTrue(sampleRevision1 == sampleRevision2)
```
theck13: The `assertTrue(sampleRevision1 == sampleRevision2)` statement can be replaced with `assertEquals(sampleRevision1, sampleRevision2)`.

example/src/test/java/org/wordpress/android/fluxc/post/PostRevisionModelTest.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        sampleRevision1.titleDiffs[0] = Diff(DiffOperations.COPY, "wrong value")
+        assertFalse(sampleRevision1 == sampleRevision2)
```
theck13: The `assertFalse(sampleRevision1 == sampleRevision2)` statement can be replaced with `assertNotEquals(sampleRevision1, sampleRevision2)`.

khaykov:
Thanks! I changed the calls to `assertEquals` and `assertNotEquals`.

PR #907, opened 2018-09-19
Issue/904 woo visitor stats

### LINKED ISSUE #904: Woo: Add visitor stats
Add support for visitor stats to FluxC, which is a 'classic' WP.com stats endpoint (and not a Store stats endpoint, like we use for the rest of the graph): `/rest/v1.1/sites/$site/stats/visits/`

ex: `/sites/$siteId/stats/visits?unit=month&quantity=1`
Ref: https://anchor.host/wordpress-com-stats-via-api/

nbradbury:
Resolves #904 - adds the endpoint to get visitor stats for the current store.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -89,6 +94,51 @@ class OrderStatsRestClient(
...
+
+        add(request)
+    }
+
+    /**
+     * Returns the number of visitors from the VisitorStatsApiResponse payload
+     */
+    private fun getVisitorsFromResponse(response: VisitorStatsApiResponse): Int {
+        try {
+            val array = response.data?.asJsonArray?.get(0) as JsonArray
```
AmandaRiu: It looks like you always grab the first element in the array of data in the results. So if grabbing the stats for "days" you grab only the visitors from today. If months, only the visitor count for this month, etc. While I actually agree with this whole-heartedly because it would make more sense to show today, this month, this week, and this year, looking at the current order stats card, it appears that the other two fields, `Orders` and `Revenue` are the sum of all orders in the range pulled.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -89,6 +94,51 @@ class OrderStatsRestClient(
...
+        unit: OrderStatsApiUnit,
+        date: String,
+        limit: Int,
+        force: Boolean = false
+    ) {
+        val url = WPCOMREST.sites.site(site.siteId).stats.visits.urlV1_1
+        val params = mapOf(
+                "unit" to unit.toString(),
+                "date" to date,
+                "limit" to limit.toString())
```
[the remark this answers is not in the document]
nbradbury: My mistake - that's supposed to be `quantity`. I'll change that and document what it does.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
nbradbury: I think you may be right here. I'll dig into this further.

nbradbury:
@AmandaRiu I've updated this PR to request the same quantity as order stats, and I'm now returning the sum. I verified that it matches the web, so ready for round two!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -126,13 +125,20 @@ class OrderStatsRestClient(
     }
 
     /**
-     * Returns the number of visitors from the VisitorStatsApiResponse payload
+     * Returns the number of visitors from the VisitorStatsApiResponse data, which is an array of items for
+     * each period, the second element of which contains the visitor count for that period
      */
     private fun getVisitorsFromResponse(response: VisitorStatsApiResponse): Int {
```
AmandaRiu: You could replace this method with:
```
private fun getVisitorsFromResponse(response: VisitorStatsApiResponse): Int {
    return try {
        response.data?.asJsonArray?.map { it.asJsonArray?.get(2)?.asInt ?: 0 }?.sum() ?: 0
    } catch (e: Exception) {
        AppLog.e(T.API, "${e.javaClass.simpleName} parsing visitor stats", e)
        0
    }
}
```

nbradbury: I keep forgetting about the magic of `map`! Updated in 04084e6.

nbradbury:
> I think there may be a bug in the api or something...when I look at the actual api response for your test `FETCH 30-DAY VISITOR STATS`, 13 months are included. I played with the API using Paw, and it seems that if you _don't_ supply a value for `date` in the request, then exactly 12 months are returned, but if you do supply it, even if it's _today_, an additional month is returned.

Hmm, I'm seeing the same thing. I checked Calypso to see what it was doing, but it shows 15 months of data so it's not the same. I did notice, though, that it appends `&stat_fields=visitors` to the request to limit the response to just visitor data (no clicks, etc). I'll make that change here, and contact someone about the 13mo mystery.

nbradbury:
@AmandaRiu I made the above changes, ready for another round!

AmandaRiu:
> that it appends &stat_fields=visitors to the request to limit the response to just visitor data (no clicks, etc) - and when I append that, I get 12 months as expected!

I'm getting 10 months. If you pull up chrome inspector and what the network tab when you run that test `FETCH 30 DAY VISITOR STATS (FORCED)`, your request looks like this:
```
https://public-api.wordpress.com/rest/v1.1/sites/140191676/stats/visits/?unit=month&date=2018-09&quantity=12&stat_fields=visitors&locale=en_US
```
but the response only has 10 months:
```
{
   "date":"2018-09-01",
   "unit":"month",
   "fields":[
      "period",
      "visitors"
   ],
   "data":[
      [
         "2017-12-01",
         5
      ],
      [
         "2018-01-01",
         5
      ],
      [
         "2018-02-01",
         23
      ],
      [
         "2018-03-01",
         6
      ],
      [
         "2018-04-01",
         0
      ],
      [
         "2018-05-01",
         2
      ],
      [
         "2018-06-01",
         75
      ],
      [
         "2018-07-01",
         36
      ],
      [
         "2018-08-01",
         12
      ],
      [
         "2018-09-01",
         0
      ]
   ]
}
```

Also, the test app, the button says 30 days, but the actual test is 12 months

AmandaRiu:
@nbradbury actually, it looks like maybe it was the test account I was using did have a full year. I tried another account, and am still seeing the 13 months in the response when hitting the `month` stat type

nbradbury:
>  I tried another account, and am still seeing the 13 months in the response

Bah, you're right. I *was* seeing 12 months in the console but turns out that only happened because I didn't pass a date. When you don't pass a date it assumes today - but not passing a date turns out not to be a reliable solution here.

Here's why: when I pass `2018-09` as the date, I get 13 months. When I pass `2018-09-01` as the date, I also get 13 months. But when I pass `2018-09-22` as the date, I get 12 as expected. I don't know why this is happening - but it's possible it's *supposed* to work this way.

Our choice here, then, is to ship this as-is on Monday and then deal with the solution once we hear from the backend devs, or hold off shipping this until we have the solution. I'm okay with shipping now if you are.

PR #909, opened 2018-09-20
Bump KtLint version to 0.28.0 and fix all the failures

planarvoid:
I've noticed the ktlint version we're using is obsolete so I'm bumping it and fixing all the ktlint failures. I was mostly using IDE reformatting tool but I went through all the changes and all seems fine.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_CacheTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_CacheTest.kt
@@ -59,7 +59,7 @@ class MockedStack_CacheTest : MockedStack_Base() {
         requestQueue.cache.clear()
 
         // Make initial request
-        with (prepareAndCreateRequest()) {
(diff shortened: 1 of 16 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_CacheTest.kt
```diff
@@ -197,9 +197,11 @@ class MockedStack_CacheTest : MockedStack_Base() {
     private fun prepareAndCreateRequest(params: Map<String, String> = mapOf()): WPComGsonRequest<*> {
         interceptor.respondWith(responseJson)
 
-        return WPComGsonRequest.buildGetRequest(requestUrl, params, Any::class.java,
+        return WPComGsonRequest.buildGetRequest(
```
aforcier: This changes to this return statement don't seem necessary for ktlint to pass, but if that's what the reformatter did we may as well keep it 🤷‍♂️

aforcier:
:shipit:!

Issue #910, opened 2018-09-20
Add endpoints for Insights 


planarvoid:
Our current data fetching and storing for Stats is handled in WPAndroid. We need to migrate it to FluxC using the new architectural patterns.

PR #911, opened 2018-09-20
Add first Insights endpoint - All Time, Most Popular and Latest post

planarvoid:
Ticket #910 
- This PR is adding first 3 endpoints for fetching All Time insights, Most Popular insights and insights for Latest Post
- I'm adding an `InsightsRestClient` and an `InsightsStore`
- As another PR we need to add the data caching and unit tests for both. As part of this PR I've added a simple Integration test (most of it is copy paste) that fetches the data a checks it's present.


0nko:
Thanks for the explanation! 👍

fluxc/src/main/java/org/wordpress/android/fluxc/model/stats/InsightsLatestPostModel.kt
```diff
@@ -0,0 +1,14 @@
+package org.wordpress.android.fluxc.model.stats
+
+import java.util.Date
+
+data class InsightsLatestPostModel(
+    val siteId: Long,
+    val postTitle: String,
+    val postURL: String,
+    val postDate: Date,
+    val postID: Long,
```
0nko: This is a nitpick - you're using "Id" in `siteId` but "ID" in `postID`.

0nko:
:shipit:

PR #914, opened 2018-09-22
Revisions DB migration script

khaykov:
Forgot to add a DB migration script, which might cause some issue when upgrading/downgrading app versions.

To test:

1. Install the current `develop` branch of wpandroid with flux-c reference pointed to this PR `74b85fe6a290278d35952394d4e749ae5962d3e6`
2. Revert the flux-c reference and install the app again.
3. Make sure app installs correctly.


theck13:
Right.  I think the testing steps should be this, which would be upgrading:
1. Install WordPress app with `2304a69c29d3728802ac827f006eb4a0d833004c` (i.e. `develop`) FluxC hash.
2. Run app without crash.
3. Install WordPress app with `74b85fe6a290278d35952394d4e749ae5962d3e6` FluxC hash.
4. Run app without crash.

But the steps listed in the pull request testing switch the order of the FluxC hash like this:
1. Install WordPress app with `74b85fe6a290278d35952394d4e749ae5962d3e6` FluxC hash.
2. Run app without crash.
3. Install WordPress app with `2304a69c29d3728802ac827f006eb4a0d833004c` (i.e. `develop`) FluxC hash.
4. Run app with crash.

I just wanted to verify I was testing correctly using the steps without the crash, which are different than the steps you listed that cause a crash.

khaykov:
Yes, my steps were in the wrong order :)

PR #915, opened 2018-09-25
Add Insight endpoints tests

planarvoid:
- add unit test for InsightsRestClient and InsightsStore
- simple coverage of the basic usecases, I expect them to grow once we add DB caching
- Instead of setting views to 0 the store now returns a failure if the second call to Post views fails on the Latest post insights


0nko:
:shipit:

PR #925, opened 2018-10-15
Use 'AndroidKeyStore' in the trust manager

jtreanor:
Currently `MemorizingTrustManager` stores SSL certs using a key store in the app's file directory with a password that is hardcoded and therefore public. Although it can be tricky to access the file, this is not a secure & robust way of keeping the cert info safe.

I have made a relatively small change here to change the `KeyStore` type to `AndroidKeyStore`. This is the new type introduced in API 18 which allows saving in a system `KeyStore`. Access is limited to the app so cannot be snooped.

Some relevant resources:

- https://developer.android.com/reference/java/security/KeyStore
- https://developer.android.com/training/articles/keystore#UsingAndroidKeyStore
- https://academy.realm.io/posts/secure-storage-in-android-san-francisco-android-meetup-2017-najafzadeh/#android-keystore-system

To test:
- I would like to do some more thorough testing with custom SSL certificates before moving forward.
- Try out migration from the existing version to make sure it works gracefully.

If we go ahead with this approach I will follow up with a change to delete `wpstore_certs_truststore.bks` if it exists. This will ensure no sensitive data is sitting around, even after updating.


jtreanor:
Quick update here:

- I have tested and confirmed that the behaviour for self signed certs is unchanged. It persists between launches.
- The migration behaviour as expected is that trust for the cert needs to be granted again since the old store is never loaded. I think this is the correct way to handle this. Given `wpstore_certs_truststore.bks` is now never loaded, I think it is fine not to delete it.

fluxc/src/main/java/org/wordpress/android/fluxc/network/MemorizingTrustManager.java
```diff
@@ -110,54 +99,11 @@ private X509TrustManager getTrustManager(@Nullable KeyStore keyStore) {
     }
 
     private KeyStore loadTrustStore() throws IOException, GeneralSecurityException {
-        File localKeyStoreFile = new File(mContext.getFilesDir(), KEYSTORE_FILENAME);
```
jtreanor: `mContext` is now unused. We could tidy up the API by removing it from the constructor but I haven't done that here to limit scope.

daniloercoli:
I read the documentation linked above, and your approach seems OK to me. (didn't test this PR with wp-android actually).
I'd like to ping other devs to take a look at this PR, since this is a sensitive update to the App/FluxC. cc @aforcier  

If we decide to go down this route, we need to migrate the wpcom token from Preferences to this new KeyStore (in another PR off course), since it's way more important to protect the access token, than a self-signed certificate...

Instead, if we agree that the current security level - that uses Preferences to store the access token - is OK, we can skip this update, and rather use Preferences to store the run-time generated password for `MemorizingTrustManager`  and remove the hard-coded password in code.
Note: Rooted devices, or setting the app debugabble, expose the prefs file to the outer world.

I also see `clearLocalTrustStore` removing all the keys/values in it. To be prepared for future improvements, I think is better to introduce a new routine that does only remove Self-signed certificates, and use it here. (Should we prepend all the self-signed cert keys with a prefix, and remove all the keys/values that start with this prefix?)

jtreanor:
Thanks for taking the time to look @daniloercoli. I absolutely agree its good to be thorough with review given the sensitive nature.

Yeah, I'm going to look a bit closer at the token case to see if a similar approach would be necessary.  Once its a bit clearer, we can discuss and I can do that if needed.

Regarding `clearLocalTrustStore`, I made a small change this morning which makes it clear only certificates from the store. This means everything that impacts `checkServerTrusted()`. Other keys would be untouched. Do you think it is needed to only remove certs that were added by `MemorizingTrustManager`? I can see reasonable arguments for both.

daniloercoli:
> Do you think it is needed to only remove certs that were added by MemorizingTrustManager?

Oh yes I saw the new code - I mentioned it by having in mind that the wpcom token will probably be the next item to be migrated to the new secure store. 
We should just make sure we can only remove the wpcom access token from it, and not the self-signed certs, on wpcom logout (Not actually sure how it will be stored there, if a new cert is needed or we can store a plain String...).

jtreanor:
Ah yes, I see what you mean now.

My understanding of how something else like the token would be stored is that we would store a key in the KeyStore and use that to encrypt the token to put it somewhere (maybe still shared prefs). This means deleting the certs would have no impact on it and it could be deleted separately.

Because of this, `clearLocalTrustStore` clears all data that influences trust (i.e. the trusted certificates), and nothing else. This feels like the correct behaviour to me. If calling `clearLocalTrustStore()` left some trusted certs in there, I feel that would be confusing.

If we decide to store the access token, I think it would make sense to manage keys somewhere else so that `MemorizingTrustManager` is just for trust management.

However, if you still think this is unclear, I am happy to rethink it and revise some of the naming.

daniloercoli:
Also tested the integration in `wp-android`, and tested that entries in the KeyStore are not accessible to other apps on the device.

PR #929, opened 2018-10-16
Fix page parent parameter

0nko:
This PR changes the page parent parameter for the XML-RPC endpoint used by the self-hosted sites without Jetpack. The original parameter was incorrect and setting/loading of page parents didn't work.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/post/PostXMLRPCClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/post/PostXMLRPCClient.java
@@ -376,7 +376,7 @@ public class PostXMLRPCClient extends BaseXMLRPCClient {
         }
 
         if (post.isPage()) {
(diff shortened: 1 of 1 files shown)


mzorz:
hi @0nko ! 👋 Could you provide steps to repro the problem and the fix here? I'm not sure what the exact scenario needs to be to this code path to be executed. We should probably add some tests here as well, if that makes sense.

Also I'm looking into the docs and it looks like if this is actually the fix then the .org API docs are outdated? https://codex.wordpress.org/XML-RPC_WordPress_API/Pages (which could be the possibility, yes). Maybe we should possibly reach out to someone caring for that doc to be updated? wdyt?

0nko:
Hi @mzorz. You can't test this by itself, but there is an [associated PR](https://github.com/wordpress-mobile/WordPress-Android/pull/8436) in wpandroid.

I'm not sure what the `wp.getPage` endpoint is for, but we're not using it. We're using `wp.getPost`.

mzorz:
> there is an associated PR in wpandroid.
Ah! thank you 👍 didn't see that earlier, sorry :)

PR #933, opened 2018-10-17
Fixing inconsistency between equals and hashCode methods of RevisionModel

khaykov:
Thanks to @malinajirka for pointing inconsistency in `equals` and `hashCode` methods of `RevisionModel`.
This PR addresses the issue and makes sure that the `equals` and `hashCode` behave symmetrically.


malinajirka:
Thanks @khaykov! LGTM

>makes sure that the equals and hashCode behave symmetrically.

You probably know this, but I feel like it's better to make sure we are on the same page. Objects which are .equals() **must** have the same .hashCode(). But objects with the same .hashcode() **doesn't** have to be .equals().

PR #937, opened 2018-10-18
Add the missing field value to the Crashylics error exception log

AmandaRiu:
Addresses #936 

Adds the missing field in question to the Crashlytics exception log for the "IllegalStateException - Missing field from stats endpoint". Having this data will allow us to better understand what needs to be fixed to prevent the exception.


nbradbury:
Looks good! :shipit:

nbradbury:
I just updated [the related WCAndroid issue](https://github.com/woocommerce/woocommerce-android/issues/447#issuecomment-464789725) with a comment showing the missing field.

PR #943, opened 2018-10-23
Show XMLRPC error when it's broken

planarvoid:
Issue: https://github.com/wordpress-mobile/WordPress-Android/issues/8459

This PR is adding one additional case when we're returning the `MISSING_XMLRPC_METHOD`. When we were able to scrape it from the the site, we already know the the site is wporg. In that case we shouldn't show the `NO_SITE_ERROR` because it is not true.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_DiscoveryTest.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_DiscoveryTest.java
@@ -452,8 +452,6 @@ public class ReleaseStack_DiscoveryTest extends ReleaseStack_Base {
     @Test
     public void testXMLRPCMissingMethodDiscovery() throws InterruptedException {
         mUrl = BuildConfig.TEST_WPORG_URL_SH_MISSING_METHODS;
-        mUsername = BuildConfig.TEST_WPORG_USERNAME_SH_MISSING_METHODS;
-        mPassword = BuildConfig.TEST_WPORG_PASSWORD_SH_MISSING_METHODS;
 
         mNextEvent = TestEvents.MISSING_XMLRPC_METHOD;
         mCountDownLatch = new CountDownLatch(1);
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_DiscoveryTest.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_DiscoveryTest.java
@@ -435,10 +435,8 @@ public class ReleaseStack_DiscoveryTest extends ReleaseStack_Base {
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


aforcier:
Looks good @planarvoid - how about we add a new discovery test for this case though? Looks like a site with `xmlrpc.php` renamed but the rsd link untouched would do the trick - I'll send you a URL you can use.

planarvoid:
@aforcier I've added the test

aforcier:
@planarvoid you need to add a placeholder value for `TEST_WPORG_URL_DELETED_XMLRPC_PHP` to `tests.properties-example` to keep things buildable for Travis.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_DiscoveryTest.java
```diff
@@ -434,11 +434,24 @@ public void testXMLRPCForbiddenDiscovery() throws InterruptedException {
         assertTrue(mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
     }
 
+    @Test
+    public void testXMLRPCForbiddenDiscoveryOnSelfHostedSiteWithMovedPhpFile() throws InterruptedException {
+        mUrl = BuildConfig.TEST_WPORG_URL_DELETED_XMLRPC_PHP;
+        mUsername = BuildConfig.TEST_WPORG_USERNAME_SH_FORBIDDEN;
```
aforcier: We shouldn't need to define `mUsername` and `mPassword` here, they're not used (and they're also currently values for a different test site).

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_DiscoveryTest.java
```diff
@@ -434,11 +434,24 @@ public void testXMLRPCForbiddenDiscovery() throws InterruptedException {
         assertTrue(mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
     }
 
+    @Test
+    public void testXMLRPCForbiddenDiscoveryOnSelfHostedSiteWithMovedPhpFile() throws InterruptedException {
```
aforcier: I would maybe call this test `testXMLRPCDeletedXmlrpcFileDiscovery` or something like that - 'forbidden' tends to refer to the xmlrpc.php file existing, but being blocked, usually via `.htaccess` rules.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_DiscoveryTest.java
planarvoid: You are right of course, I see now that I've deleted them in the following test by mistake 😢

aforcier:
:shipit:!

PR #986, opened 2018-11-21
Add PageSize parameter to the store calls that need a page size

planarvoid:
- This PR adds a page size parameter to comments, followers and tags and categories. We're loading pageSize + 1 items from the API to find out if we should show the `View more` button in the UI. We're returning only `pageSize` items to the UI.


0nko:
:shipit:

PR #987, opened 2018-11-21
Make fields in Stats nullable to avoid API errors

planarvoid:
- the API behaves randomly and I figured it's safer to keep all the fields nullable (because they can disappear at any time). This PR is making everything nullable and either shows a default (empty) value instead or filters out items that don't have the required fields.


malinajirka:
Thanks @planarvoid ! I've left two super minor comments.

Feel free to merge it!

PR #990, opened 2018-11-22
Woo: Add action for fetching the max supported WooCommerce version

aforcier:
Adds a `WooCommerceStore` action for obtaining the maximum supported version of the WooCommerce API for any Jetpack-connected, logged in site.

This happens by doing a call to the site's root `wp-json` API through the WP.com Jetpack tunnel (the site's `wp-json` root corresponds to `/` in the Jetpack tunnel). Among other things, this returns a list of namespaces supported by the site, which looks like this:

```js
{
    ...
    "namespaces": [
      "oembed/1.0",
      "akismet/v1",
      "jetpack/v4",
      "wc/v1",
      "wc/v2",
      "wc/v3",
      "wp/v2"
    ],
    ...
}
```

The callback will contain the latest API version the site supports that FluxC knows about (currently up to V3).

Intended use is to identify sites that aren't updated to WooCommerce 3.5 (and so are lacking the V3 API endpoints), and warn those users to update to continue using the WooCommerce app.

### To test
The WooCommerce fragment in the example app has a new "Log Woo API versions" button, which will log the max supported Woo API version for every WooCommerce site attached to the logged-into account.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedNetworkAppComponent.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedNetworkAppComponent.java
@@ -27,6 +27,7 @@ public interface MockedNetworkAppComponent {
     void inject(MockedStack_SiteTest object);
     void inject(MockedStack_UploadStoreTest object);
     void inject(MockedStack_UploadTest object);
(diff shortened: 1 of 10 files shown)


nbradbury:
Looks good and tests passed :shipit:

PR #992, opened 2018-11-23
Refactor Stats classes

0nko:
This PR refactors the Stats classes in preparation for use in TimeStatsStore, REST client and others.


fluxc/src/main/java/org/wordpress/android/fluxc/store/StatsStore.kt
```diff
@@ -50,7 +68,9 @@ class StatsStore
         )
     }
 
-    enum class InsightsTypes {
+    interface StatsTypes
```
planarvoid: this shouldn't be necessary (if we don't use `InsightsTypes` directly in code but use `.ordinal` instead)

PR #993, opened 2018-11-26
Woo: Migrate to V3 API

aforcier:
Replaces all uses of the `wc/v2` API with `wc/v3`. This is a non-breaking change, though this is a good time to take advantage of some improvements the V3 API offers - will be adding those in follow-up PRs.

This is targeting a feature branch, which we should hold off merging until after the next WooCommerce app release.

### To test
Run through the example app, make sure order-related calls are working as expected. Order note creation should work, but all created notes will be flagged as `system` - I'll be addressing that in a followup PR.


nbradbury:
Nice work @aforcier. Code looks good, and I ran through the example app and everything worked just fine. :shipit:

PR #995, opened 2018-11-26
Woo: Differentiate system-generated order notes

aforcier:
Updates our use of the order notes API to take advantage of some V3 API additions:

1. Fetched order notes now have an `author` field, which will either be `system` or the name of a user. Using this field we can classify private notes as being system or non-system in the UI.
2. Uploaded order notes are marked `author = system` by default, but by passing `added_by_user = true` we can flag them as user-created (all app-created notes are now flagged this way by default)

Should be merged after #993.

### To test
Run the tests, and inspect the network requests made from the example app for note fetching and uploading. Fetched notes should have `isSystemNote` reflect the status of the note in `wp-admin`, and uploaded notes should always be marked as user-generated in `wp-admin`.


nbradbury:
Looks good! :shipit:

PR #1004, opened 2018-11-27
Feature/997 woo order search

### LINKED ISSUE #997: Woo: Add order search
We hope to implement order search in the WooCommerce app, which requires FluxC work. Note that the PR should be branched from `feature/woo-v3-api` so we can use the v3 API.

nbradbury:
Resolves #997 - adds the ability to search orders. Rather than create a ton of new code I chose to piggyback on the existing `fetchOrders()` routine, as we did when we added filtering by status.


aforcier:
@nbradbury haven't dug into this too deeply yet, but are we sure we want to wipe out the database for search results? Maybe it's better to provide those directly, skipping database storage, so we don't have to do a full re-fetch when we return to the unfiltered list. WDYT?

nbradbury:
> are we sure we want to wipe out the database for search results?

Oof, hadn't noticed that was happening :( I'll fix this and then ping you when it's ready (most likely tomorrow).

nbradbury:
@aforcier I'm closing this and will open a new PR for order search shortly.

PR #1005, opened 2018-11-28
Add missing migration of refreshed stats

planarvoid:
- the migration step was missing after adding new fields to the Stats cache table
- As part of this PR I've also fixed the incorrect Enum mapping


malinajirka:
LGTM!

Feel free to merge it as soon as you fix the checkstyle issue.

PR #1006, opened 2018-11-28
Feature/997 woo order search take2

### LINKED ISSUE #997: Woo: Add order search

nbradbury:
Resolves #997 - adds WooCommerce order search. Note that for now paging isn't supported, but this will come later. I want to tackle some of the related UI work before addressing paging.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
@@ -18,6 +18,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComGson
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
 import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequest
 import org.wordpress.android.fluxc.store.WCOrderStore
+import org.wordpress.android.fluxc.store.WCOrderStore.Companion.DEFAULT_ORDER_STATUS
 import org.wordpress.android.fluxc.store.WCOrderStore.FetchHasOrdersResponsePayload
 import org.wordpress.android.fluxc.store.WCOrderStore.FetchOrderNotesResponsePayload
 import org.wordpress.android.fluxc.store.WCOrderStore.FetchOrdersCountResponsePayload
@@ -26,6 +27,7 @@ import org.wordpress.android.fluxc.store.WCOrderStore.OrderError
 import org.wordpress.android.fluxc.store.WCOrderStore.OrderErrorType
 import org.wordpress.android.fluxc.store.WCOrderStore.RemoteOrderNotePayload
 import org.wordpress.android.fluxc.store.WCOrderStore.RemoteOrderPayload
+import org.wordpress.android.fluxc.store.WCOrderStore.SearchOrdersResponsePayload
 import javax.inject.Singleton
 
 @Singleton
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -142,7 +157,13 @@ class WCOrderStore @Inject constructor(dispatcher: Dispatcher, private val wcOrd
     }
 
     // OnChanged events
-    class OnOrderChanged(var rowsAffected: Int, var statusFilter: String? = null, var canLoadMore: Boolean = false)
+    class OnOrderChanged(
```
aforcier: This might be overcomplicating `OnOrderChanged` a bit - `rowsAffected` implies a database change which isn't the case for search, and the rest of the pre-existing properties aren't used either.

What do you think about adding a new `OnChanged` instead, called something like `OnOrdersSearched` (we have a somewhat similarly named action in [OnReaderSitesSearched](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/08db5f52ca2172e4bc4d9f26064826704cd6a7b9/fluxc/src/main/java/org/wordpress/android/fluxc/store/ReaderStore.java#L112))?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -90,6 +92,41 @@ class OrderRestClient(
...
+     * Dispatches a [WCOrderAction.SEARCHED_ORDERS] action with the resulting list of orders.
+     *
+     * @param [searchQuery] the keyword or phrase to match orders with
+     */
+    fun searchOrders(site: SiteModel, searchQuery: String) {
+        val url = WOOCOMMERCE.orders.pathV3
+        val responseType = object : TypeToken<List<OrderApiResponse>>() {}.type
+        val params = mapOf(
+                "per_page" to WCOrderStore.NUM_ORDERS_PER_SEARCH.toString(),
+                "status" to DEFAULT_ORDER_STATUS,
```
aforcier: Total nitpick but - why not also use the more explicit `WCOrderStore.DEFAULT_ORDER_STATUS` like for the `per_page` field right above?

aforcier:
Looks great @nbradbury! Just a couple small comments.

nbradbury:
> this might be overcomplicating OnOrderChanged a bit...What do you think about adding a new OnChanged instead

I had been leaning that way but decided to extend `OnOrderChanged` instead, but I think you're right. It would be better to add a new event specific to search. I'll implement that and ping you when it's done.

nbradbury:
@aforcier I updated this PR to use a different event for search, as you suggested. So, ready for another round!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -246,14 +275,23 @@ class WCOrderStore @Inject constructor(dispatcher: Dispatcher, private val wcOrd
...
         }
 
         onOrderChanged.causeOfChange = WCOrderAction.FETCH_ORDERS
 
         emitChange(onOrderChanged)
     }
 
+    private fun handleSearchOrdersCompleted(payload: SearchOrdersResponsePayload) {
+        val onOrdersSearched = if (payload.isError) {
+            OnOrdersSearched("", emptyList()).also { it.error = payload.error }
```
aforcier: We could modify the error constructor for `SearchOrdersResponsePayload` to also accept a `searchQuery` argument. Since we know what the original query string was in the error case inside the `OrderRestClient`, we can add it to the payload there so it floats up here, and pass it to `OnOrdersSearched` instead of `""`. That way we'd be able to record what the failed search query was in the client, in case of multiple parallel requests.

aforcier:
Looks good @nbradbury! One last comment and it's good to go - sorry for not thinking of it earlier.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
nbradbury: > We could modify the error constructor for `SearchOrdersResponsePayload` to also accept a `searchQuery` argument.

Good call - added in 6e66e9c.

aforcier: Don't forget to also give `payload.searchQuery` to `OnOrdersSearched()` here 😀

nbradbury: Oops :) Added in caebf23.

aforcier:
:shipit:

PR #1010, opened 2018-11-29
Fix Posts and Pages endpoint for stats

planarvoid:
- fixes mapping of ViewsType
- disables caching (because caching ignores REST parameters)


aforcier:
> disables caching (because caching ignores REST parameters)

Hmm that's odd, it's not supposed to - there's even a [test checking this](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_CacheTest.kt#L127-L160). Can you share a bit more detail about what's happening?

planarvoid:
@aforcier yep, whenever I enable the cache, the same call with different query parameters returns the same result. It was very hard to find. I checked the actual calls in Stetho and there is really only one call. In this case loading data for different time periods (day/week/month/year) all returned data for a Day. I haven't investigated this issue deeper but changing the `enableCaching` to `false` fixes the issue.

planarvoid:
@aforcier I'll check what you posted. I was facing this issue in one other use case in stats (when I was loading email and wpcom followers) and it also didn't work. The thing is that enabling/disabling the caching makes it call the API once/multiple times - and this was happening in the real app, not in tests. There is not yet a PR that calls this endpoint (my wip is `feature/init_time_stats` in WPAndroid).

planarvoid:
@aforcier hey, I tried to reproduce this in the test and it never happened. Just as a sanity check I added the crash into the rest client that should never happen if the cache worked correctly. When I run the Android app, it crashes because the request and response period don't match. I assume that the test behaviour is not the same as the real app.

aforcier:
That's pretty odd @planarvoid - we're also using WP.com stats endpoints for the Woo app, and those have been caching correctly in production 🤔.

You could try calling those endpoints with your setup (you can see what we're using [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt)) - but I'm not sure how the endpoints being different could have an effect.

malinajirka:
Thanks @planarvoid ! The code LGTM. However, I'd suggest creating an issue so we don't forget to turn the cache on when we fix it or find out why it's not working.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/PostAndPageViewsRestClient.kt
```diff
@@ -67,9 +67,10 @@ constructor(
             @SerializedName("total_views") val totalViews: Int
         ) {
             data class PostViewsResponse(
-                @SerializedName("id") val id: Int,
+                @SerializedName("id") val id: Long,
```
malinajirka: Shouldn't all the fields in `PostAndPageViewsResponse` be nullable?

PR #1017, opened 2018-12-04
Properly translate remoteSiteId to localSiteId before storing notific…

### LINKED ISSUE #1016: Woo Notifications Fix: Use proper local site id
The current notifications code is using the `remote_site_id` as the `local_site_id`. Figure out the best way of mapping the `local_site_id` from `SiteModelTable` to the `remote_site_id` from the notifications api response. Store this value in the database.

AmandaRiu:
Fixes #1016 

Uses the `SiteStore` to query for the matching `local_site_id` using the remote site id. This value is then stored in the NotificationModelTable and allows for querying the database for notifications matching the current site on Woo.


nbradbury:
Looks good! :shipit:

PR #1018, opened 2018-12-04
Create referrers endpoint for Stats

### LINKED ISSUE #998: Create Referrers endpoint for Stats
- create an endpoint that returns a list of referrers
- the referrers need to be cached
- make the endpoint resistant to missing fields
- the endpoint has to return values by days, weeks, months and years

planarvoid:
Fixes #998 
This PR is adding a Referrers rest client, sql utils and Store. Most of the code are extensive tests - the logic is mostly copy&paste of the `PostAndPageViews` endpoint. I didn't expect the tests to be this long and the PR to grow this much


malinajirka:
I'll review this when we have a PR with an UI in WPAndroid.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/ReferrersRestClient.kt
```diff
@@ -0,0 +1,117 @@
...
+                FetchStatsPayload(response.data)
+            }
+            is Error -> {
+                FetchStatsPayload(response.error.toStatsError())
+            }
+        }
+    }
+
+    data class ReferrersResponse(
+        @SerializedName("period") val period: String?,
```
malinajirka: I'd suggest renaming this field to 'statsGranurality/granurality' so we don't use several terms for a single attribute as it can be confusing. If you decide to make this change, we should probably rename it everywhere so `@SerializedName("period")` is the only time it's called "period".

planarvoid: good point

malinajirka: Thanks for the change. However,  I think we should rename it everywhere - even in tests, rest client etc.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/StatsUtils.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.wordpress.android.fluxc.model.SiteModel
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+import java.util.TimeZone
+import javax.inject.Inject
+
+class StatsUtils
+@Inject constructor() {
+    fun getCurrentDateTZ(site: SiteModel): String {
```
malinajirka: We may want to write a test for this class, wdyt?

malinajirka:
I've left last two comments, but LGTM. Thanks!

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/StatsUtils.kt
planarvoid: you're right. I'm just not really sure what it does 🙈

malinajirka: :D I didn't know it was copy-pasted from the WPAndroid repo. In that case we can leave it without tests, I guess.

malinajirka:
LGTM, thanks!

PR #1023, opened 2018-12-10
Add date to stats endpoints

planarvoid:
- we need to send a date as part of requests for Granular stats endpoints in order to show different data when the user selects different date
- this PR is adding date to `PostsAndPages` endpoint and to `Referrers` endpoint. The Date is now cached in DB (so we can cache multiple days per Tab)


malinajirka:
Thanks @planarvoid ! I've reviewed the code and it looks good! I've left one minor comment. Let me know when you fix the conflict.;)

planarvoid:
@malinajirka all should be fixed now, thanks 👍

PR #1031, opened 2018-12-13
Woo: create OrderIdentifier from remoteOrderId and localSiteId

### LINKED ISSUE #1030: WooCommerce: Get order by remoteOrderId and localSiteId only
We require a way to fetch an order from the database with only the `remoteOrderId` and `localSiteId`. This is already built into the functionality of `OrderIdentifier` but currently the only way to get an instance of this class is to first have a `WCOrderModel`.

AmandaRiu:
Fixes #1030 

There are two ways to fetch an order from the database: using the local order `id`, or using the `localSiteId` with the `remoteOrderId`. Before this PR you could only get an instance of `OrderIdentifier` if you had a `WCOrderModel` fetched from FluxC first. If you only knew the `remoteNoteId` and `localSiteId` (like when dealing with Notifications), then a hack of creating a temp `WCOrderModel` was the only way to get that `OrderIdentifier` object, even though the local order `id` is not required.

This PR adds a way method for creating an `OrderIdentifier` object using only the `localSiteId` and `remoteNoteId`.


nbradbury:
:shipit:

PR #1033, opened 2018-12-13
Create authors endpoint for Stats

### LINKED ISSUE #1000: Create an Authors endpoint for Stats
- create an endpoint that returns a list of authors
- the authors need to be cached
- make the endpoint resistant to missing fields
- the endpoint has to return values by days, weeks, months and years

planarvoid:
Fixes #1000 
In this PR I'm adding an endpoint that returns a list of Authors with their top posts for Stats.
This should be tested with the WPAndroid PR

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/stats/time/AuthorsModel.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/stats/time/AuthorsModel.kt
@@ -1,0 +1,11 @@
+package org.wordpress.android.fluxc.model.stats.time
+
+data class AuthorsModel(val otherViews: Int, val authors: List<Author>, val hasMore: Boolean) {
+    data class Author(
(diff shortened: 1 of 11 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/stats/time/AuthorsModel.kt
```diff
@@ -0,0 +1,11 @@
+package org.wordpress.android.fluxc.model.stats.time
+
+data class AuthorsModel(val otherViews: Int, val authors: List<Author>, val hasMore: Boolean) {
+    data class Author(
+        val name: String,
+        val views: Int,
+        val avatar: String?,
```
0nko: Nitpick: What do you think about naming this `avatarUrl`? It's not immediately clear if it's a URL or an emoji string or something else.

PR #1038, opened 2018-12-14
Add support for marking a notification as read

### LINKED ISSUE #1037: Woo Notifications: Mark as read
Add the ability to mark a notification as read.

AmandaRiu:
This PR fixes #1037 and adds support for the following:

- The `/notifications/read/` rest api endpoint
- Notification actions for marking a notification as read
- Logic for posting the request to the api
- Saving the notification to the database on success
- Success and error response handling
- Mocked automated tests
- **New notifications test page in the example app**

### Comments
Turns out the endpoint for marking a notification as read is different from all the other endpoints in that it requires the parameters to be included in the body not as json but form-encoded. The previous implementation of `GsonRequest` specifically checked if the type of request was `POST` and if the `body` was null it would create an empty object for the body. This overrode the logic used in WPAndroid where the arguments for this endpoint are sent over as parameters in a post request. After some digging I found that in the case of a POST request, Volley will use the `params` property to populate the body if the `body` property is null. Adding a little check to make sure the `params` argument was not null or empty before creating that empty body object allowed for this built in functionality to work.

## Notifications Test fragment
I added a new notifications fragment to test all the functionality present in notifications so far as well. This became necessary since notifications are really difficult to test for release automated testing. This exercise uncovered an issue with the recently added `FETCH_NOTIFICATION` action (single notification) that has been fixed in #1046.  

### Main Fragment notifications button and Test page
screenshot-1544658220736

- **Fetch all notifications from API**: Fetches all available notifications from the api for the current user. These are populated in the database and seed the tests marked with `DB`
- **Get notifications for first site (DB)**: Pulls a list of notifications for the first site returned by the `SiteStore` for the current user.
- **Get notifications by type or subtype (DB)**: Opens a custom dialog for selecting a single type and subtype for pulling a filtered list from the database. 

- **Mark notifications seen**: Tells the api all notifications before this moment have been seen by the user.
- **Fetch notification from the API**: Grabs the `remoteNoteId` first notification in the database and fires off a request to fetch it from the api. ~~This will not completely work until #1046 is merged.~~ The request is still fired and response appropriately processed, but the missing `causeOfChange` makes it impossible for the example app to print the message that it was successful.
- **Mark read**: Grabs the first notification in the database and fires off a request to mark that notification as read.
- **Update Notification**: Grabs the first notification from the database, changes the `read` property to `false` and fires the request to update the notification.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/NotificationsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/NotificationsFragment.kt
@@ -1,0 +1,167 @@
+package org.wordpress.android.fluxc.example
+
+import android.content.Context
+import android.os.Bundle
+import android.support.v4.app.Fragment
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import dagger.android.support.AndroidSupportInjection
+import kotlinx.android.synthetic.main.fragment_notifications.*
+import org.greenrobot.eventbus.Subscribe
+import org.greenrobot.eventbus.ThreadMode.MAIN
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.action.NotificationAction.FETCH_NOTIFICATION
+import org.wordpress.android.fluxc.action.NotificationAction.FETCH_NOTIFICATIONS
+import org.wordpress.android.fluxc.action.NotificationAction.MARK_NOTIFICATIONS_SEEN
+import org.wordpress.android.fluxc.action.NotificationAction.MARK_NOTIFICATION_READ
+import org.wordpress.android.fluxc.action.NotificationAction.UPDATE_NOTIFICATION
+import org.wordpress.android.fluxc.example.NotificationTypeSubtypeDialog.Listener
+import org.wordpress.android.fluxc.generated.NotificationActionBuilder
+import org.wordpress.android.fluxc.model.notification.NotificationModel.Subkind
+import org.wordpress.android.fluxc.store.NotificationStore
+import org.wordpress.android.fluxc.store.NotificationStore.FetchNotificationPayload
+import org.wordpress.android.fluxc.store.NotificationStore.FetchNotificationsPayload
+import org.wordpress.android.fluxc.store.NotificationStore.MarkNotificationReadPayload
+import org.wordpress.android.fluxc.store.NotificationStore.MarkNotificationsSeenPayload
+import org.wordpress.android.fluxc.store.NotificationStore.OnNotificationChanged
+import org.wordpress.android.fluxc.store.SiteStore
+import org.wordpress.android.util.DateTimeUtils
+import java.util.Date
+import javax.inject.Inject
+
+class NotificationsFragment : Fragment() {
+    @Inject internal lateinit var dispatcher: Dispatcher
+    @Inject internal lateinit var notificationStore: NotificationStore
+    @Inject internal lateinit var siteStore: SiteStore
+
+    private var typeSelectionDialog: NotificationTypeSubtypeDialog? = null
+
+    override fun onAttach(context: Context?) {
+        AndroidSupportInjection.inject(this)
+        super.onAttach(context)
+    }
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? =
+            inflater.inflate(R.layout.fragment_notifications, container, false)
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        notifs_fetch_all.setOnClickListener {
+            prependToLog("Fetching all notifications from the api...")
+            dispatcher.dispatch(NotificationActionBuilder.newFetchNotificationsAction(FetchNotificationsPayload()))
+        }
+
+        notifs_fetch_for_site.setOnClickListener {
+            prependToLog("Getting all notifications for the first site...")
+            val site = siteStore.sites.first()
+            val notifs = notificationStore.getNotificationsForSite(site)
+
+            // todo amanda - fix display name
+            prependToLog("SUCCESS! ${notifs.size} pulled from the database for ${site.name}")
+        }
+
+        notifs_by_type_subtype.setOnClickListener {
+            showNotificationTypeSubtypeDialog(object : Listener {
+                override fun onSubmitted(type: String, subtype: String) {
+                    prependToLog("Fetching notifications matching $type or $subtype...")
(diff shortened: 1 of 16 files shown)


nbradbury:
@AmandaRiu I'm done with my review, just one minor thing and we can merge.

fluxc/src/main/java/org/wordpress/android/fluxc/store/NotificationStore.kt
```diff
@@ -302,7 +318,33 @@ constructor(
...
+        emitChange(onNotificationChanged)
+    }
 
+    private fun markNotificationRead(payload: MarkNotificationReadPayload) {
+        notificationRestClient.markNotificationRead(payload.notification)
+    }
+
+    private fun handleMarkedNotificationRead(payload: MarkNotificationReadResponsePayload) {
+        // Update the notification in the database
+        payload.notification?.let {
```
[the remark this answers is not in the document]
AmandaRiu: Updated in 3543cefb06562659e767327ddfaf3f0d7a12532d

AmandaRiu:
@nbradbury addressed the comment and ready for another round

example/src/main/java/org/wordpress/android/fluxc/example/NotificationsFragment.kt
```diff
@@ -0,0 +1,168 @@
...
+            // todo amanda - fix display name
+            prependToLog("SUCCESS! ${notifs.size} pulled from the database for ${site.name}")
+        }
+
+        notifs_by_type_subtype.setOnClickListener {
+            showNotificationTypeSubtypeDialog(object : Listener {
+                override fun onSubmitted(type: String, subtype: String) {
+                    prependToLog("Fetching notifications matching $type or $subtype...")
+                    val notifs = notificationStore.getNotifications(listOf(type), listOf(subtype))
+                    val groups = notifs.groupingBy {
```
nbradbury: I'm wondering if we need to replace `Subkind.UNKNOWN` with `Subkind.NONE` then make `Subkind.NONE` the default subtype in the notif model.

example/src/main/java/org/wordpress/android/fluxc/example/NotificationsFragment.kt
```diff
@@ -0,0 +1,168 @@
...
+                        it.subtype?.name?.takeIf { it != Subkind.UNKNOWN.name } ?: it.type.name
+                    }.fold(0) { acc, _ -> acc + 1 }
```
nbradbury: TIL: [fold](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/fold.html)

fluxc/src/main/java/org/wordpress/android/fluxc/action/NotificationAction.java
```diff
@@ -39,6 +43,8 @@
     FETCHED_NOTIFICATION, // Response to fetching a single notification
     @Action(payloadType = MarkNotificationSeenResponsePayload.class)
     MARKED_NOTIFICATIONS_SEEN, // Response to marking a notification as seen
```
nbradbury: I should've caught this earlier, but the comment here is misleading in that it refers to marking a single notification as seen.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
```diff
@@ -52,7 +54,11 @@ protected void deliverResponse(T response) {
 
     @Override
     public String getBodyContentType() {
-        return PROTOCOL_CONTENT_TYPE;
+        if (mBody == null) {
```
nbradbury: Just out of curiosity, I'm wondering what you encountered that led to this change?

fluxc/src/main/java/org/wordpress/android/fluxc/store/NotificationStore.kt
```diff
@@ -345,7 +367,38 @@ constructor(
...
+        }
+
+        // Create an dispatch result
+        val onNotificationChanged = if (payload.isError) {
+            OnNotificationChanged(0).apply {
+                error = payload.error
+                success = false
+            }
+        } else {
+            OnNotificationChanged(1).apply {
```
nbradbury: I'm not sure if this matters, but in the `apply` block below we're only adding the id if the notification isn't null, but here were setting `rowsAffected` to `1` regardless of whether the notification is null. This raises the possibility that `rowsAffected` will be `1` but the list of affected IDs will be empty.

nbradbury:
@AmandaRiu Done with my first round! As usual my comments are mostly nitpicks :)

fluxc/src/main/java/org/wordpress/android/fluxc/action/NotificationAction.java
AmandaRiu: Nice catch. I actually thought that's what was happening when I originally wrote the comment, but then after studying the API docs closer I corrected that assumption. Fixed in b4c84a703baec41497283512402ea6b99204bc83

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
AmandaRiu: Marking a notification as read was failing with an invalid request response from the server. I figured out that it was due to sending the params as the body, so I changed that, and then after a LOT of troubleshooting  I found it was still failing due to the `body` of the request being artificially created so the params were ignored.

fluxc/src/main/java/org/wordpress/android/fluxc/store/NotificationStore.kt
```diff
@@ -345,7 +367,38 @@ constructor(
...
+    private fun handleMarkedNotificationRead(payload: MarkNotificationReadResponsePayload) {
+        // Update the notification in the database
+        if (payload.success) {
+            payload.notification?.let {
+                it.read = true // Just in case it wasn't set by the calling client
+                notificationSqlUtils.insertOrUpdateNotification(it)
+            }
+        }
+
+        // Create an dispatch result
```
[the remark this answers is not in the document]
AmandaRiu: Fixed in 716f181ccdb49cb2afcd276e5dc112d8b72cf063

fluxc/src/main/java/org/wordpress/android/fluxc/store/NotificationStore.kt
AmandaRiu: Good catch! Fixed in 79a0900debda21a43909ed5158f8ccba0600c0bf

AmandaRiu:
@nbradbury Thank you for the very thorough review 👍 All comments addressed and ready for the next round!

nbradbury:
Looks good, let's :shipit: !

PR #1041, opened 2018-12-17
Woo Notification: Fetch and save a single notification

### LINKED ISSUE #1039: Woo Notifications: Fetch and save single notification
Add the ability to fetch a single notification from the rest api by it's `remote_note_id`, as well as the ability to get and save a single notification directly to the database.

AmandaRiu:
Fixes #1039

Covers the following:

- Adds the ability to fetch a single notification by the remote_note_id. A successful response will be parsed into a `NotificationModel` and saved to the database.
- Adds a local `NotificationAction.UPDATE_NOTIFICATION` for saving a single notification directly to the database. Will send an `OnNotificationChanged` event when complete.
- Also added a new test to `MockedStack_NotificationTest`

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_NotificationTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_NotificationTest.kt
@@ -18,7 +18,6 @@ import org.wordpress.android.fluxc.module.ResponseMockingInterceptor
 import org.wordpress.android.fluxc.network.rest.wpcom.notifications.NotificationRestClient
 import org.wordpress.android.fluxc.store.NotificationStore.FetchNotificationResponsePayload
 import org.wordpress.android.fluxc.store.NotificationStore.FetchNotificationsResponsePayload
-import org.wordpress.android.fluxc.store.NotificationStore.MarkNotificationReadResponsePayload
 import org.wordpress.android.fluxc.store.NotificationStore.MarkNotificationSeenResponsePayload
 import org.wordpress.android.fluxc.store.NotificationStore.NotificationAppKey
 import org.wordpress.android.fluxc.store.NotificationStore.RegisterDeviceResponsePayload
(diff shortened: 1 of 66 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/notification/NotificationModel.kt
```diff
@@ -5,21 +5,21 @@ import org.wordpress.android.fluxc.tools.FormattableMeta
...
-    val url: String?,
-    val title: String?,
-    val body: List<FormattableContent>?,
-    val subject: List<FormattableContent>?,
-    val meta: FormattableMeta?
+    val noteId: Int = 0,
+    val remoteNoteId: Long = 0L,
+    var localSiteId: Int = 0,
+    val noteHash: Long = 0L,
+    val type: Kind = NotificationModel.Kind.STORE_ORDER,
```
nbradbury: I'm wondering if this should default to `UNKNOWN`?

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_NotificationTest.kt
```diff
@@ -135,6 +139,28 @@ class MockedStack_NotificationTest : MockedStack_Base() {
         }
     }
 
+    @Test
+    fun testFetchNotificationSuccess() {
```
nbradbury: Should we also have a release stack test for getting a single notif?

AmandaRiu: @nbradbury I thought about that, but with notifications they don't stick around long enough to rely on them, and I'm not sure how quite to force one to come through...thoughts?

fluxc/src/main/java/org/wordpress/android/fluxc/model/notification/NotificationModel.kt
AmandaRiu: Good call! fixed in c08406994da86a1c070313aa30cae07d2ba4e118

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_NotificationTest.kt
nbradbury: Ah, good point - we can skip it.

nbradbury:
Looks good! :shipit:

PR #1042, opened 2018-12-18
Add support for vendor parameter

kwonye:
This opens up the .blog domains suggestions

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestWPCom.java
@@ -474,12 +474,12 @@ public class ReleaseStack_SiteTestWPCom extends ReleaseStack_Base {
         }
         assertEquals(TestEvents.FETCHED_WPCOM_SUBDOMAIN_SUGGESTIONS, mNextEvent);
 
-        final String dotComSuffix = ".wordpress.com";
+        final String wpcomSuffix = ".wordpress.com";
         final String dotBlogSuffix = ".blog";
         for (DomainSuggestionResponse suggestionResponse : event.suggestions) {
             String domain = suggestionResponse.domain_name;
-            assertTrue("Was expecting the domain to end in " + dotComSuffix + " or " + dotBlogSuffix,
-                    domain.endsWith(dotComSuffix) || domain.endsWith(dotBlogSuffix));
+            assertTrue("Was expecting the domain to end in " + wpcomSuffix + " or " + dotBlogSuffix,
+                    domain.endsWith(wpcomSuffix) || domain.endsWith(dotBlogSuffix));
         }
 
         mCountDownLatch.countDown();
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -373,14 +373,17 @@ public void onErrorResponse(@NonNull WPComGsonNetworkError error) {
...
-                               final int quantity) {
+                               final int quantity, final boolean includeVendorDot) {
         String url = WPCOMREST.domains.suggestions.getUrlV1_1();
         Map<String, String> params = new HashMap<>(4);
         params.put("query", query);
         params.put("only_wordpressdotcom", String.valueOf(onlyWordpressCom)); // CHECKSTYLE IGNORE
         params.put("include_wordpressdotcom", String.valueOf(includeWordpressCom)); // CHECKSTYLE IGNORE
         params.put("include_dotblogsubdomain", String.valueOf(includeDotBlogSubdomain));
         params.put("quantity", String.valueOf(quantity));
+        if (includeVendorDot) {
```
hypest: 👋 @kwonye , I can't say I have insight on what the Vendor Dot param is about so, is there anything you can share about it? Ideally, perhaps add a test for it if it can help show its use. Thanks!

kwonye: Good point! I added a test to handle the "vendor" parameter. It's not a very clear parameter in general so I don't want to add a comment to it because what it does will most likely change in the backend. The test should catch any changes however.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestWPCom.java
```diff
@@ -474,9 +474,12 @@ public void onSuggestedDomains(OnSuggestedDomains event) {
...
         assertEquals(TestEvents.FETCHED_WPCOM_SUBDOMAIN_SUGGESTIONS, mNextEvent);
 
-        final String suffix = ".wordpress.com";
+        final String dotComSuffix = ".wordpress.com";
+        final String dotBlogSuffix = ".blog";
         for (DomainSuggestionResponse suggestionResponse : event.suggestions) {
-            assertTrue("Was expecting the domain to end in " + suffix, suggestionResponse.domain_name.endsWith(suffix));
+            String domain = suggestionResponse.domain_name;
+            assertTrue("Was expecting the domain to end in " + dotComSuffix + " or " + dotBlogSuffix,
+                    domain.endsWith(dotComSuffix) || domain.endsWith(dotBlogSuffix));
```
hypest: Adding a comment now and we might want to fix it in a new PR:
The conditional seems to be able to succeed even if all results are `.com` and no `.blog` at all. Perhaps we'd like to refactor it to count for `.com` and `.blog` and assert that both are non zero or something? WDYT @kwonye ?

kwonye: Good point, @hypest! I'll get started on modifying this tomorrow!

PR #1055, opened 2018-12-29
Amanda/wc notifs sync update

### LINKED ISSUE #1054: WC Notifications: Optimize syncing logic
Notification syncing needs to be updated so it can be smarter about syncing. Currently we just pull down the list and replace all records in the db.

Endpoint: http://public-api.wordpress.com/rest/v1.1/notifications

Single atomic operation when syncing notes locally:

1. Call the remote notifications endpoint requesting only note_hash and note_id for response fields
2. Delete local notes note NOT contained in the hashes array from (1)
3. Figure out what note ids need to be updated/added based on the hashes from (1)
4. Make another call to the notifications endpoint requesting all normal fields and using the note ids from (3)

AmandaRiu:
Fixes #1054 

The `FETCH_NOTIFICATIONS` action handling has been modified for optimized performance. Instead of doing this every time notifications are synced:
1. Fetch all notifications
2. Delete notifications in db
3. Insert fetched notificatons

We now do the following: 
fluxc notifications sync logic

I did some manual benchmarking and my results show that with an average of 20 requests to sync notifications, this new logic saves:
- 4.2 seconds of network time
- 2.54 MB of network data
** This is based off an average bundle of 70 notifications and assumes half of those notifications are being dropped and pulled every time we sync. If nothing changes, those numbers are significantly better.

### Known Issues
I noticed some really odd behavior from the API during my testing:
- Call to fetch with `fields="id,note_hash,type,read,body,subject,timestamp,meta"` returns **74** results. There are more than just the specified fields included.
- Call to fetch with `fields` not specified returns exactly the same number of results, and the same fields.
- Call to fetch with `fields="id,note_hash"` returns **95** results
- Sending another call to fetch **the missing 21 notifications** returns nothing
This means that with my login there will always be an attempt to fetch more records. I'll open a separate ticket for this once I gather all the info next week.

### Tests
- Added more Mocked, Release instrumentation tests, as well as additional unit tests.
- The Notifcation test page of the Example app features two new buttons:
  - DELETE ALL NOTIFICATIONS (DB)
  - DELETE HALF OF NOTIFICATIONS (DB)
  - These are used for testing the sync functionality. If there are no notifications in the db the hashes sync is skipped and a full sync is performed immediately. Once there is data in the database, the sync hashes logic is employed. These two buttons allow for testing those two scenarios.

**Note** these changes will not break the front end, but WCAndroid will not be able to take advantage of these optimizations until it's FluxC commit hash has been updated. I'll submit a PR for that next week as well since it's so late :)


nbradbury:
@AmandaRiu Love the flowchart! What software did you use to create that?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
```diff
@@ -101,18 +105,57 @@ class NotificationRestClient constructor(
...
+     * @param remoteNoteIds Optional. A list of remote notification ids to be fetched from the remote api
+     */
+    fun fetchNotifications(siteStore: SiteStore, remoteNoteIds: List<Long>? = null) {
+        val url = WPCOMREST.notifications.urlV1_1
+        val params = mutableMapOf(
                 "number" to NOTIFICATION_DEFAULT_NUMBER.toString(),
                 "num_note_items" to NOTIFICATION_DEFAULT_NUM_NOTE_ITEMS.toString(),
                 "fields" to NOTIFICATION_DEFAULT_FIELDS)
+
+        remoteNoteIds?.let { if (remoteNoteIds.isNotEmpty()) params["ids"] = remoteNoteIds.joinToString() }
```
nbradbury: Minor nit: `remoteNoteIds` can be replaced with `it` inside the `let` clause.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
```diff
@@ -101,18 +105,57 @@ class NotificationRestClient constructor(
...
+     * API endpoint:
      * https://developer.wordpress.com/docs/api/1/get/notifications/
      */
-    fun fetchNotifications(siteStore: SiteStore) {
+    fun fetchNotificationHashes() {
         val url = WPCOMREST.notifications.urlV1_1
         val params = mapOf(
+                "number" to NOTIFICATION_DEFAULT_NUMBER.toString(),
+                "num_note_items" to NOTIFICATION_DEFAULT_NUM_NOTE_ITEMS.toString(),
+                "fields" to NOTIFICATION_SYNC_FIELDS)
```
nbradbury: Is it possible to use the `type` parameter to limit the response to include just orders or reviews? https://developer.wordpress.com/docs/api/1/get/notifications/

nbradbury:
@AmandaRiu Since I'll be gone for a while, I figured I should go ahead and merge this. It all looks good :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
AmandaRiu: I don't believe there is a way to filter at the API level, plus this is shared code. Eventually it will be used by WPAndroid as well.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
AmandaRiu: Oh good catch. I'll address it in a smaller PR.

PR #1059, opened 2019-01-08
Strip location from images on upload

planarvoid:
Add parameter to Upload Payload to strip the geolocation parameters from an uploaded image when necessary. I'm using the support library Exif interface to replace the geo location values with zeros. Setting the attributes to null didn't work. 

The related issue WPAndroid: https://github.com/wordpress-mobile/WordPress-Android/issues/8557


fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -48,7 +48,7 @@
     //
 
     /**
-     * Actions: FETCH(ED)_MEDIA, PUSH(ED)_MEDIA, UPLOAD(ED)_MEDIA, DELETE(D)_MEDIA, UPDATE_MEDIA, and REMOVE_MEDIA
+     * Actions: FETCH(ED)_MEDIA, PUSH(ED)_MEDIA, UPLOADED_MEDIA, DELETE(D)_MEDIA, UPDATE_MEDIA, and REMOVE_MEDIA
```
mzorz: this one here was probably modified involuntarily - can we get it back to `UPLOAD(ED)_MEDIA`?

fluxc/src/main/java/org/wordpress/android/fluxc/utils/MediaUtils.java
```diff
@@ -144,4 +148,30 @@ public static String getFileName(String filePath) {
...
+
+    /**
+     * Removes location from the Exif information from an image
+     *
+     * @param imagePath image file path
+     * @return success
+     */
+    public static boolean stripLocation(String imagePath) {
+        try {
+            ExifInterface exifInterface = new ExifInterface(imagePath);
```
mzorz: I can see what we're doing here is setting the information to all zeroes. I was going to suggest to _remove_ the information, but looked into the ExifInterface library source code and saw that `removeAttribute`  is private there (https://developer.android.com/reference/android/media/ExifInterface.html#public-methods).

planarvoid: I tried setting it to `null` and it didn't work so this seemed like a good enough solution.

mzorz: Yeah, setting to `null` actually just bails the method 🤷‍♂️ 
screen shot 2019-01-11 at 11 39 03

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
mzorz: ah that's right - thank you! 👍

PR #1060, opened 2019-01-08
Update Kotlin version and use Stable coroutines

planarvoid:
- Update Kotlin version to 1.3.0
- Update Coroutines to use stable version 1.1.0
- update mockito and mockito-kotlin to fix tests with the updated kotlin/coroutines


0nko:
:shipit:

PR #1062, opened 2019-01-08
Issue/1061 missing data error

### LINKED ISSUE #1061: Treat missing Overview data as an error
Treat the missing `period` and `dates` parameters as an error. The timed stats without these parameters don't make sense and passing empty data can cause unexpected errors in the client.

0nko:
Fixes #1061.


fluxc/src/main/java/org/wordpress/android/fluxc/model/stats/time/TimeStatsMapper.kt
```diff
@@ -108,7 +108,7 @@ class TimeStatsMapper
             }
         }
         if (response.data == null) {
```
planarvoid: do you think it makes sense to have here `if (response.data == null || dataPerPeriod == null)` in order to log the other case when we have only invalid in the response?

PR #1068, opened 2019-01-14
Woo: Retrieve exact order count by status

aforcier:
Updates `FETCH_ORDERS_COUNT` to use the Woo V3 API's `/reports/orders/totals/` endpoint. This one returns an exact count of orders broken down by status, which is both faster and more precise than our previous way of fetching an actual order list.

_This is a breaking change for the Woo app - a counterpart PR will be opened to update the Woo app._

### To test

The existing tests have been updated, and the count by site can be verified through the example app.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -47,15 +47,10 @@ class OrderRestClient(
...
      */
-    fun fetchOrders(site: SiteModel, offset: Int, filterByStatus: String? = null, countOnly: Boolean = false) {
+    fun fetchOrders(site: SiteModel, offset: Int, filterByStatus: String? = null) {
         // If null, set the filter to the api default value of "any", which will not apply any order status filters.
-        val statusFilter = if (filterByStatus.isNullOrBlank()) {
-            WCOrderStore.DEFAULT_ORDER_STATUS
-        } else {
-            filterByStatus!!
-        }
+        val statusFilter = filterByStatus.takeUnless { it.isNullOrBlank() } ?: WCOrderStore.DEFAULT_ORDER_STATUS
```
nbradbury: 👍

nbradbury:
Everything looks and works fine! :shipit:

Issue #1073, opened 2019-01-15
Crash: java.lang.NumberFormatException in StatsSqlUtils


jtreanor:
First crashed in 11.5-rc-1

```
Caused by java.lang.NumberFormatException: Invalid number: Jan 
       at com.google.gson.internal.bind.util.ISO8601Utils.parseInt(ISO8601Utils.java:311)
       at com.google.gson.internal.bind.util.ISO8601Utils.parse(ISO8601Utils.java:129)
       at com.google.gson.internal.bind.DateTypeAdapter.deserializeToDate(DateTypeAdapter.java:85)
       at com.google.gson.internal.bind.DateTypeAdapter.read(DateTypeAdapter.java:75)
       at com.google.gson.internal.bind.DateTypeAdapter.read(DateTypeAdapter.java:46)
       at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:131)
       at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:222)
       at com.google.gson.Gson.fromJson(Gson.java:927)
       at com.google.gson.Gson.fromJson(Gson.java:892)
       at com.google.gson.Gson.fromJson(Gson.java:841)
       at com.google.gson.Gson.fromJson(Gson.java:813)
       at org.wordpress.android.fluxc.persistence.StatsSqlUtils.select(StatsSqlUtils.kt:53)
       at org.wordpress.android.fluxc.persistence.StatsSqlUtils.select$default(StatsSqlUtils.kt:43)
       at org.wordpress.android.fluxc.persistence.InsightsSqlUtils.select(InsightsSqlUtils.kt:118)
       at org.wordpress.android.fluxc.persistence.InsightsSqlUtils.selectLatestPostDetail(InsightsSqlUtils.kt:79)
       at org.wordpress.android.fluxc.store.InsightsStore.getLatestPostInsights(InsightsStore.kt:100)
       at org.wordpress.android.ui.stats.refresh.lists.sections.insights.usecases.LatestPostSummaryUseCase.loadCachedData(LatestPostSummaryUseCase.kt:36)
       at org.wordpress.android.ui.stats.refresh.lists.sections.BaseStatsUseCase.fetch(BaseStatsUseCase.kt:53)
[...]

PR #1075, opened 2019-01-15
Use fixed date format in Stats Jsons

planarvoid:
Should fix: #1073

I think this is caused by different Locales saving and restoring the date in a wrong way. The date is stored without `AM/PM` suffix and the retrieving doesn't work if it's not there. This PR sets a constant standard date format to the Gson builder and rebuilds the DB cache.

This should be tested with the WPAndroid PR.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_TimeStatsTestJetpack.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_TimeStatsTestJetpack.kt
@@ -27,9 +27,9 @@ import org.wordpress.android.fluxc.store.stats.time.ClicksStore
 import org.wordpress.android.fluxc.store.stats.time.CountryViewsStore
 import org.wordpress.android.fluxc.store.stats.time.PostAndPageViewsStore
 import org.wordpress.android.fluxc.store.stats.time.ReferrersStore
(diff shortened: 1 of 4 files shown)


malinajirka:
Thanks @planarvoid !

When we discussed this solution I haven't realized we'd be changing the dateformatter for all the usages of Gson. I think this solution is a bit too risky. I think we should consider implementing a custom deserializer for the weird format which is crashing the app. Wdyt?

planarvoid:
@malinajirka you're right. I've moved the Gson setup to the StatsSqlUtils instead so we should be safe. The DB cache is cleared as part of this change so there shouldn't be any issues with this move.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_TimeStatsTestJetpack.kt
```diff
@@ -161,13 +161,14 @@ class ReleaseStack_TimeStatsTestJetpack : ReleaseStack_Base() {
                         true
                 )
             }
+            if (!fetchedInsights.isError) {
```
planarvoid: I'm including this because the test started failing because the endpoint started returning an error 😢

malinajirka:
LGTM! Thanks!

PR #1076, opened 2019-01-16
Add featured image to Page model

0nko:
This PR adds a featured image ID to the `PageModel` so that an image can be displayed in the list.


example/src/test/java/org/wordpress/android/fluxc/page/PageStoreTest.kt
```diff
@@ -287,6 +287,6 @@ class PageStoreTest {
 
     private fun createPageFromPost(post: PostModel, site: SiteModel, parent: PageModel? = null): PageModel {
         return PageModel(site, post.id, post.title, PageStatus.fromPost(post), Date(), post.isLocallyChanged,
-                post.remotePostId, parent)
+                post.remotePostId, parent, 0)
```
planarvoid: shouldn't this also be `post.featuredImageId`? 😉

0nko: You're right! 💯

0nko:
Oops! Sorry about that, @oguzkocer!

PR #1078, opened 2019-01-17
Use different date format when requesting stats

planarvoid:
- Previously we were using the same date format as Woo but it seems that this format is not what the API expects (and it doesn't return the correct data). This PR is fixing this issue.


malinajirka:
LGTM! Feel free to merge it.

PR #1087, opened 2019-01-22
Fixes for connectedAndroidTest suite

aforcier:
This PR fixes a collection of issues I found with the connected, live network tests.

~Currently a WIP and not ready to merge - some fixes are still in progress and at least one is pending server-side changes.~

### Checklist:

* [x] Address failing Reader search test
* [x] Address theme test switching to theme without post formats (breaking a separate test)
* [x] Fix failures in multiple upload tests
* [x] Address breaking Automated Transfer tests (pending server-side change)
* [x] Deal with memory errors when running entire suite
* [x] Possibly rework post formats test not to be theme-dependent

#### Reader

The reader search test we have was failing due to only 9 results being returned when the full 10 requested were returned - I'm not sure what changed, but for now I've switched to a keyword with more results.

#### Themes

We've had an issue before where the theme activation test was switching to a theme that doesn't support post formats, causing the separate post formats test to fail. This is happening again because we had limited the themes to the `twenty*` themes family, but the new `twentynineteen` theme doesn't support post formats.

~For simplicity I've refined the logic to exclude `twentynineteen`, but this is still brittle - I'll likely just change the test to a new set of credentials before opening this PR up for review.~

Edit: Ended up implementing the above - there's now a new test site defined that's required to be using a post formats-compatible theme, and the post formats test uses that site specifically. Also updated the theme test not to be limited to twenty* themes since it's no longer necessary.

#### Media

There were some intermittent failures with our multiple concurrent upload tests. The cause turned out to be the recently added EXIF data stripping logic (https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1059). Since the tests are using the same file multiple times, we end up with a concurrency issue of one thread reading the data to upload, while another writes blank EXIF data to the same file. This ends up causing the upload to fail.

Since the EXIF is being tested implicitly by single-upload tests anyway, I fixed the issue by just disabling location stripping for multiple upload tests.

I don't think that this issue affects production, as I believe we make a local copy of each media being uploaded.

#### Automated Transfer

A few of our AT tests are failing or causing failures due to some server-side changes - I'm holding off any changes to those until current work on the backend is complete. At that point I'll either tweak the tests or remove some of them.

Edit: Server-side changes are pending - for now I opted to comment out the breaking tests, and opened #1145 to track restoring them when the server-side changes are live.

#### Memory Errors

I was seeing the last few connected tests fail due to memory errors. After some troubleshooting, it looks like it's at least partially due to a build up of `RequestQueue`s (we spin up a new one for each test) that aren't memory collected. Calling `RequestQueue.stop()` after each test seems to have fixed the issue (see https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1087/commits/b1829c463d56cb3b13c18605020b4cec862bfa38).


aforcier:
This PR is nearly complete, just a few items left:

* [ ] Merge in latest `develop` after #1146 is merged and fixes a broken test
* [ ] Release a new `tests.properties` file with new creds
* [ ] Test changes on CircleCI

aforcier:
For reference, this is what I use to run the full set of tests on Firebase (a local physical device also works of course):

(after `./gradlew clean example:assembleDebug example:assembleDebugAndroidTest`)

```
gcloud firebase test android run \
 --type instrumentation \
 --app example/build/outputs/apk/debug/example-debug.apk \
 --test example/build/outputs/apk/androidTest/debug/example-debug-androidTest.apk \
 --device model=walleye,version=26,locale=en,orientation=portrait \
 --project api-project-108380595987 \
 --timeout 30m
```

This is using a real device instead of an emulator since experience has shown emulators aren't very reliable for running the full test suite (maybe this is better now that the request queue memory leak has been addressed, but I have my doubts - worth trying though).

aforcier:
Note: In my last few runs of the tests, I'm unfortunately getting occasional failures of various media tests on Firebase (seems okay when run directly on my device). I'll try using a more recent device on Firebase to see if that makes a difference.

It's probably worth looking into converting some of our media tests into mocked ones (the mocked configuration wasn't really up to it when those tests were written originally, but we should be in better shape now).

aforcier:
Merging, thanks @jtreanor & @jkmassel!

Issue #1067, opened 2019-01-14
Add Manage Insights order functionality to Stats


planarvoid:
We need to implement functionality to manage insights. On the FluxC part it means adding methods to modify the Insights types order and an endpoint to retrieve the types with the status for the Settings screen.

PR #1088, opened 2019-01-22
Manage insights in stats

planarvoid:
#1067 
This PR adds a Store functionality to add/remove and move up/down functionality for types in Stats. There should be tests (including Integration tests) for everything. There will be a PR in WPAndroid to test the functionality.


0nko:
:shipit:

PR #1092, opened 2019-01-23
Issue/1090 unread notifs check

### LINKED ISSUE #1090: Woo: Detect unread store notifications
In order to resolve [this Woo issue](https://github.com/woocommerce/woocommerce-android/issues/639), we need a FluxC method to detect when there are unread store notifications.

nbradbury:
Resolves #1090 - adds a method to check if a site has unread notifications, optionally matching a type and subtype. This is required in order to tackle [this Woo issue](https://github.com/woocommerce/woocommerce-android/issues/639).

screenshot_1548255403

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/NotificationSqlUtils.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/NotificationSqlUtils.kt
@@ -120,6 +120,39 @@ class NotificationSqlUtils @Inject constructor(private val formattableContentMap
                 .map { it.build(formattableContentMapper) }
     }
 
+    fun hasUnreadNotificationsForSite(
+        site: SiteModel,
+        filterByType: List<String>? = null,
+        filterBySubtype: List<String>? = null): Boolean {
+        val conditionClauseBuilder = WellSql.select(NotificationModelBuilder::class.java)
+                .where()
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/NotificationSqlUtils.kt
```diff
@@ -120,6 +120,38 @@ class NotificationSqlUtils @Inject constructor(private val formattableContentMap
                 .map { it.build(formattableContentMapper) }
     }
 
+    fun hasUnreadNotificationsForSite(
```
aforcier: @nbradbury how about adding a unit test for this function in `NotificationsSqlUtilsTest`? This will also help test out a later refactor when I add the `SELECT 1` functionality to Wellsql.

nbradbury:
Thanks for the review @aforcier. I added a test in 26f29ba.

nbradbury:
> can you move the `CHECK UNREAD NOTIFICATIONS (DB)` button down to the next section under the `Perform actions on a selected site` since it requires a site be selected first?

Good point - added in c1dc1b6. A test was added in 26f29ba so ready for another round!

nbradbury:
@AmandaRiu My apologies, totally my mistake. I corrected it in 3f54ddb

PR #1094, opened 2019-01-25
Woo: Support localized currency formatting

aforcier:
Uses the `WCSettingsModel` added in #1093 to support formatting currency values for display according to the site's settings.

See https://github.com/woocommerce/woocommerce-android/issues/148 for more background, and the WooCommerce app counterpart PR https://github.com/woocommerce/woocommerce-android/pull/730.

~_Marking as not ready to review until I'm finished with the WooAndroid-side implementation._~

A few notes:
* The currency symbol used depends on the currency code (`USD`, `CAD`, etc) provided, and on the device's primary locale. For example, CAD are shown as `$` in `en_CA`, and as `CA$` in `en_US`.
* I added a few 'mocked network' tests that involve the currency localization. These tests don't work in a unit test environment, and I also found that older Android versions have different localized currency values in some cases. I've disabled these from running on older APIs (including Travis' API 15 emulator) for now - if they cause trouble on local tests as well we can just get rid of the tests.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/TestUtils.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/TestUtils.java
@@ -1,6 +1,7 @@
 package org.wordpress.android.fluxc;
 
 import android.content.Context;
+import android.content.res.Configuration;
 import android.support.annotation.NonNull;
 import android.text.TextUtils;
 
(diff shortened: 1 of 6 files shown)


nbradbury:
@aforcier This looks good to me. I'll review the Woo PR before merging this, then merge before merging the Woo one.

nbradbury:
:shipit:

PR #1096, opened 2019-01-30
Feature/plans

khaykov:
This PR adds support for Plans to Flux-C. 

Do not be alarmed by the number of changes - there is tons of boilerplate code, models and tests. The basic logic and functionality inside Store, SQL Utils and Rest Client are really simple:
- We fetch plans using plain get request.
- We cache plans and return them to the client.
- In case of error, the client will get an error message alongside cached plans.

To distinguish Plans that you see on the Plans page from the Plan data that Site has, I named them `PlanOffers`.

A lot of extra boilerplate code comes from the fact that the data in the Plans response is not flat, and in order for us to store it in DB, it needs to go through various mappings.


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PlanOffersSqlUtils.kt
```diff
@@ -0,0 +1,131 @@
...
+        override fun getId(): Int {
+            return this.id
+        }
+
+        override fun setId(id: Int) {
+            this.id = id
+        }
+    }
+
+    @Table(name = "PlanOffer")
```
theck13: Can you help me understand why we need both the `PlanOffer` and `PlanOfferId` tables?

khaykov: Sure! Plans are weird and can have multiple ID's that come to us in an array. As there is no good way to store an array in a DB the proper course of action (at least from my understanding) is to have another table where we store the ID's and link them to a single plan with `internalPlanId`. In the future, we might need plan id's to match against the plan that customer's site has. Hope this makes sense!

khaykov:
Thanks for taking a look at this PR! `PlanOffer` refers to pieces of code that deal with a single plan (eg. Bussines) and `PlanOffers` to the one that deals with multiple. Honestly, I'm not sure if it's this is 100% linguistically correct. Does the `PlanOffers` make more sense?

theck13:
You are correct by using `PlanOffer` for a single plan and `PlanOffers` for multiple plans.  There are some parts of the code that have a mix of both (e.g. `PlanOfferAction.FETCH_PLAN_OFFERS`).  Would `PlanOffers` be confusing if we used it everywhere?  If so, we should keep the current naming.  If not, I would just use `PlanOffers` for consistency.  It's not a major issue.  I'm fine with merging this either way.  I just don't want others or our future selves to be confused reading the code.  Let me know what you think.

khaykov:
You are right, I thought about it a bit, and using `PlanOffers` doesn't take anything away, and adds consistency 👍

PR #1098, opened 2019-01-31
Add paging when loading the followers

0nko:
This PR adds paging functionality, when fetching the followers and adds a way to request how the data is loaded from the cache (`CacheMode`) and the server (`FetchMode`)

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/InsightsSqlUtils.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/InsightsSqlUtils.kt
@@ -51,8 +51,8 @@ class InsightsSqlUtils
         insert(site, TODAYS_INSIGHTS, data)
     }
 
-    fun insert(site: SiteModel, data: FollowersResponse, followerType: FollowerType, deleteOldDataFirst: Boolean) {
-        insert(site, followerType.toDbKey(), data, deleteOldDataFirst)
+    fun insert(site: SiteModel, data: FollowersResponse, followerType: FollowerType, replaceExistingData: Boolean) {
+        insert(site, followerType.toDbKey(), data, replaceExistingData)
     }
 
     fun insert(site: SiteModel, data: CommentsResponse) {
@@ -114,8 +114,8 @@ class InsightsSqlUtils
         return select(site, TAGS_AND_CATEGORIES_INSIGHTS, TagsResponse::class.java)
     }
 
-    private fun <T> insert(site: SiteModel, blockType: BlockType, data: T, deleteOldDataFirst: Boolean = true) {
-        statsSqlUtils.insert(site, blockType, INSIGHTS, data, deleteOldDataFirst)
+    private fun <T> insert(site: SiteModel, blockType: BlockType, data: T, replaceExistingData: Boolean = true) {
+        statsSqlUtils.insert(site, blockType, INSIGHTS, data, replaceExistingData)
     }
 
     private fun <T> select(site: SiteModel, blockType: BlockType, classOfT: Class<T>): T? {
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/InsightsSqlUtils.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/InsightsSqlUtils.kt
@@ -91,6 +91,10 @@ class InsightsSqlUtils
         return select(site, followerType.toDbKey(), FollowersResponse::class.java)
     }
 
+    fun selectAllFollowers(site: SiteModel, followerType: FollowerType): List<FollowersResponse> {
+        return selectAll(site, followerType.toDbKey(), FollowersResponse::class.java)
+    }
+
     fun selectPublicizeInsights(site: SiteModel): PublicizeResponse? {
         return select(site, PUBLICIZE_INSIGHTS, PublicizeResponse::class.java)
     }
@@ -116,5 +120,9 @@ class InsightsSqlUtils
 
     private fun <T> select(site: SiteModel, blockType: BlockType, classOfT: Class<T>): T? {
         return statsSqlUtils.select(site, blockType, INSIGHTS, classOfT)
+    }
+
+    private fun <T> selectAll(site: SiteModel, blockType: BlockType, classOfT: Class<T>): List<T> {
+        return statsSqlUtils.selectAll(site, blockType, INSIGHTS, classOfT)
     }
 }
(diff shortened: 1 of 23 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/InsightsSqlUtils.kt
```diff
@@ -51,8 +51,8 @@ class InsightsSqlUtils
         insert(site, TODAYS_INSIGHTS, data)
     }
 
-    fun insert(site: SiteModel, data: FollowersResponse, followerType: FollowerType) {
-        insert(site, followerType.toDbKey(), data)
+    fun insert(site: SiteModel, data: FollowersResponse, followerType: FollowerType, deleteOldDataFirst: Boolean) {
```
planarvoid: maybe a better name for the param would be `replace`?

fluxc/src/main/java/org/wordpress/android/fluxc/store/stats/InsightsStore.kt
```diff
@@ -129,38 +129,65 @@ class InsightsStore
...
-                OnStatsFetched(insightsMapper.map(response.response, followerType, pageSize))
+                sqlUtils.insert(siteModel, response.response, followerType, deleteOldDataFirst = !loadMore)
+                val allFollowers = mergeFollowersModels(siteModel, followerType, pageSize)
+                OnStatsFetched(allFollowers)
             }
             else -> OnStatsFetched(StatsError(INVALID_RESPONSE))
         }
     }
 
+    private fun mergeFollowersModels(
```
planarvoid: what do you think about moving this method to the InsightsMapper?

0nko:
Thanks @planarvoid! Ready for another check.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StatsSqlUtils.kt
```diff
@@ -22,15 +22,24 @@ class StatsSqlUtils
...
-        WellSql.delete(StatsBlockBuilder::class.java)
-                .where()
-                .equals(StatsBlockTable.BLOCK_TYPE, blockType.name)
-                .equals(StatsBlockTable.STATS_TYPE, statsType.name)
-                .equals(StatsBlockTable.DATE, date)
-                .endWhere()
-                .execute()
+        if (replaceExistingData) {
+            WellSql.delete(StatsBlockBuilder::class.java)
+                    .where()
```
malinajirka: Shouldn't we also add `.equals(StatsBlockTable.LOCAL_SITE_ID, site.id)` into the where clause?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StatsSqlUtils.kt
```diff
@@ -22,15 +22,24 @@ class StatsSqlUtils
...
+                    .equals(StatsBlockTable.BLOCK_TYPE, blockType.name)
+                    .equals(StatsBlockTable.STATS_TYPE, statsType.name)
+                    .equals(StatsBlockTable.DATE, date)
```
malinajirka: I think this will select only rows with `date == ""` if the client calls this method with the default value for "date". Is this the intended behavior or do we want to ignore value of `date` when `date==""`?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StatsSqlUtils.kt
```diff
@@ -43,6 +52,23 @@ class StatsSqlUtils
...
+        statsType: StatsType,
+        classOfT: Class<T>,
+        date: String = ""
+    ): List<T> {
+        val models = WellSql.select(StatsBlockBuilder::class.java)
+                .where()
+                .equals(StatsBlockTable.LOCAL_SITE_ID, site.id)
+                .equals(StatsBlockTable.BLOCK_TYPE, blockType.name)
+                .equals(StatsBlockTable.STATS_TYPE, statsType.name)
+                .equals(StatsBlockTable.DATE, date)
```
malinajirka: I think this will select only rows with `date == ""` if the client calls this method with the default value for "date". Is this the intended behavior or do we want to ignore value of `date` when `date==""`?

malinajirka:
Thanks @0nko! Looks good to me, I've left just couple of questions.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StatsSqlUtils.kt
0nko: Good catch.. I think it should be!

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StatsSqlUtils.kt
0nko: This should be OK. This parameter is not used for `INSIGHTS` StatsType, so the `""` default value is used when it's inserted. Which means that it's correctly selected here. For granular StatsTypes, the date parameter is actually supplied.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StatsSqlUtils.kt
0nko: Same as above, this is the case used for Insights.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StatsSqlUtils.kt
malinajirka: If it works the way I think it does, I'm a bit concerned it may be confusing. Imagine you call `store.selectAll(..)` and you don't pass "date" parameter. I expect to receive all rows no matter what's in the date column, but I'll receive just rows with date=="".

@planarvoid Can you please confirm, it works the way I think it does?

planarvoid: @malinajirka @0nko both of you are right and it might indeed be confusing. Currently you get back either Insights when you don't set the date parameter and DWMY stats if you do. I think it might be a better solution to have the `Date` nullable and not add the `equals` call when we're dealing with the Insights.

malinajirka: Thanks Vojta!

>I think it might be a better solution to have the `Date` nullable and not add the `equals` call when we're dealing with the Insights.

100% ;)

0nko:
Thanks for all the suggestions, @malinajirka! I've updated the date parameter/column to be nullable and created a migration script.

malinajirka:
Thanks @0nko! I haven't tested it yet. The code looks good overall. I just noticed you've probably forgotten to increment the db version, right?

malinajirka:
Thanks the code looks good! After a discussion with Ondrej, I'll test this PR along with `post & pages` PR when it's ready for review.

malinajirka:
Thanks @0nko! I've just tested the WPAndroid app and it seems the changes are working as expected.

I've encountered some unrelated issues though - I'll send you a gif over DM as it contains some private information.

PR #1106, opened 2019-02-06
Parcelize plan offers model and drop UNIQUE constraint from plan feature table

khaykov:
This PR parcelizes the PlanOffersModel used in the client app.

In addition, I removed UNIQUE constraint from feature id column, which was there originally with a different data structure in mind. This was causing a non-fatal error message in logcat on client app.
There is no way to drop constraint in SQLite, so I had to migrate the table's content.


theck13:
The changes look good to me.  There's a merge conflict in `WellSqlConfig.java`.  Once that's resolved, I'll merge this pull request.

PR #1116, opened 2019-02-12
Return null when order status not found

AmandaRiu:
Fixes an issue (https://github.com/woocommerce/woocommerce-android/issues/781) where if an order status key is not found in the database, a `NoSuchElementException` is thrown and the app crashes.

While I was in there I also added tests to cover such scenarios to make sure the fixes work and continue work:
- `OrderSqlUtilsTest.testGetOrderStatusOption_NotExists()`
- `OrderSqlUtilsTest.testGetOrderStatusOptions_Empty`

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
@@ -139,7 +139,7 @@ object OrderSqlUtils {
         }
     }
 
-    fun getOrderStatusOptionsForSite(site: SiteModel) =
+    fun getOrderStatusOptionsForSite(site: SiteModel): List<WCOrderStatusModel> =
             WellSql.select(WCOrderStatusModel::class.java)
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
```diff
@@ -139,7 +139,7 @@ object OrderSqlUtils {
         }
     }
 
-    fun getOrderStatusOptionsForSite(site: SiteModel) =
+    fun getOrderStatusOptionsForSite(site: SiteModel): List<WCOrderStatusModel> =
```
nbradbury: It looks like we call this from several places that may not be expecting a null result. How about we return an empty list instead of null?

AmandaRiu: Hrm...interesting. This one with the list is not returning a null.

nbradbury:
:shipit:

PR #1128, opened 2019-02-19
Anitaa/support custom order stats

### LINKED ISSUE #1007: WooCommerce: Add ability to fetch custom stats
Since the caching of stats to the database heavily relies on preset query values such as 30 DAYS, in order to support custom stat ranges new logic will need to be added to differentiate that query from the other query types. 

cc: @aforcier @astralbodies

AmandaRiu:
Fixes #1007  - The original PR from the forked repo can be found [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1103). That PR was merged into a branch in this repo for additional testing prior to the final merge into `develop`

This PR adds the following:

* Provides option to pass date range when fetching order/revenue stats
* Calculates quantity (how far back to go), based on the date range and the granularity before passing it to server api. If no date range is passed, default/static values are used.
* Parses and caches the response in the db - including the quantity and date.
*  Unit tests and mock tests included. (Added in `WCStatsStoreTest`)
* Also includes 2 new buttons on the **Woo** screen of the **example** app: **FETCH CUSTOM REVENUE STATS** and **FETCH CUSTOM REVENUE STATS (FORCED)**

## Testing Notes
*  This feature includes an alteration to existing db table: `WCOrderStatsModel` - App upgrade testing to be done as well.

**Note** Changes need to be added to the front end as well to utilise this feature in WCAndroid.


AmandaRiu:
@anitaa1990 For the visitor stats code: I recommend removing anything that was added to the Payload specifically for the example qpp that wouldn't be used by the main app since it's just extra baggage. The visitor api calls all use caching at the HTTP level so if the parameters are the same, it will just automatically use the matching response. I also recommend removing the code for shared preferences. For the example app, just having the date ranges cached for the _active session_ is enough. Does that sound good? 

The rest all looks really good 👍 

You may need to create a new PR to merge into my `anitaa/support_custom_order_stats ` branch up on FluxC for these changes so I can merge them into our current branch since I don't believe you have access to edit that PR or branch directly (although you are welcome to try).

anitaa1990:
> For the visitor stats code: I recommend removing anything that was added to the Payload specifically for the example qpp that wouldn't be used by the main app since it's just extra baggage. The visitor api calls all use caching at the HTTP level so if the parameters are the same, it will just automatically use the matching response. I also recommend removing the code for shared preferences. For the example app, just having the date ranges cached for the _active session_ is enough. Does that sound good?

@AmandaRiu , This makes sense! I have created another [PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1129) to merge into this one. This PR addresses the above issue. I have just added variables in `WooCommerceFragment` for the `startDate`, `endDate` and `granularity`. So the selection will be remembered till user closes the app.

PR #1139, opened 2019-02-27
Remove caching from stats

planarvoid:
- this PR disables caching in Comments and Tags. The cache returns the same result and ignores the query params (in this case needed for loading the next pages).


0nko:
Looks good :+1:

PR #1143, opened 2019-03-01
Add CircleCI config

jtreanor:
Adding a CircleCI config for this repo. It re-uses the configuration from https://github.com/wordpress-mobile/circleci-orbs.

I have added an optional build param, `testsMaxHeapSize `, that limits the heap size for unit tests. This is needed to prevent out of memory errors on CI.

As discussed with @aforcier, this uses Firebase Test Lab for the mocked connected tests. Although it is possible to run the tests in a headless emulator on CircleCI, it proved to be slower than Firebase. Emaulators are also [not fully supported by CircleCI](https://support.circleci.com/hc/en-us/articles/360000028928-Testing-with-Android-emulator-on-CircleCI-2-0).

To avoid duplicating boilerplate across several projects, this uses the `firebase-test` command I added in https://github.com/wordpress-mobile/circleci-orbs/pull/14.

To test:

- CircleCI checks are ✅


.circleci/config.yml
```diff
@@ -0,0 +1,84 @@
...
+    executor: 
+      name: android/default
+      api-version: "27"
+    steps:
+      - checkout
+      - android/restore-gradle-cache:
+          cache-prefix: lint
+      - copy-gradle-properties
+      - run:
+          name: Checkstyle
```
aforcier: I wonder if we should name this task something else since it's actually running both `checkstyle` and `ktlint`? Maybe `Check Style`? Or literally `Checkstyle & Ktlint`? Or `Style Lint`? WDYT?

jtreanor: Yeah, that makes sense. I'll update it now 😄

aforcier:
Looks great @jtreanor! Tested breaking tests or tripping the linters in a test branch and making sure they're reported.

It's fun to watch Travis still pending minutes after all the CircleCI tasks are done 😀 Glad to have this speed-up in place.

:shipit:

PR #1144, opened 2019-03-03
Issue/wc product images

nbradbury:
The purpose of this PR is to enable WCAndroid to display product images. The most important change is [WCProductModel.getFirstImage()](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/77ce6d4864682d383fe30bde3fb7f3ea1c7b68a0/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt#L120) which extracts the first image from a product's JSON array of images. I updated the mocked test to include `getFirstImage()`.

This PR also updates the `WCProductStore` by shortening a method name and adding a method to check whether a product exists in the db.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -119,7 +119,7 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
      */
     fun getFirstImage(): String? {
         try {
-            Gson().fromJson<JsonElement>(images, JsonElement::class.java).asJsonArray.first() { jsonElement ->
+            Gson().fromJson<JsonElement>(images, JsonElement::class.java).asJsonArray.first { jsonElement ->
```
AmandaRiu: Recommend using `firstOrNull` here just in case no product images have been set

nbradbury: > I left one tiny code comment, but the rest looks good to go 👍 Once the comment is addressed and CircleCI fixed, this will be ready for merge.

@AmandaRiu Good call on using `firstOrNull`. I added that in 39d951f.

I don't think we need to wait for CircleCI. The whole problem there is that the CircleCI config PR is waiting to be merged, and until it is any new PRs will fail.

nbradbury:
> Why break it up into two statements? You're original was really nice, you just needed to change `first` to `firstOrNull`

Changed in f08bec0

PR #1149, opened 2019-03-07
Removed Travis CI config

jtreanor:
CircleCI has been running reliably for a while now so I am removing the Travis config. We have already seen that CircleCI is more reliable than the existing Travis setup.

@aforcier If you would prefer to keep these running side by side for longer, we can wait before merging.


aforcier:
It's been working well so far and seems to get rid of the mocked test failure issue - plus we can always bring Travis back if we have an emergency with the CircleCI setup - :shipit:

PR #1154, opened 2019-03-08
Woo: Fix default stats refresh bug

AmandaRiu:
This PR fixes https://github.com/woocommerce/woocommerce-android/issues/894 by only pulling stats by date if custom stats.

This fixes a bug where the default stats were creating multiple rows because
a different date was used each time. This in turn caused the logic that
polls for the default stats to only return the top row - making it
appear that stats was not refreshing on the device. The fix was to not use a date value when polling the database for default stats so only a single row ever exists. Special thanks to Anita Murthy finding and providing the snippets used in this fix.

## To Test
1. Manually change the device date to yesterday
2. Using a clean install of the Example app, click "FETCH 30 DAY REVENUE STATS" .
3. Change the device date back to today.
4. Click "FETCH 30 DAY REVENUE STATS" in the example app.
5. verify only a single row in the database for 30 day stats using Stethos.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCStatsSqlUtils.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCStatsSqlUtils.kt
@@ -8,7 +8,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.orderstats.OrderStatsRe
 
 object WCStatsSqlUtils {
     fun insertOrUpdateStats(stats: WCOrderStatsModel): Int {
-        val statsResult = if(stats.isCustomField) {
+        val statsResult = if (stats.isCustomField) {
             WellSql.select(WCOrderStatsModel::class.java)
                     .where().beginGroup()
                     .equals(WCOrderStatsModelTable.LOCAL_SITE_ID, stats.localSiteId)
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCStatsSqlUtils.kt
```diff
@@ -8,15 +8,25 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.orderstats.OrderStatsRe
...
+                    .equals(WCOrderStatsModelTable.QUANTITY, stats.quantity)
+                    .equals(WCOrderStatsModelTable.IS_CUSTOM_FIELD, stats.isCustomField)
+                    .endGroup().endWhere()
+                    .asModel
+        } else {
+            WellSql.select(WCOrderStatsModel::class.java)
+                    .where().beginGroup()
+                    .equals(WCOrderStatsModelTable.LOCAL_SITE_ID, stats.localSiteId)
+                    .equals(WCOrderStatsModelTable.UNIT, stats.unit)
+                    .equals(WCOrderStatsModelTable.IS_CUSTOM_FIELD, stats.isCustomField)
```
nbradbury: Was dropping the match on `stats.quantity` here intentional?

nbradbury:
Both this and the related WCAndroid PR look good. I just had [one question](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1154/files#r263912914) and then I'm ready to :shipit:

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCStatsSqlUtils.kt
AmandaRiu: Yes it was...not used for default stats.

PR #1162, opened 2019-03-12
Add endpoints for stats detail

planarvoid:
- This PR adds Store for post detail. The endpoint was already implemented for loading post details in the `LatestPostSummary`. I've only changed the SqlUtils to accept a `postId` of the cached post. It should be tested with the related `WPAndroid` PR


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/InsightsSqlUtils.kt
```diff
@@ -85,8 +85,8 @@ class InsightsSqlUtils
         return select(site, LATEST_POST_DETAIL_INSIGHTS, PostResponse::class.java)
     }
 
-    fun selectLatestPostStats(site: SiteModel): PostStatsResponse? {
-        return select(site, LATEST_POST_STATS_INSIGHTS, PostStatsResponse::class.java)
+    fun selectLatestPostStats(site: SiteModel, postId: Long): PostStatsResponse? {
```
0nko: We should probably rename this to something like `selectDetailedPostStats()` now that we're storing more than one.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/InsightsSqlUtils.kt
```diff
@@ -85,8 +85,8 @@ class InsightsSqlUtils
...
+        return select(site, LATEST_POST_STATS_INSIGHTS, PostStatsResponse::class.java, postId)
```
0nko: Same here, I think the enum value should be renamed.

fluxc/src/main/java/org/wordpress/android/fluxc/store/stats/insights/LatestPostInsightsStore.kt
```diff
@@ -43,12 +43,10 @@ class LatestPostInsightsStore @Inject constructor(
     }
 
     fun getLatestPostInsights(site: SiteModel): InsightsLatestPostModel? {
```
0nko: Same here: method name.

planarvoid: I think this one is correct. This is still returning the stats for the latest post. There is another store that returns the value for a post by `id`

planarvoid:
@0nko  thanks for the review. Everything should be fixed now and there is a WPAndroid PR that can be tested.

PR #1171, opened 2019-03-15
Add details about OAuth2

drspaceboo:
This PR updates the README file to add some more details regarding OAuth2. Specifically pointing people to the page where they can get those details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,11 +33,19 @@ our example:
     $ ./gradlew testDebug # Unit tests
 
 Note: this is the default `example/gradle.properties` file. You'll have to get
-a WordPress.com OAuth2 ID and secret.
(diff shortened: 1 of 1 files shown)


aforcier:
@jtreanor it looks like the mocked connected test task on CircleCI for this PR is failing - I suspect it's to do with the Firebase config and the branch being from a fork? I think I recall us running into something like this before.

jtreanor:
@aforcier yeah, unfortunately Frirebase tests won’t work for forks. There’s no easy way to avoid that unfortunately but we should definitely make it fail with a more helpful message

drspaceboo:
Tests can probably be ignored give that the PR only changes the README

AmandaRiu:
@jtreanor this all looks good for merge, but I'm unable to do it due to the failed tests. Any recommendations?

Issue #1172, opened 2019-03-15
XMLRPC not generated after cleaning project


drspaceboo:
I was setting up today to start contributing to this project however I ran into an issue where I couldn't just load the project up in Android Studio and build the example application.

It looks like some recent changes with Gradle has caused the `XMLRPC` class to not be generated by the annotation processors. The two changes that are causing the issue seem to be:

1. [Gradle is starting to ignore annotation processors that are on the `compile` path instead of on annotation processor path](https://discuss.gradle.org/t/regarding-the-annotation-processors-on-compile-classpath-warning-in-gradle-4-6/26144)
1. [Gradle now collapses multiple warnings that are generated when compiling trying to only show the most important ones.](https://docs.gradle.org/4.10.3/userguide/command_line_interface.html#sec:command_line_warnings)

The second was hiding the first issue, and when I ran `./gradlew example:assembleDebug --warning-mode all` to get all the warnings it ran the annotation processor correctly. After running that command the missing `XMLRPC` class now exists and can be found by Android Studio.

Currently the default build and run in Android Studio doesn't work for me as it deletes those generated files each time.

aforcier:
@drspaceboo Thanks for digging into the issue! I'm not sure what the ultimate cause of it is, but we've been running into it on occasion for a while - it predates our using Gradle `4.6`, for example.

The best guess so far has been that it's an issue with `kapt` + Android Studio as it started happening sometime after we starting using `kapt` and it doesn't seem to happen when initiating clean builds from the commandline.

Usually running `./gradlew --stop` and then a `./gradlew build` from the commandline sorts things out.

Regardless though it would definitely be worth addressing that warning you mentioned - it's possible it's even related somehow even if it's been happening since before Gradle `4.6`.

I'm not sure I follow your steps for fixing it on your end though - did you do something to address the warning? Otherwise I'm not sure I follow why showing the warnings would have been enough to resolve the issue.

aforcier:
After running `./gradlew example:assembleDebug --warning-mode all` I'm seeing a warning for `AutoServiceProcessor`. It looks like they recently released a new version which splits out the annotation artifact from the processor. I'm not sure doing that will solve the issue, but it might play better with `kapt`, and certainly it's worth addressing the warning regardless.

I'll look into this and open a PR you could maybe try @drspaceboo and see if that fixes your build from Android Studio - it happens to me pretty rarely so I don't expect I'll be able to confirm if this fixes it or not.

[...]

PR #1173, opened 2019-03-15
Address AutoService-based annotation processor warnings

aforcier:
In #1172 @drspaceboo ran into the missing `XMLRPC.java` (and other annotation processor-generated files) issue, though for her it's happening very consistently. She also pointed out that as of `Gradle 4.6` there's a warning that can be surfaced about the annotation processor:

> Detecting annotation processors on the compile classpath has been deprecated. Gradle 5.0 will ignore annotation processors on the compile classpath. The following annotation processors were detected on the compile classpath: 'com.google.auto.service.processor.AutoServiceProcessor'.  Please add them to the annotation processor path instead. If you did not intend to use annotation processors, you can use the '-proc:none' compiler argument to ignore them.

I don't believe the problem is related to Gradle `4.6` (we've been running into this issue since Gradle `4.1` at least). But it turns out this warning led the `AutoService` team to resolve it a few months ago by separating their annotation processor into two separate packages: https://github.com/google/auto/commit/af1e5daff849b4762062018b9e8bce1716bfc8b0.

Addressing the warning is only supposed to potentially increase speed (and needs to be resolved before we can update to Gradle 5.0 in the future, when presumably the build will be broken as the annotation process will be ignored). However, I'm hopeful adding `ktlint` and calling the `AutoService` processor in a more canonical way is enough to resolve whatever issue `ktlint` was having all along and get rid of the missing `XMLRPC.java` issue 🤞.

I also fixed another annotation processor warning - I'm less confident this has any relation to the build issue but we may as well fix it):

> w: warning: Supported source version 'RELEASE_6' from annotation processor 'org.jetbrains.kotlin.kapt3.base.ProcessorWrapper' less than -source '1.8'

(There's one more warning: `[options] bootstrap class path not set in conjunction with -source 1.7`. I have an idea how to fix this but I'm not sure it's safe just yet - I'll look at it some more and either file an issue or open a separate PR.)

@drspaceboo could you please give this branch a try and see if you're able to build the project from Android Studio with it?

It would also be great for multiple people to give this a try in case something _else_ breaks. Pinging a few folks I know have run into this build issue recently: @oguzkocer @0nko

### DIFF
--- a/fluxc-processor/build.gradle
+++ b/fluxc-processor/build.gradle
@@ -1,5 +1,16 @@
+buildscript {
+    repositories {
+        jcenter()
+    }
+    dependencies {
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
+    }
+}
+
 apply plugin: 'java'
 apply plugin: 'maven'
+apply plugin: 'kotlin'
+apply plugin: 'kotlin-kapt'
 
 repositories {
     jcenter()
@@ -10,6 +21,7 @@ targetCompatibility = JavaVersion.VERSION_1_7
 
 dependencies {
     implementation project (':fluxc-annotations')
-    implementation 'com.google.auto.service:auto-service:1.0-rc2'
+    implementation 'com.google.auto.service:auto-service:1.0-rc4'
+    kapt 'com.google.auto.service:auto-service:1.0-rc4'
     implementation 'com.squareup:javapoet:1.7.0'
 }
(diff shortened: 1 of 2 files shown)


drspaceboo:
@aforcier I could not replicate my issue with the branch. However I couldn't replicate it again with develop after running the commands you suggested.

Still good to fix the warnings and move to a newer version of `AutoService` 😀

aforcier:
> @aforcier I could not replicate my issue with the branch. However I couldn't replicate it again with develop after running the commands you suggested.
>
> Still good to fix the warnings and move to a newer version of AutoService 😀

Thanks for checking @drspaceboo! We may not be sure for a while if this resolves the issue or not, but hopefully!

Will wait for @oguzkocer / @0nko to give this a try as well.

PR #1177, opened 2019-03-19
Fix DB migration

planarvoid:
- increase step in WellSqlConfig. This error came from wrong merge to develop.


malinajirka:
LGTM!

PR #1178, opened 2019-03-19
Run connected tests on CircleCI/Firebase every night

jtreanor:
This PR adds a scheduled CircleCI job to run the full connected test suite on Firebase every night.

We still have some remaining flakiness with the tests but this should provide a reliable way to run them. @aforcier if you would prefer to wait for more stability work on the tests before merging, I'm fine with that.

I am running the tests on an emulator on Firebase as I actually found this to be more reliable to the device we are using currently. I'm not sure why that is.

You can see an example of a successful job [here](https://circleci.com/gh/wordpress-mobile/WordPress-FluxC-Android/779).


aforcier:
@jtreanor going to merge this. Since we're still seeing some intermittent failures on Firebase via wpmt, it might be useful to in the meantime have two independent runs of the tests to compare as we work on the flakiness.

:shipit:

jtreanor:
That makes sense. Thanks @aforcier!

PR #1182, opened 2019-03-21
Add accountStore necessary for running integration tests

planarvoid:
- the integration tests need to inject the AccountStore in order to pass (to successfully login the test user). This PR adds the class and fixes the tests.


aforcier:
Confirmed the two broken test classes fully pass now 👍

:shipit:

PR #1186, opened 2019-03-22
Wc/608 shipment tracking

### LINKED ISSUE #1185: WC: Add support for Shipment Trackings Extension
Add support for working with the Shipment Trackings extension by providing a way to:
- Fetch shipment trackings for a single order via the Jetpack Tunnel
- Store shipment trackings in the database

AmandaRiu:
Fixes #1185 by adding support for the Shipment Tracking Woo extension. 

Tests for this PR include:
- A new button in the example app to fetch the shipment trackings for a single order
- Mocked test
- Sql unit tests
- Release tests for a site with the plugin installed, and a site without the plugin installed

## Notes
I chose to add this code to the `WCOrderStore` since it was tied in so closely with orders and like order notes, the table would need to be cleared when the orders table is refreshed.


example/src/main/java/org/wordpress/android/fluxc/example/WooCommerceFragment.kt
```diff
@@ -382,6 +385,25 @@ class WooCommerceFragment : Fragment(), CustomStatsDialog.Listener {
...
+                        "Enter the remoteOrderId to fetch shipment trackings:"
+                ) { editText ->
+                    editText.text.toString().toLongOrNull()?.let { remoteOrderId ->
+                        wcOrderStore.getOrderByIdentifier(OrderIdentifier(site.id, remoteOrderId))?.let { order ->
+                            prependToLog("Submitting request to fetch shipment trackings for " +
+                                    "remoteOrderId: ${order.remoteOrderId}")
+                            pendingShipmentTrackingOrder = order
+                            val payload = FetchOrderShipmentTrackingsPayload(site, order)
+                            dispatcher.dispatch(WCOrderActionBuilder.newFetchOrderShipmentTrackingsAction(payload))
+                        } ?: prependToLog("No order found in the db for remoteOrderId: $remoteOrderId")
```
nbradbury: This isn't a big deal, but it would be nice if we didn't have to rely on the order being in the db.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
```diff
@@ -154,4 +156,43 @@ object OrderSqlUtils {
...
+                .equals(WCOrderShipmentTrackingModelTable.LOCAL_SITE_ID, tracking.localSiteId)
+                .equals(WCOrderShipmentTrackingModelTable.LOCAL_ORDER_ID, tracking.localOrderId)
+                .equals(WCOrderShipmentTrackingModelTable.REMOTE_TRACKING_ID, tracking.remoteTrackingId)
+                .endGroup().endGroup().endWhere().asModel
+
+        return if (result.isEmpty()) {
+            WellSql.insert(tracking).asSingleTransaction(true).execute()
+            1
+        } else {
+            0
```
nbradbury: If we ignore updating the db when the tracking already exists, are we safe to assume other fields in the model haven't changed (like `tracking_link`)?

nbradbury:
@AmandaRiu I'm done with my first pass. Looks great, just a couple minor comments. Also, there's a conflict that needs to be addressed.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
AmandaRiu: I wondered that as well, but there is no way to update the tracking once submitted. The only options are create and delete...no update option. So it seemed safe to write it this way.

example/src/main/java/org/wordpress/android/fluxc/example/WooCommerceFragment.kt
AmandaRiu: I've updated the log message to let the test user know to fetch orders before submitting to make the test work. I prefer to keep it this way because it's true to how the implementation works in general. There would already be an order in the database that we are fetching the order shipment trackings for, and that information is needed during the process of fetching and saving the order shipment tracking info to the db. Updated the log message in e3d91b19fac39369036244d0b39fe81997cb948f

AmandaRiu:
@nbradbury Made some minor tweaks to logging and also moved the button in the test app to be with the other order-related tests. Changed the label from "Fetch Shipment Trackings" to "Fetch Order Shipment Trackings" for clarity. Comments in code also addressed. Ready for another round! ⚡️

nbradbury:
This is still failing tests, but locally tests are passing. Merging this as per our Slack discussion. :shipit:

PR #1211, opened 2019-04-12
Add support for email_login_not_allowed

hypest:
Addresses https://github.com/wordpress-mobile/WordPress-Android/issues/7922

Following a corresponding fix on the WordPress.com side (so the error type is sent properly), this PR adds support for the `email_login_not_allowed` error type. Without this change, the error will be reported as `GENERIC_ERROR`.

WPAndroid side PR: https://github.com/wordpress-mobile/WordPress-Android/pull/9601

To test:

No test steps available.

👋 @aforcier , I'm adding you as a reviewer so we to make sure the Woo app is fine with the change too.


aforcier:
Confirmed that using this FluxC branch doesn't break anything for WPAndroid/WooAndroid, since the login flow defaults to the same behavior as when this caused an `INVALID_REQUEST`.

:shipit:

PR #1218, opened 2019-04-15
Woo: order tracking shipment providers

### LINKED ISSUE #1208: Woo Shipment Tracking: Add endpoint to fetch shipment tracking providers
Add a way to fetch and store shipment providers for the Shipment Trackings WooCommerce plugin

AmandaRiu:
Fixes #1208 by adding the ability to fetch shipment providers from the installed WooCommerce Shipment Tracking plugin. 

Notes:
- The plugin currently requires an order Id be used to fetch the list of providers even though providers are not directly attached to order_ids. This is why all this code lives in `WCOrderStore` and related classes.
- Mocked, Release, and unit tests have all been added
- The shipment providers api response is in non-standard JSON. For this reason the entire document has to be manually parsed instead of using model objects to automagically parse.


nbradbury:
Looks good! :shipit:

PR #1224, opened 2019-04-17
issue/1223-wc-manage-stock 

### LINKED ISSUE #1223: WC: Product `manageStock` needs to handle `parent`
The products API endpoint field `manage_stock` can return `parent` for a product variation if stock management is enabled at the product level but not for the variation. In this situation, we should treat it as `true`.

nbradbury:
Fixes #1223 by handling `manage_stock` when it's set to `parent` for a product variation. To test:

* Enable stock management at the parent level but not the variation, fetch that variation and confirm the product model has `manageStock=true`
* Disable stock management at the parent level but enable it for the variation, fetch that variation and confirm the product model has `manageStock=true` 
* Disable stock management for both the parent and the variation, fetch that variation and confirm the product model has `manageStock=false`


AmandaRiu:
@nbradbury this is a great case for a mocked test

PR #1226, opened 2019-04-18
Woo: Add and Delete Order Shipment Tracking

AmandaRiu:
Fixes #1221 

Adds the ability to add and delete order shipment tracking records for a single order. Also supports adding tracking with a custom provider.

## Automated Tests
Unit, Mocked and Release tests have also been added

PR #1229, opened 2019-04-22
Remove size from test since they will now be variable

AmandaRiu:
Small fix for one of the current Order Shipment Tracking tests. An existing PR (#1226 ) made some changes that changed the number of records were on the test server. This was causing the test to fail. Now that tests can add and delete shipment tracking records, it's better just to make sure that there ARE tracking records and not worry about exactly how many.


aforcier:
Closing since @nbradbury merged https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1226 in the meantime 😀

PR #1231, opened 2019-04-23
issue/reset-tables

nbradbury:
Adds an alternate method for resetting the database. This will be used by both WCAndroid and WPAndroid to recreate the database when a version downgrade is detected and `BuildConfig.DEBUG` is true. The existing `reset()` method can't be used in this situation (it results in an `IllegalAccessException`).

To test, change `getDbVersion()` to return an older version then add this code:

```
@Override 
public void onDowngrade(SQLiteDatabase db, WellTableManager helper, int oldVersion, int newVersion) {
    reset(helper);
}
```

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
@@ -537,6 +537,22 @@ public class WellSqlConfig extends DefaultWellConfig {
         }
     }
 
+    /**
+     * Recreates all the tables in this database - similar to the above but can be used from onDowngrade where we can't
+     * call giveMeWritableDb (attempting to do so results in "IllegalStateException: getDatabase called recursively")
+     */
+    @SuppressWarnings("unused")
+    private void reset(WellTableManager helper) {
+        AppLog.d(T.DB, "resetting tables");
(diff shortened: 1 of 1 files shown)


jtreanor:
@nbradbury  What circumstances lead to a DB downgrade that we want to support?

A few weeks ago we had an unusual situation where we accidentally released a hotfix of WPAndroid without releasing new beta/alpha versions. This lead to beta and alpha users getting a crash on launch when they installed the hotfix version due to a DB version downgrade. We quickly released new alpha and beta versions to resolve this and once those users updated everything worked fine again, without data loss.

If I'm not mistaken this change would mean if this ever happened again, users may lose locally stored data. Is this what we would want to happen?

nbradbury:
> What circumstances lead to a DB downgrade that we want to support?

@jtreanor Sorry, should've clarified that this is part of my hack week project, which is to detect a db downgrade when `BuildConfig.DEBUG == true` (ie: it would only affect devs).

The problem we have is we often switch branches, and if we switch to a branch that uses an older db version the app crashes, forcing us to either uninstall or clear storage, then login again.

I run into this **very** often, and I assume many others do, too, so my goal is to handle this better.

jtreanor:
@nbradbury That makes sense, thanks for clarifying. Sounds like a great improvement 😄

PR #1236, opened 2019-04-26
CircleCI: More helpful Slack notifications for connected tests

jtreanor:
Slack notifications for connected tests now work as follows.
- On success:
     - Post to `$SLACK_STATUS_WEBHOOK` with this message: `🎉 FluxC Connected tests have succeeded!\nSee https://console.firebase.google.com/u/0/project/api-project-108380595987/testlab/histories/bh.fbe4c2866a46a076`.
- On failure:
    - Post to `$SLACK_STATUS_WEBHOOK` with this message: `🔴 FluxC Connected tests have failed!\nSee https://console.firebase.google.com/u/0/project/api-project-108380595987/testlab/histories/bh.fbe4c2866a46a076`.
    - Post to `$SLACK_FAILURE_WEBHOOK` with the same message as above.

`$SLACK_FAILURE_WEBHOOK` and `$SLACK_STATUS_WEBHOOK` are currently the same, but we can update them to use different channels once merged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,6 +117,7 @@ jobs:
           steps:
             - run:
                 name: Setup Slack messages
+                when: always
                 command: |
                   FIREBASE_URL="https://console.firebase.google.com/project/api-project-108380595987/testlab/histories/bh.fbe4c2866a46a076"
                   echo "export SLACK_FAILURE_MESSAGE=':red_circle: FluxC Connected tests have failed!\nSee $FIREBASE_URL'" >> $BASH_ENV
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   android: wordpress-mobile/android@0.0.23
-  slack: circleci/slack@2.0.0
+  slack: circleci/slack@2.5.0
 
 commands:
   copy-gradle-properties:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -115,7 +115,22 @@ jobs:
       - when:
           condition: << parameters.post-slack-status >>
           steps:
-            - slack/status
+            - run:
+                name: Setup Slack messages
+                command: |
+                  FIREBASE_URL="https://console.firebase.google.com/u/0/project/api-project-108380595987/testlab/histories/bh.fbe4c2866a46a076"
```
aforcier: I think the `/u/0` part of the URL should be left out, I believe it can cause an issue if your primary Google account in that browser isn't the one with Firebase access.

jtreanor: Ah, thanks!

jtreanor: Updated and confirmed the change works.

aforcier:
@jtreanor I'm not sure if the reporting is working correctly, the content seems to be missing in the Slack message:

Screen Shot 2019-04-29 at 10 04 22 AM

Am I missing something?

jtreanor:
@aforcier I think that is because two messages were trigger to the same channel so the full detail isn't shown. I'll test again now to confirm.

jtreanor:
@aforcier It turns out I was mistaken. Messages weren't working on failure because the `Setup Slack messages` step was skipped. This is now fixed:

image

aforcier:
Working great! I updated the `SLACK_FAILURE_WEBHOOK` to point to a new webhook reporting to the right channel and tested it out. :shipit:

PR #1238, opened 2019-04-26
Woo: 3 of 3 - Add updating capabilities to order list management

AmandaRiu:
PR 3 of 3. This final PR adds the logic needed to fetch orders that have been updated since last downloaded.

In this gif, observe how the status of the second record updates from `processing` to `on-hold`


AmandaRiu:
Updated branch with all the AndroidX changes from the parent branch.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
```diff
@@ -789,6 +789,7 @@ private void migrateAddOn(@AddOn String addOnName, SQLiteDatabase db, int oldDbV
                      break;
                 case 68:
                     AppLog.d(T.DB, "Migrating addon " + addOnName + " to version " + (oldDbVersion + 1));
+                    db.execSQL("ALTER TABLE WCOrderModel ADD DATE_MODIFIED TEXT");
```
malinajirka: I assume you intentionally added this into an existing case as this feature hasn't been released yet, right?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -433,7 +434,9 @@ class WCOrderStore @Inject constructor(dispatcher: Dispatcher, private val wcOrd
     }
 
     private fun fetchOrdersByIds(payload: FetchOrdersByIdsPayload) {
-        wcOrderRestClient.fetchOrdersByIds(payload.site, payload.remoteIds)
+        payload.remoteIds.chunked(NUM_ORDERS_PER_FETCH).forEach { idsToFetch ->
```
malinajirka: TIL `chunked` ;)

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
AmandaRiu: That is correct!

AmandaRiu:
Thanks for the review @malinajirka !! Merging.

PR #1239, opened 2019-04-29
Check null response object for device register api & order stats fetc…

anitaa1990:
Crash fix for woocommerce/woocommerce-android#917 in WooCommerce.

There have been 3 reports of crashes in Woo app because of the response object being `null`. This PR catches that null response and returns an error response to the client and lets the client handle it by showing an error message to the user.

There are only 2 places in the app where the crashes have occurred so added the fix to only 2 methods:
1. **when fetching order stats** - used only by Woo and ~might~ does not need an additional PR to handle this error response since Woo is handling all error responses in a generic manner while also logging the error response.
2. **When registering fcm token** - this is called by both Woo and WP and might need subsequent PRs in WP to handle this error response.

### Tests
1. Run `testStatsFetchResponseNullError` in `MockedStack_WCStatsTest` class
2. Run `testRegistrationResponseNull` in `MockedStack_NotificationTest` class.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
@@ -74,21 +74,24 @@ class NotificationRestClient constructor(
         val url = WPCOMREST.devices.new_.urlV1
         val request = WPComGsonRequest.buildPostRequest(
                 url, params, RegisterDeviceRestResponse::class.java,
-                { response ->
-                    response?.let { response.id?.takeIf { it.isNotEmpty() }?.let {
-                        val payload = RegisterDeviceResponsePayload(it)
-                        dispatcher.dispatch(NotificationActionBuilder.newRegisteredDeviceAction(payload))
+                { response: RegisterDeviceRestResponse? ->
+                    response?.let {
+                        if (!it.id.isNullOrEmpty()) {
+                            val payload = RegisterDeviceResponsePayload(it.id)
+                            dispatcher.dispatch(NotificationActionBuilder.newRegisteredDeviceAction(payload))
+                        } else {
+                            val registrationError =
+                                    DeviceRegistrationError(DeviceRegistrationErrorType.MISSING_DEVICE_ID)
+                            val payload = RegisterDeviceResponsePayload(registrationError)
+                            dispatcher.dispatch(NotificationActionBuilder.newRegisteredDeviceAction(payload))
+                        }
                     } ?: run {
-                        val registrationError = DeviceRegistrationError(DeviceRegistrationErrorType.MISSING_DEVICE_ID)
-                        val payload = RegisterDeviceResponsePayload(registrationError)
-                        dispatcher.dispatch(NotificationActionBuilder.newRegisteredDeviceAction(payload))
-                    } } ?: run {
                         AppLog.e(T.API, "Response for url $url with param $params is null: $response")
-                        val registrationError = DeviceRegistrationError(DeviceRegistrationErrorType.INVALID_RESPONSE,
-                                "Response object is null")
+                        val registrationError = DeviceRegistrationError(DeviceRegistrationErrorType.INVALID_RESPONSE)
                         val payload = RegisterDeviceResponsePayload(registrationError)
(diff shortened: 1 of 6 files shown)


anitaa1990:
@AmandaRiu I checked WP and Woo and if a notification registration fails, we don't inform the user. [This](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/push/GCMRegistrationIntentService.java#L59) is the line in WP where the FCM registration takes place.

And [this](https://github.com/woocommerce/woocommerce-android/blob/develop/WooCommerce/src/main/kotlin/com/woocommerce/android/push/FCMRegistrationIntentService.kt#L67) is the line in Woo where the FCM registration takes place.

Just wondering if perhaps add logs to both apps that might help narrow down the root of the problem? Or is adding logs to FluxC enough to find the root cause?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
```diff
@@ -73,15 +75,20 @@ class NotificationRestClient constructor(
...
+                    response?.let { response.id?.takeIf { it.isNotEmpty() }?.let {
                         val payload = RegisterDeviceResponsePayload(it)
                         dispatcher.dispatch(NotificationActionBuilder.newRegisteredDeviceAction(payload))
                     } ?: run {
                         val registrationError = DeviceRegistrationError(DeviceRegistrationErrorType.MISSING_DEVICE_ID)
                         val payload = RegisterDeviceResponsePayload(registrationError)
                         dispatcher.dispatch(NotificationActionBuilder.newRegisteredDeviceAction(payload))
-                    }
-                },
+                    } } ?: run {
```
AmandaRiu: np: I love how concise this is but the multiple nested null checks is difficult to follow and could be error prone if someone needs to change it later. Can make this easier to read and follow? It may not be as pretty, but it will avoid confusion later I think.

aforcier: Agreed that this might be a little _too_ concise. 👍

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
```diff
@@ -73,15 +75,20 @@ class NotificationRestClient constructor(
...
+                        AppLog.e(T.API, "Response for url $url with param $params is null: $response")
+                        val registrationError = DeviceRegistrationError(DeviceRegistrationErrorType.RESPONSE_NULL,
```
aforcier: We typically have used `INVALID_RESPONSE` for this kind of error - it's not a big deal but it might be worth calling it that for consistency. See for example:

https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/39a08b3419f2fedfec741ea7fa7fc4dd7b3078ae/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooCommerceRestClient.kt#L114-L118

aforcier:
@anitaa1990 @AmandaRiu I gave this a look and it looks good with a couple of minor comments.

> I will see if someone in WPAndroid will also take a look at this since it does touch WP code as well.

I don't believe this has any effect on WPAndroid as we're not using `NotificationStore` there yet - is there something else you noticed that leads you to think this may break something?

AmandaRiu:
@aforcier 

> I don't believe this has any effect on WPAndroid as we're not using NotificationStore there yet - is there something else you noticed that leads you to think this may break something?

Nope, just wanted to reach out just in case. Thank you for reviewing this!

anitaa1990:
Thanks @aforcier and @AmandaRiu for reviewing the changes. 

> np: I love how concise this is but the multiple nested null checks is difficult to follow and could be error prone if someone needs to change it later. Can make this easier to read and follow? It may not be as pretty, but it will avoid confusion later I think.

> We typically have used INVALID_RESPONSE for this kind of error - it's not a big deal but it might be worth calling it that for consistency. See for example:

Agreed on both counts 👍I cleaned up the code a bit for better readability. @AmandaRiu it's good for another review 👍

PR #1240, opened 2019-04-29
Fix a couple of unescaped display names that might need to be HTML unescaped

maxme:
Related to this issue: https://github.com/wordpress-mobile/WordPress-Android/issues/9747

Note: I checked that `StringEscapeUtils.unescapeHtml4(null)` returns `null`.


malinajirka:
Thanks @maxme !

The changes look good to me. Should we update FluxC hash in WPAndroid?

maxme:
> The changes look good to me. Should we update FluxC hash in WPAndroid?

Yes, I'll check recent FluxC changes and create a PR for that.

PR #1241, opened 2019-04-30
Temporarily disable post list tests as they keep failing

malinajirka:
Temporarily disables PostList tests as they keep failing due to a race condition.

It seems they don't take into account following constraints
- there is no guarantee event bus events will be consumed in the same order as they where emitted when they are consumed on a background thread
- LiveData don't propagate all the events to their observables if more changes happen within a short timeframe (which is desired in the UI, but our tests don't take it into account - we wait for all the events, but some of them just never arrive)

These changes https://github.com/wordpress-mobile/WordPress-FluxC-Android/compare/do-not-merge/fix-post-list-tests?expand=1 fix the issues, but it's not something we can merge into develop. I pushed them just to gather ideas how to fix it properly.


aforcier:
Thanks @malinajirka! :shipit:

malinajirka:
While reviewing Woo changes, I've noticed only tests with `testMode = MultiplePages` are flaky.

PR #1243, opened 2019-05-01
Add PostStore.getLocalDrafts

shiki:
This adds a `PostStore.getLocalDrafts` method. This is a companion PR to wordpress-mobile/WordPress-Android#9774.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
@@ -422,6 +422,9 @@ public class PostStore extends Store {
         return getUploadedPagesForSite(site).size();
     }
 
+    /**
+     * Returns all posts and pages that are local drafts for the given site.
+     */
     public List<PostModel> getLocalDrafts(@NonNull SiteModel site) {
         return PostSqlUtils.getLocalDrafts(site.getId());
     }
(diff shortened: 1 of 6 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -422,6 +422,13 @@ public int getUploadedPagesCountForSite(SiteModel site) {
         return getUploadedPagesForSite(site).size();
     }
 
+    /**
+     * Returns all posts and pages that are local drafts for the given site.
+     */
+    public List<PostModel> getLocalDrafts(@NonNull SiteModel site) {
```
malinajirka: We might want to consider creating two separate methods for post and pages. I feel like the fact we use PostStore for both post and pages is already confusing. Getting both post and pages with `PostStore.getLocalDrafts()` might be confusing even to developers who are familiar with the PostStore, not mentioning others. Btw do we really want to upload even local page drafts as part of these two PRs? How do you feel about it?

Related comment: `PageStore` was introduced during Pages list refactoring. If you decide to create two separate methods, we might want to add the one for pages to the `PageStore.

shiki: I'm not ashamed to admit that I did not notice `PageStore`. 😄 Separating them makes sense in that case. 

> Btw do we really want to upload even local page drafts as part of these two PRs? 

I realized it was such a small change that I included it. But since we're going to exclude the Pages from this query, then it will not be included. I'll work on the Pages in a separate PR.

shiki: Split them up in 5aad0a29a0ae1df58cc4cb2df6fe40d878deb9d4 and ea5c2a122c07114aff2cde80a73cdb0d4bb7e721.

shiki:
@malinajirka This is now ready for another look. 🙂

malinajirka:
Thanks ;)! LGTM.

PR #1246, opened 2019-05-07
Fix failing insights test

planarvoid:
`testAddAndRemoveInsights` started failing because now we have 5 default Insights blocks, not 4. This PR is increasing the size to fix the test.


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ManageInsightsTestJetpack.kt
```diff
@@ -65,20 +65,20 @@ class ReleaseStack_ManageInsightsTestJetpack : ReleaseStack_Base() {
         val emptyStats = runBlocking { statsStore.getAddedInsights(site) }
 
         // Starts with 4 default blocks
```
malinajirka: This comment seems out-of-date now.

planarvoid:
@malinajirka thanks for the check, I've updated the comment

PR #1248, opened 2019-05-13
Add transaction to prevent other threads to overriding current account

planarvoid:
Fixes: https://github.com/wordpress-mobile/WordPress-Android/issues/8403
The crash is caused by multiple threads selecting and inserting data into the DB. I think in this case it's valid to have the select/insert in a transaction (I don't think it will change the performance much, both of the operations should be really fast). The transaction should block the access to the writeable database and should fix the crash in the issue. Let me know what you think!


aforcier:
Change makes sense 👍 Also ran the relevant tests and tested out login from WPAndroid with this branch, all looks good.

:shipit:

PR #1250, opened 2019-05-14
Added logic to fetch and cache a store's country name in WcSettingsModel

### LINKED ISSUE #1249: Woo: fetch and cache a store's country name
In reference to this woocommerce/woocommerce-android#1058 issue, we need to fetch the store's country information along with the site settings in `WCSiteSettings` model.

anitaa1990:
Fixes #1249 

This PR adds logic to fetch a store's country code and maps the country code to a readable country name.

### Changes
* Added a enum class `SiteCountry` to map the country code provided to a readable country name, in order to match it with the current list of order shipment tracking providers.
* Mocked tests have also been added


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSettingsModel.kt
```diff
@@ -18,4 +19,321 @@ data class WCSettingsModel(
             fun fromString(type: String?) = CurrencyPosition.reverseMap[type?.toUpperCase(Locale.US)] ?: LEFT
         }
     }
+
+    enum class SiteCountry(val countryName: String) {
```
nbradbury: How is this going to be used? I ask because I'm wondering if these strings will need to be localized.

anitaa1990: So when fetching a store's settings, we store the country code as well in `WcSettingsModel` in this format: `US:NY`. Then when fetching the country name for a store, we use this country code to fetch the corresponding country name i.e. United States. 

This country name will be fetched when displaying the list of shipment tracking providers to check if it matches any one of the country names in the provider list. If it does, then that country will be displayed at the top of the list.

Correct me if I am wrong, but if we need to localise this, then the mapping has to take place in Woo and not in FluxC?

nbradbury: Okay, so if the country names are for display, then we'll definitely them to be localized. I haven't tried this, but I'm wondering if we can rely on [getAvailableLocales](https://developer.android.com/reference/java/util/Locale.html#getAvailableLocales()) to give us a set of locales, and then use each locale's `getDisplayCountry()` method to get the country name.

anitaa1990: @nbradbury, removed the country mapping logic from FluxC. Will add it to Woo in a separate PR. PR ready for another review 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -141,7 +140,7 @@ class WooCommerceStore @Inject constructor(
      */
     fun getStoreCountry(site: SiteModel): String? {
```
nbradbury: This is minor, but how about we rename this to `getStoreCountryCode()` to better reflect what it does?

anitaa1990: 👍 Done in a88243d

nbradbury:
Looks great - nice work, @anitaa1990! :shipit:

PR #1256, opened 2019-05-16
Adds message when authentication fails when running WPCom tests

malinajirka:
Adds explicit message to facilitate debugging when WPCom tests fail due to invalid credentials in tests.properties.


malinajirka:
:( I'm so sorry... I had 3 commits in a branch and I felt like having 3 commits for such a simple change doesn't make any sense. So I created a new branch and copy-pasted the line - but I must have copy-pasted it by mistake into the wrong method:X. 

This is from local the AS history a minute before I pushed the changes 😞 
Screenshot 2019-05-17 at 09 22 34

Sorry :( - fixed in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1256/commits/a44cc96820f56dcd2789572c1b7da4f365849aae

PR #1269, opened 2019-05-22
Add test for page remote-auto-save

malinajirka:
I wanted to add support for "Page remote-auto-save" into PageStore. However, after looking at the code and it's usage I don't think it makes sense. All code related to uploading/fetching Page content/title etc is located in the PostStore so it makes sense to have there even the code for remote-auto-save. 
The Remote-Auto-Save action already works with Pages so I just added an integration test to make sure the API handles it correctly. 

To test
1. Run `testAutoSavePost` and `testAutoSavePage`

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
@@ -662,9 +662,26 @@ public class ReleaseStack_PostTestWPCom extends ReleaseStack_WPComBase {
 
     @Test
     public void testAutoSavePost() throws InterruptedException {
+        boolean testPage = false;
+        //noinspection ConstantConditions
+        testAutoSavePostOrPage(testPage);
+    }
+
+    @Test
+    public void testAutoSavePage() throws InterruptedException {
+        boolean testPage = true;
+        //noinspection ConstantConditions
+        testAutoSavePostOrPage(testPage);
+    }
+
+    private void testAutoSavePostOrPage(boolean testPage) throws InterruptedException {
         // Arrange
         PostModel post = createNewPost();
         setupPostAttributes(post);
+
+        if (testPage) {
+            post.setIsPage(true);
+        }
 
         post.setStatus(PostStatus.PUBLISHED.toString());
 
@@ -673,7 +690,7 @@ public class ReleaseStack_PostTestWPCom extends ReleaseStack_WPComBase {
         PostModel uploadedPost = mPostStore.getPostByLocalPostId(post.getId());
 
         // Act
-        uploadedPost.setContent("post content edited");
+        uploadedPost.setContent("content edited");
         remoteAutoSavePost(uploadedPost);
 
         // Assert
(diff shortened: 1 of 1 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
```diff
@@ -662,18 +662,35 @@ public void testRestoreRemotePost() throws InterruptedException {
 
     @Test
     public void testAutoSavePost() throws InterruptedException {
+        boolean testPage = false;
+        //noinspection ConstantConditions
```
shiki: Weird. The inspection looks like a bug. It looks like you can avoid the inspection by making `testPage` `final` though. 

```java
final boolean testPage = false;
```

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
```diff
@@ -662,18 +662,35 @@ public void testRestoreRemotePost() throws InterruptedException {
...
+        testAutoSavePostOrPage(testPage);
+    }
+
+    private void testAutoSavePostOrPage(boolean testPage) throws InterruptedException {
         // Arrange
         PostModel post = createNewPost();
         setupPostAttributes(post);
 
+        if (testPage) {
```
shiki: Passing `PostModel` to this method instead of `testPage` would be ideal so we have better control over what object is being uploaded. But considering this is pretty small, it's fine. My only request is we **don't trust** `createNewPost()` to always set `isPage` to `false`:

```java
    private PostModel createNewPost() {
        PostModel post = mPostStore.instantiatePostModel(sSite, false);
```

But instead enforce it ourselves:

```java
    private void testAutoSavePostOrPage(boolean isPage) throws InterruptedException {
        // Arrange
        PostModel post = createNewPost();
        setupPostAttributes(post);

        post.setIsPage(isPage); // <--- this
        post.setStatus(PostStatus.PUBLISHED.toString());
```

If somebody changes `createNewPost()` to set `isPage` to `true`, the `testAutoSavePost()` test will still pass but it will be testing the wrong scenario.

What say you? 😄

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
malinajirka: Fixed in d3d9f8b71ca0702edfa3580d6790f535a185bebc

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
malinajirka: >Passing PostModel to this method instead of testPage would be ideal so we have better control over what object is being uploaded.

I had it like that during a first iteration, but I think the current approach makes it pretty clear that the only thing that changes between the two tests is the "isPage" parameter. 

>My only request is we don't trust createNewPost() to always set isPage to false:

💯 :D I'm not sure why I added the `if` statement. Fixed in d3d9f8b71ca0702edfa3580d6790f535a185bebc

malinajirka:
Thanks @shiki! It's ready for another round;).

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
shiki: > So if it's ok with you, I'd rather keep it.

That's totally fine.

PR #1283, opened 2019-06-13
Distinguish 2FA auth errors when fetching account settings

aforcier:
Adds a distinct error type for the `reauthorization_required` API error, so we can identify it from the host app.

Part of the fix for https://github.com/wordpress-mobile/WordPress-Android/issues/8754. There will be a counterpart WPAndroid issue making use of the change.

We get this error for 2FA accounts when using a non-production WordPress.com OAuth client. Essentially, some APIs around account management are disabled in those cases for security reasons.

The error is a bit generic from the server-side - it essentially means the user isn't privileged to do the action (in this case, calling the `/me/settings/` endpoint) and needs to reauthorize. For bearer token-based login, there is no escalation of privileges possible, so the request just fails at that point. If using production credentials, this error should never be seen. This is pretty deeply embedded and I don't think it's worth changing server-side for a development-only issue.

### To test
Verify the connected account tests are still passing. There isn't otherwise a great way to test this in an automated way due to 2FA being required, but you can test this from the example app:
1. Generate an OAuth client id and secret for an account that has 2fa enabled ([instructions](https://github.com/wordpress-mobile/WordPress-Android/#oauth2-authentication)).
2. Update `wp.OAUTH.APP.ID` and `wp.OAUTH.APP.SECRET` in `example/gradle.properties` to those values.
3. Run the example app, and log into the account the credentials belong to.
4. Confirm that you see a `SETTINGS_FETCH_REAUTHORIZATION_REQUIRED_ERROR` in the log, but otherwise things are working.

(These steps are similar to the WPAndroid ones I'll be sharing in the counterpart PR.)

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
@@ -959,6 +959,11 @@ public class AccountStore extends Store {
 
             AccountErrorType errorType;
             if (payload.error.apiError.equals("reauthorization_required")) {
+                // This error will always occur for 2FA accounts when using a non-production WordPress.com OAuth client.
+                // Essentially, some APIs around account management are disabled in those cases for security reasons.
+                // The error is a bit generic from the server-side - it essentially means the user isn't privileged to
+                // do the action and needs to reauthorize. For bearer token-based login, there is no escalation of
+                // privileges possible, so the request just fails at that point.
                 errorType = AccountErrorType.SETTINGS_FETCH_REAUTHORIZATION_REQUIRED_ERROR;
             } else {
                 errorType = AccountErrorType.SETTINGS_FETCH_GENERIC_ERROR;
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
@@ -954,19 +954,15 @@ public class AccountStore extends Store {
             mAccount.copyAccountSettingsAttributes(payload.account);
             updateDefaultAccount(mAccount, AccountAction.FETCH_SETTINGS);
         } else {
-            if (payload.error != null) {
-                OnAccountChanged accountChanged = new OnAccountChanged();
-                AccountErrorType errorType;
-                if (payload.error.apiError.equals("reauthorization_required")) {
-                    errorType = AccountErrorType.SETTINGS_FETCH_REAUTHORIZATION_REQUIRED_ERROR;
-                } else {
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
```diff
@@ -946,20 +947,25 @@ private void handleFetchAccountCompleted(AccountRestPayload payload) {
...
         } else {
-            if (payload.error != null) {
-                OnAccountChanged accountChanged = new OnAccountChanged();
-                accountChanged.error = new AccountError(AccountErrorType.SETTINGS_FETCH_ERROR, payload.error.message);
-                emitChange(accountChanged);
+            OnAccountChanged accountChanged = new OnAccountChanged();
+            accountChanged.causeOfChange = AccountAction.FETCH_SETTINGS;
+
+            AccountErrorType errorType;
+            if (payload.error.apiError.equals("reauthorization_required")) {
```
shiki: On the previous version, we had a `null` check for `payload.error`, should we keep that in here?  Or are we sure that `payload.error` and `payload.error.apiError` will never be `null`?

shiki: Is it worth it to add a comment here for what "reauthorization_required" means? Your description of it in the PR description is really good.

aforcier: We should be okay to drop the `null` check for `payload.error`, as we already checked it - if you look a bit further up, we're calling `checkError()`, which is calling `payload.isError()`, which itself is just a `null` check for `payload.error`.

We can be pretty confident `apiError` is not null because of the way it's initialized and used:

https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/961609facf48a24a2cc8ca185d27fa5f499a47c9/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequest.java#L50-L53

https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/961609facf48a24a2cc8ca185d27fa5f499a47c9/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequest.java#L147-L176

True, it's not immutable, but in practice `WPComGsonRequest` is never modified once it's delivered to the individual network client functions - it's either passed up the chain like it is here, or interpreted into custom error classes.

aforcier: Probably a good idea 👍 Added in 0cc15b1.

shiki: Thank you for the explanation!

PR #1289, opened 2019-06-20
Add new field into MediaModel

malinajirka:
Corresponding WPAndroid PR - https://github.com/wordpress-mobile/WordPress-Android/pull/10079

This PR adds a new field into the MediaModel so we can differentiate regular media uploads from featured image uploads.


maxme:
I was wondering if we should also update the `MediaRestClient.pushMedia()` function, but since it won't be used in wpandroid now (we're only allowing local media to be added), I think it's fine keep it that way.

LGTM but a new conflicting DB version has been added in `develop`, can you merge it?

malinajirka:
Thanks for the review @maxme! I've just merged `develop` into the branch;).

maxme:
:shipit:

PR #1303, opened 2019-07-08
Add methods for getting locally changed Posts and Pages

shiki:
This is a dependency of https://github.com/wordpress-mobile/WordPress-Android/issues/10177. 

This adds the methods `PostStore.getLocallyChangedPosts` and `PageStore.getLocallyChangedPages`. 

## Testing 

- Please validate the unit and integration tests.
- Probably better to test this along with https://github.com/wordpress-mobile/WordPress-Android/issues/10177 too. 

## Reviewing 

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/page/PageStoreLocalChangesTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/page/PageStoreLocalChangesTest.kt
@@ -18,6 +18,7 @@ import org.wordpress.android.fluxc.persistence.PostSqlUtils
 import org.wordpress.android.fluxc.store.PageStore
 import org.wordpress.android.fluxc.test
 import java.util.UUID
+import kotlin.random.Random
 
 @RunWith(RobolectricTestRunner::class)
 class PageStoreLocalChangesTest {
@@ -50,16 +51,14 @@ class PageStoreLocalChangesTest {
 
         val baseTitle = "Voluptatem harum repellendus"
         val expectedPages = List(3) {
-            createLocalDraft(localSiteId = site.id, baseTitle = baseTitle).apply {
-                setIsPage(true)
-            }
+            createLocalDraft(localSiteId = site.id, baseTitle = baseTitle, isPage = true)
         }
 
         val unexpectedPosts = listOf(
                 // local draft post
-                createLocalDraft(localSiteId = site.id).apply { setIsPage(false) },
+                createLocalDraft(localSiteId = site.id, isPage = false),
                 // other site page
-                createLocalDraft(localSiteId = 4_000),
+                createLocalDraft(localSiteId = 4_000, isPage = true),
                 // uploaded page
                 PostModel().apply {
                     title = "Title"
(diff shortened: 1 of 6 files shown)


example/src/test/java/org/wordpress/android/fluxc/page/PageStoreLocalChangesTest.kt
```diff
@@ -0,0 +1,146 @@
+package org.wordpress.android.fluxc.page
```
shiki: This was renamed from `PageStoreLocalDraftTest`.

example/src/test/java/org/wordpress/android/fluxc/page/PageStoreLocalChangesTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        val localDraftPages = pageStore.getLocalDraftPages(site)
+
+        // Assert
+        assertThat(localDraftPages).hasSize(3)
+        assertThat(localDraftPages).allMatch { it.title.startsWith(baseTitle) }
+        assertThat(localDraftPages.map { it.id }).isEqualTo(expectedPages.map { it.id })
+    }
+
+    @Test
+    fun `getLocallyChangedPages returns local draft and locally changed pages only`() = test {
```
shiki: This is the new test that was added.

shiki:
Leaving this open even if https://github.com/wordpress-mobile/WordPress-Android/pull/10177 was canceled. We might need this in the near future.

malinajirka:
Adding "Do not merge" label as the whole "auto-upload/save" feature has been put on hold until we decide how to proceed.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java
```diff
@@ -147,6 +147,18 @@ public int insertOrUpdatePostOverwritingLocalChanges(PostModel post) {
                       .getAsModel();
     }
 
+    public List<PostModel> getLocallyChangedPosts(@NonNull Integer localSiteId, boolean isPage) {
+        return WellSql.select(PostModel.class)
+                      .where()
+                      .equals(PostModelTable.IS_PAGE, isPage)
+                      .equals(PostModelTable.LOCAL_SITE_ID, localSiteId)
+                      .beginGroup()
+                      .equals(PostModelTable.IS_LOCAL_DRAFT, true).or().equals(PostModelTable.IS_LOCALLY_CHANGED, true)
```
malinajirka: I think we try (in most cases at least) to differentiate between localDraft and locallyChanged post - `isLocalDraft = true && isLocallyChanged == true` should never be true. Therefore I believe having a method `getLocallyChangedPosts` which returns even `localDrafts` can be a bit misleading. I'd consider returning only `locallyChanged` post and using the `getLocalDrafts` method for local drafts. Another option which might be even better 🤷‍♂  is to rename the method to something else eg. "getNotSyncedPosts" or similar. Wdyt?

shiki: I thought that `getLocallyChangedPosts` made sense because these existing methods do similar query conditions:

https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/663ec407fbd32ef07ce959b12e7a06fbda7f4e80/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java#L241-L249

https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/663ec407fbd32ef07ce959b12e7a06fbda7f4e80/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java#L230-L239

As for the `getLocalDrafts`, I was planning on deleting that since it's no longer used. 

Your suggestion makes sense though and I agree. The question is, are we okay with having these slight differences with the existing methods (above)?

malinajirka: Ohh, good point. I'd consider renaming it to `getPostsWithLocalChanges`. It wouldn't contain `locallyChanged` in its name so it would be at least a bit less confusing. Wdyt?

shiki: @malinajirka Renamed in df6ad0999cf774d9f264ec8e3e68b79b660401c7.

PR #1309, opened 2019-07-17
Post scheduling reminder store

planarvoid:
This PR adds a store that keeps the information about the scheduling reminders set from Edit post settings. It lets you schedule a notification, delete all scheduled notifications and get all notifications from the DB. 

It should be tested with the related WPAndroid PR.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/store/PostSchedulingNotificationStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/PostSchedulingNotificationStoreTest.kt
@@ -34,8 +34,8 @@ class PostSchedulingNotificationStoreTest {
 
     @Test
     fun `schedule deletes previous notification and inserts update when set`() {
-        periodMappings.entries.forEach { (domainModel, dbModel) ->
-            store.schedule(postId, domainModel)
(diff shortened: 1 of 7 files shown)


example/src/test/java/org/wordpress/android/fluxc/store/PostSchedulingNotificationStoreTest.kt
```diff
@@ -0,0 +1,89 @@
...
+    )
+
+    @Before
+    fun setUp() {
+        store = PostSchedulingNotificationStore(sqlUtils)
+    }
+
+    @Test
+    fun `schedule deletes previous notification and inserts update when set`() {
+        periodMappings.entries.forEach { (domainModel, dbModel) ->
```
khaykov: Not really sure why the SchedulingReminderModel is named `domainModel` :)

planarvoid:
thanks for the review @khaykov !

> Also, what do you think about adding a test for SQL Utils? If something goes wrong with those, the store test will still pass, and we will be in the dark.

Good point! I'm not a big fan of mocking a database so I've created a release (integration) test for the store which tests both the store and the database.

Issue #1251, opened 2019-05-15
PostModel breaks the equals/hashcode contract


malinajirka:
PostModel.java overrides the equals method, but does not override the hashcode. This breaks the equals-hashcode contract and can results in hard to find bugs. 

Consider using EqualsVerifier or a similar library.

PR #1317, opened 2019-07-24
Add author into PostModel

### LINKED ISSUE #1316: Add author id and name into PostModel
Add three new fields into PostModel - see https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/%24post_ID/
- authorFirstName
- authorLastName
- authorId

Ignore author name fields on self-hosted sites as the API returns only the ID.

malinajirka:
Fixes #1316

This PR adds 3 new fields into the PostModel
- ~authorFirstName~
- ~authorLastName~
- authorDisplayName
- authorId

Notes:
- I considered removing all posts without localChanges (locallyChanged, localDrafts) in order to make sure we fetch the new empty fields. However, I decided it's not worth users' data and the risk of (not)removing something important (I wasn't sure if I need to clear some other dependant tables or not). So the fields will be filled only on new posts or posts which were recently modified. Let me know what you think about it.
- We fetch author fields only for WPCom sites. The core endpoint contains only authorId and fetching the additional fields from another endpoint isn't worth it.

To test:
https://github.com/wordpress-mobile/WordPress-Android/pull/10276

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
@@ -53,8 +53,7 @@ public class PostModel extends Payload<BaseNetworkError> implements Cloneable, I
     @Column private double mLongitude = PostLocation.INVALID_LONGITUDE;
 
     @Column private long mAuthorId;
-    @Column private String mAuthorFirstName;
-    @Column private String mAuthorLastName;
+    @Column private String mAuthorDisplayName;
 
     // Page specific
     @Column private boolean mIsPage;
@@ -290,20 +289,12 @@ public class PostModel extends Payload<BaseNetworkError> implements Cloneable, I
         this.mAuthorId = authorId;
     }
 
-    public String getAuthorFirstName() {
-        return mAuthorFirstName;
+    public String getAuthorDisplayName() {
+        return mAuthorDisplayName;
     }
 
-    public void setAuthorFirstName(String authorFirstName) {
-        this.mAuthorFirstName = authorFirstName;
-    }
-
-    public String getAuthorLastName() {
-        return mAuthorLastName;
-    }
-
-    public void setAuthorLastName(String authorLastName) {
-        this.mAuthorLastName = authorLastName;
+    public void setAuthorDisplayName(String authorDisplayName) {
+        mAuthorDisplayName = authorDisplayName;
     }
 
     public boolean isPage() {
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostWPComRestResponse.kt
```diff
@@ -46,4 +47,9 @@ data class PostWPComRestResponse(
         @SerializedName("edit_post") val editPost: Boolean = false,
         @SerializedName("delete_post") val deletePost: Boolean = false
     )
+    data class Author(
```
shiki: ℹ️  I have a concern about this. At first, I thought that there was just something wrong with what I'm doing because I couldn't get the name to show up on the Post List. Then I realized, my account does not have a first and last name. 🤦‍♂ 

Untitled Paw Document 2019-07-24 12-20-26

And then I found out that iOS probably uses a different method because my username is shown in the Post List. 

iPhone Xs — 12 2 2019-07-24 12-33-24

It looks like iOS just uses the `name` part of the `author` object.

https://github.com/wordpress-mobile/WordPressKit-iOS/blob/c8b0633282c8e2dc8923ec379472b10756700140/WordPressKit/PostServiceRemoteREST.m#L384-L390

Should we follow suit?

malinajirka: Nice find! You are right, I should be using `name`. I remember that it was initially my plan (4 months ago), but I forgot about it:X.

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -370,6 +398,7 @@ public boolean equals(Object other) {
         return getId() == otherPost.getId() && getLocalSiteId() == otherPost.getLocalSiteId()
                 && getRemoteSiteId() == otherPost.getRemoteSiteId() && getRemotePostId() == otherPost.getRemotePostId()
                 && getFeaturedImageId() == otherPost.getFeaturedImageId()
+                && getAuthorId() == otherPost.getAuthorId()
```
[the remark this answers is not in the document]
malinajirka: In general, when `equals` method changes, we should make sure to also update `hashcode`. Otherwise, some classes which rely on this contract might not work correctly - eg. HashMap. 

However, we weren't overriding hashcode method even before this change - I actually created an issue a few months ago https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1251.

Other than that, I don't think there is anything to worry about as long as this is true `If the author fields are not equal the objects are not equal`.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostWPComRestResponse.kt
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1317/commits/114cffafe0155a154c6d3f0f014432180a8cac2d

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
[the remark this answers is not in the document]
shiki: Thank you for the explanation! 🙂

PR #1320, opened 2019-07-26
Convert flaky fetch sites + logout test to mocked network

aforcier:
The test `testWPComSiteFetchAndLogoutCollision` started to occasionally fail on automated test runs.

This test logs in, dispatches an action to fetch the site list, and immediately logs the user out. This was written as a regression test a while back when we found that under these circumstances it was possible for the site list to be stored after the user had been logged out. (This no longer happens and the test validates that.)

It looks like the test was occasionally failing when the `SIGN_OUT` action got handled before the site fetch network request was assembled, clearing the token and causing the site fetch request to fail due to an authentication error.

Since the test doesn't need to be live, I moved it to the mocked network suite so that we're assured of a (mocked) successful response and the test can proceed.

### To test
1. Run the  tests in`MockedStack_SiteTest` and the `ReleaseStack_SiteTestWPCom`, make sure everything passes (you may get an unrelated failure in `ReleaseStack_SiteTestWPCom#testFetchSiteEditors` - that's a separate issue that's being addressed by @daniloercoli).
2. You can check that the test is still validating the right functionality by checking the logs. You should see this warning:
```
W/WordPress-DB: Can't insert WP.com site https://en.wordpress.com, missing user account
```


maxme:
I ran the following tests - all green:
```
./gradlew cAT -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.release.ReleaseStack_SiteTestWPCom
```

The new test also works fine:
```
./gradlew cAT -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.release.ReleaseStack_SiteTestWPCom#testWPComSiteFetchAndLogoutCollision
```

(got `Can't insert WP.com site https://en.wordpress.com, missing user account`).

:shipit:

PR #1321, opened 2019-07-26
Fix ReleaseStack_SiteTestWPCom#testFetchSiteEditors

daniloercoli:
The test `testFetchSiteEditors` started to fail on automated test runs, since the backend changed the default returned value recently.

This PR fixes the mobile editor preference test by including n the pool of the expected responses the new default from the server. 
Honestly this is not a great test anyway. We should change it in the future, or at least add another test that sets and verifies the new response.


aforcier:
Looks good! :shipit:

PR #1322, opened 2019-07-29
Fix for returning correct editor details when fetching site details

daniloercoli:
It seems to me that the code that fetches site details from the backend doesn't proper set the local ID of the [new SiteModel](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/7de37009030be38cd9e23072e0bc154f321de2cb/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java#L191) when the response is receive from the network, resulting in dispatching a `newUpdateSiteAction` with a SiteModel that doesn't have a local ID set. 

On the other side of the code, before emitting a `OnSiteChanged` event, we need to make sure the value for mobile_editor (and maybe later for web_editor) are copied from the DB to this new model. So the localID was necessary to correctly retrieve it from DB.

I'm not quite sure if we introduce the duplicate rows issue when added the editor items, but here https://github.com/wordpress-mobile/WordPress-FluxC-Android/compare/fix/return-fresh-mobile-editor-value-on-site-fetch?expand=1#diff-b53e7ac94134ff56178e112330b7464eR1536 it seems a new row was added instead of upgraded, and the host app receiving empty editor(s).

Edited: This PR does also write the new editor value to the DB before starting network calls. https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1322/commits/a8b6845345820da6082e381a48ae79d2b9366040

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
@@ -183,6 +183,7 @@ public class SiteRestClient extends BaseWPComRestClient {
                     public void onResponse(SiteWPComRestResponse response) {
                         if (response != null) {
                             SiteModel newSite = siteResponseToSiteModel(response);
+                            newSite.setId(site.getId());
                             mDispatcher.dispatch(SiteActionBuilder.newUpdateSiteAction(newSite));
                         } else {
                             AppLog.e(T.API, "Received empty response to /sites/$site/ for " + site.getUrl());
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
@@ -183,11 +183,6 @@ public class SiteRestClient extends BaseWPComRestClient {
                     public void onResponse(SiteWPComRestResponse response) {
                         if (response != null) {
                             SiteModel newSite = siteResponseToSiteModel(response);
-                            // The REST API doesn't return info about the editor(s). Make sure to copy old values
-                            // otherwise the apps will receive an update site without editor prefs set.
-                            // The apps will dispatch the action to update editor(s) when necessary.
-                            newSite.setMobileEditor(site.getMobileEditor());
-                            newSite.setWebEditor(site.getWebEditor());
                             mDispatcher.dispatch(SiteActionBuilder.newUpdateSiteAction(newSite));
                         } else {
                             AppLog.e(T.API, "Received empty response to /sites/$site/ for " + site.getUrl());
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
@@ -183,7 +183,11 @@ public class SiteRestClient extends BaseWPComRestClient {
                     public void onResponse(SiteWPComRestResponse response) {
                         if (response != null) {
                             SiteModel newSite = siteResponseToSiteModel(response);
-                            newSite.setId(site.getId());
+                            // local ID is not copied into the new model, let's make sure it is
+                            // otherwise the call that updates the DB can add a new row?
+                            if (site.getId() > 0) {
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -183,11 +183,11 @@ public void fetchSite(final SiteModel site) {
                     public void onResponse(SiteWPComRestResponse response) {
                         if (response != null) {
                             SiteModel newSite = siteResponseToSiteModel(response);
-                            // The REST API doesn't return info about the editor(s). Make sure to copy old values
-                            // otherwise the apps will receive an update site without editor prefs set.
```
hypest: Just to be sure @daniloercoli , can you elaborate on why we don't need to manually copy the editor value from `site` here anymore?

daniloercoli: We were using the value of `mobile_editor` read from the `SiteModel` passed to the request in the payload. (it's a fetch site details call). It can happen that another request updates the mobile_editor value right before this request finishes, and then the site details can contain a wrong value of mobile editor fields.

We're now using the fresh copy of the editor that's actually read from the DB here: https://github.com/wordpress-mobile/WordPress-Android/pull/10254/commits/74a3b822b48f36cb61875915380608661a6b72a5
just before emitting `OnSiteChanged`.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -1526,6 +1526,13 @@ private void updateSite(SiteModel siteModel) {
             event.error = SiteErrorUtils.genericToSiteError(siteModel.error);
         } else {
             try {
+                // The REST API doesn't return info about the editor(s). Make sure to copy current values
+                // available on the DB. Otherwise the apps will receive an update site without editor prefs set.
+                // The apps will dispatch the action to update editor(s) when necessary.
+                SiteModel freshSiteFromDB = getSiteByLocalId(siteModel.getId());
+                if (freshSiteFromDB != null) {
+                    siteModel.setMobileEditor(freshSiteFromDB.getMobileEditor());
```
hypest: WDYT about copying the web setting too, just so the two flags are consistent (not equal, just copied as a pair)?

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -1680,20 +1687,21 @@ private void fetchSiteEditors(SiteModel site) {
...
-                event.rowsAffected = SiteSqlUtils.insertOrUpdateSite(site);
-            } catch (Exception e) {
-                event.error = new SiteEditorsError(SiteEditorsErrorType.GENERIC_ERROR);
-            }
-            emitChange(event);
         }
+
+        // Update the editor pref on the DB, and emit the change immediately
+        SiteModel site = payload.site;
+        site.setMobileEditor(payload.editor);
```
hypest: Like above, WDYT about copying the web setting too, just so the two flags are consistent (not equal, just copied as a pair)?

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
daniloercoli: Fixed in 14afb86c00ee6883e4860551c82aec9f6eae1aa4

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
daniloercoli: This is the call made to upgrade the mobile editor setting.  Not sure we need to pass the web editor here and upgrade it to the DB.

If we want to change this, we need to patch also` wp-android` and make sure to not override the web setting.

hypest: Hmm, I see. the `payload.site` is actually getting mutated here and that's what will be sent in the network call, even though the schedule of the call happened on earlier lines of code. Tricky.

But yeah, it's fine as it is since we don't want to potentially modify the remote value of the web side flag. So, nothing to do here 👍.

daniloercoli:
Nice, will merge this, create the tag, and update `wp-android`. So other devs and testers can check the migration code.

PR #1327, opened 2019-07-30
Fix PostModel.dateLocallyChanged and Comment.datePublished timezone

### LINKED ISSUE #1326: PostModel.dateLocallyChanged and Comment.datePublished are using invalid timezone
The `DateTimeUtils.nowUTC()` doesn't work as expected and as a result we are setting a wrong date to `PostModel.dateLocallyChanged` and `Comment.datePublished`  More info https://github.com/wordpress-mobile/WordPress-Utils-Android/pull/25

malinajirka:
Fixes #1326

The `DateTimeUtils.nowUTC()` and `DateTimeUtils.localDateToUTC` don't work as expected and as a result we are setting a wrong date to PostModel.dateLocallyChanged and Comment.datePublished More info wordpress-mobile/WordPress-Utils-Android#25

This should be merged at the same time as https://github.com/wordpress-mobile/WordPress-Android/pull/10306

Note:
- I realize users already have the wrong dates stored in their local databases. I couldn't think of any scenario under which it might be a problem. Wdyt?

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
@@ -181,7 +181,7 @@ public class ReleaseStack_PostTestWPCom extends ReleaseStack_WPComBase {
 
         // Wait one sec
         Thread.sleep(1000);
-        Date testStartDate = DateTimeUtils.localDateToUTC(new Date());
+        Date testStartDate = new Date();
 
         // Check local change date is set and before "right now"
         assertNotNull(mPost.getDateLocallyChanged());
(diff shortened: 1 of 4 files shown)


maxme:
Looks good :shipit: 

> I realize users already have the wrong dates stored in their local databases. I couldn't think of any scenario under which it might be a problem. Wdyt?

- The datePublished field is overridden by the API on the POST comment endpoint, this field will be updated on comment list refresh
- For `PostModel.dateLocallyChanged` field, I check in FluxC and WPandroid. It's used for pending draft notifications. We're checking draft that haven't been published in +1 day/+1 week/+1 month and tell the user about it via a notification. This might delay or advance the first notification significantly (less significantly on the second and third one), since we picked an arbitrary timeframe in the first place, there is no problem IMO.

PR #1328, opened 2019-08-01
Feautre/introduce new fields into post model v2

malinajirka:
First task from https://github.com/wordpress-mobile/WordPress-Android/issues/10174

Introduces
_changesConfirmedContentHashcode_ - local field which will help us determine which locally modified posts can be safely auto-uploaded. 

The idea is simple - when the user clicks on a confirmation button (publish/schedule/update/save/sync) we store the current `contentHashcode()` of the post. When we upload the post we compare the current `contentHashcode()` with the stored hashcode and if they are equal we are 100% no-one edited the post since the user confirmed the action.

To test:
There isn't much to test. This PR simply adds a new field but the code doesn't use it yet.


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
```diff
@@ -554,6 +554,7 @@ public void onUpgrade(SQLiteDatabase db, WellTableManager helper, int oldVersion
                            + "TITLE TEXT,FORMATTABLE_BODY TEXT,FORMATTABLE_SUBJECT TEXT,FORMATTABLE_META TEXT)");
                 oldVersion++;
             case 74:
+                AppLog.d(T.DB, "Migrating to version " + (oldVersion + 1));
```
maxme: I noticed that one as well, added in the merge with our feature branch ;)

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -415,6 +431,36 @@ public boolean equals(Object other) {
                 && StringUtils.equals(getDateLocallyChanged(), otherPost.getDateLocallyChanged());
     }
 
+    public int contentHashcode() {
```
maxme: Can you add a comment explaining it's the standard implementation of `hashCode()` function, and that we can't use `Objects.hash()` because minSdk < 19?

I was wondering if we could reuse it in `equals()` so it would be a bit less prone to error when we add a new field "content" field in this class, but I fear that collisions might introduce weird problems in the app.

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -415,6 +431,36 @@ public boolean equals(Object other) {
...
+        result = 31 * result + (mPostFormat != null ? mPostFormat.hashCode() : 0);
+        result = 31 * result + (mSlug != null ? mSlug.hashCode() : 0);
+        temp = Double.doubleToLongBits(mLatitude);
+        result = 31 * result + (int) (temp ^ (temp >>> 32));
+        temp = Double.doubleToLongBits(mLongitude);
+        result = 31 * result + (int) (temp ^ (temp >>> 32));
+        result = 31 * result + (mIsPage ? 1 : 0);
+        result = 31 * result + (int) (mParentId ^ (mParentId >>> 32));
+        result = 31 * result + (mParentTitle != null ? mParentTitle.hashCode() : 0);
+        return result;
```
maxme: I think the Author Id is missing. I don't think we can change a post's author in the app, but in case we add that functionality, we should check it.

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
malinajirka: Fixed in f9e7fbae2479ad71bd2d1c7039f6f2bbbcc9444d

>I was wondering if we could reuse it in equals() so it would be a bit less prone to error when we add a new field "content" field in this class, but I fear that collisions might introduce weird problems in the app.

I don't think chances of a collision are real. However, I intentionally didn't override Object.hashcode() but created a separate method. Since Object.hashcode() should check all the fields in the PostModel, not just the ones which are editable/publishable by the user -> same applies to equals. So ideally I think we should also introduce generated hashcode method which will use the same fields as the equals method. So the equals-hashcode contract is followed. I actually submitted an issue a while back - #1251.

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
malinajirka: Fixed in f9e7fbae2479ad71bd2d1c7039f6f2bbbcc9444d

malinajirka:
Thanks @maxme. It's ready for another round.

PR #1329, opened 2019-08-01
Fix remote auto save issues

malinajirka:
Fixes a couple of issues with remote-auto-save feature.

1. Adds `REMOTE_AUTO_SAVED_POST` UploadAction - so we count number of attempts to autosave + we clear the upload model from the database.
2. Enables remote-auto-save endpoint for all PostTypes
3. Adds a workaround for an issue where s conflict was detected after we invoked remote-auto-save on a draft.: `When `/autosave` endpoint is invoked on a draft, the server automatically updates the post content and clears autosave object instead of just updating the autosave object.`
4. Removes auto-save object conflict detection -> We decided not to handle auto-save conflicts in https://github.com/wordpress-mobile/WordPress-Android/issues/10008#issuecomment-511799037.
5. Emits `newRemoteAutoSavedPostAction` with error when remote-auto-save invoked on a self-hosted site instead of throwing a RunTimeException.

~I'll turn this into a real PR when:~(I unintentionally created a real PR instead of a draft :( )
 This will be ready for review and merge when: 
- ~https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1328 is reviewed and merged~
- ~develop is merged into `master-remote-auto-save`~
- a PR is created in WPAndroid so the changes can be properly tested


fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -974,7 +974,12 @@ private void remoteAutoSavePost(RemotePostPayload payload) {
         if (payload.site.isUsingWpComRestApi()) {
             mPostRestClient.remoteAutoSavePost(payload.post, payload.site);
         } else {
-            throw new NotImplementedException("RemoteAutoSave is not supported in XML-RPC api.");
+            PostError postError = new PostError(
```
malinajirka: Log this into Sentry in WPAndroid.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -704,14 +703,25 @@ private void handleFetchedPostList(FetchPostListResponsePayload payload) {
...
+                // Check if the post's last modified date or status have changed.
                 // We need to check status separately because when a scheduled post is published, its modified date
                 // will not be updated.
                 boolean isPostChanged =
                         !post.getLastModified().equals(item.lastModified)
-                        || !post.getStatus().equals(item.status)
-                        || isAutoSaveChanged;
+                        || !post.getStatus().equals(item.status);
+
+                /*
```
shiki: Can we also explain the effect if we don't do this? For example, it will cause a false-positive conflict in the UI.

malinajirka: Good idea;)

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
malinajirka: Fixed

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
malinajirka: Fixed in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1329/commits/b5c2d42cc8e55779690ab3f40fdc75c47efdef3f

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
maxme: 👍

maxme:
:shipit:

PR #1332, opened 2019-08-05
Add the ability to set the mobile editor preference on all sites with one call 

daniloercoli:
This PR adds the ability to bulk set the mobile editor preference on all sites with just one network call to a new WordPress.com REST API Endpoint (`https://public-api.wordpress.com/wpcom/v2/me/gutenberg/`). 

The endpoint is not deployed yet, and the response format is still in brainstorming, but we can start coding the request/response flow.

wp-android side PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10332

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
@@ -1742,7 +1742,7 @@ public class SiteStore extends Store {
                 wpcomPostRequestRequired = true;
             }
             try {
-                rowsAffected = SiteSqlUtils.insertOrUpdateSite(site);
+                rowsAffected += SiteSqlUtils.insertOrUpdateSite(site);
             } catch (Exception e) {
                 event.error = new SiteEditorsError(SiteEditorsErrorType.GENERIC_ERROR);
             }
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -1723,6 +1777,35 @@ private void updateSiteEditors(FetchedEditorsPayload payload) {
...
+        OnAllSitesMobileEditorChanged event = new OnAllSitesMobileEditorChanged();
+        if (payload.isError()) {
+            event.error = payload.error;
+        } else {
+            // Loop over the returned sites and make sure we've the fresh values for editor prop stored locally
+            for (Map.Entry<String, String> entry : payload.editors.entrySet()) {
+                SiteModel currentModel = getSiteBySiteId(Long.parseLong(entry.getKey()));
+
+                if (currentModel == null) {
+                    // this should never happen normally
```
malinajirka: We might want to consider either 
- throwing an exception
- letting the app crash on NPE
- logging a message

 Otherwise, the debugging might be a nightmare. Wdyt?

daniloercoli: Nice call, I've added a log to make sure to keep track the edge case where the remote server returns info for a site, but the site is still not in the app site store. (created/added in another app or on the web). 4fb66b46ff04d02535f6e8a329c788782b54cb9b

PR #1336, opened 2019-08-06
Issue/1335 v4 visitor stats changes

### LINKED ISSUE #1335: Woo: add changes to visitor stats needed for the new v4 stats improvements
When testing the scrubbing interaction for v4 stats, I noticed that the current implementation of fetching visitor stats based on the `StatsGranularity` would need to be modified in order to support the new v4 stats UI changes.

Current implementation is:
- When `DAYS` is selected, the `OrderStatsApiUnit` field passed to the visitor stats api is `DAY`
- When `WEEKS` is selected, the `OrderStatsApiUnit` field passed to the visitor stats api is `WEEK`
- When `MONTHS` is selected, the `OrderStatsApiUnit` field passed to the visitor stats api is `MONTH`
- When `YEARS` is selected, the `OrderStatsApiUnit` field passed to the visitor stats api is `YEAR`

Based on the design changes for the new v4 stats improvements [here](https://github.com/woocommerce/woocommerce-android/issues/1199), we would need to get the data in `DAYS` for the current week and current month. So the new implementation should support the following changes:
- `Today` tab is selected: `OrderStatsApiUnit` field passed to the visitor API should be `DAY`
- `This week` tab is selected: `OrderStatsApiUnit` field passed to the visitor API should be `DAY`
- `This month` tab is selected: `OrderStatsApiUnit` field passed to the visitor API should be `DAY`
- `This year` tab is selected: `OrderStatsApiUnit` field passed to the visitor API should be `MONTH`.

Since we are supporting both version of stats endpoints, the changes in this issue addresses only the v4 stats changes. This requires:
- [ ] Creating new `WcStatsAction` for fetching visitor stats with a different `StatsGranularity` and handling visitor stats response
- [ ] Creating new model class and table to store the visitor stats for v4 separately.
- [ ] Creating a new button in the example app to get visitor stats data v4 separately.

anitaa1990:
Fixes #1335. This PR adds logic to get the visitor stats data for the new date ranges as defined in [here](https://github.com/woocommerce/woocommerce-android/issues/1199). We would use the existing API endpoint `/rest/v1.1/sites/{{site_id}}/stats/visits`. But as described [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1335), the `quantity`, `unit` and `date` fields would need to be modified to satisfy the new v4 stats UI.

Since we will be supporting both stats UI for the foreseeable future, I thought it made sense to implement the v4 changes for the visitor stats separately for better code readability and maintenance.

### Changes:
- Created a new table and model class - `WCNewVisitorStatsModel`. - The only difference between this model class and the `WcVisitorStatsModel` class is that we are storing the `StatsGranularity` to the `WCNewVisitorStatsModel` instead of `OrderStatsApiUnit`. This is because the current visitor stats implementation fetches the visitor data by `WEEKS` or `MONTHS`. In the new stats UI, we would need to fetch the visitor stats in `DAYS` for the current week or month. 
- Adds an action to `WCStatsAction`: `FETCH_NEW_VISITOR_STATS` and `FETCHED_NEW_VISITOR_STATS`  - for fetching visitor stats with a different `StatsGranularity` and `quantity` and handling visitor stats response. 
- Adds two payload classes to `WcStatsStore`: `FetchNewVisitorStatsPayload` and `FetchNewVisitorStatsResponsePayload`. 
- Adds a new method in `OrderStatsRestClient`: `fetchNewVisitorStats` to store the response data separately. This will be used by the new stats UI.
- Adds new buttons to the `WooRevenueStatsFragment` to test the visitor stats api for the current day, week, month and year and verify that the data is consistent with the date range.
- Added mocked tests to `MockedStack_WCStatsTest`
- Keeping in mind @nbradbury 's suggestion [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1325#pullrequestreview-268432017), created a new `StoreSelectorDialog` that displays only WooCommerce sites :)  

### Testing:
- Run all tests in `MockedStack_WCStatsTest`
- Run all tests in `WCStatsStoreTest`
- Run all tests in `WCVisitorStatsSqlUtilsTest`.
- Verify that the visitor stats fields are consistent with the `StatsGranularity` selected. Also note that the size of the visitors data map fetched from the local cache should match the `quantity` field:

| `TAB`  | `quantity` field | `date` field | `unit` field |
| ------------- | ---------------- | ---------------- | ---------------- |
| TODAY  | 1  | current date i.e. `2019-08-06` |  `OrderStatsApiUnit.DAY`  |
| This WEEK  | 3  | current date i.e. `2019-08-06` |  `OrderStatsApiUnit.DAY` (since we need the data in days i.e. for each day from the first of this week till the current date)  |
| This MONTH  | 6  | current date i.e. `2019-08-06` |  `OrderStatsApiUnit.DAY` (since we need the data in days i.e. for each day from the first of this month till the current date)  |
| This YEAR  | 8  | current date i.e. `2019-08-06` |  `OrderStatsApiUnit.MONTH` (since we need the data in months) - The API currently responds with 9 quantities, even though we have specified it as 8.  |

- Verify that the current implementation of visitor stats is working correctly, as a sanity check.
- Verify that upgrading to the changes in this PR does not lead to a crash.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCStatsTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCStatsTest.kt
@@ -18,6 +18,7 @@ import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.module.ResponseMockingInterceptor
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.orderstats.OrderStatsRestClient
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.orderstats.OrderStatsRestClient.OrderStatsApiUnit
+import org.wordpress.android.fluxc.store.WCStatsStore.FetchNewVisitorStatsResponsePayload
 import org.wordpress.android.fluxc.store.WCStatsStore.FetchOrderStatsResponsePayload
 import org.wordpress.android.fluxc.store.WCStatsStore.FetchRevenueStatsAvailabilityResponsePayload
 import org.wordpress.android.fluxc.store.WCStatsStore.FetchRevenueStatsResponsePayload
@@ -278,6 +279,54 @@ class MockedStack_WCStatsTest : MockedStack_Base() {
             assertNotNull(error)
             assertEquals(siteModel, site)
             assertEquals(OrderStatsApiUnit.MONTH, apiUnit)
+            assertNull(stats)
+            assertEquals(OrderStatsErrorType.INVALID_PARAM, error.type)
+        }
+    }
+
+    @Test
+    fun testNewVisitorStatsSuccess() {
+        interceptor.respondWith("wc-visitor-stats-response-success.json")
+        orderStatsRestClient.fetchNewVisitorStats(
+                siteModel, OrderStatsApiUnit.MONTH, StatsGranularity.YEARS, "2019-08-06", 8, true
+        )
+
+        countDownLatch = CountDownLatch(1)
+        assertTrue(countDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS.toLong(), MILLISECONDS))
+
+        assertEquals(WCStatsAction.FETCHED_NEW_VISITOR_STATS, lastAction!!.type)
+        val payload = lastAction!!.payload as FetchNewVisitorStatsResponsePayload
+        with(payload) {
+            assertNull(error)
+            assertEquals(siteModel, site)
+            assertEquals(StatsGranularity.YEARS, granularity)
+            assertNotNull(stats)
+            assertNotNull(stats?.data)
+            assertEquals(stats?.dataList?.size, 12)
+        }
+    }
+
+    @Test
+    fun testNewVisitorStatsError() {
+        val errorJson = JsonObject().apply {
+            addProperty("error", "rest_invalid_param")
+            addProperty("message", "Invalid parameter(s): date")
+        }
+
+        interceptor.respondWithError(errorJson)
+        orderStatsRestClient.fetchNewVisitorStats(
+                siteModel, OrderStatsApiUnit.MONTH, StatsGranularity.YEARS, "2019-08-06", 8, true
+        )
+
+        countDownLatch = CountDownLatch(1)
+        assertTrue(countDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS.toLong(), TimeUnit.MILLISECONDS))
+
+        assertEquals(WCStatsAction.FETCHED_NEW_VISITOR_STATS, lastAction!!.type)
+        val payload = lastAction!!.payload as FetchNewVisitorStatsResponsePayload
+        with(payload) {
+            assertNotNull(error)
+            assertEquals(siteModel, site)
+            assertEquals(StatsGranularity.YEARS, granularity)
             assertNull(stats)
             assertEquals(OrderStatsErrorType.INVALID_PARAM, error.type)
         }
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectorDialog.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectorDialog.kt
@@ -1,0 +1,83 @@
+package org.wordpress.android.fluxc.example.ui
+
+import android.app.AlertDialog
+import android.app.Dialog
+import android.content.Context
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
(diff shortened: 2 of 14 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCStatsTest.kt
```diff
@@ -283,6 +284,54 @@ class MockedStack_WCStatsTest : MockedStack_Base() {
         }
     }
 
+    @Test
+    fun testNewVisitorStatsSuccess() {
```
nbradbury: May I suggest we use `v4` instead of `new` throughout this PR? At some point these won't be new endpoints and will be replaced with even newer ones, which makes this terminology tricky.

example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectorDialog.kt
```diff
@@ -0,0 +1,83 @@
...
+import android.widget.ArrayAdapter
+import android.widget.TextView
+import androidx.fragment.app.DialogFragment
+import dagger.android.support.AndroidSupportInjection
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.store.WooCommerceStore
+import javax.inject.Inject
+
+class StoreSelectorDialog : DialogFragment() {
```
nbradbury: Thank you for adding this!

nbradbury:
In the PR description it says the interval (unit) for getting today's stats should be `OrderStatsApiUnit.DAY` but shouldn't that be `OrderStatsApiUnit.HOUR`? That's what's being sent in the request so I'm assuming that was just a typo.

example/src/main/java/org/wordpress/android/fluxc/example/ui/stats/WooRevenueStatsFragment.kt
```diff
@@ -176,5 +244,9 @@ class WooRevenueStatsFragment : Fragment() {
         fetch_current_month_revenue_stats_forced.isEnabled = enabled
         fetch_current_year_revenue_stats.isEnabled = enabled
         fetch_current_year_revenue_stats_forced.isEnabled = enabled
+        fetch_current_day_visitor_stats.isEnabled = enabled
```
nbradbury: Just a suggestion - feel free to ignore - but what if we named the linear layout which contains the buttons to `buttonContainer` and then do this:

```
private fun toggleSiteDependentButtons(enabled: Boolean) {
    for (i in 0 until buttonContainer.childCount) {
        val child = buttonContainer.getChildAt(i)
        if (child is Button) {
            child.isEnabled = enabled
        }
    }
}
```

That way we don't have to keep adding buttons to this code.

nbradbury:
@anitaa1990 This looks great! I left a few minor comments but I'm done with this first round.

example/src/main/java/org/wordpress/android/fluxc/example/ui/stats/WooRevenueStatsFragment.kt
anitaa1990: Nice idea!! 💯 addressed (by directly stealing your code) in d11fab0 😄

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCStatsTest.kt
anitaa1990: I thought of using v4 vs new but as you mentioned if we at some future point change the endpoints from v4 to v5, it wouldn't make sense to keep these named v4. To be honest, I couldn't think of anything other than these two 😕and I defer to you and @AmandaRiu on the naming :)

anitaa1990:
Thanks for the quick review @nbradbury!!

> In the PR description it says the interval (unit) for getting today's stats should be `OrderStatsApiUnit.DAY` but shouldn't that be `OrderStatsApiUnit.HOUR`? That's what's being sent in the request so I'm assuming that was just a typo.

The visitor stats API does not support the unit `HOUR`. So for the day stats, we would fetch the data with unit as `OrderStatsApiUnit.DAY` with the `quantity` as 1. Umm, I'm not able to find the visitor stats API request that sends the `unit` as `OrderStatsApiUnit.HOUR` from testing the example app though. Could you provide the steps to test this?

nbradbury:
> Umm, I'm not able to find the visitor stats API request that sends the `unit` as `OrderStatsApiUnit.HOUR`

Apologies, I was looking at the request to fetch revenue stats. Nevermind :)

PR #1343, opened 2019-08-10
Woo: Add support for product reviews

### LINKED ISSUE #1334: Woo: Add support for product reviews endpoint
Add the WooCommerce [product review endpoints](https://woocommerce.github.io/woocommerce-rest-api-docs/#product-reviews) for:

- Fetch product reviews list
- Fetch single product review
- Update (moderate) product review
- Delete product review

AmandaRiu:
This PR fixes #1334  by adding support for the Woo Product Reviews API, with the following actions:
- Fetch product reviews for site w/ available filters:
  - reviews matching a list of product id's
  - reviews matching a list of review statuses
- Fetch a single product review by remote id
- Update the status of a product review (for moderation)

NOTE: This PR includes a couple additions to the `tests.properties` file:
`TEST_WC_PRODUCT_REVIEW_ID = FIXME`
`TEST_WC_PRODUCT_WITH_REVIEWS_ID = FIXME`
These have been added to the `test.properties-example` file, as well as to the mobile secrets repository :) Also added to the `tests.properties.enc` file to run CircleCI tests.

Also included:
- Unit tests: `ProductSqlUtilsTest` and `WCProductReviewModelTest`
- Instrumented Tests: `ReleaseStack_WCProductTest` and `MockedStack_WCProductsTest`
- Example app tests:

## Recommended Test Scenarios

- Upgrade from previous version - make sure the product review table is properly created
- Run unit tests (ProductSqlUtilsTest, WCProductReviewModelTest)
- Run instrumented tests (MockedStack_WCProductTests, ReleaseStack_WCProductTests)
- Play with the new buttons in the example app

**Only 1 developer is needed to approve this PR**


anitaa1990:
Update: the CircleCI tests are passing now! PR is ready for review. 

I added the `Not Ready for Merge` label though because one of the tests in the `ReleaseStack_WCProductTest` is failing with the following exception, when trying to insert product reviews for a site:
`android.database.sqlite.SQLiteConstraintException: FOREIGN KEY constraint failed (code 787 SQLITE_CONSTRAINT_FOREIGNKEY)`.`

I spent an inordinate amount on time trying to fix ☝️test. It runs successfully when running the test individually but fails when running the whole class. Removing the `RawConstraints` from the `WcProductReviewModel` from [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/8f889d2d9a1f301ff629c5f4845a80bda3b1c0f3/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductReviewModel.kt#L15) fixes the test but I’m not sure if this is a valid solution. 

Hoping one of our developers can lend a hand to fix this :)  

Cc @nbradbury and @0nko for your inputs on this please 🙏

nbradbury:
@anitaa1990 The problem is that when running all tests, something is removing the site that `testFetchProductReviews` relies on from the database. This results in the constraint exception because the review's `LOCAL_SITE_ID` doesn't exist in the site table.

You can verify this by adding this to the start of that test:

```
val site = siteStore.getSiteBySiteId(sSite.siteId)
assertNotNull("site is null", site)
```

When you run that single test it passes, but if you run all the tests you get the "site is null" exception.

anitaa1990:
> The problem is that when running all tests, something is removing the site that testFetchProductReviews relies on from the database. 

Thank you @nbradbury!! I spent a lot of time last night and figured this out by checking the `Resources` tab in Sethos. The `SiteModel` table is empty 🤦‍♀. Turns out we reset the db before starting each test [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/8954d4098e20b1fc8ecff0a2fb76a0ffb0b62676/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_Base.java#L53). But there is logic in place to authenticate the user account and fetch the resulting sites only if the `sSite` is not previously authenticated [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/8954d4098e20b1fc8ecff0a2fb76a0ffb0b62676/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WPComBase.java#L78). I should have seen this before 🤦‍♀. 

I added logic to not reset the db before each test, just for this class: `ReleaseStack_WCProductTest` and it's working now 🎉

anitaa1990:
PR ready for review 👍

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCProductTest.kt
```diff
@@ -40,11 +45,14 @@ class ReleaseStack_WCProductTest : ReleaseStack_WCBase() {
         remoteProductId = BuildConfig.TEST_WC_PRODUCT_WITH_VARIATIONS_ID.toLong()
         dateCreated = "2018-04-20T15:45:14Z"
     }
+    private val remoteProductWithReviewsId = BuildConfig.TEST_WC_PRODUCT_WITH_REVIEWS_ID.toLong()
```
nbradbury: This variable and the related `BuildConfig` value appear to be unused.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -129,4 +132,81 @@ object ProductSqlUtils {
...
+    }
+
+    fun insertOrUpdateProductReview(productReview: WCProductReviewModel): Int {
+        val result = WellSql.select(WCProductReviewModel::class.java)
+                .where().beginGroup()
+                .equals(WCProductReviewModelTable.ID, productReview.id)
+                .or()
+                .beginGroup()
+                .equals(WCProductReviewModelTable.REMOTE_PRODUCT_REVIEW_ID, productReview.remoteProductReviewId)
+                .equals(WCProductReviewModelTable.LOCAL_SITE_ID, productReview.localSiteId)
```
nbradbury: I think this should include `REMOTE_PRODUCT_ID` since it's part of the unique key defined [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1343/files#diff-7b4c47eb749776091657a82024f162ceR922).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -129,4 +132,81 @@ object ProductSqlUtils {
...
+    }
+
+    fun getProductReviewByRemoteId(
+        localSiteId: Int,
+        remoteReviewId: Long
+    ): WCProductReviewModel? {
+        return WellSql.select(WCProductReviewModel::class.java)
+                .where()
+                .beginGroup()
+                .equals(WCProductReviewModelTable.LOCAL_SITE_ID, localSiteId)
```
nbradbury: Same comment here - this should include `REMOTE_PRODUCT_ID`.

nbradbury:
@anitaa1990 I'm finished with my first round!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
anitaa1990: Goof point! I was under the impression that the `PRODUCT_REVIEW_ID` would be unique across all products but I could be wrong on that. I've added the fix in 2d7698d

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
anitaa1990: I've added the fix in 2d7698d

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCProductTest.kt
anitaa1990: 👍 Done in 2d7698d

anitaa1990:
Thank you @nbradbury!! I've addressed the code changes and also added an option to select a site before testing the products in 4fa0f96 (since I couldn't find any products/reviews for the first site that was selected by default).

PR ready for another round 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -129,4 +132,84 @@ object ProductSqlUtils {
...
+            1
+        } else {
+            // Update
+            val oldId = result.id
+            WellSql.update(WCProductReviewModel::class.java).whereId(oldId)
+                    .put(productReview, UpdateAllExceptId(WCProductReviewModel::class.java)).execute()
+        }
+    }
+
+    fun getProductReviewByRemoteId(
```
nbradbury: This is really nitpicky so apologies in advance :) The order of the params here seems off to me - I think it should be:

```
fun getProductReviewByRemoteId(
        localSiteId: Int,
        remoteProductId: Long,
        remoteReviewId: Long
```

Does that make sense?

anitaa1990: :) No worries 😄 That makes sense! I had to verify a couple of times to check that I hadn't implemented it that way. I thought I _had_ added the `product_id` as the second param 🤦‍♀. Fixed in f5c601a

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
AmandaRiu: FYI, product review IDs are unique across sites. I thought the same as both of you when first writing these changes which is why you were seeing the unused `BuildConfig.TEST_WC_PRODUCT_WITH_REVIEWS_ID` :)

PR #1347, opened 2019-08-16
Add generated serialVersionUID to all serializable classes

malinajirka:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/5677

This PR adds generated `serialVersionUID` to all classes marked as `Serializable`. This is a best-practice recommended in the official docs `However, it is strongly recommended that all serializable classes explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations, and can thus result in unexpected InvalidClassExceptions during deserialization.`. https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Serializable.html

It should also fix [this](https://github.com/wordpress-mobile/WordPress-Android/issues/5677) rare crash (InvalidClassException).


malinajirka:
@maxme @hypest @aforcier The crash has very low impact ~8users in 90 days. Should we just close this PR without merging? Wdyt?

hypest:
Thanks for the bump @malinajirka !

I think this should be trivial to merge actually, sorry it took us that long to circle back to it.

I wonder though, should we also update it to cover more classes? For example, I see that the `fluxc/model/AccountModel` doesn't have a `serialVersionUID` set. WDYT?

malinajirka:
Thanks for looking into it @hypest! No worries at all ;)

>I wonder though, should we also update it to cover more classes? For example, I see that the fluxc/model/AccountModel doesn't have a serialVersionUID set. WDYT?

I believe we should add it only to classes which implement `Serializable` interface. `AccountModel` doesn't implement that interface so I didn't add it there. Wdyt?

PR #1349, opened 2019-08-16
Clean up after post integration tests

malinajirka:
Remove posts and pages created during integration tests.

To test:
1. Run the tests and verify all posts&pages created during the tests were permanently deleted. 

SideNote:
`ReleaseStack_MediaTestXMLRPC.testUploadMultipleImagesAndCancel()` keeps failing locally for me. Can you please check if it works with your setup. I'll look into it next week.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestWPCom.java
@@ -116,6 +116,9 @@ public class ReleaseStack_PostTestWPCom extends ReleaseStack_WPComBase {
 
         // The site should automatically assign the post the default category
         assertFalse(uploadedPost.getCategoryIdList().isEmpty());
+
+        // Clean up
+        permanentlyDeletePost(uploadedPost);
     }
 
     @Test
@@ -148,6 +151,9 @@ public class ReleaseStack_PostTestWPCom extends ReleaseStack_WPComBase {
         // The date created should not have been altered by the edits
         assertFalse(finalPost.getDateCreated().isEmpty());
         assertEquals(dateCreated, finalPost.getDateCreated());
+
+        // Clean up
+        permanentlyDeletePost(uploadedPost);
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


aforcier:
@malinajirka I haven't dived in too deeply yet, but I wanted to raise a concern/ask for some clarification.

I'm a bit concerned that adding a post deletion + verification step to every post test is adding a lot of extra time to run the connected tests overall without being vital. These tests aren't uploading any media so they shouldn't affect the storage capacity of a test site (if any of them are uploading media we should delete the media item).

Can you share a bit more on your reasoning/concerns here?

aforcier:
> ReleaseStack_MediaTestXMLRPC.testUploadMultipleImagesAndCancel() keeps failing locally for me. Can you please check if it works with your setup. I'll look into it next week.

This is passing for me. I have some stashed changes I've been meaning to get back to which converts these tests to mocked network, but they've been stable on CI for a long time so it hasn't been a priority.

As for why they're failing for you - one guess is that you're supplying your own local test image instead of using the one bundled with the tests (see `TestUtils#getSampleImagePath()` and `BuildConfig.TEST_LOCAL_IMAGE`) and that it's a really small file. That might mean the uploads are completing before they're cancelled and the test is failing. I would try unsetting `TEST_LOCAL_IMAGE` first to see if they pass then.

Assuming this is the problem, I think I'll find my changes and open a PR next week finally moving these multi-upload+cancel tests to the mocked set.

malinajirka:
>I'm a bit concerned that adding a post deletion + verification step to every post test is adding a lot of extra time to run the connected tests overall without being vital. 

You are completely right and I shouldn't work on Friday's after 4pm 😞 

>As for why they're failing for you - one guess is that you're supplying your own local test image instead of using the one bundled with the tests 

It doesn't seem to be the case. I even tried to slow down emulator's connection and it didn't help. However, thanks for the tip. I'll leave it for now since it's working for you and on CircleCI.

aforcier:
Just noting that I got around to the changes I mentioned above about moving some of the flaky upload tests to the mocked stack (they weren't nearly as close to complete as I remembered): 

https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1462

Hoping to finish the last few pieces and open the PR properly in the next few weeks.

PR #1351, opened 2019-08-18
Add usernameCanBeChanged to AccountModel

### LINKED ISSUE #1346: Add usernameCanBeChanged to AccountModel
To support the implementation of [changing usernames from the settings screen](https://github.com/wordpress-mobile/WordPress-Android/issues/8002) in the Android app a new column `usernameCanBeChanged` will be added to the `AccountModel`. 
The API property `user_login_can_be_changed` lives at the endpoint [me/settings](https://developer.wordpress.com/docs/api/1.1/get/me/settings/) and it will be used to provide the data for the `AccountModel`. The property doesn't exist at that URL because it seems the documentation hasn't been updated to reflect it but an API call returns it.

jd-alexander:
This PR fixes #1346 by adding `usernameCanBeChanged` to the `AccountModel` and updating the  `AccountRestClient` with the associated `user_login_can_be_changed` property.

**Testing**

- [x]  Unit tests :`AccountModelTest` and `AccountSqlUtilsTest` were updated with the new column and all their tests passed. 
- [x]  Instrumented Tests : Ran `MockedStack_AccountTest` and all tests passed. Run `ReleaseStack_AccountTest` to verify the account api layer still functions accurately. - Having an `AuthenticationError` so I am digging into that to resolve it. 
- [x] Example app tests: Run the `AccountFragment` to fetch sample account settings. N.B example app was not updated to show this property since it doesn't seem necessary to do so. 
- [x] Test in WPAndroid - called `accountStore.getAccount().getUsernameCanBeChanged()` and got back a result.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/AccountModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/AccountModel.java
@@ -36,6 +36,7 @@ public class AccountModel extends Payload<BaseNetworkError> implements Identifia
     @Column private boolean mPendingEmailChange;
     @Column private String mWebAddress; // WPCom rest API: user_URL
     @Column private boolean mTracksOptOut;
+    @Column private boolean mUsernameCanBeChanged;
 
     public AccountModel() {
         init();
@@ -76,7 +77,8 @@ public class AccountModel extends Payload<BaseNetworkError> implements Identifia
                && getPendingEmailChange() == otherAccount.getPendingEmailChange()
                && StringUtils.equals(getWebAddress(), otherAccount.getWebAddress())
                && getHasUnseenNotes() == otherAccount.getHasUnseenNotes()
-               && getTracksOptOut() == otherAccount.getTracksOptOut();
+               && getTracksOptOut() == otherAccount.getTracksOptOut()
+               && getUsernameCanBeChanged() == otherAccount.getUsernameCanBeChanged();
     }
 
     public void init() {
@@ -98,6 +100,7 @@ public class AccountModel extends Payload<BaseNetworkError> implements Identifia
         mPendingEmailChange = false;
         mWebAddress = "";
         mTracksOptOut = false;
+        mUsernameCanBeChanged = false;
     }
 
     /**
@@ -134,6 +137,7 @@ public class AccountModel extends Payload<BaseNetworkError> implements Identifia
         setTracksOptOut(other.getTracksOptOut());
         setWebAddress(other.getWebAddress());
         setDisplayName(other.getDisplayName());
+        setUsernameCanBeChanged(other.getUsernameCanBeChanged());
     }
 
     public long getUserId() {
@@ -286,5 +290,13 @@ public class AccountModel extends Payload<BaseNetworkError> implements Identifia
 
     public void setTracksOptOut(boolean tracksOptOut) {
         mTracksOptOut = tracksOptOut;
+    }
+
+    public boolean getUsernameCanBeChanged() {
+        return mUsernameCanBeChanged;
+    }
+
+    public void setUsernameCanBeChanged(boolean usernameCanBeChanged) {
+        mUsernameCanBeChanged = usernameCanBeChanged;
     }
 }
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
@@ -1130,9 +1130,6 @@ public class AccountRestClient extends BaseWPComRestClient {
         if (from.containsKey("primary_site_ID")) {
             accountModel.setPrimarySiteId(((Double) from.get("primary_site_ID")).longValue());
         }
-        if (from.containsKey("user_login_can_be_changed")) {
-            accountModel.setUsernameCanBeChanged((Boolean) from.get("user_login_can_be_changed"));
-        }
         return !old.equals(accountModel);
     }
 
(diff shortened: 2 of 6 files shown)


jd-alexander:
@maxme I am not sure why the `Mocked Connected Tests` are failing. I even triggered the build again and the same result.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
```diff
@@ -929,6 +929,10 @@ private void migrateAddOn(@AddOn String addOnName, SQLiteDatabase db, int oldDbV
                 case 82:
                     AppLog.d(T.DB, "Migrating addon " + addOnName + " to version " + (oldDbVersion + 1));
                     db.execSQL("ALTER TABLE WCOrderModel ADD COLUMN DATE_PAID TEXT NOT NULL DEFAULT '';");
+                    break;
+                case 83:
+                    AppLog.d(T.DB, "Migrating addon " + addOnName + " to version " + (oldDbVersion + 1));
```
maxme: You added this migration step to `migrateAddOn` which is specific to "WooCommerce Addons", you should move that to the main migration function [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/a1b9f269a02468650a2146a68b4e6fff7ba550e4/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java)

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
```diff
@@ -929,6 +929,10 @@ private void migrateAddOn(@AddOn String addOnName, SQLiteDatabase db, int oldDbV
...
+                    db.execSQL("ALTER TABLE AccountModel ADD USERNAME_CAN_BE_CHANGED boolean");
```
maxme: Looks good:
- I checked the type (`integer` is used in the table creation code, but `boolean` is an alias to `integer` IIRC in sqlite and is used elsewhere) 
- I checked `USERNAME_CAN_BE_CHANGED` in the generated code `AccountModelTable.java`

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
```diff
@@ -1129,6 +1130,9 @@ public static boolean updateAccountModelFromPushSettingsResponse(AccountModel ac
         if (from.containsKey("primary_site_ID")) {
             accountModel.setPrimarySiteId(((Double) from.get("primary_site_ID")).longValue());
         }
+        if (from.containsKey("user_login_can_be_changed")) {
```
maxme: This setting can't be changed now, unless we update the API in the future. I'd personally remove this, but that's not a big deal if we keep it.

fluxc/src/main/java/org/wordpress/android/fluxc/model/AccountModel.java
```diff
@@ -287,4 +291,12 @@ public boolean getTracksOptOut() {
     public void setTracksOptOut(boolean tracksOptOut) {
         mTracksOptOut = tracksOptOut;
     }
+
+    public boolean getUsernameCanBeChanged() {
```
maxme: You're right about this getter/setter names, I just want to comment that we can't use `canUsernameBeChanged` here because WellSql expects get/set names.

jd-alexander: Yes, I actually tried to use that and `WellSql` didn't allow.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
jd-alexander: > I checked the type (integer is used in the table creation code, but boolean is an alias to integer IIRC in sqlite and is used elsewhere)

Yes, you are correct. I used it here because it communicates the intent and maps back to the definition within the entity. 

> I checked USERNAME_CAN_BE_CHANGED in the generated code AccountModelTable.java

Yes, the `createStatement` function has all the boolean type properties as integers. A good point to note 👍

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/account/AccountRestClient.java
jd-alexander: I agree with removing it. If it's ever necessary it can be easily readded. Fixed in [7ecbe83](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1351/commits/7ecbe83e20e4e4f05f0bb7c94a53004ca68fdac4)

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.java
jd-alexander: Good catch @maxme  😄 thank you for spotting this. Resolved in [7ecbe83](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1351/commits/7ecbe83e20e4e4f05f0bb7c94a53004ca68fdac4)

maxme:
LGTM, :shipit: (I'll merge it to a feature branch here as well, I named it `issue/wpandroid-8002-change-username`)

jd-alexander:
@maxme quick question about the feature branch. Something I noticed while working today is that the `feature branch` was behind `develop` a bit and when I tried to utilize it with `WPAndroid` I got several errors because the `WPAndroid` feature branch had in code that referenced FluxC that wasn't yet merged to this `feature branch`. 

My working solution was to create a fork of both feature branches since I can't update it myself, merge `develop` into it and use the hash from my repo. Is that implementation okay for now? Any thoughts?

maxme:
@jd-alexander Yes that's right, `develop` was updated right after I forked, I'll merge it to the feature branch.

PR #1357, opened 2019-08-20
Add new autosave fields to the PostModel

maxme:
Add 3 new fields to the PostModel + migration plan, to be used in wpandroid for restoring title/content/excerpt.

Also updated the `PostStore` and `PostRestClient` to update the AutoSaveRevisionId when fetching the post list (needed to detect unhandled auto saves in the post list).

Autosave restoration logic will be done on the client side.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
@@ -126,12 +126,14 @@ public class PostRestClient extends BaseWPComRestClient {
                         List<PostListItem> postListItems = new ArrayList<>(response.getPosts().size());
                         for (PostWPComRestResponse postResponse : response.getPosts()) {
                             String autoSaveModified = null;
+                            long autoSaveRevisionId = 0;
                             if (postResponse.getPostAutoSave() != null) {
+                                autoSaveRevisionId = postResponse.getPostAutoSave().getRevisionId();
                                 autoSaveModified = postResponse.getPostAutoSave().getModified();
                             }
                             postListItems
                                     .add(new PostListItem(postResponse.getRemotePostId(), postResponse.getModified(),
-                                            postResponse.getStatus(), autoSaveModified));
+                                            postResponse.getStatus(), autoSaveModified, autoSaveRevisionId));
                         }
                         boolean canLoadMore = postListItems.size() == pageSize;
                         FetchPostListResponsePayload responsePayload =
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
@@ -47,6 +47,7 @@ import org.wordpress.android.fluxc.store.PostStore.FetchRevisionsResponsePayload
 import org.wordpress.android.fluxc.store.PostStore.PostDeleteActionType;
 import org.wordpress.android.fluxc.store.PostStore.PostError;
 import org.wordpress.android.fluxc.store.PostStore.PostListItem;
+import org.wordpress.android.fluxc.store.PostStore.PostListItem.PostListAutoSave;
 import org.wordpress.android.fluxc.store.PostStore.RemoteAutoSavePostPayload;
 import org.wordpress.android.fluxc.store.PostStore.RemotePostPayload;
 import org.wordpress.android.util.StringUtils;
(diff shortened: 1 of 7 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -708,7 +711,8 @@ private void handleFetchedPostList(FetchPostListResponsePayload payload) {
                 // will not be updated.
                 boolean isPostChanged =
                         !post.getLastModified().equals(item.lastModified)
-                        || !post.getStatus().equals(item.status);
+                        || !post.getStatus().equals(item.status)
+                        || item.autoSaveRevisionId > 0;
```
malinajirka: Can you pls elaborate why we compare `item.autoSaveRevisionId > 0` instead of comparing `item.autoSaveRevisionId != post.autoSaveRevisionId`? It seems the code will mark the post as "potential conflict" whenever the post contains a remoteAutoSave object. Is that what we want?

Imagine the following scenario
1. We edit a published post and leave the editor
2. The app remote-auto-saves it
3. We fetch the post list -> `item.autoSaveRevisionId > 0` is true since the post has unpublished revision -> a conflict is detected even though the current local copy of the post is equal to the `AutoSaved` copy.

Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -736,6 +740,7 @@ private void handleFetchedPostList(FetchPostListResponsePayload payload) {
                         // both locally and on the remote), so flag the local version of the Post so the
                         // hosting app can inform the user and the user can decide and take action
                         post.setRemoteLastModified(item.lastModified);
+                        post.setAutoSaveRevisionId(item.autoSaveRevisionId);
```
malinajirka: I'd suggest introducing a new field for this scenario -> something like `autoSaveRemoteRevisionId` or another options is to update all the auto-save fields here. As otherwise we might end up in an inconsistent state. 
Eg. we would update `autoSaveRevisionId` but all the other fields of the AutoSave object would still contain content related to the previous revisionId. Wdyt?

malinajirka: Btw I think we might remove `mRemoteAutoSaveModified` since we don't use it for conflict detection anymore.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
maxme: Right, I was a bit confused, the idea was to always get the "unhandled auto save" unless we have some local modifications (which happens when you push a remote-auto-save from the app). 

I'll follow your suggestion, this won't change the current behavior, but this might be less confusing if we update the logic later.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
maxme: Ok, do you see any future usage of `mRemoteAutoSaveModified`? For a version 2 of the conflict dialog?

maxme: Sidenote: would be great to add tests for this.

malinajirka: >Sidenote: would be great to add tests for this.
We have some tests for this method
https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/71513ff9b476c2b9b704c17fea445d6c899c9211/example/src/test/java/org/wordpress/android/fluxc/post/PostStoreTest.kt#L32. It should be pretty straightforward to add tests for the new logic.

>Ok, do you see any future usage of mRemoteAutoSaveModified? For a version 2 of the conflict dialog?

Yes, I think we'll need it for v2.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
```diff
@@ -125,13 +126,17 @@ public void fetchPostList(final PostListDescriptorForRestSite listDescriptor, fi
                     public void onResponse(PostsResponse response) {
                         List<PostListItem> postListItems = new ArrayList<>(response.getPosts().size());
                         for (PostWPComRestResponse postResponse : response.getPosts()) {
-                            String autoSaveModified = null;
+                            PostListAutoSave postListAutoSave = new PostListAutoSave();
                             if (postResponse.getPostAutoSave() != null) {
-                                autoSaveModified = postResponse.getPostAutoSave().getModified();
```
malinajirka: .

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -736,6 +744,11 @@ private void handleFetchedPostList(FetchPostListResponsePayload payload) {
                         // both locally and on the remote), so flag the local version of the Post so the
                         // hosting app can inform the user and the user can decide and take action
                         post.setRemoteLastModified(item.lastModified);
+                        post.setAutoSaveRevisionId(item.autoSave.revisionId);
```
malinajirka: .

fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java
```diff
@@ -721,8 +730,7 @@ private void handleFetchedPostList(FetchPostListResponsePayload payload) {
                   * was updated. However, if we know the last modified date is equal to the date we have in local
                   * autosave object we are sure that our invocation of /autosave updated the post directly.
                  */
-                if (isPostChanged && item.lastModified.equals(post.getAutoSaveModified())
-                    && item.autoSaveModified == null) {
+                if (isPostChanged && item.lastModified.equals(post.getAutoSaveModified())) {
```
malinajirka: Is this change intentional?

maxme: Not intentional 🤦‍♂ but I actually found why it was removed (removed a field from PostListItem). Did an extra commit (quite useless since you'll open a followup)  57cc768a2edc27022412829ecbd70ba20af8016c

Issue #1277, opened 2019-05-30
Fetch only autosave.modifiedDate object instead of the whole meta object


malinajirka:
The introduction of the List Management component changed the way we fetch posts from the server. We basically first fetch just list of ids, modifiedDates and statuses -> if the PostModel in our local database has the same status and modified date, we don't fetch its detailed data. This approach saves tons of data and is also more performant.

However, we've started fetching the whole "meta" object [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/1ec4e335986162af07901b124a032833a176ed3e/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java#L113) - this change was introduced in [this PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1255) as we didn't come up with a better solution.

Ideal fix for this issue is to add support for requesting specific nested fields to the API -> we'd request just "meta.data.autosave.modified" instead of the whole meta object.

cc @shiki @maxme

shiki:
Sounds good to me. @malinajirka, have you measured how big the `meta` objects can be?

malinajirka:
The difference is significant. This is just an example + if you take into account the `autosave.content`  can be huge, I believe this ticket should have at least a medium priority.
<img width="453" alt="Screenshot 2019-05-30 at 19 04 18" src="https://user-images.githubusercontent.com/2261188/58650008-eae1f080-830d-11e9-86d1-6a97a233679a.png">

vs
<img width="349" alt="Screenshot 2019-05-30 at 19 05 04" src="https://user-images.githubusercontent.com/2261188/58650010-ee757780-830d-11e9-9e2c-f8bc014cfa7c.png">

shiki:
Thank you. I agree. That is significant.

maxme:
Sounds good to me. I had a look at the [API doc](https://developer.wordpress.com/docs/api/1.2/get/sites/%24site/posts/) and I'm not sure how this is doable.

I tried `meta=data.autosave.modified` and `meta=autosave.modified` but it doesn't work.

malinajirka:
Yes, sorry. I should have been more explicit. This is not supported on the API at the moment. I created this ticket so we don't forget about it and propose the change to the API team.

malinajirka:
The api change request is being discussed here - p3hLNG-14E-p2.

shiki:
Fixed in #1379

PR #1364, opened 2019-08-23
Optimize fetch post list call

malinajirka:
This PR contains a couple of changes:
- We stop storing most of AutoSave object subfields into PostListItem. The main reason for this change is that PostListItem should request only vital fields for determining whether the object was modified in the remote (we want to fetch the full object only when it was modified to save users' data). We currently request the whole AutoSave object, but we plan to optimize the call and request only `autosave.lastModified` very soon - https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1277 (PR on the API is almost ready to be merged).
- It fixes an issue in `handleFetchedPostList` where  `isPostChanged` field was comparing `item.autoSave.revisionId != post.getAutoSaveRevisionId()`. The issue was that `autoSaveRevisionId` doesn't change when the content is overridden with newer autosave. It's safer to compare `autosave.lastModified` instead which is always updated whenever something changes in the Autosave object.
- It "simplifies" `handleFetchedPostList` -> the logic is following
1. The post is not locally changed -> if the post or its autosave object changes, we want to fetch it.
2. The post is locally changed and the post itself is changed -> we don't want to fetch the post and we want to update remoteLastModified (aka mark it as "in conflict with remote")
3. The post is locally changed and its autosave object changes -> we don't want to fetch the post (this scenario will be handled in conflict resolution v2). We can simply ignore this case since the UI shows the Unpublished revision dialog only when there are no local changes.
4. The post is locally changed and both the post itself and its autosave object changes -> We can proceed with `2.` (the autosave conflict resolution will be handled in v2)

https://github.com/wordpress-mobile/WordPress-Android/pull/10426 is targeting this branch so we'll test these changes as part of the PR in WPAndroid.


malinajirka:
>>wordpress-mobile/WordPress-Android#10426 is targeting this branch so we'll test these changes as part of the PR in WPAndroid.
>Sorry, I'm having trouble understanding this. Should I wait to review until #10426 is merged? Or the other way around?

You can test the functionality of this PR in https://github.com/wordpress-mobile/WordPress-Android/pull/10426. However, this PR should be merged first.

>I'm hoping to see those first before diving in since, at first glance, this seems like a complicated change and the tests would help me understand this better. 😬

It's actually not that complicated. I discussed this change with Maxime, but unfortunately he didn't have time to review it.
We merged https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1357/files a PR last week and this PR is just modifying some of the code which was added there. I'll add some tests, but they'll basically tests scenarios 1-4 described above.

Let me know if you rather want to jump on a quick call.



>

malinajirka:
Added unit tests in d4ceb54dd4d76a64a2f9caa2c4f35aa78e1242a3

PR #1375, opened 2019-08-29
issue/1373-woo-stats-timezone-fix

### LINKED ISSUE #1373: Woo stats: incorrect start date calculated for current month due to timezone
While testing the new stats v4 features, I noticed that `This month` TAB displays data from the previous month as well. This seems to be because of the way the start date is calculated because of the timezone of the site. 

This issue seems to resolve itself after a couple of hours.
(LEFT: tested around 10:30 IST . RIGHT: tested around 2:00 IST):
 .

anitaa1990:
Fixes #1373. We need to pass a `before` and `after` param to the v4 stats API request. Currently, the `after` i.e. start date is calculated based on the `StatsGranularity`.

- `DAYS` -> current date
- `WEEKS` -> the first day of the current week
- `MONTHS` -> the first day of the current month
- `YEARS` -> the first day of the current year

The logic was to fetch the the above dates and then convert it according to the site's timezone. This was causing problems with timezones of the site and in some timezones, the start date was calculated as:
- `WEEKS` -> the last day of the previous week
- `MONTHS` -> the last day of the previous month
- `YEARS` -> the last day of the previous year

This PR modifies the logic to calculate the start date by first fetching the current date according to the site's timezone and then modifying the date based on the `StatsGranularity`. 

### Testing
- It would be good to test with multiple timezones to verify if the dates passed to the API request is correct. I changed the timezone of my testing device to GMT+14 and fetched the revenue stats in the example app for the current day, this week, this month and this year. 
- Using Sethos, notice that the `after` param passed to the revenue stats API request for `This Week`, `This month` and `This year` is incorrect.
- Pull changes from the PR and notice that the `after` param value for `This Week`, `This month` and `This year` is correct. 
- Run the tests in `DateUtilsTest`

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/utils/DateUtilsTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/utils/DateUtilsTest.kt
@@ -5,6 +5,9 @@ import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.utils.DateUtils
 import org.wordpress.android.fluxc.utils.SiteUtils
 import java.text.SimpleDateFormat
+import java.time.LocalDate
+import java.time.ZoneId
+import java.time.temporal.WeekFields
 import java.util.Calendar
 import java.util.Date
 import java.util.Locale
@@ -13,6 +16,7 @@ import kotlin.test.assertEquals
 class DateUtilsTest {
     companion object {
         private const val DATE_FORMAT_DAY = "yyyy-MM-dd"
+        private const val DATE_TIME_FORMAT_START = "yyyy-MM-dd'T'00:00:00"
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


AmandaRiu:
Testing note - an emulator can be started from the commandline set to a specific timezone:
```
emulator @Pixel_3_XL_API_28 -timezone [timezone] -debug-timezone
```
- `timezone` value must be in this format: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
- `-debug-timezone` prints timezone related messages to the console

AmandaRiu:
@anitaa1990 I also went ahead and checked out the branch in woo-mobile for using this FluxC branch to see how it all looked in the stats ui. I opened an emulator for the `Asia/Sakhalin` timezone. Then I submitted 3 orders on my test site. Notice that even though "today" on this device is **Friday, August 30th**, all 3 of my new orders are showing up for **Thursday, August 29th**:

 

I ran the same test on a device with the timezone set to `America/Los_Angeles` and there is no difference in how the data is presented:

anitaa1990:
@AmandaRiu thank you so much for testing this yesterday! 

> emulator @Pixel_3_XL_API_28 -timezone [timezone] -debug-timezone

This is super great!! Thanks for the tip!

> I tried using the FETCH CUSTOM REVENUE STATS in the Example app to test pulling stats for today, and week..etc, but I'm not seeing anything in stethos in regard to an after param so I'm thinking I'm doing something wrong. 

Sorry for not including detailed testing steps. I've used the below steps to reproduce the issue:
- I changed the timezone on my emulator to `GTM+14`.
- Click on the `Woo` button in the sample app. 
- Click on `Revenue Stats` button.
- Select a site that has `wc-admin` installed.
- Open `chrome://inspect` and click on the `Network` tab.
- Click on `Fetch current month revenue stats (Forced)`.
- Verify that the API request sent from Network tab is `/wc/v4/reports/revenue/stats`.
- Before pulling the changes from this PR, notice that the `after` param for the current month is sent as `2019-07-31` and after pulling the changes from this PR the current month is sent as `2019-08-01`.

#### Sample request for current month API request:
```
/wc/v4/reports/revenue/stats/&_method=get&interval=day&after=2019-08-01T00%3A00%3A00&before=2019-08-30T23%3A59%3A59&per_page=36&order=asc
```

#### Expected Behaviour


#### Actual Behaviour


> I also went ahead and checked out the branch in woo-mobile for using this FluxC branch to see how it all looked in the stats ui. I opened an emulator for the Asia/Sakhalin timezone. Then I submitted 3 orders on my test site. Notice that even though "today" on this device is Friday, August 30th, all 3 of my new orders are showing up for Thursday, August 29th

I believe this would be because the stats are displayed according to the timezone from the site and not from the timezone of the device. Usually when I place an order today, since my store is set in the US, I usually see the order for the previous day as well. I believe there was some discussion around this recently on whether to display the stats according to the site's timezone or the device's timezone. @jaclync wrote a post on whether we should support both. I think it was decided to keep the site's timezone for now since this is what WP.com is doing at the moment but if get a lot of users worried about it, we can certainly think about changing this. 

Adding the relevant discussion here from Kylea and Aaron:
```
IMO it makes the most sense to use the client app timezone otherwise it’s confusing when a new order comes through and the time doesn’t match the current time. But, if we are using the store/site timezone in WP.com and in core we should probably stay consistent.

I feel that it’s probably fine to wait on this and if we get a lot of confused users, we can worry about it then. In the meantime maybe add it to the feature backlog in Monday
```
 
Let me know if the above makes sense!

AmandaRiu:
Thank you for all the details @anitaa1990! I'll go through and retest these out today.

PR #1379, opened 2019-08-30
Request only the required fields in fetchPostList request

### LINKED ISSUE #1277: Fetch only autosave.modifiedDate object instead of the whole meta object

malinajirka:
Fixes #1277 

~Note: This is just a draft PR since the change on the API hasn't been merged yet `D29715-pha`.~

Modifies the fetchPostList request so we don't waste users' data. We were previously fetching the whole meta object, because further filtering wasn't supported on the API level. However, filtering support was added in `D29715-pha`.

To test
1. Use this version of FluxC in WPAndroid
2. Open the Post list
3. Use for example Stetho to check details of the request
4. Make sure the meta.autosave object doesn't contain any other data but "modified".

cc @jkmassel

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
@@ -114,7 +114,7 @@ public class PostRestClient extends BaseWPComRestClient {
                         listDescriptor.getOrderBy().getValue(), listDescriptor.getSearchQuery());
 
         // We want to fetch only the minimal data required in order to save users' data
-        params.put("meta_fields","autosave.modified");
+        params.put("meta_fields", "autosave.modified");
 
         final boolean loadedMore = offset > 0;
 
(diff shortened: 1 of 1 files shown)


shiki:
@malinajirka Since `D29715-pha` was merged, I guess we can open this now?

PR #1382, opened 2019-09-19
Update Flux link

designsimply:
Update the Flux docs link in README.md as it appears to have moved and was 404ing.


aforcier:
:shipit:

PR #1405, opened 2019-10-17
Gateways endpoint

0nko:
This PR adds a client for fetching data from the gateway REST API endpoint, which is required for automatic refunds. It also includes unit tests and example app functionality for testing.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGatewayStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGatewayStore.kt
@@ -1,31 +1,15 @@
 package org.wordpress.android.fluxc.store
 
 import kotlinx.coroutines.withContext
-import org.wordpress.android.fluxc.Payload
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.gateways.GatewayMapper
 import org.wordpress.android.fluxc.model.gateways.WCGatewayModel
-import org.wordpress.android.fluxc.network.BaseRequest.GenericErrorType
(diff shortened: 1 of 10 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGatewayStore.kt
```diff
@@ -0,0 +1,123 @@
...
+        TIMEOUT,
+        API_ERROR,
+        INVALID_GATEWAY_ID,
+        GENERIC_ERROR,
+        INVALID_RESPONSE,
+        AUTHORIZATION_REQUIRED
+    }
+}
+
+fun WPComGsonNetworkError.toGatewayError(): GatewayError {
```
anitaa1990: The error scenarios listed here seem to be a duplicate of the error scenarios in `WcRefundsStore` [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/22405120eef8dd2ea5b295b27ee7aa8a21568c4d/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCRefundStore.kt#L149). Do you think we can reuse the one already available?

0nko: I wanted to do that but didn't get around to it. It makes sense, though :+1:.

0nko:
Thanks for the review, @anitaa1990! I've made the requested changes and it's ready for another pass. I haven't been able to reproduce the `INVALID_RESPONSE` error, though.. I works both on my phone and the emulator (I can see the gateways in the result box):

image

PR #1410, opened 2019-10-26
Woo: support for hiding future-dated orders

### LINKED ISSUE #1409: Woo: Option to not include future orders when fetching orders list
Need an option to not include future orders (orders with a `dateCreated` > today). For this to work properly, the code to fetch order status counts should also be updated somehow to not include orders that have a future create date.

AmandaRiu:
This PR fixes #1409 and includes the following:

- Adds a new `excludeFutureOrders` parameter to the `WCOrderListDescriptor` 
- Updated the "Order LIst" view in the example app to include a new option to **exclude future orders** (see screenshot), as well as a new future orders time group section to be displayed in the list.

**NOTE**: I'm excluding orders programmatically in the `WCOrderListItemDataSource` because the API to fetch orders does not have an option for excluding orders dated for the future. Since the new list management design uses the results of the API call for everything, I had to move this logic into the datasource.

### Screenshots

New "Exclude Future Orders" checkbox and the "GROUP_FUTURE" section to display orders dated for the future:

Excluding filter orders:

### Recommended Test Scenarios

- Manually modify the created date of an order to change it to a date in the future. 
- Verify it shows up in the "FUTURE" section of the list. 
- check "Exclude Future Orders" - verify the future orders are not included in the list results

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -28,6 +28,7 @@
       <option name="XML_LEGACY_SETTINGS_IMPORTED" value="true" />
     </XML>
     <codeStyleSettings language="JAVA">
+      <option name="RIGHT_MARGIN" value="120" />
       <option name="ALIGN_MULTILINE_CHAINED_METHODS" value="true" />
       <option name="ALIGN_MULTILINE_BINARY_OPERATION" value="true" />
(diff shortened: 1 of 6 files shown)


AmandaRiu:
@anitaa1990 That test seems to be really flaky. Anyway, I reran the Circle CI tests and they all passed. Ready for merge 👍

PR #1426, opened 2019-11-12
Issue/1425 wc remove product image

### LINKED ISSUE #1425: Woo: add method to remove a product image from the db
WCAndroid is adding the ability to remove an image. We're using the existing code to update a product's image list to handle the network side of removing a product image, but we still need a way to delete an image from SQLite.

nbradbury:
Closes #1425 - adds a method to delete a single image from a product, with a related test.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
@@ -270,7 +270,7 @@ object ProductSqlUtils {
         return insertOrUpdateProduct(product)
     }
 
-    fun removeProductImage(site: SiteModel, remoteProductId: Long, remoteMediaId: Long): Boolean {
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -251,4 +254,37 @@ object ProductSqlUtils {
     }
 
     fun deleteAllProductReviews() = WellSql.delete(WCProductReviewModel::class.java).execute()
+
+    fun updateProductImages(site: SiteModel, product: WCProductModel, imageList: List<WCProductImageModel>): Int {
```
anitaa1990: np: It doesn't look like the `SiteModel` passed to this method is used anywhere. Can we remove it?

nbradbury: Done in e903413.

PR #1432, opened 2019-11-20
Woo: add methods to update a product

### LINKED ISSUE #1431: Woo: support updating a product
The existing FluxC code currently includes the endpoint needed to update a product: This is done via PUT `/wp-json/wc/v3/products/<id>`, but we still need separate actions/methods to update product info.

anitaa1990:
Fixes #1431. This PR adds the ability to update a product description, name + sku. 

### Changes
- Adds new actions for updating a product: `UPDATE_PRODUCT` and `UPDATED_PRODUCT`.
- Adds new payload classes for updating a product: `UpdateProductPayload`, `RemoteUpdateProductPayload`.
- Adds a new fragment to the example app to test updating a product description, name and sku.
- Adds new tests to `MockedStack_WCProductsTest` and `ReleaseStack_WCProductTest`.

### Screenshots
 .

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -82,6 +82,11 @@ class WCProductStore @Inject constructor(dispatcher: Dispatcher, private val wcP
         var imageList: List<WCProductImageModel>
     ) : Payload<BaseNetworkError>()
 
+    class UpdateProductPayload(
+        var site: SiteModel,
+        val product: WCProductModel
+    ) : Payload<BaseNetworkError>()
+
     enum class ProductErrorType {
         INVALID_PARAM,
         INVALID_REVIEW_ID,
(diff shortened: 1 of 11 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -290,6 +315,8 @@ class WCProductStore @Inject constructor(dispatcher: Dispatcher, private val wcP
                 updateProductReviewStatus(action.payload as UpdateProductReviewStatusPayload)
             WCProductAction.UPDATE_PRODUCT_IMAGES ->
                 updateProductImages(action.payload as UpdateProductImagesPayload)
+            WCProductAction.UPDATE_PRODUCT ->
```
nbradbury: We can do this in a separate PR, but I'm wondering if we should drop `UPDATE_PRODUCT_IMAGES` and rely solely on `UPDATE_PRODUCT`?

anitaa1990: I think that's a good idea, especially since we would now be updating images along side all other items in a product only when **publish/update** is initiated from the UI.

PR #1434, opened 2019-11-26
Fix crash in PageStore - continuation already resumed

malinajirka:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/8401

I'm not sure how to reproduce this, but the change should be easy to review. We always stored a continuation into a local variable even when we synchronously resumed it since the post wasn't found -> I believe we want to store it into a local variable only when we emit a bus event.


malinajirka:
Thanks @planarvoid! I've created a draft PR in WPAndroid - https://github.com/wordpress-mobile/WordPress-Android/pull/10851

PR #1441, opened 2019-12-07
WCOrderSummaryModel records should match WCOrderModel

### LINKED ISSUE #1440: Woo: Cached orders not being displayed when device is offline in some scenarios
While documenting and testing my order list project, I found that the current implementation of using the `WCOrderSummaryModel` table for tracking order identifiers had some smells:
- The whole table was being cleared whenever a new batch of `WCOrderSummaryModels` were fetched from the API. 
- That same table is being used by the `ListDescriptor` for orders to create the identifiers for the initial loading of a list. 

The above works fine as long as you have network connectivity, but when the device goes offline and attempts to load a new `pagedListWrapper` with different filter params, the list will show no data is available, even though there are records in `WCOrderModel` - they just don't exist in `WCOrderSummaryModel`. 

~After further evaluation I've come to the conclusion that we can remove the `WCOrderSummaryModel` table altogether. I'm not 100% sure why it was determined we needed it back when Oguz and I originally implemented this project, but after really getting a better understanding of the `ListStore` design and working the full implementation, I think we should be fine with just converting that to a POJO for in-memory comparisons and passing around smaller objects for the identifier creation process.~ This table is actually needed to populate the order list with the order identifiers grouped by dates.

AmandaRiu:
Fixes #1440 

It took me a while but I finally figured out why we needed the `WCOrderSummaryModel` table for the order list's `ListStore` implementation. It was because we use it to get the date of orders that are being fetched so we can show the loading view for that order while it’s being fetched.  So despite what the ticket says, we will be keeping the `WCOrderSummaryModel` table.

To address the original reason for the ticket, I removed the code to delete all records in `WCOrderSummaryModelTable` before inserting the latest batch. This is to fix an issue where orders that have been downloaded and cached for a particular `ListDescriptor` on the device won't show up on the device when it's offline because it doesn't have a matching `WCOrderSummaryModel` record.

### To Test
1. Log into example app
2. Go to Woo -> Orders -> Fetch Order List
3. Verify orders load properly. 
4. Filter the order list by an order status that is known to have matching orders.
5. Put the device in Airplane Mode to set it offline.
6. Clear the filter. If the cached orders are returned, then the fix is successful.

NOTE: This only fixes displaying lists that have already been used before going offline. So if you try to filter by a different status than the one used in the test above while offline, you will get no records back - even if you know there are cached records in the db. This is because the `ListStore` keeps a list of items in a different table (`ListItemModel`) to track items downloaded for each listDescriptor. So if you filter by "processing" while online, the `ListStore` will know about the results, but if you have never filtered by "processing" on this device, then the `ListStore` will have no record of the results. This is a workaround for this bit it needs to be addressed on the `ListItemDatasourceInterface` implementation - which lives on the client.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderSummaryModel.kt
@@ -5,8 +5,18 @@ import com.yarolegovich.wellsql.core.annotation.Column
 import com.yarolegovich.wellsql.core.annotation.PrimaryKey
 import com.yarolegovich.wellsql.core.annotation.RawConstraints
 import com.yarolegovich.wellsql.core.annotation.Table
+import org.wordpress.android.fluxc.model.list.datasource.ListItemDataSourceInterface
 import org.wordpress.android.fluxc.persistence.WellSqlConfig
 
+/**
+ * Class represents the bare minimum fields needed to determine if an order is outdated and
+ * needs to be fetched. It's also important to store this information in the database so it
+ * can be used by the order list's implementation of [ListItemDataSourceInterface] to create a list
+ * of existing orders in the [WCOrderModel] table, as well as a list of orders being fetched by the
+ * API because they do not yet exist. Normally we wouldn't need this extra step to work with the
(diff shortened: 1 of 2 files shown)


AmandaRiu:
This is now ready for review

Issue #1442, opened 2019-12-09
Woo - Products: Add endpoints to fetch tax_class and shipping_class options


anitaa1990:
As part of editing a product, we need to provide an option for users to update the `tax_class` and `shipping_class` of a product. This requires adding the option to fetch the `tax_class` and `shipping_class` options associated with a site.

anitaa1990:
Closed via #1456 and #1458

PR #1443, opened 2019-12-09
Woo: Update product fields

### LINKED ISSUE #1435: Woo Poducts: add support to update empty product description
PR #1432 added support to update product description. But there is no way for the user to update a blank product description at the moment. This is supported by `wp-admin` and needs to be added to the app

anitaa1990:
Fixes #1435. This PR adds logic to update the following product fields to the API.

#### Changes
- Adds two columns/fields `dateOnSaleFrom` and `dateOnSaleTo` to `WCProductModel` to allow for scheduling product sales.
- Adds logic to update product only if updated product fields do not match the stored product fields. This is to allow the option for users to update empty product descriptions (should they choose to do so) and also, to ensure that we do not update product fields that do not contain any changes. 
- Testing the API, I found that we cannot update an existing SKU value from another product to the product we are trying to update. So added a new error type in `ProductErrorType` if user tries to update the SKU of the product by entering an existing value.
- Added enum classes for displaying product tax status options (`CoreProductTaxStatus`), product back order options (`CoreProductBackOrders`) and product stock status options (`CoreProductStockStatus`).
- Added a new page in the example app to test updating multiple product fields.

#### Testing
- Run tests in `MockedStack_WCProductsTest`, `WCProductStoreTest`.
- Since there is a migration script added to the PR, it would be good to verify if app can be updated to this PR without any issues.
- Testing in the example app:
  - Open the app, click on `Woo` (after login, of course) -> `Products` and select a site.
  - Click on `Update Product` button.
  - Enter a product ID.
  - Modify fields displayed in the UI and verify that only the fields updated in the UI are updated to the product.
  - Verify that an error is displayed in the example app when trying to update an existing SKU (from another product).

#### Screenshots
 . 

#### Notes 
- There is no option to test updating `tax_class` and `shipping_class` fields to a product. This is because we need to add a separate API endpoints to FluxC before testing this. Added issue [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1442) and here to address this separately.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -535,6 +535,8 @@ class ProductRestClient(
 
             dateCreated = response.date_created ?: ""
             dateModified = response.date_modified ?: ""
+            dateOnSaleFrom = response.date_on_sale_from ?: ""
+            dateOnSaleTo = response.date_on_sale_to ?: ""
 
             type = response.type ?: ""
             status = response.status ?: ""
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -228,15 +228,6 @@ class ProductRestClient(
         val responseType = object : TypeToken<ProductApiResponse>() {}.type
         val body = productModelToProductJsonBody(storedWCProductModel, updatedProductModel)
 
-        if (body.isEmpty()) {
-            val payload = RemoteUpdateProductPayload(
-                    ProductError(ProductErrorType.EMPTY_REQUEST_BODY, "No updates found!"),
-                    site, WCProductModel().apply { this.remoteProductId = remoteProductId }
-            )
-            dispatcher.dispatch(WCProductActionBuilder.newUpdatedProductAction(payload))
-            return
-        }
-
         val request = JetpackTunnelGsonRequest.buildPutRequest(url, site.siteId, body, responseType,
                 { response: ProductApiResponse? ->
                     response?.let {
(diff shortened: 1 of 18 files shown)


nbradbury:
This doesn't need to be done in this PR, but it would be nice if this screen relied on labels above the TextViews rather than hints. It's not clear what each fields is when filled in. For example, I don't know what the "2", "3", and "4" fields are here.

Screenshot_1575899170

nbradbury:
If you rotate the device while the list dialog is showing, it loses its list items.

Screenshot_1575899339

nbradbury:
If I rotate the device while the date picker is showing, the date picker disappears and I'm left with a screen where all the fields are disabled.

Screenshot_1575899532

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -420,19 +425,93 @@ class ProductRestClient(
...
+            if (storedWCProductModel.stockStatus != updatedProductModel.stockStatus) {
+                body["stock_status"] = updatedProductModel.stockStatus
+            }
+            if (storedWCProductModel.stockQuantity != updatedProductModel.stockQuantity) {
+                body["stock_quantity"] = updatedProductModel.stockQuantity
+            }
+            if (storedWCProductModel.backorders != updatedProductModel.backorders) {
+                body["backorders"] = updatedProductModel.backorders
+            }
+            if (storedWCProductModel.backorders != updatedProductModel.backorders) {
```
nbradbury: Backorders is duplicated here.

nbradbury:
@anitaa1990 Code looks good - just a few minor issues to fix and I'm good to merge.

anitaa1990:
Thanks for the quick review @nbradbury!

> It would be nice if this screen relied on labels above the TextViews rather than hints.  

Good point! Fixed in 66cd2b0

> If you rotate the device while the list dialog is showing, it loses its list items.

Fixed in 762ea3a

> If I rotate the device while the date picker is showing, the date picker disappears and I'm left with a screen where all the fields are disabled.

Fixed in 1cc77ea

Ready for another round 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
anitaa1990: Fixed in 6bbdd44

PR #1446, opened 2019-12-11
Woo v4 Stats: Change gross_revenue to total_sales to match changes to wc-admin

AmandaRiu:
This field was changed in wc-admin v0.23. We only support the very latest version of the v4 Revenue Stats API since this is a beta feature so anyone running a store with the wc-admin plugin less than v0.22 will not see the gross sales total.

## TO TEST
On a store running the older version of wc-admin:
1. Run the example app
2. WOO -> REVENUE STATS
3. Select a site and then run FETCH CURRENT REVENUE STATS (FORCED)
4. The output in the log will say the total is null.
5. Update the wc-admin plugin to the latest version (currently 0.23.1)
6. Redo steps 1 - 3
7. The output should now show a total.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCRevenueStatsModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCRevenueStatsModel.kt
@@ -52,7 +52,7 @@ data class WCRevenueStatsModel(@PrimaryKey @Column private var id: Int = 0) : Id
         @SerializedName("orders_count")
         val ordersCount: Int? = null
         @SerializedName("total_sales")
-        val grossRevenue: Double? = null
+        val totalSales: Double? = null
     }
 
     /**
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCRevenueStatsModel.kt
```diff
@@ -51,7 +51,7 @@ data class WCRevenueStatsModel(@PrimaryKey @Column private var id: Int = 0) : Id
     class Total {
         @SerializedName("orders_count")
         val ordersCount: Int? = null
-        @SerializedName("gross_revenue")
+        @SerializedName("total_sales")
```
nbradbury: Shouldn't this also be renamed from `grossRevenue`?

nbradbury: Also, assuming that's changed, there are a lot of tests that should be renamed to us the phrase "totalSales" rather than "grossRevenue`.

Screen Shot 2019-12-11 at 2 11 39 PM

AmandaRiu: @nbradbury I updated the label for `WCRevenueStatsModel.Total` but left the rest alone as I'd like to keep the hotfix changes as narrow as possible to limit possible side effects. I'll open a ticket to update the rest of the labels.

AmandaRiu:
@nbradbury Ready for another round!

PR #1451, opened 2019-12-13
Issue/1450 product variants sorting

### LINKED ISSUE #1450: Product variations should be sorted by title ASC
When we request a product's variations, we don't pass an `orderby` param so the default sorting (date DESC) is used. We should sort them the same as we do products (title ASC), and likewise we should request them from SQLite in this same order.

anitaa1990:
Fixes #1450. This PR adds logic to sort the product variants using the `menu_order` field from the GET product variations API response.

#### Changes
- Adds a new column to `WCProductVariationModel` to store `menu_order` field.
- Adds logic to sort the list of `WCProductVariationModel` using `menu_order` when fetching product variations from the local db.

#### Testing
- Run `ReleaseStack_WCProductTest` and verify that all tests pass.
- Since there is a migration script running, it would be good to test app update. 
- Open the example app. Click on `Woo` -> `Products` -> Select site -> `Fetch product variations`. Verify that the ids are sorted according to the `menu_order` and that it matches the sorting in `wp-admin`.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
@@ -27,7 +27,7 @@ open class WellSqlConfig : DefaultWellConfig {
     annotation class AddOn
 
     override fun getDbVersion(): Int {
-        return 93
+        return 94
     }
 
     override fun getDbName(): String {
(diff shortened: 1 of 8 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
```diff
@@ -1021,6 +1021,9 @@ open class WellSqlConfig : DefaultWellConfig {
                     db.execSQL("ALTER TABLE WCProductModel ADD DATE_ON_SALE_FROM TEXT")
                     db.execSQL("ALTER TABLE WCProductModel ADD DATE_ON_SALE_TO TEXT")
                 }
+                93 -> migrateAddOn(ADDON_WOOCOMMERCE, version) {
```
AmandaRiu: Recommend dropping this table and recreating it so the records are guaranteed to have this new column populated with data.

anitaa1990: There is another F[luxC PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1453) opened, that deletes previous records from the db, before adding new items. So I figured we don't need to drop the table and recreate it again. WDYT?

nbradbury:
@anitaa1990 When we request the product list we should `order` and `orderby` rather than `menu_order`, and I think we want to be consistent. Which do you think is correct?

PR #1455, opened 2019-12-17
Issue/1454 product variants pagination

### LINKED ISSUE #1454: Support paging of product variations
As mentioned [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1453#pullrequestreview-331808533), right now we don't support paging when we fetch product variations. This means we'll never show more than the default of 10 variations for a product, regardless of how many variations the product has.

anitaa1990:
Fixes #1454. This PR adds logic to support pagination when fetching product variations.

#### Changes
- Adds a default page size when fetching variations, which is 25.
- Adds an `offset` field to the API request when fetching variations.
- Adds a sort order to the API request, which is `asc`. This essentially sorts the variations by `menu_order` ASC.
- Adds a new button in the example app to test the pagination.

#### Testing
- In the example app, click on `Woo` -> `Products` - > `Select Site` - > `Fetch product variations`.
- Verify that the variations are fetched correctly and the sorting order is maintained. i.e the variations are fetched by `menu_order` ascending..
- Click on `Lore more variations` and verify that the variations are fetched correctly.
  - If there is no more variation data found in the API response, the `Load more variations` button should be displayed.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
@@ -260,6 +260,7 @@ class WooProductsFragment : Fragment() {
                     if (event.canLoadMore) {
                         pendingFetchSingleProductVariationOffset += event.rowsAffected
                         load_more_product_variations.visibility = View.VISIBLE
+                        load_more_product_variations.isEnabled = true
                     } else {
                         pendingFetchSingleProductVariationOffset = 0
                         load_more_product_variations.isEnabled = false
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
@@ -50,7 +50,9 @@ class WooProductsFragment : Fragment() {
     private var selectedPos: Int = -1
     private var selectedSite: SiteModel? = null
     private var pendingFetchSingleProductRemoteId: Long? = null
+
     private var pendingFetchSingleProductVariationRemoteId: Long? = null
+    private var pendingFetchSingleProductVariationOffset: Int = 0
 
     override fun onAttach(context: Context?) {
         AndroidSupportInjection.inject(this)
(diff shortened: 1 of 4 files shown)


example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
```diff
@@ -241,12 +255,14 @@ class WooProductsFragment : Fragment() {
...
-                        pendingFetchSingleProductVariationRemoteId = null
-                        val variations = wcProductStore.getVariationsForProduct(site, remoteId)
-                        variations.forEach { variant ->
-                            prependToLog("Variations: ${variant.remoteVariationId}")
-                        }
+                    prependToLog("Fetched ${event.rowsAffected} product variants. " +
+                            "More variants available ${event.canLoadMore}")
+                    if (event.canLoadMore) {
+                        pendingFetchSingleProductVariationOffset += event.rowsAffected
+                        load_more_product_variations.visibility = View.VISIBLE
```
nbradbury: The button also needs to be enabled here.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -182,10 +183,19 @@ class ProductRestClient(
...
+        pageSize: Int = DEFAULT_PRODUCT_VARIATIONS_PAGE_SIZE,
+        offset: Int = 0
+    ) {
         val url = WOOCOMMERCE.products.id(productId).variations.pathV3
         val responseType = object : TypeToken<List<ProductVariationApiResponse>>() {}.type
-        val params = emptyMap<String, String>()
+        val params = mutableMapOf(
+                "per_page" to pageSize.toString(),
+                "offset" to offset.toString(),
+                "order" to "asc")
```
nbradbury: Does `orderby` also need to be included here?

anitaa1990: So I found out yesterday that the variations are by default sorted by `menu_order` with `order` = `desc`. There is no option to pass `menu_order` via the API. It _is_ supported by default on the backend but is not one of the supported parameters for the client side. 

But when testing with multiple pages of variations where a variation in a later page has a lower `menu_order`, the sorting keeps changing in the app. So we need to manually pass `order` here to `asc` but without `orderby`

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
anitaa1990: Good point! Fixed in 36cfb7d

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
nbradbury: Okay, can you add a comment to the code which explains this? I think future devs will be thankful for it :)

anitaa1990: For sure :D Fixed in 7e47a77

nbradbury:
@anitaa1990 I just [checked out the docs](http://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-product-variations), and they state that the default order is by date. Are you sure it's defaulting to ordering by `menu_order`?

PR #1456, opened 2019-12-17
Woo/fetch product shipping class

anitaa1990:
Fixes Part 1 of #1442 by adding support to fetch `shipping_class` list for a site.

#### Changes
- Adds new endpoint `/wc/v3/products/shipping_classes`. Since this is part of the products feature, added all subsequent logic under the products package, even though the `shipping_classes` are not product, but site, specific.
- Added new `WCProductAction` and payload classes.
- Added new `WCProductShippinClassModel` model class and a migration script to create table.
- Added methods to `WCProductRestClient`, `WCProductStore` and `ProductSqlUtils` to fetch, store, update, delete shipping class list for a site. 
- Added mocked, release and unit tests.
- Added a new button to the example app to fetch shipping class list for a site.

#### Testing
- Since this PR adds a migration script, it would be nice to test updating the app.
- Run `MockedStack_WCProductsTest`, `ReleaseStack_WCProductTest` and `ProductSqlUtilsTest`.
- In the example app, click on `Woo` -> `Products` -> `Select Site` -> `Fetch Product shipping classes` and verify that the shipping class list for the site is fetched successfully.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -13,6 +13,7 @@ import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.WCProductImageModel
 import org.wordpress.android.fluxc.model.WCProductModel
 import org.wordpress.android.fluxc.model.WCProductReviewModel
+import org.wordpress.android.fluxc.model.WCProductShippingClassModel
 import org.wordpress.android.fluxc.model.WCProductVariationModel
 import org.wordpress.android.fluxc.network.UserAgent
 import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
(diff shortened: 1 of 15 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -51,6 +53,42 @@ class ProductRestClient(
     accessToken: AccessToken,
     userAgent: UserAgent
 ) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
+    /**
+     * Makes a GET request to `GET /wp-json/wc/v3/products/shipping_classes` to fetch
+     * prodict shipping classes for a site
```
nbradbury: np: typo - `prodict`

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -51,6 +53,42 @@ class ProductRestClient(
...
+     *
+     * Dispatches a WCProductAction.FETCHED_PRODUCT_SHIPPING_CLASS_LIST action with the result
+     *
+     * @param [site] The site to fetch product shipping class list for
+     */
+    fun fetchProductShippingClassList(site: SiteModel) {
+        val url = WOOCOMMERCE.products.shipping_classes.pathV3
+        val responseType = object : TypeToken<List<ProductShippingClassApiResponse>>() {}.type
+        val params = emptyMap<String, String>()
+        val request = JetpackTunnelGsonRequest.buildGetRequest(url, site.siteId, params, responseType,
```
nbradbury: This doesn't support pagination, nor does it specify a page size, so it will open fetch the first 10 shipping classes for a site. That can be tackled in a separate PR, if necessary.

anitaa1990: Good point! I believe there might be some more work around this and extensive testing would be involved and also adding support for deleting those shipping classes, if they are not available in the web anymore. So I would like to take care of this in another PR, if that's ok with you :)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
anitaa1990: Fixed in c0c3927

anitaa1990:
Good point about adding pagination support to shipping class list. I have opened an [issue](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1457) to address that separately. PR ready for another round 👍

PR #1464, opened 2019-12-26
Pass the last day of the week as end date when fetching revenue stats api

### LINKED ISSUE #1463: Woo Stats: provide option to display all the days of the week when fetching stats
The x axis in the Revenue data graph for “This Week” needs to display all the day of the week and the column width of the daily revenue should stay consistent along the week (currently changes depending on how many days have passed in the week)

anitaa1990:
Fixes #1463 - This PR adds the option to pass the end of the week date as the end date, when fetching revenue stats from the `/wc/v4/reports/revenue/stats` endpoint.

#### To Test
- Click on `Woo` -> `Revenue Stats` -> `Select Site` -> `Fetch Current week revenue stats`.
- Verify that the end date for the weekly stats is the future date for the end of the week instead of the current date. i.e. for this week end date should be `28-12-2019`.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt
@@ -641,7 +641,7 @@ class WCStatsStore @Inject constructor(
 
     private fun fetchRevenueStats(payload: FetchRevenueStatsPayload) {
         val startDate = getStartDateForRevenueStatsGranularity(payload.site, payload.granularity, payload.startDate)
-        val endDate = getEndDateForRevenueStatsGranularity(payload.site, payload.granularity, payload.startDate)
+        val endDate = getEndDateForRevenueStatsGranularity(payload.site, payload.granularity)
         val perPage = getRandomPageIntForRevenueStats(payload.forced)
         wcOrderStatsClient.fetchRevenueStats(
                 payload.site,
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt
```diff
@@ -674,6 +674,27 @@ class WCStatsStore @Inject constructor(
...
     }
 
+    /**
+     * Given a [endDate], formats the date based on the site's timezone in format yyyy-MM-dd'T'hh:mm:ss
+     * If the endDate date is empty, fetches the date based on the [granularity]
+     */
+    private fun getEndDateForRevenueStatsGranularity(
+        site: SiteModel,
+        granularity: StatsGranularity,
+        endDate: String?
```
AmandaRiu: What is the argument `endDate` for? Doesn't appear to be used.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt
```diff
@@ -674,6 +674,27 @@ class WCStatsStore @Inject constructor(
         }
     }
 
+    /**
+     * Given a [endDate], formats the date based on the site's timezone in format yyyy-MM-dd'T'hh:mm:ss
```
AmandaRiu: Seems like two different jobs were intended for this method.  Comment probably needs to be updated.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/DateUtils.kt
```diff
@@ -314,6 +314,13 @@ object DateUtils {
         return formatDate(DATE_TIME_FORMAT_START, cal.time)
     }
 
+    fun getLastDayOfCurrentWeekBySite(site: SiteModel): String {
```
AmandaRiu: np: I expect this method to return a mapping of last day by site due to the naming. Recommend changing to `getLastDayOfCurrentWeekForSite()`.

anitaa1990: 👍 Done in b7700e3

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt
anitaa1990: The idea was that if we allow the option to pass `endDate` in the UI side in the future, we can use that `endDate`, rather than having to make changes in FluxC side again. But it's not really used at the moment as you said and I foresee some changes we would need to make anyway to accommodate those changes so removed the redundant `endDate` from the method now in a530d04.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt
anitaa1990: Since I removed the `endDate`, the comment was updated as well in a530d04

anitaa1990:
> found that the end date being sent has a time of 00:00:00 - shouldn't that be 23:59:59 so the results include the full last day of the week?

Good catch! Fixed in b7700e3

Fixed all code comments and ready for another round 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt
AmandaRiu: @anitaa1990 Definitely better, but the comments make it sound like all that's happening is formatting a date, but the method is doing something different. Would it be more accurate to say this method does the following?
```
Returns the appropriate end date for the [site] and [granularity] provided to use for fetching revenue stats.
```

anitaa1990: 👍 Fixed in 2cbf507

AmandaRiu:
@anitaa1990 Looks like there are some failing tests.

anitaa1990:
@AmandaRiu this looks like it was due to a flaky test. I re-ran CircleCI and it's working now. PR ready for another round 👍

PR #1466, opened 2019-12-30
Add a setOnlyIfEmpty option to the DesignateMobileEditorForAllSites payload

maxme:
Make sure to replace "CHANGEME" by an actual Bearer token.

To test:

- Open a user report card, on at least one site, turn `mobile-editor` to `aztec`.
- Run one of the following command.
- Refresh the user report card. Check the `mobile-editor` setting on that site.

3 commands, one is not working as expected (the second one):

- `set_only_if_empty: true`, this won't change the `mobile-editor` setting unless it's "empty" (ie. unset or null).
    ```sh
    curl 'https://public-api.wordpress.com/wpcom/v2/me/gutenberg/' -H 'Authorization: Bearer CHANGEME' -H 'Content-Type: application/json; charset=utf-8' --data '{"editor":"gutenberg","set_only_if_empty":"true","platform":"mobile"}'
    ```

- `set_only_if_empty: false`, this should change the `mobile-editor` whatever current value is, but this doesn't work.
    ```sh
    curl 'https://public-api.wordpress.com/wpcom/v2/me/gutenberg/' -H 'Authorization: Bearer CHANGEME' -H 'Content-Type: application/json; charset=utf-8' --data '{"editor":"gutenberg","set_only_if_empty":"false","platform":"mobile"}'
    ```

- Omit `set_only_if_empty` - This will change the `mobile-editor` value whatever current value is.
    ```sh
    curl 'https://public-api.wordpress.com/wpcom/v2/me/gutenberg/' -H 'Authorization: Bearer CHANGEME' -H 'Content-Type: application/json; charset=utf-8' --data '{"editor":"gutenberg","platform":"mobile"}'
    ```

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
@@ -332,12 +332,12 @@ public class SiteRestClient extends BaseWPComRestClient {
         add(request);
     }
 
-    public void designateMobileEditorForAllSites(final String mobileEditorName) {
+    public void designateMobileEditorForAllSites(final String mobileEditorName, final boolean setOnlyIfEmpty) {
         Map<String, Object> params = new HashMap<>();
         String url = WPCOMV2.me.gutenberg.getUrl();
         params.put("editor", mobileEditorName);
         params.put("platform", "mobile");
-        params.put("set_only_if_empty", "true");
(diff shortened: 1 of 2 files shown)


hypest:
Verified:

The `set_only_if_empty: true` command is not changing the remote setting if the remote setting is `aztec`.

The `set_only_if_empty: false` command is not changing the remote setting but the opposite would be expected.

The "Omit `set_only_if_empty`" command is indeed changing the remote setting of the sites not already on (mobile) Gutenberg.

PR #1467, opened 2020-01-06
Don't remove "p" and "br" tags when uploading local drafts on self hosted sites

maxme:
This fixes https://github.com/wordpress-mobile/WordPress-Android/issues/11030 and https://github.com/wordpress-mobile/gutenberg-mobile/issues/1728

I tried to find the origin of this code and I think it was imported from the old network code when we created FluxC. 

I guess we wanted to clean things up from the legacy editors before pushing things to the site. This doesn't make sense to have this on the network side. If we have an editor that creates unwanted `<p>` or `<br>` tags, we should do this when we get the post content from the editor and not here when we upload the post.

To test:
- Checkout e8adbc22bc3a7eb72b098a812a2f38c3b93c4b72 (contains the new test but not the fix) and run the test `ReleaseStack_PostTestXMLRPC.testUploadNewPostWithGutenbergContent`. Test will fail. 
```sh
./gradlew cAT -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.release.ReleaseStack_PostTestXMLRPC\#testUploadNewPostWithGutenbergContent
```
- Checkout this branch and run the same test. Test will succeed.
```sh
./gradlew cAT -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.release.ReleaseStack_PostTestXMLRPC\#testUploadNewPostWithGutenbergContent
```

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestXMLRPC.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestXMLRPC.java
@@ -109,6 +109,31 @@ public class ReleaseStack_PostTestXMLRPC extends ReleaseStack_XMLRPCBase {
     }
 
     @Test
+    public void testUploadNewPostWithGutenbergContent() throws InterruptedException {
+        String postContent = "<!-- wp:paragraph -->\n"
+                             + "<p>Paragraph 1</p>\n"
+                             + "<!-- /wp:paragraph -->";
+        // Instantiate new post
+        createNewPost();
+        mPost.setTitle("Gutenberg content");
+        mPost.setContent(postContent);
+
+        // Upload new post to site
+        uploadPost(mPost);
+
+        PostModel uploadedPostBeforeFetch = mPostStore.getPostByLocalPostId(mPost.getId());
+        // Make sure the uploaded post content is the same as the one we uploaded
+        assertEquals(postContent, uploadedPostBeforeFetch.getContent());
+
+        // Fetch the uploaded post from the server
+        fetchPost(mPost);
+
+        PostModel uploadedPostAfterFetch = mPostStore.getPostByLocalPostId(mPost.getId());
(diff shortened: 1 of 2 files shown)


maxme:
We would like to merge this before wpandroid 13.9 release.

maxme:
Also cc @aforcier if you remember why we added this (related to the hybrid editor maybe?).

aforcier:
> Also cc @aforcier if you remember why we added this in the first place (related to the hybrid editor maybe?).

I didn't really remember but I poked around a bit.

Looks like it did come along with the rest of the post upload logic we imported from WPAndroid, but the original code already existed over 7 years ago - must have been legacy editor related:

https://github.com/wordpress-mobile/WordPress-Android/blame/203a7ed19b1d070d5ba1a705a943b398f46ae3c6/WordPress/src/main/java/org/wordpress/android/ui/posts/services/PostUploadService.java#L311

maxme:
Thanks @aforcier , seems safe to remove it.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_PostTestXMLRPC.java
```diff
@@ -108,6 +108,31 @@ public void testUploadNewPost() throws InterruptedException {
...
     }
 
+    @Test
+    public void testUploadNewPostWithGutenbergContent() throws InterruptedException {
+        String postContent = "<!-- wp:paragraph -->\n"
+                             + "<p>Paragraph 1</p>\n"
+                             + "<!-- /wp:paragraph -->";
+        // Instantiate new post
+        createNewPost();
+        mPost.setTitle("Gutenberg content");
```
mchowning: Most just curious, but why did you add a title to the post for this test?

maxme: It's not required. It can be useful for debugging when you need to search for a specific post on a test site.

PR #1485, opened 2020-01-23
Add defences for bad APIs

0nko:
This PR makes all gateway response parameters nullable and even handles possible empty strings in place of integers.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt
```diff
@@ -75,10 +75,10 @@ constructor(
         @SerializedName("id") val gatewayId: String,
         @SerializedName("title") val title: String?,
         @SerializedName("description") val description: String?,
-        @SerializedName("order") val order: Int,
-        @SerializedName("enabled") val enabled: Boolean,
+        @SerializedName("order") val order: String?,
```
nbradbury: @0nko I feel like I need some background here. Why are we treating "order" as a string when the API docs say it's an integer?

0nko: @nbradbury I should have linked this, sorry - https://github.com/woocommerce/woocommerce-android/issues/1840

nbradbury: @0nko What will happen now if the response contains "order" as an int?

0nko: It'll work automatically, Gson will treat it as a string. I tested it to make sure but I haven't created a separate client PR because there's another FluxC PR I want to include in the hotfix. You can just reference this commit to test it yourself.

PR #1491, opened 2020-02-05
issue/1490-woo-product-sale-dates-gmt 

### LINKED ISSUE #1490: Woo: parse and persist GMT sale dates
Related to [this WCAndroid issue](https://github.com/woocommerce/woocommerce-android/issues/1790), we need to store the GMT product sale dates for proper localization.

nbradbury:
Closes #1490 by adding the GMT versions of product sale dates. Before this is merged, I think we need to discuss how to name date variables because we're inconsistent. For example, the order model has `dateCreated` and `dateModified` fields which are UTC dates, whereas the product model has these same fields but they're _not_ UTC.

My vote is to append `Gmt` to the field names to make it explicit. I've started a thread on Slack to discuss this.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,8 +1,5 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
-    <AndroidXmlCodeStyleSettings>
-      <option name="ARRANGEMENT_SETTINGS_MIGRATED_TO_191" value="true" />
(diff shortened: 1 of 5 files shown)


nbradbury:
> I noticed that when updating/displaying a product we are still using `date_on_sale_from` and `date_on_sale_to` instead of `date_on_sale_from_gmt` and `date_on_sale_to_gmt` . Will that change take place in another PR?

Yes, a separate WCAndroid PR is in the works for this.

PR #1505, opened 2020-02-26
Add bold type to formattable range types

renanferrari:
Related [WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/11382).

This PR adds the bold type to the list of supported formattable range types. 

This allows us to parse and show up-to-date definitions for bold text in notifications, such as the likes count (e.g. "You've received **5** likes on **Example Blog**").

Internal reference: `pauD4L-z9-p2#comment-567`


renanferrari:
Hi @develric. Thanks for reviewing this!

I had not added a test before because I felt it would be a bit redundant given the existing ones, but I guess it doesn't hurt trying to cover the new case we're introducing here, so I took your suggestion and created a new test for that. 👍

PR #1506, opened 2020-02-26
Fixing CI related to PR #1460.

develric:
This PR fixes an issue with test 

`ReleaseStack_ManageInsightsTestJetpack > testAddAndRemoveInsights` 

after we changed the `DEFAULT_INSIGHTS` list in [1460 PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1460/files)

This PR aligns the expected number of elements in the list due to the above mentioned modification.

Everyone can review this and 1 reviewer is enough. (thanks @aforcier for pinging on this 🙇‍♂️)

### To test
Run the ReleaseStack_ManageInsightsTestJetpack tests (I've done it locally) and check it passes.

### Note
Do not know why tbh (and possibly I'm just missing something obvious 🤷‍♂️), but rerunning the workflow from Circle CI console yesterday I got the CI suite passed, this is why I did not dig into (my bad) and supposed it had been a flaky test/timing. Leaving this as a note in case someone with more knowledge than me on this can have some thoughts eventually.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ManageInsightsTestJetpack.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ManageInsightsTestJetpack.kt
@@ -17,6 +17,7 @@ import org.wordpress.android.fluxc.store.AccountStore
 import org.wordpress.android.fluxc.store.AccountStore.AuthenticatePayload
 import org.wordpress.android.fluxc.store.AccountStore.OnAccountChanged
 import org.wordpress.android.fluxc.store.AccountStore.OnAuthenticationChanged
+import org.wordpress.android.fluxc.store.DEFAULT_INSIGHTS
 import org.wordpress.android.fluxc.store.SiteStore
 import org.wordpress.android.fluxc.store.SiteStore.OnSiteChanged
 import org.wordpress.android.fluxc.store.SiteStore.SiteErrorType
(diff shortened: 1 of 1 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ManageInsightsTestJetpack.kt
```diff
@@ -65,20 +65,20 @@ class ReleaseStack_ManageInsightsTestJetpack : ReleaseStack_Base() {
         val emptyStats = runBlocking { statsStore.getAddedInsights(site) }
 
         // Starts with 5 default blocks
-        assertEquals(emptyStats.size, 5)
+        assertEquals(emptyStats.size, 4)
```
aforcier: It might make sense to just use `DEFAULT_INSIGHTS.size` here so we're more flexible against this kind of change in the future - WDYT?

(And in the `statsWithFollowers` case below, `DEFAULT_INSIGHTS.size + 1`.)

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ManageInsightsTestJetpack.kt
```diff
@@ -65,20 +65,20 @@ class ReleaseStack_ManageInsightsTestJetpack : ReleaseStack_Base() {
         val emptyStats = runBlocking { statsStore.getAddedInsights(site) }
 
         // Starts with 5 default blocks
```
aforcier: Should update this comment too. 😀

aforcier:
> Do not know why tbh (and possibly I'm just missing something obvious 🤷‍♂️), but rerunning the workflow from Circle CI console yesterday I got the CI suite passed, this is why I did not dig into (my bad) and supposed it had been a flaky test/timing. Leaving this as a note in case someone with more knowledge than me on this can have some thoughts eventually.

Hmm 🤔 which workflow failed/did you rerun? The connected tests aren't run on every PR but instead are a job running once nightly. Did something else fail and then pass CI?

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ManageInsightsTestJetpack.kt
develric: Makes sense 👍! Changed here ef3554c364eb1e3913ad6d1e477cd24b8d6a7401

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ManageInsightsTestJetpack.kt
develric: Done here ef3554c364eb1e3913ad6d1e477cd24b8d6a7401 😊

develric:
Hi @aforcier 👋! Thanks for the review! I followed up to your suggestions, let me know thanks 😊.

> Hmm 🤔 which workflow failed/did you rerun? The connected tests aren't run on every PR but instead are a job running once nightly. Did something else fail and then pass CI?

I think you are right 👍! I was looking to this two run below:

| image | image |
|---|---|

So the temporary issue was with `mocked connected test` and not the nightly `connected test`, good catch. 
I was assuming all connected tests were running on PRs, thanks for clarification. I will run local tests next time 😊.

aforcier:
> So the temporary issue was with mocked connected test and not the nightly connected test

Interesting! That's a recently added mocked test that flaked, I'll check on it.

aforcier:
Working well, :shipit: - thanks @develric

PR #1508, opened 2020-03-05
Unescape author display name from the REST response

maxme:
This will fix at least one issue in WPAndroid: https://github.com/wordpress-mobile/WordPress-Android/issues/10410 -

I also checked if there was other fields we could unescape in this response, I think only the author display name should be unescaped.

Note: the XMLRPC response doesn't contain the author display name, so nothing to do there.


malinajirka:
LGTM, thanks!

PR #1510, opened 2020-03-12
[Site Creation] Step 1: Remove SegmentPromptModel 

ashiagr:
Since we are removing verticals header and search input segments from Site Creation Step 1 in https://github.com/wordpress-mobile/WordPress-Android/pull/11434, as part of  https://github.com/wordpress-mobile/WordPress-Android/pull/11421  we won't need `SegmentPromptModel` anymore. 

This PR removes verticals segment prompt endpoint, model, action and rest API call. It also removes corresponding tests from `ReleaseStack_VerticalTest`.

I am creating this as a draft PR as https://github.com/wordpress-mobile/WordPress-Android/pull/11421 is not merged yet. We need to remove everything related to Verticals before merging this PR in FluxC.


malinajirka:
I've updated the target branch to `site_creation_verticals_site_info_removal` so this can be merged now.

PR #1525, opened 2020-03-24
Handle authentication for self-hosted sites in ReactNativeStore

mchowning:
Adding authentication handling for the rest API using a [nonce endpoint](https://developer.wordpress.org/reference/functions/wp_ajax_rest_nonce/). The flow looks like this:

1. If we have a rest endpoint saved use that, otherwise use discovery to find the rest endpoint;
2. Attempt to make the requested call;
3. If the call fails because the endpoint is "not found" and we are using a previously saved rest endpoint, clear the previously saved rest endpoint from storage and use discovery to find the rest endpoint and try the call again;
4. If the call fails because of a lack of authorization, clear any saved nonce and try to get a new nonce from the nonce endpoint. If the attempt to get a new nonce endpoint fails, use `wp-json/` as a fallback endpoint, and try again the call.

As part of this PR, I have simplified the external interface for these calls. Instead of having separate methods for WP.com and self-hosted sites, we now just have a single method that accepts a site and a path; the logic for appropriately handling that is now on the FluxC side (as a result, the [WPAndroid side](https://github.com/wordpress-mobile/WordPress-Android/pull/11522) is much simpler now).

#### Only try nonce endpoint if WP ≥ 5.3

It would be good if we only tried to use the nonce endpoint if the version of WordPress running on the site was ≥ 5.3 since that endpoint doesn't exist in WP < 5.3. I couldn't see an obvious way to check this however (for example, [`SiteModel.mSoftwareVersion`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/7e50af2dc11cd5587dc497826c1dd6994c18ced4/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java#L63) looked like it might be promising, but it was always `null` in my limited testing). Under the current implementation, after we fail to fetch a nonce for any reason other than a lack of an internet connection we won't try again in the next 5 minutes (or until the app is restarted since I'm just storing the nonce in memory).

Although not a part of this PR, in the future when WP < 5.3 (or when the nonce endpoint fails) we could try to scrape the nonce from the editor page. This is obviously a pretty big hack 😬 , but it has been tested out with some success [on iOS](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/202/files#diff-23bcb2622a43c4aa81dc5ab5c4540fccR204-R215).

#### Handling persistent failures to fetch nonce

~If our attempts to fetch a nonce for a particular site consistently fail, this implementation will currently try to retrieve the nonce again with every new request. This obviously is not a good user experience. iOS is currently addressing this by [only storing the nonce in memory for the life of the app](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/202/files#diff-23bcb2622a43c4aa81dc5ab5c4540fccR131), so only the first call after an app restart will try to fetch the nonce. This seems like a reasonable solution (especially since it is relatively simple), but it does require special handling if the attempt to retrieve the nonce fails because there is no internet connection (and maybe other scenarios I haven't thought of).~

~I haven't moved forward with adding any special retry handling yet because I would like to get some more thoughts on the best way to handle persistence and retrying.~ I'm moving forward with the same in-memory cache approach iOS has used because I'm noticing that sometimes get a 403 Forbidden response when trying to reuse the nonce after restarting the app.

#### Insure new e2e tests pass

Run `./gradlew cAT -Pandroid.testInstrumentationRunnerArguments.class=org.wordpress.android.fluxc.release.ReleaseStack_ReactNativeWPAPIRequestTest`

#### Test in WPAndroid

1. Check out the related WPAndroid PR (https://github.com/wordpress-mobile/WordPress-Android/pull/11522)
2. Open a new post with the block editor
3. Add an image block
4. Add an image to that block from your WordPress Media Library
5. Open the settings for the image block
6. Insure that you can change the image size (thumbnail images should appear smaller and the block's HTML should reflect the updated image size and image URL). If the call fails, there will be no option to change the image size.


fluxc/src/main/java/org/wordpress/android/fluxc/store/ReactNativeStore.kt
```diff
@@ -21,19 +24,152 @@ class ReactNativeStore
...
+                    ?: "${site.url}/wp-json/" // fallback to ".../wp-json/" default if discovery fails
+        }
+
+        val fullRestUrl = "$baseRestUrl/$path".replace("/+".toRegex(), "/")
+        val response = executeFetch(site, baseRestUrl, fullRestUrl, params, site.cookieNonce)
+
+        return@withContext when (response) {
+            is Success -> response
+
+            is Error -> when (response.statusCode()) {
```
mchowning: Instead of testing the response's status code directly here, I considered just checking the error type ([`GenericErrorType`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/0b047695e902489dec1ec4f09712c67d6c247e35/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java#L99-L120)). I was concerned about that making it less clear exactly which errors were being handled though, so I just went with the more direct (and clearer to me at least) approach of using the status code directly. Let me know if you think that's not the best approach.

mchowning:
:wave: @maxme  and @planarvoid ! You may want to hold off on giving this any in-depth review at this time because I'm working on one (maybe two) reasonably signification changes to this PR:

1. Switching to only persisting the nonce in memory (instead of in the db). I think I have this pretty much done.
2. Solving the problem of authenticating with sites that have a redirect. The redirect works fine with unauthenticated requests, and the redirect to the proper `/wp-login.php` location is working, but the encoded body (with the username and password) is not getting sent with the redirect so the authentication is failing for me. This seems like a tricky problem, so I may end up leaving redirect handling to be addressed in a follow-up PR.

 I'll ping you when I'm done with my changes.

mchowning:
:wave: @maxme and @planarvoid ! This is ready for review now. 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/reactnative/ReactNativeWPAPIRestClient.kt
```diff
@@ -1,37 +1,101 @@
...
     @Named("regular") requestQueue: RequestQueue,
-    userAgent: UserAgent
+    userAgent: UserAgent,
+    private val currentTimeMillis: () -> Long
 ) : BaseWPAPIRestClient(dispatcher, requestQueue, userAgent) {
+    @Inject constructor(
+        wpApiGsonRequestBuilder: WPAPIGsonRequestBuilder,
+        wpApiEncodedBodyRequestBuilder: WPAPIEncodedBodyRequestBuilder,
+        dispatcher: Dispatcher,
+        @Named("regular") requestQueue: RequestQueue,
```
maxme: Note that we use the `@Named("custom-ssl")` provider to support self signed SSL in `ReleaseNetworkModule` (the client builder providers are defined in `ReleaseOkHttpClientModule`). I guess WPAPI requests on self hosted with a self signed certificate won't work.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ReactNativeWPComRequestTest.kt
```diff
@@ -19,22 +19,26 @@ class ReleaseStack_ReactNativeWPComRequestTest : ReleaseStack_WPComBase() {
...
     @Test
     fun testWpComCall_fails() {
-        val url = "https://public-api.wordpress.com/wp/v2/sites/${sSite.siteId}/an-invalid-extension"
-        val response = runBlocking { reactNativeStore.performWPComRequest(url, emptyMap()) }
-
+        val response = runBlocking { reactNativeStore.executeRequest(sSite, "an-invalid-extension") }
         val assertionMessage = "Call should have failed with a 404, instead response was $response"
         val actualStatusCode = (response as? Error)?.error?.volleyError?.networkResponse?.statusCode
-        Assert.assertEquals(assertionMessage, actualStatusCode, 404)
+        assertEquals(assertionMessage, actualStatusCode, 404)
```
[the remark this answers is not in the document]
mchowning: You're right! Done in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1525/commits/e63f4f5e040a1c5eefaf75e703521151a7c46f20.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ReactNativeStore.kt
```diff
@@ -18,22 +26,166 @@ import javax.inject.Singleton
...
+    }
+
+    private suspend fun executeWPAPIRequest(
+        site: SiteModel,
+        path: String,
+        params: Map<String, String>
+    ): ReactNativeFetchResponse {
+        val usingSavedRestUrl = site.wpApiRestUrl != null
+        if (!usingSavedRestUrl) {
+            site.wpApiRestUrl = discoveryWPAPIRestClient.discoverWPAPIBaseURL(site.url) // discover rest api endpoint
```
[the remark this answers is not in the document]
maxme: Looks good, I also double checked with Stetho that we were not doing extra discovery requests.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/reactnative/ReactNativeWPAPIRestClient.kt
maxme: I updated the site version and everything works fine on the example app, but the test is failing. 

I think we need to add a step in that test case to add the self signed certificate to our local db (this step is done manually on the example app via a dialog). I have to double check and see how we can change that test.

maxme: I had a look at the XMLRPC test suite. We have to do a first request to get the ssl error, then call the [MemorizingTrustManager to store the certicate](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/26de55dab75d0b9026107916aa0d6a1070d94717/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestXMLRPC.java#L120) (from the previous error) in our whitelist, after this, we can do a second request and it should work fine.

I actually tried that in `testAuthenticatedCallToSelfSignedSslSite` but I get an authentication issue on the second call. I'll investigate more later.

maxme: So I spent way too much time on this, but this works now :). I updated the test in 9cf460b1f - but it wasn't working.

First, the property `TEST_WPORG_URL_SH_SELFSIGNED_SSL` should be updated to `https://do2.wpmt.co/sslselfsigned/`, because vanilla is configured to redirect to the correct `do.wpmt.co` domain. This wasn't an issue with the existing test but will be a problem for this one.

Secondly, there was something wrong on https://do2.wpmt.co/sslselfsigned/, the REST API was not working even after upgrading to WordPress v5.4).

Screenshot 2020-04-08 at 11 55 06

I installed a new WordPress instance, and I still get a 404 on https://do2.wpmt.co/sslselfsigned/wp-json/ - **but** the discovery process is smarter than me and found that the url was `https://do2.wpmt.co/sslselfsigned/index.php/wp-json/`, the test now passes.

mchowning: Tested this out on my end with the updated tests.properties and everything is working well for me. 🎉 

Interestingly, I was having some trouble when I was using the trust manager yesterday, but I figured out today that it was something with my Charles Proxy configuration that was causing the problem, and just turning off the proxy got my test to pass.

maxme: Ah yep, you might not get the "invalid_ssl_certificate" error from the network level when you use Charles. Not that you can also use Stetho to look at network requests (but it's a pain with e2e tests because you need to set a breakpoint wait for the debugger to stop on it, then connect Chrome to the Stetho instance, continue on the debugger).

mchowning: That's a handy trick for using Stetho with e2e tests. I just gave up and used Charles. Thanks for sharing!

PR #1534, opened 2020-04-07
Add API tests against bleeding edge WC Rest API

bsessions85:
This PR adds in 27 new API tests that run against a site that is using the latest development version of the `woocommerce-rest-api`.

The tests will run daily by scheduled job that is configured in CircleCI config. The use Rest Assured to handle the calls and validation.

I think I hit every api call that is used in this project that hits the wc endpoints, but if you see any that I missed, please let me know.

### DIFF
--- a/tests/api/src/test/java/APITesting_WCOrder.java
+++ b/tests/api/src/test/java/APITesting_WCOrder.java
@@ -1,0 +1,316 @@
+import org.json.JSONObject;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.restassured.http.ContentType;
+import io.restassured.specification.RequestSpecification;
+import io.restassured.builder.RequestSpecBuilder;
+
+import static io.restassured.RestAssured.given;
+import static io.restassured.RestAssured.oauth2;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.hasSize;
+
+public class APITesting_WCOrder {
+    private RequestSpecification requestSpec;
+    private String ORDER_FIELDS = "id,number,status,currency,date_created_gmt,total,total_tax,shipping_total," +
+            "payment_method,payment_method_title,prices_include_tax,customer_note,discount_total," +
+            "coupon_lines,refunds,billing,shipping,line_items,date_paid_gmt,shipping_lines";
+    private String TRACKING_FIELDS = "tracking_id,tracking_number,tracking_link,tracking_provider,date_shipped";
+
+    @Before
+    public void setup() {
+        Map<String, String> pathParams = new HashMap<String, String>();
+        pathParams.put("json", "true");
+        pathParams.put("locale", "en_US");
+        pathParams.put("status", "any");
+
+        RequestSpecBuilder requestBuilder = new RequestSpecBuilder().
+            setBaseUri("https://public-api.wordpress.com").
+            setBasePath("rest/v1.1/jetpack-blogs/173063404/rest-api/").
+            addQueryParams(pathParams).
+            setAuth(oauth2(System.getenv("API_TEST_OAUTH_KEY")));
+        this.requestSpec = requestBuilder.build();    
+    }
+
+    @Test
+    public void canGetAllOrders() {
+        given().
+            spec(this.requestSpec).
+            queryParam("path", "/wc/v4/orders&per_page=50&offset=0&_fields=" + ORDER_FIELDS).
+        when().
+            get().
+        then().
+            statusCode(200).
+            body("data", hasSize(50));
+    }
+
+    @Test
+    public void canGetOrderListSummary() {
+        given().
+            spec(this.requestSpec).
+            queryParam("path", "/wc/v4/orders&per_page=50&offset=0&status=failed&_fields=id,"
+                + "date_created_gmt,date_modified_gmt").
+        when().
+            get().
+        then().
+            statusCode(200).
+            body("data", hasSize(10));
+    }
+
+    @Test
+    public void canGetOrdersByID() {
+        given().
+            spec(this.requestSpec).
+            queryParam("path", "/wc/v4/orders&per_page=50&include=625,628,618").
+        when().
+            get().
+        then().
+            statusCode(200).
+            body("data", hasSize(3),
+                "data.find { it.id == 618 }.status", equalTo("processing"),
+                "data.find { it.id == 625 }.status", equalTo("failed")
+            );
+    }
+
+    @Test
+    public void canGetOrderStatusOptions() {
+        given().
+            spec(this.requestSpec).
+            queryParam("path", "/wc/v4/reports/orders/totals").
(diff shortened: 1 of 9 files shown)


rachelmcr:
What do you think of adding something like a Slack notification, at least when tests fail? That could help with visibility of the results from these daily runs.

anitaa1990:
@bsessions85, I noticed a few endpoints missing from the tests so thought I would add them here:

- [Update Product](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/a9561376c1093cd36113278fecbd2307cbe79312/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt#L358-L357) 
- [GET all refunds for an order](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/72b2b781475c830c24ba8044094472f5332913ea/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt#L112)
- [GET a single refund by ID](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/72b2b781475c830c24ba8044094472f5332913ea/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt#L88)
- [Refund an order](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/72b2b781475c830c24ba8044094472f5332913ea/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt#L65)
- [GET all payment gateways supported for the site.](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/73fb317c2083c6d3afcb90e883e04a3819b1df4e/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt#L52)
- [GET a single payment gateway by ID](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/73fb317c2083c6d3afcb90e883e04a3819b1df4e/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt#L29)

Stats endpoints were missing as well but we use the V4 endpoints and not sure if that comes under the scope of this PR.

bsessions85:
> What do you think of adding something like a Slack notification, at least when tests fail? That could help with visibility of the results from these daily runs.

@rachelmcr  Good idea. Do you have a place set up already where it would make sense to send the notification?

rachelmcr:
> Do you have a place set up already where it would make sense to send the notification?

Since these tests are specific to WooCommerce it might make sense to send them to the `#mobile-woo` channel. @anitaa1990 would that make sense to you or is there a better channel (or should we create a new channel) for that kind of notification?

bsessions85:
@anitaa1990 I have added those missing tests. Thanks for letting me know they were missing!

anitaa1990:
>  @anitaa1990 would that make sense to you or is there a better channel (or should we create a new channel) for that kind of notification?

💯agree that #mobile-woo would be great in this case 👍

rachelmcr:
Thanks, Anitaa! I set up a webhook saved in the CircleCI env var `SLACK_WOO_WEBHOOK` that we can use for that notification.

bsessions85:
I have added the slack notification for failures. Thanks for setting up the var @rachelmcr !

rachelmcr:
@bsessions85 Sorry for not getting back to this one sooner! I just realized it's still open and added myself as a reviewer so I won't lose track of it again. :)

Quick question: Is there a straightforward way to kick off a local test run for these API tests? I was thinking I should re-run them now to make sure they're still passing, but it seems like it'd also be useful for debugging failing tests in the future.

bsessions85:
@rachelmcr To run the tests locally, you need to set the `API_TEST_OAUTH_KEY` and `API_TEST_SITE_URL` env variables and then run the tests using `./gradlew :tests:api:test`

Values for the env variables can be found in ss `Mobile API Test key`

tests/api/src/test/java/APITesting_WCOrder.java
```diff
@@ -0,0 +1,314 @@
...
+            body("data", hasSize(1),
+                "data[0].tracking_number", equalTo("324567890234567")
+            );
+    }
+
+    @Test
+    public void canGetOrderShipmentProviders() {
+        given().
+            spec(this.mRequestSpec).
+            queryParam("path", "/wc/v4/orders/635/shipment-trackings/providers").
```
rachelmcr, line 258: As with the `orders/<id>/shipment-trackings` endpoint, it looks like `orders/<id>/shipment-trackings/providers` also isn't supported in `v4`. It would be good to double check the expected behavior but for now using `v3` resolves the failure I'm seeing:

```suggestion
            queryParam("path", "/wc/v3/orders/635/shipment-trackings/providers").
```

tests/api/src/test/java/APITesting_WCProduct.java
```diff
@@ -0,0 +1,368 @@
...
+        given().
+            spec(this.mRequestSpec).
+            header("Content-Type", ContentType.JSON).
+            queryParam("path", path).
+            queryParam("_method", method).
+            body(jsonObj.toString()).
+        when().
+            post().
+        then().
+            statusCode(200);
```
rachelmcr, line 174: This test is failing at this point for me, with the following error:

`Expected status code <200> but was <400>.`

Nothing is jumping out at me immediately as an issue here; do you see the same error on your end?

tests/api/src/test/java/APITesting_WCRefund.java
```diff
@@ -0,0 +1,120 @@
...
+        given().
+            spec(this.mRequestSpec).
+            header("Content-Type", ContentType.JSON).
+            queryParam("path", path).
+            queryParam("_method", method).
+            body(jsonObj.toString()).
+        when().
+            post().
+        then().
+            statusCode(200);
```
rachelmcr, line 118: This test is failing at this point for me, with this error:

`Expected status code <200> but was <500>.`

Have you seen that failure? I'm not sure what's going on there.

tests/api/src/test/java/APITesting_WCOrder.java
```diff
@@ -0,0 +1,314 @@
...
+        when().
+            post().
+        then().
+            statusCode(200).
+        extract().
+            path("data.tracking_id");
+        
+        // Delete tracking number    
+        method = "delete";
+        path = "/wc/v4/orders/634/shipment-trackings/" + tracking + "&_method=" + method;
```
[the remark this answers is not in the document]
bsessions85: yes, good idea

bsessions85:
Thanks @rachelmcr . I knew most of these were failing on v4, because they aren't in the schema, so I switched them to v3 now. I also updated the path variable as suggested and the other failures you were seeing were I think issues with needing two paths in the same test to set up data. Right now they are all passing for me.

PR #1536, opened 2020-04-09
Remove usage of global scope

planarvoid:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/9527

It's not advised to use GlobalScope in your code so in this PR I'm replacing it with a coroutine scope of the Coroutine engine. In the end it behaves the same way - when CoroutineEngine (and the Store that owns it) gets killed, so gets killed the coroutine. Let me know what you think of this approach.

This could be tested with the WPAndroid PR - https://github.com/wordpress-mobile/WordPress-Android/pull/11657


fluxc/src/main/java/org/wordpress/android/fluxc/store/VerticalStore.kt
```diff
@@ -10,26 +8,27 @@ import org.wordpress.android.fluxc.action.VerticalAction
...
     @Subscribe(threadMode = ThreadMode.ASYNC)
     override fun onAction(action: Action<*>) {
         val actionType = action.type as? VerticalAction ?: return
 
-        GlobalScope.launch(coroutineContext) {
-            val onChanged = when (actionType) {
-                VerticalAction.FETCH_SEGMENTS -> fetchSegments()
+        when (actionType) {
+            VerticalAction.FETCH_SEGMENTS -> {
+                coroutineEngine.launch(AppLog.T.API, this, "FETCH_SEGMENTS") {
```
planarvoid, line 28: I think the previous was wrong to wrap the whole `when` in the GlobalScope.

PR #1538, opened 2020-04-10
Issue/1535 filter products

anitaa1990:
Fixes #1535 by adding option to filter products. Product filter options currently supported are: product status, product type and stock status.

#### Screenshots
. 

#### To test
- Run `ProductSqlUtilsTest`
- Run `WCProductStoreTest`
- Run the example app, click on `Woo` -> `Products` - `Fetch products with filters` and select filters and click on `Filter Products`.
- Verify that the products fetched match the filters on the web.

PR #1542, opened 2020-04-13
issue/product-filter-sorted

nbradbury:
In #1541 we added case-insensitive sorting of products, but #1538 was created before that and needed to be updated to support case-insensitive sorting of filtered products.

This small PR adds that.


nbradbury:
Closing this in favor of another PR which handles non-alpha sorting.

PR #1543, opened 2020-04-13
Issue/woo product sorting take2

nbradbury:
This PR updates the recently-merged product filtering (#1538) to support case-insensitive sorting by name.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
@@ -178,9 +178,9 @@ class MockedStack_WCProductsTest : MockedStack_Base() {
         assertEquals(productsFromDb.size, 3)
 
         // verify that the products are correctly sorted
-        assertEquals(productsFromDb.get(0).name, "aaa test product")
-        assertEquals(productsFromDb.get(1).name, "Booklet")
-        assertEquals(productsFromDb.get(2).name, "Test Product")
+        assertEquals("aaa test product", productsFromDb.get(0).name)
+        assertEquals("Booklet", productsFromDb.get(1).name)
+        assertEquals("Test Product", productsFromDb.get(2).name)
     }
 
     @Test
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
@@ -178,9 +178,9 @@ class MockedStack_WCProductsTest : MockedStack_Base() {
         assertEquals(productsFromDb.size, 3)
 
         // verify that the products are correctly sorted
-        assertEquals("aaa test product", productsFromDb.get(0).name)
-        assertEquals("Booklet", productsFromDb.get(1).name)
-        assertEquals("Test Product", productsFromDb.get(2).name)
+        assertEquals("aaa test product", productsFromDb[0].name)
+        assertEquals("Booklet", productsFromDb[1].name)
+        assertEquals("Test Product", productsFromDb[2].name)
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -101,10 +101,28 @@ object ProductSqlUtils {
...
             DATE_ASC, DATE_DESC -> WCProductModelTable.DATE_CREATED
         }
-        return queryBuilder
+
+        val products = queryBuilder
                 .endGroup().endWhere()
                 .orderBy(sortField, sortOrder)
                 .asModel
+
+        return if (sortType == TITLE_ASC || sortType == TITLE_DESC) {
```
AmandaRiu, line 110: It won't let me comment on it since it wasn't included in the changes, but shouldn't `TITLE_ASC` and `TITLE_DESC` be removed from the when statement starting on line 96 above? Wouldn't it cause more work to happen for the database to first sort it, and then sort it again in the code below?

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
```diff
@@ -176,6 +176,11 @@ class MockedStack_WCProductsTest : MockedStack_Base() {
         val productsFromDb = ProductSqlUtils.getProductsForSite(siteModel)
         assertNotNull(productsFromDb)
         assertEquals(productsFromDb.size, 3)
+
+        // verify that the products are correctly sorted
+        assertEquals(productsFromDb.get(0).name, "aaa test product")
+        assertEquals(productsFromDb.get(1).name, "Booklet")
+        assertEquals(productsFromDb.get(2).name, "Test Product")
```
AmandaRiu, line 183: np: The arguments for assert are backwards [`assertEquals(expected, actual)`] and we should use the index syntax to fetch the product from the collection. 
Example - replace:
```
assertEquals(productsFromDb.get(0).name, "aaa test product")
```
with:


```
assertEquals("aaa test product", productsFromDb[0].name)
```
Otherwise when it fails the fail message doesn't make much sense.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
nbradbury: I wondered about that, too, but I figured SQLite is faster at sorting than our code is. I don't think we're going to see a ton of product names that start with a lowercase letter, in which case the products will already be sorted. Which means our sorting will be unlikely to do much work other than simply scan the list of products.

But I'm happy to change this - either way works for me.

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
nbradbury: Whoops. Fixed in 0d29342.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
AmandaRiu: Oh that's a good point 👍

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
planarvoid: It's not my place to comment here but I recommend using `org.assertj.core.api.Assertions.assertThat` instead of all the asserts, it gives you nice clarity :-)

PR #1550, opened 2020-04-17
Issue/1549 woo enable reviews and purchase note

### LINKED ISSUE #1549: Woo: add reviews_allowed when updating a product
In order to support [this WCAndroid issue](https://github.com/woocommerce/woocommerce-android/issues/2271), we need to add reviews_allowed when updating a product. We also need to add the purchase note to support [this WCAndroid issue](https://github.com/woocommerce/woocommerce-android/issues/2272).

nbradbury:
Closes #1549 - adds `reviews_allowed` and `purchase_note` when updating a product. These two fields have also been added to the update product test and update product fragment.


nbradbury:
> It would be great to update review allowed and purchase note when [fetching product detail in `WooUpdateProductFragment`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/906f6d2a2f8855798afe7df1e849894233d09897/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt#L258)

Done in 02a84e2.

PR #1555, opened 2020-04-21
Add URL to CommentModel

zwarm:
This PR supports [#8319](https://github.com/wordpress-mobile/WordPress-Android/issues/8319) by adding URL to the CommentModel and updating the CommentRestClient and CommentXMLRPCClient  with the associated URL property.

Testing
- Unit test:  `CommentStoreUnitTest` was updated with the new column and all its tests passed.
- Instrumented Tests : Added new tests to `ReleaseStack_CommentTestWPCom` and `ReleaseStack_CommentTestXMLRPC`  to verify that the api layer still functions accurately and returns the URL. All tests passed. 
-  Example app tests: Updated  `CommentsFragment` to show/log the comment URL.
-  Tested in WPAndroid using hashed version of this branch and received URL in the response.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_CommentTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_CommentTestWPCom.java
@@ -395,7 +395,7 @@ public class ReleaseStack_CommentTestWPCom extends ReleaseStack_WPComBase {
         createNewComment();
 
         // Edit comment instance
-        mNewComment.setContent("Trying with: " + (new Random()).nextFloat() * 10 + " gigawatts");
+        mNewComment.setContent("Trying with 10,000 gigawatts");
 
         // Create new Comment
         mNextEvent = TestEvents.COMMENT_CHANGED;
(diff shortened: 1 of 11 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_CommentTestWPCom.java
```diff
@@ -389,6 +389,27 @@ public void testEditInvalidComment() throws InterruptedException {
         assertTrue(mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
     }
 
+    @Test
+    public void testCommentResponseContainsURL() throws InterruptedException {
+        // New Comment
+        createNewComment();
+
+        // Edit comment instance
+        mNewComment.setContent("Trying with: " + (new Random()).nextFloat() * 10 + " gigawatts");
```
malinajirka, line 398: Just sharing:  I like to keep the tests as determinate as possible. Using a random value (Random, currentTime etc) makes the test indeterminate. Integrations tests are indeterminate by their nature, so it's not a big deal. This practice make sense especially in unit tests.

zwarm: You're right and that's good advice and in this case I can use a static comment. I'll take out the random.

PR #1563, opened 2020-04-22
Issue/1562 password protected products

### LINKED ISSUE #1562: Woo: enable password-protecting products
This isn't supported in the WooCommerce API but it can be done via the WordPress API.

nbradbury:
Closes #1562 - adds the ability to set a password for a product via the WP.com API. Tests are included.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -354,7 +354,7 @@ class WCProductStore @Inject constructor(dispatcher: Dispatcher, private val wcP
 
     class OnProductPasswordChanged(
         var remoteProductId: Long,
-        var password: String
+        var password: String?
     )  : OnChanged<ProductError>() {
         var causeOfChange: WCProductAction? = null
     }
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -167,6 +178,36 @@ class WCProductStore @Inject constructor(dispatcher: Dispatcher, private val wcP
         }
     }
 
+    class RemoteProductPasswordPayload(
+        val remoteProductId: Long,
+        val site: SiteModel,
+        val password: String
```
AmandaRiu, line 184: Would we just set this as an empty string to remove the password? Or should this be nullable?

nbradbury: Yes, just set this to an empty string to remove the password.

nbradbury:
> Changes _look_ good, but it'd be really helpful to have a way to test this in the example app so I can play with it under various scenarios

@AmandaRiu I added the password to the update product fragment. I'm not sure why the test is failing - it passes here. Do you know whether it's the fetch password or update password that's failing?

Screen Shot 2020-04-23 at 8 29 51 AM

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
AmandaRiu: Cool. Just wanted to make sure 👍

nbradbury:
@AmandaRiu I think we're good to merge this. As we discovered in our Slack conversation, the problem was due to the product status being private. We'll just have to make sure that  WCAndroid changes the status if a user enters a password.

PR #1573, opened 2020-05-06
Fix geolocation disappearance from app on save 

ashiagr:
This PR sets post latitude, longitude from the `metadata` param if post `geo` param is null to fix https://github.com/wordpress-mobile/WordPress-Android/issues/7626

### To Test
0. Use commit hash from this PR for `fluxCVersion` in [WordPress-Android](https://github.com/wordpress-mobile/WordPress-Android), build and install the app on device.
1. Create a new post in the mobile app on a Jetpack site.
2. Go to post settings and set the location for that post.
3. Save the post as a draft.
4. When you get back to post settings, the geolocation data should be displayed.

location

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
@@ -52,6 +52,8 @@ import org.wordpress.android.fluxc.store.PostStore.PostError;
 import org.wordpress.android.fluxc.store.PostStore.PostListItem;
 import org.wordpress.android.fluxc.store.PostStore.RemoteAutoSavePostPayload;
 import org.wordpress.android.fluxc.store.PostStore.RemotePostPayload;
+import org.wordpress.android.util.AppLog;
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
```diff
@@ -426,6 +427,21 @@ private PostModel postResponseToPostModel(PostWPComRestResponse from) {
...
             post.setLatitude(from.getGeo().latitude);
             post.setLongitude(from.getGeo().longitude);
+        } else {
+            List<PostMetaData> metaDataList = from.getMetadata();
+            if (metaDataList != null) {
+                for (PostMetaData metaData : metaDataList) {
+                    String key = metaData.getKey();
+                    if (key != null && metaData.getValue() != null) {
+                        if (key.equals("geo_longitude")) {
+                            post.setLongitude(Double.parseDouble(metaData.getValue()));
```
planarvoid, line 437: What do you think about being extra cautious here and catching here the `NumberFormatException` in case the value format changes?

PR #1581, opened 2020-05-15
Fix wpcom endpoint url

cameronvoell:
### Description
Change enables us to reach endpoint "/wpcom/v2/gutenberg/available-extensions" which is used in WPAndroid for checking which jetpack blocks are enabled for a given site. 

### Related PR's
WPAndroid PR:
Gutenberg-Mobile PR:

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/ReactNativeStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/ReactNativeStore.kt
@@ -183,6 +183,7 @@ class ReactNativeStore
                 val url = path?.let {
                     val newPath = it
                             .replace("wp/v2".toRegex(), "wp/v2/sites/$wpComSiteId")
+                            .replace("wpcom/v2/", "wpcom/v2/sites/$wpComSiteId/")
                     slashJoin(WPCOM_ENDPOINT, newPath)
                 }
                 Pair(url, params)


fluxc/src/main/java/org/wordpress/android/fluxc/store/ReactNativeStore.kt
```diff
@@ -183,6 +183,7 @@ class ReactNativeStore
                 val url = path?.let {
                     val newPath = it
                             .replace("wp/v2".toRegex(), "wp/v2/sites/$wpComSiteId")
+                            .replace("wpcom/v2/", "wpcom/v2/sites/$wpComSiteId/")
```
mchowning, line 186: Not a big deal at all, but it might be worth making the two `replace` calls consistent with respect to whether they require and replace the trailing slash (i.e., `wp/v2` vs `wpcom/v2/`. Either works fine, I just think it would be nice if they are consistent.

mchowning (approved):
Works great! Thanks @cameronvoell ! :shipit: 

I went ahead and updated this branch to have your commit on top of `develop`. Hope you don't mind.

enejb:
I just wanted to note that core since 5.5 should now also have a generic endpoint that lets you see what blocks are available. See `/wp/v2/block-types` endpoint. 

What is the difference between the two? 

As I understand it, the core one (wp/v2/block-types) is more detailed. In that it provides information about all the different blocks. Whiles the Jetpack endpoint provides more info about blocks that the user is able to use. So it contains information about "Real" availability. In the sense that if the user doesn't have a specific plan they will not be able to really use that block until they upgrade to to a specific plan.

PR #1583, opened 2020-05-19
Add site homepage settings

planarvoid:
This functionality could be tested with the [WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/11951)

This PR sets up the domain layer for setting the Homepage settings. We use the `site/%id/homepage` endpoint to set the settings and retrieve it. There are 3 fields that we can set. The first one is `showOnFront` or `is_page_on_front`. It is called `showOnFront` in the `SiteModel` response and it has two values - `page` and `posts`. The `is_page_on_front` represents exactly the same thing but it's a boolean and it's `true` when the `showOnFront` is `page`. 

The two other fields `pageOnFront` (called `page_on_front_id` when setting the value) and `pageForPosts` (called `page_for_posts_id`). 

Overall, if you have the `showOnFront` set to `posts`, it means that your homepage is the list of your posts as in most of the blogs. 
If you have the `showOnFront` set to `page`, you can set the page on front which represents your static homepage and page for posts which is your new page that will contains your list of posts. 

As part of this PR I've added two things. One is a rest client and a store that allow you to set the values. The other part is that I've added the fields to the `SiteModel` to retrieve the value of all the three fields when we're fetching the site. 

This functionality cannot be remotely set on self-hosted sites yet so they are disabled.


develric:
Just for the record, tested it again after the merge from develop and confirms it works as expected 👍!

PR #1599, opened 2020-06-05
Woo/get product category by id

### LINKED ISSUE #1598: Woo: Product Categories - minor changes required
While working on the woocommerce/woocommerce-android#2187, I noticed that we need to:
- **change the default page size we use to fetch categories from 25 to 100**. This is because the API does not provide a way to order the categories that is needed by the mobile app i.e. there are instances when subcategories are returned in a different page of the request and it would not be a good user experience. The number 100 was chosen because it is the maximum number of categories we can fetch in a single request without throwing an error on the API side.
- Add a method to `WCProductStore` to fetch a category by it's ID. This is already available in `ProductSqlUtils` so we just need to add the method to `WCProductStore` so that it's available to the client side.

anitaa1990:
Fixes #1598 by adding the following the minor changes:
- **change the default page size we use to fetch categories from 25 to 100**. This is because the API does not provide a way to order the categories that is needed by the mobile app i.e. there are instances when subcategories are returned in a different page of the request and it would not be a good user experience. The number 100 was chosen because it is the maximum number of categories we can fetch in a single request without throwing an error on the API side. Fixed in d6a6122
- **Add a method to `WCProductStore` to fetch a category by it's ID.** This is already available in `ProductSqlUtils` so we just need to add the method to `WCProductStore` so that it's available to the client side. Fixed in 9f95e44
- **Added method to `WCProductStore` to fetch a category by it's name**: Fixed in 1398b42 and added an example in 44c567b. To test, click on `Woo` -> `Products` -> `Add Product Category` and verify that the category name you entered is displayed in the logs once the add category API is successful.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
@@ -70,6 +70,8 @@ class WooProductsFragment : Fragment() {
     private var pendingFetchProductShippingClassListOffset: Int = 0
     private var pendingFetchProductCategoriesOffset: Int = 0
 
+    private var enteredCategoryName: String? = null
+
     override fun onAttach(context: Context?) {
         AndroidSupportInjection.inject(this)
         super.onAttach(context)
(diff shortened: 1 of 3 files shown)


anitaa1990:
Ready for review @0nko 👍

PR #1603, opened 2020-06-11
Add support for users/%s/auth-options endpoint

renanferrari:
### Description

This PR introduces an implementation for the `users/%s/auth-options` endpoint. 

This endpoint accepts either a username or an email address and, given it matches an existing account, a response with two boolean fields is returned:
- `passwordless`: indicates if the user has set a password for their account or not.
- `email_verified`: indicates if the account's associated email is verified or not.

If no matching account is found, a `unknown_user` error is returned instead.

We are mostly interested in the `passwordless` field, as it will allow us to further optimize our login flows on WPAndroid, specially the ones where we currently ask for a password even when the user haven't set one for their account.

### To test

You can either run `ReleaseStack_AccountTest` or use the following steps on the example app:
- On the initial screen, tap **Signed Out Actions**.
- On the next screen, tap **Fetch Auth Options**.
- On the dialog, enter an username or email address associated with an existing WP.com account and tap **Ok**.
- Check if the information printed in the console matches the account.

Feel free to try out different account combinations or check `ReleaseStack_AccountTest` for the most common scenarios.


renanferrari:
I've added two new properties to `example/tests.properties-example` but I guess I should provide real values somewhere (mobile secrets?) so the CI tests can pass. I'm not familiar with this process, so any help here is appreciated.

fluxc/src/main/java/org/wordpress/android/fluxc/action/AccountAction.java
```diff
@@ -71,6 +72,8 @@
     UPDATE_SUBSCRIPTION_NOTIFICATION_POST,
     @Action
     FETCH_DOMAIN_CONTACT,
+    @Action(payloadType = String.class)
+    FETCH_AUTH_OPTIONS,
```
khaykov, line 76: While having just a String payload is absolutely fine, I feel like `newFetchAuthOptionsAction` doesn't really tell you what kind of input it expects. For example, looking at IS_AVAILABLE_EMAIL and `newIsAvailableEmailAction` you can tell that it expects email as a parameter. What do you think about something like `FETCH_AUTH_OPTIONS_FOR_EMAIL_OR_USERNAME`? It can be anything, as long as it hints on the content of the payload. Or we can have a payload class with descriptive variable names. Either way is fine :)

renanferrari: Hey @khaykov thanks for reviewing this. More descriptive names are always a good thing, so I totally agree with your suggestion and have now introduced a payload class with a descriptive parameter name. 

To follow the naming conventions we have in place, I first renamed the response payload from `AuthOptionsPayload` to `FetchAuthOptionsResponsePayload` and then created a new payload class for the fetch action called `FetchAuthOptionsPayload`, that receives a String parameter named `emailOrUsername`. Everything else remains the same.

renanferrari:
Merged the `develop` branch to fix some conflicts.

PR #1620, opened 2020-06-25
[Gutenberg] Adjust the fetching of custom colors

chipsnyder:
Fixes: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2432
WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/12289

To Test:
1. Log in to the demo app for a self-hosted site
2. Set a theme with custom colors and/or gradients on your site
3. In the demo app select Fetch theme from first site
**Expect** to see a log of colors and/or gradients
4. Set a theme with out custom colors and/or gradients on your site
5. In the demo app select Fetch theme from first site
**Expect** to see a log of `[null]` for colors and gradients

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/EditorThemeSqlUtils.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/EditorThemeSqlUtils.kt
@@ -92,15 +92,17 @@ class EditorThemeSqlUtils {
         override fun getId() = mId
 
         fun toEditorTheme(
-            storedColors: List<EditorThemeElementBuilder>,
-            storedGradients: List<EditorThemeElementBuilder>
+            storedColors: List<EditorThemeElementBuilder>?,
+            storedGradients: List<EditorThemeElementBuilder>?
         ): EditorTheme {
-            val colors = if (storedColors.count() > 0) storedColors.mapNotNull { it.toEditorThemeElement() } else null
-            val gradients: List<EditorThemeElement>?
-            if (storedColors.count() > 0) {
+            var colors: List<EditorThemeElement>? = null
+            if (storedColors != null && storedColors.count() > 0) {
+                colors = storedColors.mapNotNull { it.toEditorThemeElement() }
+            }
+
+            var gradients: List<EditorThemeElement>? = null
+            if (storedGradients != null && storedGradients.count() > 0) {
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/EditorThemeSqlUtils.kt
```diff
@@ -92,11 +92,17 @@ class EditorThemeSqlUtils {
...
 
         fun toEditorTheme(
-            storedColors: List<EditorThemeElementBuilder>?,
-            storedGradients: List<EditorThemeElementBuilder>?
+            storedColors: List<EditorThemeElementBuilder>,
+            storedGradients: List<EditorThemeElementBuilder>
         ): EditorTheme {
-            val colors = storedColors?.mapNotNull { it.toEditorThemeElement() }
-            val gradients = storedGradients?.mapNotNull { it.toEditorThemeElement() }
+            val colors = if (storedColors.count() > 0) storedColors.mapNotNull { it.toEditorThemeElement() } else null
```
mkevins, line 98: I wonder if we should keep this optional and also keep the guard, as well as checking the count, since we don't yet know why WellSql is occasionally sending null or empty. Wdyt?

chipsnyder: Yeah that would probably be the safer path. I’ll make that tweak

mkevins: I tested, and everything is working as expected now. The only thing I'm wondering about is, do we have a risk of NPE dereferencing the values returned from WellSql? (i.e. I think `storedColors` or `storedGradients` can be null, since the inferred platform type is `(MutableList<EditorThemeSqlUtils.EditorThemeElementBuilder!>..List<EditorThemeSqlUtils.EditorThemeElementBuilder!>?)`).

https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1620/files/20782cf123eda3e81d4812cd4dbcc360a4b6551a#diff-7279efeeae7d8b75a9f9ead365830ba1R50

I looked a bit deeper into the WellSql source, but couldn't be sure. I didn't see any nullability annotations though. So what I mean is, maybe we need to keep `storedColors` and `storedGradients` optional?

PR #1634, opened 2020-07-20
Support correct mime types

planarvoid:
This PR could be tested with [this WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/12488)

The reason for this PR is that the WordPress mime types used were not correctly matching the MimeType values for Android. That's why some of the file uploads were failing (for example audio files and documents).

This PR compares the WordPress supported file types for upload https://wordpress.com/support/accepted-filetypes/ and translates them to the Android supported mime types https://android.googlesource.com/platform/frameworks/base/+/cd92588/media/java/android/media/MediaFile.java) 
The resulting mimetypes should be the overlap for images, videos, audios and documents. 
I've extracted the mime types into a separate class with comments to make it easier to understand later. 
The mimetypes do not match the file extensions so this PR now differentiates between these two things.

This is a draft because I still have to run the tests and test it with WPAndroid changes. There is no need to review this PR now.


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/MediaSqlUtils.java
```diff
@@ -47,7 +47,7 @@
         return WellSql.select(MediaModel.class)
                 .where().beginGroup()
                 .equals(MediaModelTable.LOCAL_SITE_ID, site.getId())
-                .contains(MediaModelTable.MIME_TYPE, MediaUtils.MIME_TYPE_IMAGE)
+                .contains(MediaModelTable.MIME_TYPE, Type.IMAGE.getValue())
```
develric, line 50: Here and below we were matching on a slightly more restricting `LIKE`; that is (for example for audio) `audio/` vs current `audio`. It should not make much of a real difference in my understanding but I thought better to share in case it triggers a different opinion in you.

planarvoid: You're right and I think this approach is OK. The new code is little less restricting so it shouldn't affect the current behaviour. The only thing that might make this break is if the mime type was only "image" string. I think that's highly unlikely and I like the approach without the "/" better.

planarvoid:
> One more general thing I was thinking is about binary compatibility of this fluxc version with other apps that can use it (Woo and don't know if others). I'm thinking to public fields or methods like

> public static final String MIME_TYPE_IMAGE = "image/";
> in MediaUtils for example.

> Wondering if you already checked we are not breaking anything in Woo (or whatever other app can use this lib if it exists 😊 )

This is a great point! I haven't considered it. I'll ping someone from Woo to check whether they use the MediaStore and MediaUtils. Is that something you use @AmandaRiu ? Thanks!

ThomazFB:
Hey @planarvoid! Thank you so much for reaching out about this PR. I've done some analysis around how these changes will impact the Woo app, so far, there's only one code point where it would require us to adjust in order to handle the MimeType parameter changes [right here](https://github.com/woocommerce/woocommerce-android/blob/12d49dca63804f4b03497bff54a4291fa67e3702/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/wpmediapicker/WPMediaPickerRepository.kt#L57). 

Do you think would be possible for you to let us know when PR gets released so I can address a quick issue to properly reference this new implementation inside the Woo app?

planarvoid:
thanks so much @ThomazFB , we'll merge this, create a tag and I'll ping you so you can make the change 👍

PR #1641, opened 2020-07-24
Revert "Fix the failing API unit test"

0nko:
This reverts commit 3f6f5cb86ed74c825e2c41caa6146484fbd3dd31, which will fix the failing API Test.

@bsessions85 Do you know why the response keeps changing? Is this site (woomobileapitesting.mystagingwebsite.com) being used for anything else besides API unit tests?


bsessions85:
@0nko No this site isn't used for anything other than these tests. I am not sure how it would be giving a different result other then maybe something changed in the API and was changed back. We pull the plugin directly from Master branch of the API so it is changing quickly

PR #1642, opened 2020-07-25
Fix auth-options test failing on the release stack

renanferrari:
First, some context: the `testFetchAuthOptionsForUserWithUnverifiedEmail()` test was failing on the release stack with an `EMAIL_LOGIN_NOT_ALLOWED` error, which asks for an username to be provided instead of an email. This can happen under certain circumstances involving an unverified email as a security measure by the `users/%s/auth-options` endpoint, which will classify that email as "suspicious".

This PR does two things to address that:
1. It moves the `testFetchAuthOptionsForUserWithUnverifiedEmail()` test to the mocked stack.
1. It adds a new `testFetchAuthOptionsForUserWithSuspiciousEmail()` test to the mocked stack, in order to properly handle the `EMAIL_LOGIN_NOT_ALLOWED` error.

## To test

You can either run `MockedStack_AccountTest` and verify all tests pass or use the following steps on the example app:
1. On the initial screen, tap **Signed Out Actions**.
1. On the next screen, tap **Fetch Auth Options**.
1. On the dialog, enter an unverified email address and tap **Ok**.
1. Verify the app logs a successful response in the console and that the field `Is email verified` is `false`.
1. Tap **Fetch Auth Options** again.
1. On the dialog, enter an email address that is classified as "suspicious" and tap **Ok**.
1. Verify the app logs a failed response in the console and that the error is `EMAIL_LOGIN_NOT_ALLOWED`.


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_AccountTest.java
```diff
@@ -605,8 +597,9 @@ public void onAuthOptionsFetched(OnAuthOptionsFetched event) {
                     assertEquals(mNextEvent, TestEvents.FETCH_AUTH_OPTIONS_USER_WITH_PASSWORD);
                     mCountDownLatch.countDown();
                 } else {
-                    assertEquals(mNextEvent, TestEvents.FETCH_AUTH_OPTIONS_UNVERIFIED_EMAIL);
-                    mCountDownLatch.countDown();
+                    // We test for unverified emails in the mocked stack, otherwise we might run into a security issue
```
aforcier, line 600: We could explicitly fail here since we don't expect to ever visit this branch during tests. That way also the test would fail would a clear error instead of due to `mCountDownLatch` timing out.

Maybe something like

```kotlin
throw new AssertionError("Unexpectedly unverified e-mail");
```

renanferrari:
@aforcier Thank you very much for reviewing this! I have just updated the PR to address both of your suggestions.

aforcier:
Looks good @renanferrari! I see the unverified email property was removed from the secrets store as well 👍  :shipit:

PR #1646, opened 2020-07-29
[Feature] Add Search Filter for XML RPC Requests

ParaskP7:
Feature requirement for [WordPress-Android - #12058](https://github.com/wordpress-mobile/WordPress-Android/issues/12058)

Description
-----
This PR is the first step towards solving the aforementioned [WordPress-Android](https://github.com/wordpress-mobile/WordPress-Android) issue. Briefly, it enables `search` for XML RPC requests, just like it is already enabled for Rest.

As such, when this PR gets merged with `develop` it will unlock the second step, which is to make search on blog posts available for self-hosted sites.

Test
-----
1. Build the FluxC `example` app
1. Sing in and fetch your self-hosted site (click on the `SIGN IN AND FETCH SITES` button)
1. Navigate to the post screen (click on the `POST` button)
1. Try fetching the available posts from your self-hosted site (click on the `FETCH POST FROM FIRST SITE` button)
1. Observer the output, example: `Fetched 20 posts from: Traditional Wolf`
1. Find the `PostXMLPRCClient.java` class and add a breakpoint at line 628 (see `if (!TextUtils.isEmpty(searchQuery)) { ... }`)
1. Attach a debugger and try fetching the available posts from the self-hosted site again
1. When the debugger blocks, find the `searchQuery` variable. It should be `null`. Replace it with a search term of your own, like `"tags"`. Click enter to replace `null` and resume the debugger.
1. Observer the new output, example: `Fetched 3 posts from: Traditional Wolf`
1. The above means that while without a search query the list of posts returned was 20, when that search query got updated to `tags` (normally done through the UI), then the resulting list of posts is different. In my case, because my self-hosted site has 3 posts matching the `tags` search query it returned those (that is, instead of the default 20 post, which is the page size).

PR Structure
-----
- Multiple commits are included in this PR, both 'feature' and 'refactor' commits,
- Each commit tackles a specific problem, be it formatting the code, adding `search` capabilities, etc,
- Each commit has a `title`, an empty line and a optional `description` afterwards. The line length of a commit message does not exceed the 72 characters limit.
- Test where included, both unit and connected tests. Where possible a TDD approach was followed,
- Both, `checkstyle` and `ktlint` were triggered before pushing the branch. `checkstyle` had already 43 warnings (it is probably out of scope for this PR to fix those). However, for those files that were touched, some warnings was fixed.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/post/PostXMLRPCClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/post/PostXMLRPCClient.java
@@ -143,18 +143,11 @@ public class PostXMLRPCClient extends BaseXMLRPCClient {
         SiteModel site = listDescriptor.getSite();
         List<String> fields = Arrays.asList("post_id", "post_modified_gmt", "post_status");
         final int pageSize = listDescriptor.getConfig().getNetworkPageSize();
-        List<Object> params = createFetchPostListParameters(
-                site.getSelfHostedSiteId(),
-                site.getUsername(),
-                site.getPassword(),
-                false,
-                offset,
-                pageSize,
-                listDescriptor.getStatusList(),
-                fields,
-                listDescriptor.getOrderBy().getValue(),
-                listDescriptor.getOrder().getValue(),
-                listDescriptor.getSearchQuery());
+        List<Object> params =
+                createFetchPostListParameters(site.getSelfHostedSiteId(), site.getUsername(), site.getPassword(), false,
+                        offset, pageSize, listDescriptor.getStatusList(), fields,
+                        listDescriptor.getOrderBy().getValue(), listDescriptor.getOrder().getValue(),
+                        listDescriptor.getSearchQuery());
         final boolean loadedMore = offset > 0;
 
         final XMLRPCRequest request = new XMLRPCRequest(site.getXmlRpcUrl(), XMLRPC.GET_POSTS, params,
@@ -187,18 +180,9 @@ public class PostXMLRPCClient extends BaseXMLRPCClient {
 
     public void fetchPosts(final SiteModel site, final boolean getPages, List<PostStatus> statusList,
                            final int offset) {
-        List<Object> params = createFetchPostListParameters(
-                site.getSelfHostedSiteId(),
-                site.getUsername(),
-                site.getPassword(),
-                getPages,
-                offset,
-                PostStore.NUM_POSTS_PER_FETCH,
-                statusList,
-                null,
-                null,
-                null,
-                null);
+        List<Object> params =
+                createFetchPostListParameters(site.getSelfHostedSiteId(), site.getUsername(), site.getPassword(),
+                        getPages, offset, PostStore.NUM_POSTS_PER_FETCH, statusList, null, null, null, null);
 
         final XMLRPCRequest request = new XMLRPCRequest(site.getXmlRpcUrl(), XMLRPC.GET_POSTS, params,
                 new Listener<Object[]>() {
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/post/PostXMLRPCClient.java
```diff
@@ -143,10 +143,18 @@ public void fetchPostList(final PostListDescriptorForXmlRpcSite listDescriptor,
         SiteModel site = listDescriptor.getSite();
         List<String> fields = Arrays.asList("post_id", "post_modified_gmt", "post_status");
         final int pageSize = listDescriptor.getConfig().getNetworkPageSize();
-        List<Object> params =
-                createFetchPostListParameters(site.getSelfHostedSiteId(), site.getUsername(), site.getPassword(), false,
-                        offset, pageSize, listDescriptor.getStatusList(), fields,
-                        listDescriptor.getOrderBy().getValue(), listDescriptor.getOrder().getValue());
+        List<Object> params = createFetchPostListParameters(
```
malinajirka, line 146: NitPick: I'm not sure about this. Even the state before the change was adhering to the checkstyle rules. I think it's more about what a person is used to read than one being more readable than the other. Moreover, I feel like we rarely put a parameter per line in Java code so it breaks consistency a bit. Wdyt?

ParaskP7: Thanks for the comment @malinajirka!

Yes, you are right, the state before was adhering to the checkstyle rules. I can definitely revert this change if that is the agreed convention set by the team on such cases. I wasn't sure and to be honest, I was just pushing the boundaries to see what's possible.

Starting slowly with Kotlin and the style guide it suggests here and there in such cases, I am finding myself porting that to Java sometimes, most probably this is one of this times. As such, I have been trained lately to have the parameters per line, when those parameters exceed one line, but again this is coming from Kotlin. I also did that so that my next change, adding the `searchQuery` becomes more explicit when diffing the commit individually, I usually try to do that.

But, I totally agree that it is better to keep the style as is, especially if that doesn't break the tooling. Please do let me know and I can quickly revert that. Also, I didn't change the `PostRestClient` equivalent before I get reassured that what I am doing is correct.

malinajirka: >Starting slowly with Kotlin and the style guide it suggests here and there in such cases, I am finding myself porting that to Java sometimes, most probably this is one of this times.

I would lie if I said I've never done it as well :D. However, I think it'd be better to revert it in this instance for the sake of consistency.

We don't adhere to this rule even in Kotlin. We do put one method parameter definition per line when the number of parameters is >2. But we don't do it on the calling site just on the method declaration.

ParaskP7: Got it!

Again thanks for the clarification, I will now push a commit that reverts that (or else fixes that on top).

ParaskP7: This is now done, please see [46d332ea](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1646/commits/46d332ea04fd00c36a764559fc7d2617fd9ccf74).

malinajirka: Awesome, thank you!

PR #1647, opened 2020-07-29
Bugfix: Expected BEGIN_OBJECT but was BEGIN_ARRAY at path $[0]

chipsnyder:
**Fixes:** https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1644
**Related PR:** https://github.com/wordpress-mobile/WordPress-Android/pull/12557

### To Test:
1.) Install a theme with an invalid configuration such as:

```
add_theme_support( 'editor-gradient-presets', array(array()));
```

or 

```
add_theme_support( 'editor-color-palette', array(array()));
```

You can use [twentytwenty-copy-empty-array.zip](https://github.com/wordpress-mobile/WordPress-FluxC-Android/files/4997581/twentytwenty-copy-empty-array.zip) which is already setup.

2.) In the demo app log into your site
3.) Select "Editor Theme"
4.) Select "Fetch Theme from first site"

Expect app not crash.

This can also be tested in WPAndroid with the build here:

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/EditorThemeStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/EditorThemeStore.kt
@@ -79,7 +79,18 @@ class EditorThemeStore
 
         when (response) {
             is Success -> {
-                val responseTheme = response.result?.asJsonArray?.firstOrNull() ?: return
+                val noThemeError = OnEditorThemeChanged(EditorThemeError("Response does not contain a theme"), action)
+                if (response.result == null || !response.result.isJsonArray) {
+                    emitChange(noThemeError)
+                    return
+                }
+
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/EditorThemeStore.kt
```diff
@@ -79,7 +79,18 @@ class EditorThemeStore
 
         when (response) {
             is Success -> {
-                val responseTheme = response.result?.asJsonArray?.firstOrNull() ?: return
+                val noThemeError = OnEditorThemeChanged(EditorThemeError("Response does not contain a theme"), action)
+                if (response.result == null || !response.result.isJsonArray) {
```
mchowning, line 83: Nice catch adding the `isJsonArray` check! 👍

PR #1657, opened 2020-08-07
Media upload error codes alignment - WP.COM/JP

develric:
This PR does the following related to the WP.COM/JP scenario:

- Adds `MediaErrorType.BAD_REQUEST`
- Adds the `locale` query param to the media upload endpoint request
- Adds a `getApiUserMessageIfAvailable` method to make it easier to extract the User Message in the API BAD_REQUEST response if they are present (like `rest_upload_user_quota_exceeded|You have used your space quota. Please delete files before uploading. Back`)

### To test
Follow the instructions in the companion [WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/12620)

### Note
Since we added the `locale` query param, now the error messages that comes back from the upload media endpoint are localized. Before to merge this PR in develop I think would be good (as we did for a previous one) to have someone from Woo cross-check here in case they should rely in any way on the content of the error message not being localized


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
```diff
@@ -125,9 +129,23 @@ private void addLocaleToRequest(BaseRequest request) {
...
+            String localeParamName = getLocaleParamName(url);
             request.addQueryParameter(localeParamName, LanguageUtils.getPatchedCurrentDeviceLanguage(mAppContext));
         }
     }
+
+    private @NotNull String getLocaleParamName(@NotNull String url) {
+        return url.contains(WPCOM_V2_PREFIX) ? LOCALE_PARAM_NAME_FOR_V2 : LOCALE_PARAM_NAME_FOR_V1;
+    }
+
+    protected @Nullable HttpUrl getHttpUrlWithLocale(@NotNull String url) {
```
planarvoid, line 141: I'm wondering if this doesn't introduce a new pattern of adding the locale, would it maybe be cleaner to call the existing `addLocaleToRequest` method instead?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
```diff
@@ -191,6 +192,17 @@ public void uploadMedia(final SiteModel site, final MediaModel media) {
                 .post(body)
                 .build();
 
+        // Try to add locale query param
+        HttpUrl httpUrl = getHttpUrlWithLocale(url);
+
+        if (null != httpUrl) {
+            request = request.newBuilder()
```
planarvoid, line 199: We could call the `addLocaleToRequest` method on this request, right?

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -249,6 +249,24 @@ public static MediaError fromIOException(IOException e) {
 
             return mediaError;
         }
+
+        public String getApiUserMessageIfAvailable() {
```
planarvoid, line 253: Do you think you can add tests for this method?

planarvoid (commented):
Thanks for the PR! I have a few questions/comments (I haven't tested the WPAndroid part yet)

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java
develric: Hey @planarvoid, I started by going into the path you are suggesting, but currently the MediaStoreClient is using a `WPComGsonRequest.buildPostRequest` in general but not in the `uploadMedia` method. I tried to look into refactoring that method to align the approaches but it seems not trivial so I went into the path we have here. We can evaluate a refactor of the MediaStoreClient as a separate task I think (have created a GH issue here #1661 and added it to the Media Picker consolidation project so we do not forget). Let me know wdyt 🙇‍♂️

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java
develric: See previous [comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1657#discussion_r468025885)

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
develric: Good point 👍! Added it here 6c8e0395d645dac1c6992e1fa37f068c917d6c9a

develric:
Hi @planarvoid , I get your point about the `BaseWPComRestClient` and followed up in the comments above. Also noticing that here 9dc62b2b200fc0c142ebb46e5f72d9b7aa8dfe38 I added some logic to remove the final ` Back` suffix from the API user message (if present) since I think it's not part of the message to the user. Let me know wdyt, thanks 🙇‍♂️!

planarvoid (approved):
Thanks for the changes @develric and for testing the other approach. I've tested this with the WPAndroid app and it works well. I'm not merging this PR because of the alignment with Woo

develric:
Thanks for the review @planarvoid 🙇‍♂️!

@ThomazFB, @AmandaRiu I wonder if someone from your team could have a look to this PR to cross-check if it can break anything in Woo Mobile. Main reason I can think of taken from the PR description is:

> Since we added the locale query param, now the error messages that comes back from the upload media endpoint are localized. Before to merge this PR in develop I think would be good (as we did for a previous one) to have someone from Woo cross-check here in case they should rely in any way on the content of the error message not being localized

I tried to make a quick check on the woo codebase and in my understanding we should be fine with the modifications here, but of course I'm not much familiar with Woo, so expert eyes on it from someone of your team are much needed, thanks 🙇‍♂️!

ThomazFB:
Hey @develric, thanks for checking in about the impact on the Woo changes! I did a checkout on these changes into the Woo app to see if anything gets some side effect, and it looks fine 👍. Just to be sure, from my understanding it looks like these changes would affect the error message when a media upload fails, correct? Right now we don't use the failure message from the API but show our own translatable message, so from my point of view, the Woo app is totally fine.

develric:
Hi @planarvoid, as anticipated on slack I'm updating the getApiUserMessageIfAvailable method changing strategy on how to clean from " Back" suffix since from further checks, that suffix string can be localized so we cannot directly match on that 🙇‍♂️. A part from reviewing that method code changes, testing steps should remain the same.

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -260,11 +260,21 @@ public String getApiUserMessageIfAvailable() {
...
+                    }
 
                     // Remove the final " Back" if present
-                    if (userMessage.endsWith(suffix)) {
-                        userMessage = userMessage.substring(0, userMessage.length() - suffix.length());
+                    // Note that the real string depends on current locale; this is not optimal but
+                    // the two "." sentences separators seem to be always present in this error category messages
+                    String[] splitUserMsg = userMessage.split("\\.");
+
+                    if (null != splitUserMsg && splitUserMsg.length > 2) {
```
planarvoid: Do you think you could add a unit test for this logic? Something that tests the translations and covers more cases

planarvoid (approved):
The changes look good 👍  I'll test the PR with the WPAndroid PR

develric:
Hey @planarvoid, noting here what we further discussed on slack; in 
f4258f8 I deleted the hack that was trying to remove the final (and localized!) " Back" string since we judged it better and more reliable to try to clean it backend side and eventually have that in the user message more than risk to remove meaningful parts of the message. We'll try to follow up with backend to get it sorted.

planarvoid (approved):
LGTM 👍

PR #1658, opened 2020-08-07
Media upload error codes alignment - self-hosted

develric:
This PR does the following related to the self-hosted sites scenario:

- Adds 401 and 500 error codes. 
- Maps 401 and 500 xmlrpc error codes to MediaErrorType.

### To Test
Follow instructions in the companion WPAndroid PR

As for the #1657 we are adding a couple of MediaErrorType that could maybe be relevant for the Woo App. @ThomazFB would be great if you could have a quick look also into this, I can assume you are not using at all the XML-RPC self-hosted API so possibly it's a really quick check and no problem at all on Woo (thanks for confirming 🙇‍♂️)

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
@@ -366,6 +366,8 @@ public class MediaStore extends Store {
         SERVER_ERROR, // this is also returned when PHP max_execution_time or memory_limit is reached
         TIMEOUT,
         BAD_REQUEST,
+        XMLRPC_UNAUTHORIZED,
+        XMLRPC_INTERNAL_SERVER_ERROR,
 
         // logic constraints errors
         INVALID_ID,
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
@@ -366,8 +366,8 @@ public class MediaStore extends Store {
         SERVER_ERROR, // this is also returned when PHP max_execution_time or memory_limit is reached
         TIMEOUT,
         BAD_REQUEST,
-        XMLRPC_UNAUTHORIZED,
-        XMLRPC_INTERNAL_SERVER_ERROR,
+        XMLRPC_OPERATION_NOT_ALLOWED,
+        XMLRPC_UPLOAD_ERROR,
 
         // logic constraints errors
         INVALID_ID,
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java
@@ -563,7 +563,7 @@ public class MediaXMLRPCClient extends BaseXMLRPCClient implements ProgressListe
         if (exception instanceof XMLRPCFault) {
             switch (((XMLRPCFault) exception).getFaultCode()) {
                 case 401:
-                    mediaError.type = MediaErrorType.XMLRPC_UNAUTHORIZED;
+                    mediaError.type = MediaErrorType.XMLRPC_OPERATION_NOT_ALLOWED;
                     break;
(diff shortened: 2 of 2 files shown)


ThomazFB:
Hey @develric, thanks for checking in! The Woo app is 100% attached to the Jetpack in order to properly work, so these changes shouldn't affect the app. I did some tests to see anyway if this can affect in any unexpected way, and it looks fine. From what I see and know, Woo app only uses XML-RPC communication during the login flow, so I think we're safe here.

@AmandaRiu, do you agree? Am I missing some point here?

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
```diff
@@ -366,6 +366,8 @@ public OnStockMediaUploaded(@NonNull SiteModel site, @NonNull UploadStockMediaEr
         SERVER_ERROR, // this is also returned when PHP max_execution_time or memory_limit is reached
         TIMEOUT,
         BAD_REQUEST,
+        XMLRPC_UNAUTHORIZED,
+        XMLRPC_INTERNAL_SERVER_ERROR,
```
planarvoid: I'm wondering if we shouldn't reuse the existing `AUTHORIZATION_REQUIRED` and `SERVER_ERROR` types and add something like `isXmlrpc` flag to the `MediaError`. I think it might be a little bit confusing on the FluxC part why we have two `INTERNAL_SERVER_ERROR` types. I understand we need to differentiate between XMLRPC error and the WP.com error in the client but here it feels a little bit weird :-)

planarvoid (commented):
Looks good and works well 👍  Consider my suggestion but feel free to ignore it :-)

fluxc/src/main/java/org/wordpress/android/fluxc/store/MediaStore.java
develric: Hi @planarvoid 👋; for a self-hosted, behind the 401 error we can have the following:

- __( 'Sorry, you are not allowed to upload files.' )
- __( 'Sorry, you have used your space allocation of %s. Please delete some files to upload more files.' )
- __( 'Sorry, you are not allowed to edit this post.' )

behind the 500 error we can have:

- IXR_Error( 500, $upload_err ) -> an upload error of some type
- __( 'Could not write file %1$s (%2$s).' )

In a way (especially for the 401) they seem to merge a broader set or errors.

As said the best to me would have been to leverage the localized user messages from the API but in my understanding we cannot without some modifications also API side (cannot know the effort on that tbh). Given the string error messages above, maybe could be more appropriate to cover the two categories with something like:

- XML_RPC_OPERATION_NOT_ALLOWED (401 type)
- XML_RPC_UPLOAD_ERROR (500 type)

wdyt?

planarvoid: thanks for the reply! I think the names you're suggesting are better and more explicit :-), (XML_RPC_OPERATION_NOT_ALLOWED (401 type), XML_RPC_UPLOAD_ERROR (500 type))

develric:
Hi @planarvoid, changed the error type naming, hope it's better now 😊!

Hi @ThomazFB thanks for looking into this 🙇‍♂️!

> The Woo app is 100% attached to the Jetpack in order to properly work, so these changes shouldn't affect the app. I did some tests to see anyway if this can affect in any unexpected way, and it looks fine. From what I see and know, Woo app only uses XML-RPC communication during the login flow, so I think we're safe here.
@AmandaRiu, do you agree? Am I missing some point here?

AFAIU that seems reasonable; @AmandaRiu does the above makes sense to you? If so, I will proceed to merge and tag this one. Thanks for a feedback 🙇‍♂️!

AmandaRiu:
> @AmandaRiu, do you agree? Am I missing some point here?

That looks correct to me! AFAIK and as far as I can tell it looks like XML-RPC is only used during the login process.

develric:
Hey @planarvoid , I'm just triggering your attention on this PR to say this should be fine to merge together with the companion [WPAndroid PR ](https://github.com/wordpress-mobile/WordPress-Android/pull/12625); thanks 🙇‍♂️

planarvoid (approved):
LGTM 👍

PR #1660, opened 2020-08-10
Release/1.6.18

loremattei:
Merge release/1.6.18 into Master


mokagio (commented):
The connected tests failed; I triggered a new build and it failed, too.

What do we do? Is it something known that we can skip, or something that should be addressed?

mokagio (approved):
**Update** I suspect the connected test failure had little to do with this branch and all to do with the fact that the sha on which the nightly build run is the same as the tip of this branch: a749ee481.

If that's the case, I think this can be merged.

PR #1663, opened 2020-08-11
[WooCommerce] Handling downloadable files

hichamboushaba:
This PR is to allow retrieving the downloadable files of products, and updating them.
This is needed for the tickets: https://github.com/woocommerce/woocommerce-android/issues/2708, https://github.com/woocommerce/woocommerce-android/issues/2709, https://github.com/woocommerce/woocommerce-android/issues/2710 and https://github.com/woocommerce/woocommerce-android/issues/2710

This PR changes:

- Added [WCProductFileModel](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/41579061e5d3eb1b251520d48e6150af9f3a64cc/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductFileModel.kt) to hold the downloadable files properties.
- Updated `ProductRestClient` to allow updating the `downloadable`, `downloads`, `download_limit` and `download_expiry` attributes of a product.
- Updated the mocked and connected tests to test the changes.
- Updated the example app to allow updating those attributes:
- Updated the default value of `downloadLimit` and `downloadExpiry` to -1 to align with the API https://woocommerce.github.io/woocommerce-rest-api-docs/#product-properties

Screen Shot 2020-08-12 at 15 25 16     Screen Shot 2020-08-12 at 15 25 26


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -1025,6 +1030,18 @@ class ProductRestClient(
...
+            val updatedFiles = updatedProductModel.getDownloadableFiles()
+            body["downloads"] = JsonArray().apply {
+                updatedFiles.forEach { file ->
+                    add(file.toJson())
+                }
+            }
+
+            if (updatedFiles.isNotEmpty()) {
+                body["downloadable"] = true
+            }
```
hichamboushaba: Is there any condition required before setting `downloadable` to true? on wp-admin, we can set it only if the product is a Simple product.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductFileModel.kt
```diff
@@ -0,0 +1,11 @@
+package org.wordpress.android.fluxc.model
+
+import com.google.gson.JsonObject
+
+data class WCProductFileModel(val id: String? = null, val name: String = "", val url: String = "") {
```
hichamboushaba, line 12: The id is nullable, because when creating a new downloadable file, we send the payload without the ID, and we get the new ID on the response.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -216,16 +216,26 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
         return attrList
     }
 
-    fun getDownloadableFiles(): List<String> {
-        val fileList = ArrayList<String>()
+    fun getDownloadableFiles(): List<WCProductFileModel> {
```
hichamboushaba, line 219: This is a breaking change, and will require an update to the woocommerce app.
My plan is to make a small PR addressing just this incompatibility to be able to merge it ASAP, and keeping the bigger changes for other PRS.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
anitaa1990: From testing the [update product API](https://woocommerce.github.io/woocommerce-rest-api-docs/#update-a-product), I noticed that there isn't any restriction on the API side to add/update a downloadable file for any of the product types. I think the `wp-admin` UI takes care to display the downloadable file section only if the `downloadable` flag is enabled. 

The designs follow a similar approach where we display the Downloadable files section only for Simple products. So it makes sense to handle the validation on the UI side in the Woo app and not add any restrictions on the fluxC side. WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -1025,6 +1030,18 @@ class ProductRestClient(
...
+        if (!storedWCProductModel.hasSameDownloadableFiles(updatedProductModel)) {
+            val updatedFiles = updatedProductModel.getDownloadableFiles()
+            body["downloads"] = JsonArray().apply {
+                updatedFiles.forEach { file ->
+                    add(file.toJson())
+                }
+            }
+
+            if (updatedFiles.isNotEmpty()) {
+                body["downloadable"] = true
```
anitaa1990: This raises a good point. I noticed in the web that the downloadable files section is only visible, if the `downloadable` flag is enabled. Perhaps we should follow a similar approach in the app as well and provide a switch in the Woo app under the Product Settings page, to update if a product is downloadable or not. I'll add this as a discussion point under the implementation plan and rope in our designer to get her thoughts on that as well.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -216,16 +216,26 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
...
         try {
             Gson().fromJson(downloads, JsonElement::class.java).asJsonArray.forEach { jsonElement ->
-                jsonElement.asJsonObject.getString("file")?.let {
-                    fileList.add(it)
+                with(jsonElement.asJsonObject) {
+                    fileList.add(
+                            WCProductFileModel(
+                                    id = this.getString("id"),
+                                    name = this.getString("name") ?: "",
+                                    url = this.getString("file")!!
```
anitaa1990: Rather than marking this as mandatory and adding a catch exception, which will result in an empty array being sent to the UI, what do you think about making it empty and we don't need to display the url, if it is blank. That way, atleast the user experience is not compromised. (similar to how we are handling "name" in the above scenario).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductFileModel.kt
anitaa1990: 👍 Maybe we could add this as a comment to the `WCProductFileModel` class so that in the future, people are aware of the reason behind this. WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
anitaa1990: I think it'd be fine to keep all the changes in this PR as is. Once we start working on the Woo side, I will create a separate PR from this PR and provide you with the commit hash. That can be used to reference the changes from this PR in the Woo app. If we make sure to not merge this PR, till the Woo PR is merged, we should be good. 

(Also marked the PR as `not ready for merge` in the meantime) Sounds good?

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt
```diff
@@ -373,13 +407,16 @@ class WooUpdateProductFragment : Fragment() {
             this.selectedCategories = data?.getParcelableArrayListExtra(ARG_SELECTED_PRODUCT_CATEGORIES)
         } else if (requestCode == PRODUCT_TAGS_REQUEST_CODE) {
             this.selectedTags = data?.getParcelableArrayListExtra(ARG_SELECTED_PRODUCT_TAGS)
+        } else if (requestCode == PRODUCT_DOWNLOADS_REQUEST_CODE) {
+            this.selectedProductDownloads = data!!.getParcelableArrayListExtra(ARG_PRODUCT_DOWNLOADS)
```
anitaa1990, line 428: np: we could follow the convention here and make `data` as optional? Something like:
`data?.getParcelableArrayListExtra(ARG_PRODUCT_DOWNLOADS)`

anitaa1990 (commented):
Really nice work on this @hichamboushaba 🥇 I've tested all the changes are everything works as expected!

I just left a few code comments for your review. You also mentioned that the PR is not yet ready. So we'll wait for completion before :shipit: but nice work so far 😄

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductFileModel.kt
hichamboushaba: Yes, I'll add it 👍

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt
hichamboushaba: if we make it optional, and if for any case it's null, the code will fail silently (the update to the downloadable files will be ignored), IMO it's better to have a crash here (especially that the app is just an example), to be able to identify the cause.
PS: I have to add a check on the resultCode in the condition, to make sure that the action was not cancelled.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
hichamboushaba: My idea (to comply with the Figma designs) was to make this change automatically, and that's why I'm setting the `downloadable` to true here.
So, when the user adds any files, we don't ask them to make the product `downloadable` first, but make the change automatically.
There is one issue with this approach, we can't mark a file as non-`downloadable`.
As this is a UX decision, we'll continue the discussion in Slack 🙂

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
hichamboushaba: For this one, I'm still thinking about the best approach to do it 🙂 .
For your remark, I think It will be an empty array only if the list has only this file, but in other cases we will send the other files.

And I did this choice now because from testing the API, when updating a downloadable file, if the `file` attribute is null, the backend deletes the whole file, which means that the attribute is mandatory.

hichamboushaba: I changed this to follow your suggestion now 🙂 
After more testing of the API, I found that sending an empty string in the `file` has the same behavior as null, so there is no use of this check.

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt
```diff
@@ -472,6 +536,13 @@ class WooUpdateProductFragment : Fragment() {
             return
         }
         prependToLog("Product updated ${event.rowsAffected}")
+
+        selectedRemoteProductId?.let {
+            selectedCategories = null
+            selectedTags = null
+            selectedProductDownloads = null
+            updateSelectedProductId(it)
+        }
```
hichamboushaba, line 546: @anitaa1990 I made this change to make sure that the UI is showing the updated product that was fetched after the update.
In some cases the returned product from the call has different attributes than what we had, for example when setting `downloadable` to false, the `downloads` array will be emptied in the response. WDYT about this change?

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt
anitaa1990: Sounds good  👍

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt
anitaa1990: Yup, this works! Nice fix 😄

anitaa1990 (approved):
Nice job on the PR @hichamboushaba :shipit: I'll wait to merge this PR till we open a related PR in Woo since there is a breaking change in this PR 👍

ThomazFB:
Hey @anitaa1990, I'll be opening all these changes discussed here on a different PR since this one is from a forked repository and I couldn't push any changes and code updates directly to here. Let me know If you think we should keep this PR and I'll reopen it!

PR #1665, opened 2020-08-13
Add Proguard Rules for Encrypted Logging

jkmassel:
Adds proguard rules to support encrypted logging – prevents a crash from happening on release builds.

Added per the docs at https://github.com/java-native-access/jna/blob/master/www/FrequentlyAskedQuestions.md#jna-on-android

### DIFF
--- a/fluxc/proguard-rules.pro
+++ b/fluxc/proguard-rules.pro
@@ -15,3 +15,9 @@
 #-keepclassmembers class fqcn.of.javascript.interface.for.webview {
 #   public *;
 #}
+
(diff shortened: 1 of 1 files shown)


loremattei (approved):
This is consistent with the doc and the changes are the same tested in https://github.com/wordpress-mobile/WordPress-Android/pull/12674, so I think it's 👍

oguzkocer (approved):
I am not entirely sure if we need this since clients have their own Proguard rules, but it wouldn't hurt to have it I don't think so :shipit:

PR #1666, opened 2020-08-14
Add missing "video/x-msvideo" mime type

planarvoid:
This PR adds support for the AVI files with the mime type `video/x-msvideo`. This could be tested on the corresponding WPAndroid PR


develric (approved):
Code LGTM. I tested it with the companion WPAndroid and works as described there.

PR #1667, opened 2020-08-15
Fixed failing test by adding type field when updating a product

anitaa1990:
This PR introduced the option to update the product type. The `testUpdateProduct` test in `ReleaseStack_WCProductTest` was not updated after this change and this was causing the test to fail. This PR updates the test, which should fix the failing test. 

#### To test
- Run `testUpdateProduct` in `ReleaseStack_WCProductTest` and verify that the test is passing.


aforcier:
Working well, :shipit: Thanks for fixing @anitaa1990!

PR #1668, opened 2020-08-17
Woo/grouped products small fixes

anitaa1990:
While working on woocommerce/woocommerce-android#2630, I noticed a few issues when displaying a group of products. This PR attempts to fix those issues.

- The list of `groupedProductIds` should be `Long` since the `remoteProductId` is `Long` everywhere in the app. So modified `WCProductModel` to reflect this change in e2b7a4f.
- When fetching the product details for a list of `groupedProductIds`, I noticed that we delete the previously stored products from the local db before adding these products. This means in the woo app, the Products tab only displays the groupedProducts and we need to manually refresh the app to fetch the list of products again. So I have added logic to delete the locally stored products only if the `remoteProductIds` are null.

#### To test
- Verify that Fetching products, fetching a single product and updating grouped products works as expected.


0nko (approved):
:shipit:

PR #1670, opened 2020-08-18
Added null check to groupedProductIds before parsing the json field

anitaa1990:
@AmandaRiu noticed a crash when testing this PR to display/update grouped products. I wasn't able to reproduce the crash but this PR just adds a null check to the `groupedProductIds`, which was missed in previous implementations.


AmandaRiu (approved):
Tested in WCAndroid and works like a charm. Thanks @anitaa1990! :shipit:

PR #1675, opened 2020-08-25
Woo/add grouped product fixes

anitaa1990:
Adds support for woocommerce/woocommerce-android#2630 by adding the option to exclude `remoteProductIds` when fetching and searching for a product.

#### Changes
- Added `excludedProductIds` list when fetching products and searching for products from the request API.
- Added `excludedProductIds` list when fetching products and searching for products from the local db.
- Added logic to delete the products stored locally only if `excludedProductIds` is null. This prevents the excluded products from being deleted from the local db. (This was causing the excluded products to disappear from the Products tab in the Woo app).

#### Testing
- Smoke test fetching products and searching products to ensure that everything is working correctly.
- Run `ProductSqlUtilsTest`.


0nko (approved):
LGTM 👍

PR #1677, opened 2020-09-01
Adding list of video and image supported types only

mzorz:
Title has it - this PR only adds utility methods in `MimeTypes` that provides a list of `image/* ` and `video/*` mime types (in conjunction or separately) to be used elsewhere for example as filters when picking media from the device.

#### To test
See the WPAndroid related PR https://github.com/wordpress-mobile/WordPress-Android/pull/12861


develric (approved):
Hi @mzorz 👋! Tested this with the companion WPAndroid PR and it's working as described. LGTM :shipit:

PR #1679, opened 2020-09-07
Migrated all v4 endpoints to v3 in API tests

anitaa1990:
There are some WooCommerce API tests that are failing since they are pointing to `v4` endpoint, which is no longer used. This PR changes all the API test endpoints to use `v3` instead of `v4`.

#### To test
To run the tests locally, you need to set the `API_TEST_OAUTH_KEY` and `API_TEST_SITE_URL` env variables and then run the tests using `./gradlew :tests:api:test` Values for the env variables can be found in `ss Mobile API Test key`.


nbradbury (approved):
:shipit:

PR #1681, opened 2020-09-09
Woo: Add product endpoint

0nko:
This PR is the result of a successful trial. It adds the REST client and DB support for the add-product endpoint.

It's already been reviewed but I'd like to get a final quick look before we merge it to `develop`.

**The original description:**

## RESOURCES
[The api definition](https://woocommerce.github.io/woocommerce-rest-api-docs/#create-a-product)

## CHANGES / DESCRIPTION

This PR will contain the work to allow a remote communication for the **creation of a new product**. This chunk of work meant:
- add new product actions in `WCProductAction`
- handle of new actions and new event `OnProductCreated` in `WCProductStore`
- add new product function in `ProductRestClient`
- added tests to the new request in **APITesting_WCProduct**, **ReleaseStack_WCProductTest** and **WCProductStoreTest.kt**

Following the `README`, I added the initial action in the `WCProductAction`; I called them:
- **ADD_NEW_PRODUCT** 
- **ADDED_NEW_PRODUCT**

In the `WCProductStore`, the `onAction` method will be listening to these actions that, upon trigger will behave as such:

**1)** `ADD_NEW_PRODUCT`
Will call the `createNewProduct` of the `ProductRestClient` that will make the remote connection and respond with success/error, mapping the response model accordingly. This done, it will trigger the correct action for completion, with the data;

**2)** `ADDED_NEW_PRODUCT`
Will call the `handleAddNewProduct`, from a previous completion of a request, and create the success/error scenarios accordingly, using the `OnProductCreated` for proper communication of the result.

This will get emitted to whoever is listening and trigger the UI updates (example app has this integration 👍 also, in screenshots bellow)

Let me know your thoughts on it 🙏 

## SCREENSHOTS
Product created
-
 

New Product fetched
-

Tests
-
Screenshot 2020-08-25 at 23 56 17

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCProductTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCProductTest.kt
@@ -39,6 +39,7 @@ import org.wordpress.android.fluxc.store.WCProductStore.FetchSingleProductShippi
 import org.wordpress.android.fluxc.store.WCProductStore.FetchSingleVariationPayload
 import org.wordpress.android.fluxc.store.WCProductStore.OnProductCategoryChanged
 import org.wordpress.android.fluxc.store.WCProductStore.OnProductChanged
+import org.wordpress.android.fluxc.store.WCProductStore.OnProductCreated
 import org.wordpress.android.fluxc.store.WCProductStore.OnProductImagesChanged
 import org.wordpress.android.fluxc.store.WCProductStore.OnProductPasswordChanged
 import org.wordpress.android.fluxc.store.WCProductStore.OnProductReviewChanged
@@ -47,6 +48,7 @@ import org.wordpress.android.fluxc.store.WCProductStore.OnProductTagChanged
 import org.wordpress.android.fluxc.store.WCProductStore.OnProductUpdated
 import org.wordpress.android.fluxc.store.WCProductStore.OnVariationChanged
 import org.wordpress.android.fluxc.store.WCProductStore.OnVariationUpdated
+import org.wordpress.android.fluxc.store.WCProductStore.RemoteAddNewProductPayload
 import org.wordpress.android.fluxc.store.WCProductStore.UpdateProductImagesPayload
 import org.wordpress.android.fluxc.store.WCProductStore.UpdateProductPasswordPayload
 import org.wordpress.android.fluxc.store.WCProductStore.UpdateProductPayload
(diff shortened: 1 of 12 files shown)


nbradbury:
In the example app, product width and height are duplicated and are sent to the backend as weight and length.

Screenshot_1599662538

nbradbury:
@0nko In the example app, many of the settings (product type, status, manage stock, etc) are disabled. Is this intentional?

nbradbury:
Also in the example app, I added a tag and a category but neither were sent to the backend.

Screenshot_1599663543

Screen Shot 2020-09-09 at 10 58 25 AM

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCProductTest.kt
```diff
@@ -891,4 +929,23 @@ class ReleaseStack_WCProductTest : ReleaseStack_WCBase() {
...
+    @Subscribe(threadMode = ThreadMode.MAIN)
+    fun onProductCreated(event: OnProductCreated) {
+        event.error?.let {
+            throw AssertionError("OnProductCreated has unexpected error: " + it.type)
+        }
+
+        lastAddNewProductEvent = event
+
+        when (event.causeOfChange) {
+            WCProductAction.ADD_PRODUCT -> {
```
nbradbury: This test is failing for me because `event.causeOfChange` is `ADDED_PRODUCT` rather than `ADD_PRODUCT`.

Screen Shot 2020-09-09 at 11 21 17 AM

0nko:
Thanks for the testing, @nbradbury! 

Seems like the field naming issues were already there and the other problems were related to initialization. I've addressed all issues, ready for another round.

nbradbury:
In the example app, if I tap the "Select Categories" or "Select Tags" buttons, any information I previously typed into the Add Product screen is erased when I return to it.

nbradbury:
I typed in a category and a tag, but neither is sent to the backend.

Screenshot_1600079938

0nko:
> In the example app, if I tap the "Select Categories" or "Select Tags" buttons, any information I previously typed into the Add Product screen is erased when I return to it.

This seems to be a problem if you have Don't keep activities enabled. Seems like an existing problem with the Update flow, as well. I updated the re-initialization logic so this should be fixed now.

> I typed in a category and a tag, but neither is sent to the backend.

That's not how it works. First, you need to fetch the Tags/Categories on the previous screen. Then you need to tap on the Tags/Categories button, select items from the list and tap on the Done button.

nbradbury:
@0nko Changes look good, but there's a failing test in `DateUtilsTest`. It doesn't look like this PR made any changes to that, and the test passes locally, so I'm not sure why it's failing.

nbradbury (approved):
Looks good! :shipit:

PR #1684, opened 2020-09-11
Issue/package endpoint resource files

khaykov:
This PR packages resources used to build our endpoints, so we can access them regardless of their path, using `getClass().getClassLoader().getResourceAsStream` or `getClass().getClassLoader().getResource()`.

To do this, we have to move all the `txt` files to one location inside the fluxc processor module.


oguzkocer (approved):
@khaykov Thank you very much for this PR!

I've verified that updating the txt files and rebuilding the project created the endpoints as expected. I've also verified that this fix works for the composite builds. 🥳

PR #1687, opened 2020-09-15
Make sure methods with "get" prefix do not collide with field names

malinajirka:
This PR fixes "reference to xyz is ambiguous" error.

Corresponding PR in Woo: https://github.com/woocommerce/woocommerce-android/pull/2849

**What is the root cause of the issue:**
Some models have getters (methods with "get" prefix) which seem like they are getters for fields but their return type is not the same as the type of the fields.

**Example:**
A field is called "tags" and its type is "String". However, "getTags" method doesn't return "String" but it returns "ArrayList<WCProductImageModel>". This is not allowed in Java-Kotlin code. The reason is simple - the compiler doesn't know if you want to access the field or the method. It seems JDK8 uses a default behavior which accesses the field when "o.field" syntax is used and it accesses the method when "o.getField()" syntax is used.

**Why we need to update this code**
The issue isn't breaking the build on JDK8 - it's reported by AS as an error but it doesn't fail the build. However, with JDK11 this issue breaks the build. Since CircleCI dropped support for JDK8 we need to update all our projects to JDK11 and this issue is blocking the migration.

Screenshot 2020-09-15 at 13 58 54

To Test:
Smoke test the Woo app -

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
@@ -87,13 +87,13 @@ class MockedStack_WCProductsTest : MockedStack_Base() {
         with(payload) {
             assertNull(error)
             assertEquals(remoteProductId, product.remoteProductId)
-            assertEquals(product.getCategoriesList().size, 2)
-            assertEquals(product.getTagsList().size, 2)
-            assertEquals(product.getImagesList().size, 2)
+            assertEquals(product.getCategoryList().size, 2)
+            assertEquals(product.getTagList().size, 2)
+            assertEquals(product.getImageList().size, 2)
             assertNotNull(product.getFirstImageUrl())
-            assertEquals(product.getAttributesList().size, 2)
-            assertEquals(product.getAttributesList().get(0).options.size, 3)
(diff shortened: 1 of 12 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
```diff
@@ -87,13 +87,13 @@ class MockedStack_WCProductsTest : MockedStack_Base() {
         with(payload) {
             assertNull(error)
             assertEquals(remoteProductId, product.remoteProductId)
-            assertEquals(product.getCategories().size, 2)
-            assertEquals(product.getTags().size, 2)
-            assertEquals(product.getImages().size, 2)
+            assertEquals(product.getCategoriesList().size, 2)
```
0nko: Just a nit-pick, but with all the `List`s, I'd drop the plural in the name, so `getCategoryList()` instead of `getCategoriesList()`, etc. because it sounds more natural.

malinajirka: Good call, fixed in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1687/commits/a107aa70044042b642059434fc997cabc8ac4a3c

malinajirka:
@0nko Thanks for the review ;). I've updated the names in both FluxC and Woo repos. It's ready for another round.

0nko (approved):
:shipit:

Issue #919, opened 2018-10-09
Resolve self-hosted SSL + OkHttp 3.9.1+ issue


aforcier:
It looks like OkHttp `3.9.1` or later breaks support for self-hosted SSL sites (at least, it does for our shared test site). We're currently on `3.9.0` for this reason - if/when we want to update `OkHttp`, we'll need to understand and resolve the issue.

See https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/918 for some background.

jtreanor:
This issue is now manifesting in WPAndroid (https://github.com/wordpress-mobile/WordPress-Android/issues/9032) due Gradle resolving to `3.11.0`.

jtreanor:
Update: I have merged https://github.com/wordpress-mobile/WordPress-Android/pull/9040 to remove the immediate impact on WPAndroid.

The cause may be related to square/okhttp#4053.

PR #1691, opened 2020-09-18
Update harmless libraries

planarvoid:
This PR updates:
- test libraries - this should be OK if all the tests pass
- WordPress Utils + minSdkVersion - I don't think there was a reason why we had min SDK set to 16. WP Utils requires at least 18

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,5 +77,5 @@ ext {
     arch_lifecycle_version = '1.1.1'
     arch_core_version = '2.0.1'
     appcompat_version = '1.0.2'
-    mockitoVersion = '2.28.2'
+    mockitoVersion = '3.3.3'
 }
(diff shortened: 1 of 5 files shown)


aforcier:
Hey @planarvoid 👋 I haven't tested this PR but AFAIK OkHttp 3.9.1 or later will break self-signed ssl site support still (it did in 3.14.1) - see https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/919. Running the self-hosted connected tests mentioned in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1655 should confirm that.

planarvoid:
I've reverted the OKHTTP bump. We should really do something about this but it's out of scope of this PR

malinajirka (approved):
LGTM, thanks!

PR #1693, opened 2020-09-21
Move stock media store and rest client to kotlin

planarvoid:
In order to implement stock media library cache I'm moving the store and the rest client to Kotlin. This is the first PR that will enable me to implement the calls using coroutines. It would be better to test the following PRs that include these changes


develric (approved):
Hi @planarvoid, this conversion to kotlin LGTM 👍 ! :shipit:

PR #1694, opened 2020-09-21
Introduce cache to stock media store

planarvoid:
This PR introduces a cache for the stock media requests. The cache is mostly helpful for pagination since we always clear it when the search term changes. As part of this PR I've added coroutines to handle the rest calls and DB cache (I've kept the previous event bus approach). 

The related WPAndroid PR is https://github.com/wordpress-mobile/WordPress-Android/pull/13009


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StockMediaSqlUtils.kt
```diff
@@ -0,0 +1,123 @@
...
+        WellSql.delete(StockMediaBuilder::class.java)
+                .where()
+                .not()
+                .equals(StockMediaTable.FILTER, filter)
+                .endWhere().execute()
+        WellSql.delete(StockMediaPageBuilder::class.java)
+                .where()
+                .not()
+                .equals(StockMediaPageTable.FILTER, filter)
+                .endWhere().execute()
```
develric: I would consider wrapping this in a transaction as well, even if not sure about best way to manage an eventual fail here tbh, wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StockMediaSqlUtils.kt
```diff
@@ -0,0 +1,123 @@
...
+                            itemId = it.id,
+                            name = it.name,
+                            title = it.title,
+                            url = it.url,
+                            date = it.date,
+                            thumbnail = it.thumbnail
+                    )
+                }
+        ).execute()
+        writableDb.endTransaction()
```
develric: Was wondering if we should wrap transactions (at least inserts and deletes) with try/catch and have an endTransaction in case of errors (or as a finally block in general)

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StockMediaSqlUtils.kt
```diff
@@ -0,0 +1,123 @@
...
+        WellSql.delete(StockMediaPageBuilder::class.java)
+                .where()
+                .not()
+                .equals(StockMediaPageTable.FILTER, filter)
+                .endWhere().execute()
+    }
+
+    fun clear() {
+        WellSql.delete(StockMediaBuilder::class.java).execute()
+        WellSql.delete(StockMediaPageBuilder::class.java).execute()
```
develric: same as for `deleteAllWithoutFilter` if that makes sense

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StockMediaSqlUtils.kt
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Table(name = "StockMediaPage")
+    data class StockMediaPageBuilder(
```
develric, line 75: Maybe stupid question sorry 🙃; but I'm missing where we are creating this table.

develric (commented):
Hi @planarvoid , went over the code and left a couple of comments; I think I will move to the next linked FluxC PR and make testing for this and the next PR using the wordpress-mobile/WordPress-Android#13009 before we final approve this one also 👍 . Let me know wdyt, thanks 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StockMediaSqlUtils.kt
planarvoid: that's a good point 👍

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/StockMediaSqlUtils.kt
planarvoid: I missed to create it in this PR, I think I've fixed that in the following one but good catch :-)

planarvoid:
hey @develric , thanks for the check! I've addressed your comments and moved some of the changes from the following PR. I hope it looks OK now :-)

develric (approved):
Thanks for the changes @planarvoid 🙇 . LGTM 👍

PR #1698, opened 2020-09-23
Issue/1695 woo delete product

nbradbury:
This PR adds the ability to delete/trash a WooCommerce product. The default is to trash, and WCAndroid will only be using that functionality, but I figured I'd support permanent deletion since the endpoint offers that.

Tests have been added, and the Example app has been updated with a "Delete Product" button.

example

Note: Only one reviewer is necessary.


AmandaRiu (approved):
Nice work Nick! :shipit:

PR #1699, opened 2020-09-24
Fix the failing Products API test

0nko:
Recently-added `canAddProduct()` test had an adverse effect on `canGetAllProducts()` and made it fail, because it changed the number of products on the site. This PR makes sure to delete the last product before creating a new one so that the total number stays the same.

**To test:** Run the `APITesting_WCProduct` tests and make sure all of them pass.


nbradbury (approved):
Looks good! :shipit:

PR #1701, opened 2020-09-24
issue/1700-wc-add-delete-product-db-only

### LINKED ISSUE #1700: Woo: add the ability to delete a product from the DB only, and add a product to the DB only
For [this WCAndroid issue](https://github.com/woocommerce/woocommerce-android/issues/2889), we need to be able to delete a product from the database only, and add it back. This is for the ability to Undo trashing a product.

nbradbury:
Closes #1700 - this simple PR enables adding or removing a product from the database only. This is to be used by WCAndroid when we give users the ability to trash a product and then restore it.

Note that I haven't added tests for this PR because it's simply a wrapper for two functions in `ProductSqlUtils` that are already heavily tested.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -9,16 +9,7 @@
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value>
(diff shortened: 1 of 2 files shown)


nbradbury:
I decided to close this since I found a better way to handle trashing in the Woo app that doesn't require any changes to FluxC.

PR #1702, opened 2020-09-25
Extends FETCH_BLOCK_LAYOUTS action for self-hosted sites

antonis:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2455

**Description**
This PR fetches the *Block Layouts* for self-hosted sites from `/wpcom/v2/common-block-layouts`

**To test**
This can be tested with https://github.com/wordpress-mobile/WordPress-Android/pull/13017
or by running the `testFetchBlockLayouts()` test.

PR #1703, opened 2020-09-25
Adds supported blocks parameter in FETCH_BLOCK_LAYOUTS action

antonis:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2444

**Depends on:** https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1702

**Description**
This PR fetches adds supported blocks parameter in fetch block layouts call both for WPCom and self hosted site

**To test**
This can be tested with https://github.com/wordpress-mobile/WordPress-Android/pull/13018
or by running the `testFetchBlockLayouts()` tests.


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestXMLRPC.java
```diff
@@ -238,10 +239,18 @@ public void testFetchBlockLayouts() throws InterruptedException {
         fetchSites(BuildConfig.TEST_WPORG_USERNAME_SH_SIMPLE,
                 BuildConfig.TEST_WPORG_PASSWORD_SH_SIMPLE,
                 BuildConfig.TEST_WPORG_URL_SH_SIMPLE_ENDPOINT);
-
         SiteModel firstSite = mSiteStore.getSites().get(0);
+        List<String> supportedBlocks =
```
malinajirka, line 243: Will we need to update this list whenever we add a support for a new block? 

We might want to consider passing just three basic blocks. We are testing that the API is reachable and the payload is formatted as expected. Not that the result for all these blocks is what we expect it to be, right? Wdyt?

malinajirka (approved):
LGTM, thanks! I've left one minor comment.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestXMLRPC.java
antonis: Hello @malinajirka 👋 
I agree that this test might be fragile. It will break if we remove all the layouts that contain only a subset of the `supportedBlocks`.
To my understanding adding new blocks will not break the test unless we include those in layouts and the above condition is not longer true.
I'm not sure that there is a layout that includes just three blocks.

malinajirka: I see. Makes sense, let's keep it as is and modify it only if we find the need. Thanks!

PR #1708, opened 2020-10-02
issue/1707-wc-trashed-product-status 

### LINKED ISSUE #1707: Woo: Set status of trashed products
When we request a product from the backend, the response shows `status: "trash"` but in WCAndroid the status is `null`. We need to set the status correctly so we can show a "Trashed" badge in WCAndroid.

[Related WCAndroid issue](https://github.com/woocommerce/woocommerce-android/issues/2949).

nbradbury:
Resolves #1707 - this simple PR adds the missing `TRASH` status to `CoreProductStatus`. To test:

* Open the example app
* Go to Woo > Products > Fetch products with filters
* Choose Filter by status and verify that Trash is one of the selections
* Verify that only trashed products are returned when you apply the Trash filter


0nko (approved):
Looks good! 👍

PR #1714, opened 2020-10-12
Use fixed referrer URL instead of fetched from unstable API

planarvoid:
The previous test was failing because it marked a fetched referrer as a spam but didn't revert the change at the end of the test. This is now fixed and instead of fetched referrer we're marking as a spam a fixed URL. This way we can be sure we can always unmark the referrer and we don't have to rely on the API.


malinajirka (approved):
LGTM, thanks!

PR #1718, opened 2020-10-13
Implements FETCH_BLOCK_LAYOUTS response CACHING

antonis:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2456

**Description**
This PR caches the fetch block layouts response

**To test**
This can be tested with https://github.com/wordpress-mobile/WordPress-Android/pull/13105

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/layouts/GutenbergLayoutModel.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/layouts/GutenbergLayoutModel.kt
@@ -1,0 +1,50 @@
+package org.wordpress.android.fluxc.model.layouts
+
+import com.wellsql.generated.GutenbergLayoutModelTable
+import com.yarolegovich.wellsql.WellSql
+import com.yarolegovich.wellsql.core.Identifiable
+import com.yarolegovich.wellsql.core.annotation.Column
+import com.yarolegovich.wellsql.core.annotation.PrimaryKey
+import com.yarolegovich.wellsql.core.annotation.Table
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.rest.wpcom.site.GutenbergLayout
+
+@Table
+class GutenbergLayoutModel(
+    @PrimaryKey @Column private var id: Int = 0,
+    @Column var slug: String = "",
+    @Column var siteId: Int = 0, // Foreign key
+    @Column var title: String = "",
+    @Column var preview: String = "",
+    @Column var content: String = ""
+) : Identifiable {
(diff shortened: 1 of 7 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/layouts/GutenbergLayoutModel.kt
```diff
@@ -0,0 +1,50 @@
...
+        preview = preview,
+        content = content
+)
+
+fun GutenbergLayoutModel.transform(categories: List<GutenbergLayoutCategoryModel>) = GutenbergLayout(
+        slug = slug,
+        title = title,
+        preview = preview,
+        content = content,
+        categories = categories.transform()
```
mkevins, line 41: Not suggesting a change here, but curious why we need the calls to the empty transform methods?

antonis: Hello Matthew 👋 ,
thank you for reviewing this 🙇 

The empty method transforms `List<GutenbergLayoutCategoryModel>` to `List<GutenbergLayoutCategory>` by calling the extension function [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1718/files#diff-64a480c60fc30a79d1d4ceb89e6919ba16be48344238049a51eef928b7965e4fR43).

Tbh I'm not sure if having two sets of models is the best approach on this but I couldn't find a better way to implement the relationship between `GutenbergLayout` and `GutenbergLayoutCategory` in the DB.

mkevins: Thanks for the explanation. :+1: 

> Tbh I'm not sure if having two sets of models is the best approach on this but I couldn't find a better way to implement the relationship between GutenbergLayout and GutenbergLayoutCategory in the DB.

I thought about that as well, and I agree. Not sure if there's a better way to go about it.

Fwiw, I also considered an alternative to deleting the connections and repopulating every time (something like a compound key).. but it's probably not worth it since a) I'm not sure WellSQL supports that, so it may require a custom query and b) on reflecting on it more, I guess we don't really _want_ to keep connections around, and instead should treat a successfully parsed response as fully authoritative. Also, the way you have it implemented, it's only deleting the records for the selected site, so I think it's correct as is. Nice work!

mkevins (approved):
This looks good, nice work Antonis!

All unit tests are passing, and I also tested via: https://github.com/wordpress-mobile/WordPress-Android/pull/13105#pullrequestreview-510195397, and everything is working as described.

PR #1727, opened 2020-10-20
Fix visibility of created sites

malinajirka:
This PR fixes an issue with the "Create new site" endpoint. We were sending `"public": "PUBLIC"` parameter instead of `"public":"1"`. 

Corresponding PR in WPAndroid is [here](https://github.com/wordpress-mobile/WordPress-Android/pull/13182).

More info p4a5px-2DW-p2

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
@@ -224,7 +224,7 @@ public class SiteRestClient extends BaseWPComRestClient {
         Map<String, Object> body = new HashMap<>();
         body.put("blog_name", siteName);
         body.put("lang_id", language);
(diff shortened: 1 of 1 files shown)


maxme:
Code looks good to me. I also checked running the sample app and opened Flipper to intercept the network call and it looked good `"public":"1"`.

```
{"client_secret":"xxx","public":"1","blog_name":"test420201020","client_id":"xxx","lang_id":"en","validate":"1"}
```

Then I tried changing the code and forced the value `PRIVATE`, network code looked fine (`"public":"-1"`):
```
{"client_secret":"xxx","public":"-1","blog_name":"test420201020","client_id":"xxx","lang_id":"en","validate":"1"}
```
---

I did one last check using a sandbox with the upcoming change in the API (with `validate: 0`), the created sites had the right visibility (I tried all states)

- Private: test420201020.wordpress.com
- Hidden: test520201020.wordpress.com
- Public: test620201020.wordpress.com

maxme (approved):

PR #1728, opened 2020-10-20
Add unparsed referrers object to lower db load

### LINKED ISSUE #10057: SQLiteBlobTooBigException: Row too big to fit into CursorWindow requiredPos=0, totalRows=1
https://sentry.io/share/issue/b99a555f8217430aa5c6b2246659dfe5/

planarvoid:
Fixes https://github.com/wordpress-mobile/WordPress-Android/issues/10057

This is one of the steps to decrease the size of the Referrers JSON object saved into the database. Previously (because of our API) we were saving both the unparsed JSON object and the parsed referrers. This is fixed now by introducing an `UnparsedReferrersResponse` which gets parsed into `ReferrersResponse` which is now a little simplified. This should help reducing the DB load by half. As additional step I'll decrease the number of referrers loaded in the client. 

This PR can be tested using this WPAndroid PR - https://github.com/wordpress-mobile/WordPress-Android/pull/13183 

I've made some small changes like made some nullable fields non-null. Because I'm changing the response object, the Stats DB needs to be cleared.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/stats/time/ReferrersModel.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/stats/time/ReferrersModel.kt
@@ -8,13 +8,13 @@ data class ReferrersModel(val otherViews: Int, val totalViews: Int, val groups: 
         val url: String?,
         val total: Int?,
         val referrers: List<Referrer>,
-        var markedAsSpam: Boolean? = false
+        var markedAsSpam: Boolean = false
     )
     data class Referrer(
         val name: String,
         val views: Int,
         val icon: String?,
         val url: String?,
-        var markedAsSpam: Boolean? = false
+        var markedAsSpam: Boolean = false
     )
 }
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/ReferrersRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/ReferrersRestClient.kt
@@ -17,6 +17,8 @@ import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Error
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Success
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import org.wordpress.android.fluxc.network.rest.wpcom.stats.time.ReferrersRestClient.ReferrersResponse.Group
+import org.wordpress.android.fluxc.network.rest.wpcom.stats.time.ReferrersRestClient.ReferrersResponse.Referrer
 import org.wordpress.android.fluxc.network.utils.StatsGranularity
 import org.wordpress.android.fluxc.network.utils.getInt
 import org.wordpress.android.fluxc.store.StatsStore.FetchStatsPayload
@@ -56,14 +58,24 @@ class ReferrersRestClient
                 this,
                 url,
                 params,
-                ReferrersResponse::class.java,
+                UnparsedReferrersResponse::class.java,
                 enableCaching = false,
                 forced = forced
         )
         return when (response) {
             is Success -> {
-                response.data.groups.values.forEach { it.groups.forEach { group -> group.build(gson) } }
-                FetchStatsPayload(response.data)
+                val firstGroup = response.data.groups.values.firstOrNull()
+                val parsedGroups = firstGroup?.groups?.map {
+                    it.parse(gson)
+                } ?: listOf()
+                FetchStatsPayload(
+                        ReferrersResponse(
+                                response.data.statsGranularity,
+                                firstGroup?.otherViews,
+                                firstGroup?.totalViews,
+                                parsedGroups
+                        )
+                )
             }
             is Error -> {
                 FetchStatsPayload(response.error.toStatsError())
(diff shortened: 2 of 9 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/stats/time/ReferrersStore.kt
```diff
@@ -98,35 +98,43 @@ class ReferrersStore
...
+                val children = referrer.children?.map { child ->
+                    if (child.url == domain) {
+                        // Setting child.spam as true
+                        child.copy(markedAsSpam = spam)
+                    } else {
+                        child
                     }
                 }
+                val referrerMarkedAsSpam = if (referrer.url == domain) {
+                    // Setting referrer.spam as true
```
mchowning: Is this comment (and the "// Setting child.spam as true`) accurate? It seems like it's updating the spam fields--not necessarily setting them to true.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/ReferrersRestClient.kt
```diff
@@ -56,14 +58,24 @@ class ReferrersRestClient
...
-                ReferrersResponse::class.java,
+                UnparsedReferrersResponse::class.java,
                 enableCaching = false,
                 forced = forced
         )
         return when (response) {
             is Success -> {
-                response.data.groups.values.forEach { it.groups.forEach { group -> group.build(gson) } }
-                FetchStatsPayload(response.data)
+                val firstGroup = response.data.groups.values.firstOrNull()
```
mchowning: My understanding is that we're just persisting the first group, and that this is not really a change in the logic (we were just doing after persistence in the `TimeStepMapper::map` function before instead of here). It's not clear to me why we can discard the other groups though. If you wouldn't mind giving me a bit of the context around that just so I can understand better I'd appreciate it. 🙇 

I do like the fact that we're discarding the other groups earlier (before persisting) now--in addition to helping us ease the load on the db it also makes our intent a bit clearer imo.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/ReferrersRestClient.kt
```diff
@@ -121,39 +133,57 @@ class ReferrersRestClient
         }
     }
 
-    data class ReferrersResponse(
+    data class UnparsedReferrersResponse(
         @SerializedName("period") val statsGranularity: String?,
         @SerializedName("days") val groups: Map<String, Groups>
     ) {
         data class Groups(
```
mchowning: Having `Groups` under `UnparsedReferrersReponse` and `Group` under `ReferrerResponse` has tripped me up a couple of times when reviewing this PR. I also stumbled over the `ReferrersResponse.groups` field, which is not a `Groups` object like I initially assumed, but rather is a `List<Group>`. 

I still don't really understand the data model we're getting back in the API response, but it would be nice if we could make the names a bit more distinct on our side. Perhaps something along the lines of changing `Group` to `ParsedGroup` to reflect that it matches up with `UnparsedGroup` (have to admit I don't love `ParsedGroup` as a name idea) and/or changing the `ReferrersResponse.groups` field to something like `groupList`.

I do not feel strongly about this change, it was just a thought I had while I was reviewing this.

fluxc/src/main/java/org/wordpress/android/fluxc/model/stats/time/ReferrersModel.kt
```diff
@@ -8,13 +8,13 @@ data class ReferrersModel(val otherViews: Int, val totalViews: Int, val groups:
...
-        var markedAsSpam: Boolean? = false
+        var markedAsSpam: Boolean = false
     )
     data class Referrer(
         val name: String,
         val views: Int,
         val icon: String?,
         val url: String?,
-        var markedAsSpam: Boolean? = false
+        var markedAsSpam: Boolean = false
```
mchowning, line 18: ❤️

mchowning (commented):
Code looks good and it works great. Nice work @planarvoid ! 👍 

I left a few comments, but mostly they're just questions I had. I don't see any blocking issues.

fluxc/src/main/java/org/wordpress/android/fluxc/store/stats/time/ReferrersStore.kt
planarvoid: you're right. I've left the previous comment but they don't make sense

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/ReferrersRestClient.kt
planarvoid: Absolutely! Sorry for not sharing the context before. AFAIK there are no other groups in the response we're getting (as you noticed - we were taking only the first object in the mapper anyway), both the lists contain only one item. 
I think there is a way to get more groups if you pass more request parameters (you can for example pass in more periods and this way you'd receive data for all the periods). The API has no good description so I can't say why the response looks this way - when implementing this I was basically following what I received back and I noticed the groups always only has one item. I felt like this is a better place to make this improvement since we're already mapping the object anyway. Let me know what you think.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/ReferrersRestClient.kt
planarvoid: > I still don't really understand the data model we're getting back in the API response, but it would be nice if we could make the names a bit more distinct on our side. Perhaps something along the lines of changing Group to ParsedGroup to reflect that it matches up with UnparsedGroup (have to admit I don't love ParsedGroup as a name idea) and/or changing the ReferrersResponse.groups field to something like groupList.

You're absolutely right. It's very confusing! I'll try to make it a bit better. I noticed that there is a parameter `summarize` on the endpoint and it makes the response look a bit better (calypso is using it). I might try it to simplify things.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/ReferrersRestClient.kt
planarvoid: I've improved the naming in `72e6899fdbacd07e44123c29cbbf07a5e15c5e82`. Using the summarized endpoint doesn't work (for some reason it only works if you want data for the past N days, not if you want data for the whole year from a certain date. I've also noticed that we're not using anything from the `Child` object except the URL so I've removed that part as well.

planarvoid:
thanks for the review @mchowning ! I've addressed all of your feedback (🤞 ) and updated the WPAndroid PR. Let me know what you think :-)

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/time/ReferrersRestClient.kt
mchowning: I think those changes definitely help make the code easier to understand. Thanks for taking the time to do that (and explain some of the context to me as well 🙂 ).

mchowning (approved):
Retested and everything works great! :shipit: 

Noticed one of the connected tests failed. It passes for me locally, so I've restarted it on CI. 🤞

PR #1731, opened 2020-10-22
Encode filename when uploading media

planarvoid:
Fixes wordpress-mobile/WordPress-Android#13194

In order to fix upload of files in different encodings (like Arabic) we need to encode the file names in UTF-8 to remove a crash in OKHTTP.

This change could be tested with a related WPAndroid PR.


develric (approved):
Hey @planarvoid, tested this with the companion PR and fixed the issue as per test steps 👍 . Some more details on what I checked in the WP Android PR.

PR #1736, opened 2020-10-27
Issue/1735 wc search product tags

### LINKED ISSUE #1735: Woo: add "search" parameter when fetching product tags
We want to add the ability to filter the list of product tags, but the "search" parameter isn't passed when we fetch them.

nbradbury:
Closes #1735 - adds the "search" parameter when fetching product tags so WCAndroid can enabling filtering the tag list.

To test:

- Open the example app
- Woo > Products > Fetch product tags
- Enter a search query
- Verify the resulting tags match that query


anitaa1990 (approved):
LGTM :shipit:

PR #1739, opened 2020-10-28
Update ReactNative API requests to respond with all available information (body, headers, status)

enejb:
Currently we only forward the body of api requests back to gutenberg but in order for the requests to work as expected we need to make sure that we send over the headers. 

The headers contain information such as next links as well as if particular actions are allowed. See `canUser()` 

To tests. 
Run the tests? Do they work as expected?

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/ReactNativeRequest.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/ReactNativeRequest.kt
@@ -1,0 +1,39 @@
+package org.wordpress.android.fluxc.network.rest
+
+import com.android.volley.NetworkResponse
+import com.android.volley.Response
+import com.google.gson.Gson
+import com.google.gson.JsonElement
+import com.google.gson.JsonObject
+
+object ReactNativeRequest {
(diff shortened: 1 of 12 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/ReactNativeRequest.kt
```diff
@@ -0,0 +1,39 @@
...
+    ): Response<JsonElement> =
+            // If response has a result, update the response to have the result under the "body" key and include
+            // additional information regarding the network response under relevant keys
+            if (response.result == null || networkResponse == null) {
+                response
+            } else {
+                val json = JsonObject().apply {
+                    add(ResponseKey.body, response.result)
+                    add(ResponseKey.headers, parseHeaders(networkResponse.headers))
+                    addProperty(ResponseKey.networkTimeMs, networkResponse.networkTimeMs)
```
enejb, line 30: From our testing `networkResponse.networkTimeMs` seems to be set to 0. So it doesn't seem that accurate. 
As far as I saw the `networkResponse.notModified` key is also not currently being used anywhere. 

Maybe we can remove it for now?

PR #1740, opened 2020-10-28
Woo: Parse parentName in LineItems and handle empty attibutes

### LINKED ISSUE #1738: Woo: Add parent_name for product variations in order line items
Add support for the new `parent_name` element added in [this PR to the Woo API.](https://github.com/woocommerce/woocommerce/pull/28080)

AmandaRiu:
Closes #1738 by adding the ability to parse the new `parent_name` attribute in the `line_items` portion of the order response from the API. [Here's a link to the corresponding PR on the Woo API.](https://github.com/woocommerce/woocommerce/pull/28080). Also included in this PR are:
- Updated `LineItem.getAttributesAsString()` method to skip empty and null attributes
- Updated `LineItem.getAttributesAsString()` to not include values where the `key` is `_reduced_stock`. This is a temporary workaround until we can get a `type` element added to items in `meta-data`. This is necessary to ensure we are only including `meta-data` items that are product variation attributes. (see p1602796453088400-slack-mobile-woo-async-mtg)
- Added additional tests to `WCOrderModelTest`

### To Test
- Run tests in WCOrderModelTest

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -9,16 +9,7 @@
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value>
-          <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
-        </value>
-      </option>
-      <option name="PACKAGES_IMPORT_LAYOUT">
-        <value>
-          <package name="" alias="false" withSubpackages="true" />
-          <package name="java" alias="false" withSubpackages="true" />
-          <package name="javax" alias="false" withSubpackages="true" />
-          <package name="kotlin" alias="false" withSubpackages="true" />
-          <package name="" alias="true" withSubpackages="true" />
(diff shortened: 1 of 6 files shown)


ThomazFB (changes requested):
Hey @AmandaRiu, it's working just fine and the code looks good! I've just one suggestion since we're adding to the `WCOrderModelTest`: Do you think it would be valid to add a test to ensure that the `_reduced_stock` attribute rule under the `takeWhile` filter is always there and working? I see that we will change this in the future when the `type` is added and this rule is important so we avoid breaking the Order Details view information. WDYT?

AmandaRiu:
@ThomazFB Sounds good. I added an `_reduced_stock` element to the test json to ensure it's properly stripped in the test. Ready for another round!

ThomazFB (approved):
Awesome! I've done some additional tests to see how the tests are behaving and it's working just fine and breaking when we it should, thanks, Amanda! :shipit:

PR #1741, opened 2020-10-29
Woo/product image nullpointer fix

anitaa1990:
Fixes part of woocommerce/woocommerce-android#3083 by adding a null check when fetching the first image url of a product. I was able to reproduce this crash by SSP into the user's site from the app. I tested the changes in this PR in the Woo app after SSP to ensure the app was not crashing.

### To test
- Run `MockedStack_WCProductsTest.kt`.


nbradbury (approved):
:shipit:

PR #1742, opened 2020-10-29
Woo: Woo Services plugin availability check

0nko:
This PR adds the ability to request information about the WooCommerce Services plugin, which is necessary for implementing the shipping label creation functionality.

**Unit tests:**
- Run `WooCommerceStoreTest`
- Run `WCPluginSqlUtilsTest`

**Example app:**
1. Run the example app
2. Go to Woo -> Shipping labels
3. Select a site with the WooCommerce Services plugin installed and activated
4. Tap on the Get shipping plugin info button
5. Notice the plugin information is logged
6. Disable the plugin (or choose a different site)
7. Request the shipping plugin info
8. Notice the plugin is reported as disabled
9. Delete the plugin (or choose a different site)
10. Request the shipping plugin info
11. Notice the plugin is reported as missing

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedWCNetworkModule.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedWCNetworkModule.kt
@@ -6,11 +6,16 @@ import dagger.Module
 import dagger.Provides
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.network.UserAgent
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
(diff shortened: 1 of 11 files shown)


nbradbury:
@0nko I have a self-hosted site that had the Woo services plugin, but I deleted it and the WooCommerce plugin and the log still shows that it's installed (but not active).

Screenshot_20201029_094930

The response shows that the WooCommerce plugin isn't installed.

Screen Shot 2020-10-29 at 10 03 47 AM

It looks like `fetchWooCommerceServicesPluginInfo` simply inserts the plugins into SQLite, but it should first delete all the existing ones.

0nko:
🤦‍♂️ Thanks @nbradbury, great catch! I had a bug in the logic that deletes the data. It's fixed now and I added a test for it. Ready for another round.

nbradbury (approved):
:shipit:

PR #1745, opened 2020-10-29
Add support for filters to FetchActivityLog request

### LINKED ISSUE #1744: Add support for after, before and activity type filters to fetch activity log endpoint
Goal of this ticket is to add support for `after`, `before` and `groups` query parameters for `/v2/sites/[id]/activity`.

`after` (date) - will be used for the date range filter
`before` (date) - will be used for the date range filter
`groups` (array, String) - will be used for the activity type filter

malinajirka:
Fixes https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1744

Adds support for `after`, `before` and `groups` query parameters for `/v2/sites/[id]/activity`.

`after` (date) - will be used for the date range filter
`before` (date) - will be used for the date range filter
`groups` (array, String) - will be used for the activity type filter

The API accepts `groups` parameter in the following format "[url]?groups=XYZ&groups=ABC&groups=...
Our WPComGsonRequest accepts params as a Map which doesn't support key-duplicates. I came up with a workaround, but I don't like it at all. Let me know if you have a better idea:).

To test:
The parameters aren't being used yet, so there is nothing to test.


ashiagr (commented):
Changes look good to me 👍. 

Regarding multiple query params of same name, we can probably use this format:
`/sites/$SITE_ID/activity?group[0]=user&group[1]=plugin`

More details in the [volley-users google group discussion](https://groups.google.com/g/volley-users/c/tFRclnEbpAk/m/uiC2f9nAIgkJ).

malinajirka:
Thanks for the review @ashiagr! Thanks for the tips @ParaskP7!

I've 
- Updated the params structure in e2524db and  690f9b1.
- Extracted parameters creation into it's own method in 821b5de
- Added connected tests which compare that the response from the server differs when the filter params differ. It required using a jetpack site with a plan so I introduced a new testing site with Complete Jetpack Plan in 2218a36 and aa714c5.

It's ready for another round. If you want I can introduce the new testing site in a separate PR. Just let me know.

P.S. You'll need to update tests.properties in order to run the tests.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
```diff
@@ -82,9 +84,25 @@ class ReleaseStack_ActivityLogTestJetpack : ReleaseStack_Base() {
         assertEquals(fetchActivities.rowsAffected, activityLogForSite.size)
     }
 
+    @Test
+    fun testFetchActivitiesActivityTypeFilter() {
```
ParaskP7, line 88: ⚠️ I cannot test that as I don't know where to get the actual test values to replace the FIXME within `test.properties-example`. Can you assist here? I am getting the: `Getting Unexpected error occurred with type: INCORRECT_USERNAME_OR_PASSWORD`

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
```diff
@@ -82,9 +84,25 @@ class ReleaseStack_ActivityLogTestJetpack : ReleaseStack_Base() {
...
+        val rewindPayload = ActivityLogStore.FetchActivityLogPayload(site, groups = listOf("rewind"))
+        val userPayload = ActivityLogStore.FetchActivityLogPayload(site, groups = listOf("user"))
+
+        runBlocking { activityLogStore.fetchActivities(rewindPayload) }
+        val rewindActivities = activityLogStore.getActivityLogForSite(site)
+
+        runBlocking { activityLogStore.fetchActivities(userPayload) }
+        val userActivities = activityLogStore.getActivityLogForSite(site)
+
+        assertTrue(userActivities != rewindActivities)
```
ParaskP7, line 100: 🔍  Usually I tend to keep variable and their usage as close as possible to each other, for example for the above I would have do this instead:

```
        val rewindPayload = ActivityLogStore.FetchActivityLogPayload(site, groups = listOf("rewind"))
        runBlocking { activityLogStore.fetchActivities(rewindPayload) }
        val rewindActivities = activityLogStore.getActivityLogForSite(site)

        val userPayload = ActivityLogStore.FetchActivityLogPayload(site, groups = listOf("user"))
        runBlocking { activityLogStore.fetchActivities(userPayload) }
        val userActivities = activityLogStore.getActivityLogForSite(site)
```

Some static code analysis tools actually complain about that sometimes. 😊 

PS: The 🔍 icon means that this is 100% minor, just a trigger for discussion. Obviously, feel free to ignore those especially if the PR is ready for merge.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
```diff
@@ -272,4 +289,18 @@ class ReleaseStack_ActivityLogTestJetpack : ReleaseStack_Base() {
         assertTrue(mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS.toLong(), TimeUnit.MILLISECONDS))
         assertTrue(siteStore.sitesCount > 0)
     }
+
+    private sealed class Sites(val wpUserName: String, val wpPassword: String, val siteUrl: String) {
```
ParaskP7, line 293: ⭐ Really nice!

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
```diff
@@ -64,18 +69,51 @@ class ActivityLogRestClientTest {
                 requestQueue,
                 accessToken,
                 userAgent)
+        whenever(requestPayload.site).thenReturn(site)
     }
 
     @Test
     fun fetchActivity_passesCorrectParamToBuildRequest() = test {
         initFetchActivity()
+        val afterMillis = 1603860428000
```
ParaskP7, line 78: 🔍 I would have added an empty line below the `initFetchActivity()` just so that I separate the `Arrange` from the `Act` part of the test. Obviously, the same applies to the other test this one was cloned from.

example/tests.properties-example
```diff
@@ -195,6 +195,16 @@ TEST_WPCOM_PASSWORD_JETPACK_BETA_SITE = FIXME
 TEST_WPCOM_USERNAME_HAS_POST_FORMATS = FIXME
 TEST_WPCOM_PASSWORD_HAS_POST_FORMATS = FIXME
 
+# WP.com - Account connected to the following Jetpack sites
```
ParaskP7, line 198: ❔ Can you provide the actual test values that replace FIXME below?

example/tests.properties-example
```diff
@@ -195,6 +195,16 @@ TEST_WPCOM_PASSWORD_JETPACK_BETA_SITE = FIXME
...
+TEST_WPCOM_USERNAME_JETPACK = FIXME
+TEST_WPCOM_PASSWORD_JETPACK = FIXME
+
+# Self Hosted - Jetpack site with Complete Jetpack plan
```
ParaskP7, line 202: ❔ 3/4 of the below test values are unused, are they not? Can you help me figure out why?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -108,6 +109,25 @@ constructor(
         }
     }
 
+    private fun buildActivityParams(
```
ParaskP7: ⭐ Many thanks for extracting this into its own function!

🔍 🔍 Again, please treat this comment as very minor (double 🔍 icons). I would have named this function plainly as `buildParams(...)` since there is no other function like that in this class, and the `ActivityLog` can be inferred from the class itself. However, if you really need the name to include a description of what it builds, I suggest giving it the full name, like `ActivityLog` instead of `Activity` since a) the `Activity` can be confusing and `ActivityLog` makes it consistent with the class naming.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -108,6 +109,25 @@ constructor(
...
+    ): MutableMap<String, String> {
+        val pageNumber = offset / number + 1
+        val params = mutableMapOf(
+                "page" to pageNumber.toString(),
+                "number" to number.toString()
+        )
+
+        payload.after?.let { params["after"] = DateTimeUtils.iso8601FromDate(it) }
+        payload.before?.let { params["before"] = DateTimeUtils.iso8601FromDate(it) }
+        payload.groups.forEachIndexed { index, value ->
```
ParaskP7, line 125: ⭐ Really nice!
❤️ Goes to @ashiagr for the suggestion!

ParaskP7 (commented):
⭐ Great job @malinajirka, thanks for all the changes!

Since I am still spying on the PRs, I would neither `Approve` nor `Request changes`, instead I would just `Comment`.

I want to take this opportunity and let you know why I am using an emoji in front of my event PR comment. I attended this talk some while ago and in addition to all the really nice ideas that this talk gave it provided a really nice way of leaving comments in code reviews (see slide 27 of 32 in [this link](https://www.slideshare.net/ThiagoPorcincula/the-art-of-code-reviews)). It is a great way of organizing your comment and giving a way to the author to prioritize what to do first or even what to ignore. I would love to hear your and the team's thoughts on that! PS: Me and my previous team used that for almost a year and we ended up super happy about it.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
malinajirka: I completely agree that this approach is great in production code. However, in tests I personally prefer to section tests into "Arrange/Act/Assert" sections (even connected tests where possible). Having said that, perhaps I should change my current approach as your proposal make sense in connected tests. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
malinajirka: > I would have named this function plainly as buildParams(...) since there is no other function like that in this class

I'm a bit worried that it might not be as easily extensible - if we ever need to build params eg in `rewind`, it'll be difficult to come up with a good name. Wdyt? 

The method in which `buildActivityParams` is used is called `fetchActivity` so I tried to be consistent with it. I thought it'd be clearer that the build**Activity**Params method is tied to the fetch**Activity** method. Wdyt?

P.S. Having said that we might want to consider renaming even the method - I would have named it "fetchActivities" or "fetchActivityLogItems". But that would result in renaming even the event bus events so I'm not sure it's worth it.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
malinajirka: Fixed in dc7f2e58c8b8a57897fa9d3c33771c2f4dc62efc

example/tests.properties-example
malinajirka: We don't need them in this specific test, but I felt like it's better to keep them in the configuration file. AFAIK we don't have any other place where we store the credentials than tests.properties. If anyone needs them in the future it'll be much easier for them just to look at the tests.properties than to try to derive them from the server/reset password/...

Wdyt?

malinajirka:
❤️ Thanks so much for the review @ParaskP7! 

> I want to take this opportunity and let you know why I am using an emoji in front of my event PR comment. I attended this talk some while ago and in addition to all the really nice ideas that this talk gave it provided a really nice way of leaving comments in code reviews (see slide 27 of 32 in this link). It is a great way of organizing your comment and giving a way to the author to prioritize what to do first or even what to ignore. I would love to hear your and the team's thoughts on that! PS: Me and my previous team used that for almost a year and we ended up super happy about it.

I really like emojis on PR reviews in general so I really like even this idea! Would we need to use it across all PRs (all contributors) or do you think it's ok if one of the sides (author/reviewer) doesn't understand what the emojis mean? My only concern is that the reviews might be a bit confusing for contributors who don't understand meanings of the emojis. Wdyt?

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
malinajirka: Resolved on slack

example/tests.properties-example
malinajirka: Resolved on slack

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
malinajirka: What if we named it "buildFetchActivityParams"?

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
ParaskP7: 😄 yeap, I totally see your point! Actually, I still haven't figure out to 100% how the `Arrange` and `Act` are being used, since the line is not clear to me yet. I am used to have everything in `Arrange` and only one liner as the `Act` (at least usually). But reading through the current tests that we have, unit and connected test, I am seeing a different pattern, which is to have as part of the `Act` all the parameters of the action that causing the action itself.

I am not sure if I managed to describe my thoughts in a way that you understand me, maybe we can have a chat first so I can explain. Then, I can come back to this thread and post a summary.

example/tests.properties-example
ParaskP7: Totally agree, and thanks for the explanation! 🌟 I was just trying to make sure that I didn't miss anything here.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
ParaskP7: Yeap, totally agree that it is not worthy to rename the whole `fetchActivity`. Having it as your suggested `buildFetchActivityParams` is a good name, I like it.

PS: About the extensibility aspects, I tend to always worry about it when it will be indeed needed, seems sometime that day might not even come. Obviously, it always depends, but in this case I would have had it as simple as possible and then, when that day come where I would need to have two (or more) of those functions I would refactor accordingly. Till then, it personally helps me to keep it as simple as possible.

ParaskP7:
> My only concern is that the reviews might be a bit confusing for contributors who don't understand meanings of the emojis. Wdyt?

I totally agree on the contributors not understanding the meaning of emojis. Although those can be as simple as a total of six and self explanatory, external contributors would need a guide to understand it. This could work like that, but (as discussed) I am still not sure if this is too much.

> Would we need to use it across all PRs (all contributors) or do you think it's ok if one of the sides (author/reviewer) doesn't understand what the emojis mean?

Yeap, generally this would work well if it is used across all PRs, but it might probably be still okay for teams to use it independency, since those just add extra information while not removing anything. However, anyway this need to be communicated first with the whole team and maybe even the whole mobile department. People will discuss and decide if they like or not the idea. I just did it once in the PR for demonstrative purposes so please take this as such. Until this is further discussed I am going to add my comments as usual (no emojis).

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
malinajirka: > PS: About the extensibility aspects, I tend to always worry about it when it will be indeed needed, seems sometime that day might not even come. Obviously, it always depends, but in this case I would have had it as simple as possible and then, when that day come where I would need to have two (or more) of those functions I would refactor accordingly. Till then, it personally helps me to keep it as simple as possible.

I usually try to follow this principle as well. I agree that in this case renaming "buildParams" to something else will be a very simple task. I've renamed it to `buildParams` in aec5182a70b99ff937bd713f82dd1318a0743e65. Thanks for the suggestion and for baring with me!

ParaskP7 (commented):
🟢 from my side!

ashiagr (approved):
LGTM 👍
Thanks @malinajirka , @ParaskP7

PR #1748, opened 2020-11-04
Update okhttp version

### LINKED ISSUE #919: Resolve self-hosted SSL + OkHttp 3.9.1+ issue

planarvoid:
Fixes #919 

In this PR I'm updating OkHTTP version and implementing a fix that enables logging in with a self-signed certificate once it's been approved by the user. This should fix the issue linked in this PR. Let me know if this approach makes sense to you. I was inspired by the newer implementation of the `MemorizingTrustManager`. It seems that adding the `MemorizingHostnameVerifier` does the trick. 

This could be tested using the self-signed site from tests.properties using [this WP Android PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13291).

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -141,12 +141,7 @@ dependencies {
     // Debug dependencies
     debugImplementation 'com.facebook.flipper:flipper:0.51.0'
     debugImplementation 'com.facebook.soloader:soloader:0.9.0'
-    debugImplementation ('com.facebook.flipper:flipper-network-plugin:0.51.0') {
-        // Force Flipper to use the okhttp version defined in the fluxc module
-        // okhttp versions higher than 3.9.0 break handling for self-signed SSL sites
-        // See https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/919
-        exclude group: 'com.squareup.okhttp3'
-    }
+    debugImplementation ('com.facebook.flipper:flipper-network-plugin:0.51.0')
 
     // Coroutines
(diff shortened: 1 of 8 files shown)


aforcier:
🎉 thanks for taking this on @planarvoid ! I haven't tested WPAndroid yet, but I confirmed that I can log in and make requests to a self-signed SSL test site from the example app (and that I can't if I just update okhttp without your changes).

However the connected tests that use the self-signed ssl site (`ReleaseStack_ReactNativeWPAPIRequestTest#testAuthenticatedCallToSelfSignedSslSite` and ` ReleaseStack_SiteTestXMLRPC#testXMLRPCSelfSignedSSLFetchSites`) are failing for me, presumably some extra config is needed for the debug/test network stack to work properly?

planarvoid:
good catch @aforcier ! I've only added the hostname verifier to the debug okhttp builder, I've added it to release builder as well in e1d277c so all should be good now 👍

aforcier:
Working well now @planarvoid 👍 This seems good to go, though getting one more person to test/run the connected tests would probably be a good idea since updating okhttp is such a deep change.

malinajirka (approved):
Thanks @planarvoid! I've tested the changes on 4 different android version and it all works as expected. Great job!

Feel free to merge this PR and create a new tag.

PR #1750, opened 2020-11-05
issue/1749-wc-empty-list-crash

### LINKED ISSUE #1749: WC: Crash in getProductsByFilterOptions when empty list passed
If you pass an empty list of excluded product IDs to `getProductsByFilterOptions`, the app crashes with a SQL exception. This is because [an empty query parameter](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/92e8b84806b2c18edf083d2160e05a31fbe93851/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt#L119-L121) is used when the list isn't null.

```
SQLiteException: near ")": syntax error (code 1 SQLITE_ERROR): , while compiling: 
SELECT * FROM WCProductModel WHERE ((LOCAL_SITE_ID = ? 
AND REMOTE_PRODUCT_ID NOT IN )))
ORDER BY NAME ASC
```

nbradbury:
Closes #1749 - Fixes a crash in `getProductsByFilterOptions` if an empty (rather than null) list of excluded product IDs is passed.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
@@ -117,7 +117,9 @@ object ProductSqlUtils {
         }
 
         excludedProductIds?.let {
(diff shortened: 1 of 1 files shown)


anitaa1990 (approved):
:shipit:

PR #1751, opened 2020-11-06
Update example test properties documentation

ParaskP7:
This PR adds an extra comment on the `test.properties-example` file for the self-hosted section.

This comment will inform developers that they need to include the extra `/xmlrpc.php` suffix on almost all the self-hosted endpoint URL paths in order for them to work properly while testing. This is especially important for external contributors, trialmatticians or newcomers.


malinajirka (approved):
LGTM, thanks! ;)

PR #1755, opened 2020-11-09
Add more logs to media errors

planarvoid:
This PR adds 2 more fields to the MediaError which we'll be tracking in the WPAndroid app to analyze media upload errors. These fields only need to be filled in in the `OnMediaUploaded` events which are the only ones relevant for the media upload.  I was trying to add a meaningful message and differentiate between XMLRPC and WP.com The first field is a log message and the second field is the response status code. This will tell us whether we're handling all the possible response codes from the backend. 

You can test this PR with this WPAndroid PR - https://github.com/wordpress-mobile/WordPress-Android/pull/13325


renanferrari (approved):
I tested this PR with the WPAndroid one and everything seems to be working as expected.

LGTM :shipit: 

PS: @planarvoid I'll leave the merge to you, as I'm not sure how you're planning to do with the versioning and all that.

PR #1760, opened 2020-11-10
ScanStore: Introduce skeleton classes

ashiagr:
Issue: https://github.com/wordpress-mobile/WordPress-Android/issues/13326

This PR adds skeleton classes for the `ScanStore`, `ScanAction`,`ScanRestClient`, `ScanSqlUtils` that will be used  
for the network and persistence part of the Scan feature.

To test: 
Nothing to test, just see that basic skeleton classes are added similar to other stores (e.g. `ActivityLogStore`).

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
@@ -1,0 +1,21 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.scan
+
+import android.content.Context
+import com.android.volley.RequestQueue
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.network.UserAgent
+import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import javax.inject.Singleton
+
+@Singleton
+class ScanRestClient
+constructor(
+    dispatcher: Dispatcher,
+    private val wpComGsonRequestBuilder: WPComGsonRequestBuilder,
+    appContext: Context?,
+    requestQueue: RequestQueue,
+    accessToken: AccessToken,
+    userAgent: UserAgent
+) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent)
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
@@ -10,10 +10,9 @@ import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
 import javax.inject.Singleton
 
 @Singleton
-class ScanRestClient
-constructor(
-    dispatcher: Dispatcher,
+class ScanRestClient(
     private val wpComGsonRequestBuilder: WPComGsonRequestBuilder,
+    dispatcher: Dispatcher,
     appContext: Context?,
     requestQueue: RequestQueue,
     accessToken: AccessToken,
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
@@ -12,8 +12,7 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-class ScanStore
-@Inject constructor(
+class ScanStore @Inject constructor(
     private val scanRestClient: ScanRestClient,
     private val scanSqlUtils: ScanSqlUtils,
     private val coroutineEngine: CoroutineEngine,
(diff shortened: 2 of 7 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import javax.inject.Singleton
+
+@Singleton
+class ScanRestClient
+constructor(
+    dispatcher: Dispatcher,
+    private val wpComGsonRequestBuilder: WPComGsonRequestBuilder,
```
ParaskP7, line 14: Minor: I understand that this is how the skeleton is build everywhere, but why not having the `wpComGsonRequestBuilder` field on top of the rest? This would be the only one which will actually be used within this class, not a pass through field like the rest, right? For example, `ScanStore` and the rest of the stores are actually having those on top.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.network.UserAgent
+import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import javax.inject.Singleton
+
+@Singleton
+class ScanRestClient
+constructor(
```
ParaskP7: Minor: Why is it that the `constructor` keyword is on another line? I understand that this is how it is everywhere else as well, I just wanted to maybe understand the reasoning behind it. PS: Same applies for the `ScanStore` class.

Warning: Using the `constructor` keyword explicitly means that probably this is done because an `@Inject` annotation will come next, but in this case there is no such annotation. I am also seeing the exact same thing is not happening on other rest client classes as well. I don't know whether this is a mistake or by design, but if it is the later and that indeed the `@Inject` annotation is not required maybe the `constructor` keyword can be emitted as well. Wdyt Ashita?

ParaskP7 (commented):
Thank you so much for including me in this PR Ashita! 🌟 

It is so nice to see what such a skeleton PR should look like. 🙏 

Only minor comments from my side, just for my own educational purposes...

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ashiagr: > why not having the wpComGsonRequestBuilder field on top of the rest

Makes sense 👍. I just went with the way this parameter was positioned in the rest of clients but definitely it can go to the top.
Changed here: a43afac

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ashiagr: > Why is it that the constructor keyword is on another line?

I saw both conventions in the rest of the code - constructor on the same line and on the next. I have no strong opinion about its placement 🤷‍♀️. Moved it next to class name here: e3de6f6

> Using the constructor keyword explicitly means that probably this is done because an @Inject annotation will come next, but in this case there is no such annotation. I am also seeing the exact same thing is not happening on other rest client classes as well.  I don't know whether this is a mistake or by design, but if it is the later and that indeed the @Inject annotation is not required maybe the constructor keyword can be emitted as well. 

Nice catch! 
Removed the `constructor` keyword for `ScanRestClient`  here: a43afac

ashiagr:
Thank you for the review @ParaskP7 🙏.
I've addressed your comments. Let me know if anything is missing.

ParaskP7:
> Thank you for the review @ParaskP7 🙏.
I've addressed your comments. Let me know if anything is missing.

Thank YOU @ashiagr 🌟 
Just checked your updates, all good from my side, nothing is missing!

fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
```diff
@@ -0,0 +1,28 @@
...
+@Singleton
+class ScanStore @Inject constructor(
+    private val scanRestClient: ScanRestClient,
+    private val scanSqlUtils: ScanSqlUtils,
+    private val coroutineEngine: CoroutineEngine,
+    dispatcher: Dispatcher
+) : Store(dispatcher) {
+    @Subscribe(threadMode = ThreadMode.ASYNC)
+    override fun onAction(action: Action<*>) {
+    }
```
malinajirka, line 23: If you plan to do it on the upcoming PRs, please feel free to ignore this comment.

We need to make sure not to forget to filter out actions which are not related to this store. Something like `val actionType = action.type as? ScanAction ?: return`.

malinajirka (approved):
Thanks @ashiagr! LGTM - I've left one comment but feel free to merge it.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
ashiagr: Yep - it is coming in next PR, thank you!

PR #1761, opened 2020-11-10
Add types request to rewind endpoint

### LINKED ISSUE #1759: Add support to the rewind endpoint to accept a request payload
The goal of this ticket is to support for restoring  specific areas of a site. A request object will be added to `https://public-api.wordpress.com/wpcom/v1/activity-log/[id]/rewind/to/[id]` 

 The request will take the following form:
`
{
  "types": {
    "themes": false,
    "plugins": false,
    "uploads": true,
    "sqls": false,
    "roots": false,
    "contents": false
  }
}`

zwarm:
Fixes #1759 

This PR adds `types: Map<String, Boolean` parameter to `RewindPayload` which supports individual features when requesting a rewind (restore) action. The default is a mapOf(). This value is passed along to`ActivityLogRestClient`, which will add it to the request body when it is not empty.

**To test**
The types parameter is not in use yet; however you can test WPAndroid still works as expected using commit c52dc11 locally.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
@@ -108,7 +108,7 @@ class ActivityLogStoreTest {
 
     @Test
     fun onRewindActionCallRestClient() = test {
-        whenever(activityLogRestClient.rewind(eq(siteModel), any())).thenReturn(
+        whenever(activityLogRestClient.rewind(eq(siteModel), any(), any())).thenReturn(
                 RewindResultPayload(
                         "rewindId",
                         null,
@@ -117,7 +117,7 @@ class ActivityLogStoreTest {
         )
 
         val rewindId = "rewindId"
-        val payload = RewindPayload(siteModel, rewindId)
+        val payload = RewindPayload(siteModel, rewindId, mutableMapOf())
         val action = ActivityLogActionBuilder.newRewindAction(payload)
         activityLogStore.onAction(action)
 
@@ -218,7 +218,7 @@ class ActivityLogStoreTest {
         val payload = ActivityLogStore.RewindResultPayload(rewindId, restoreId, siteModel)
         whenever(activityLogRestClient.rewind(siteModel, rewindId)).thenReturn(payload)
 
-        activityLogStore.onAction(ActivityLogActionBuilder.newRewindAction(RewindPayload(siteModel, rewindId)))
+        activityLogStore.onAction(ActivityLogActionBuilder.newRewindAction(RewindPayload(siteModel, rewindId, mutableMapOf())))
 
         val expectedChangeEvent = ActivityLogStore.OnRewind(rewindId, restoreId, ActivityLogAction.REWIND)
         verify(dispatcher).emitChange(eq(expectedChangeEvent))
@@ -246,6 +246,46 @@ class ActivityLogStoreTest {
 
         assertEquals(activityLogModel, returnedItem)
         verify(activityLogSqlUtils).getActivityByActivityId(rewindId)
+    }
+
+    @Test
+    fun onRewindActionWithTypesCallRestClient() = test {
+        whenever(activityLogRestClient.rewind(eq(siteModel), any(), any())).thenReturn(
+                RewindResultPayload(
+                        "rewindId",
+                        null,
+                        siteModel
+                )
+        )
+
+        val rewindId = "rewindId"
+        val payload = RewindPayload(siteModel, rewindId, mutableMapOf())
+        val action = ActivityLogActionBuilder.newRewindAction(payload)
+        activityLogStore.onAction(action)
+
+        verify(activityLogRestClient).rewind(siteModel, rewindId)
+    }
+
+    @Test
+    fun emitsRewindResultWhenSendingTypes() = test {
+        val rewindId = "rewindId"
+        val restoreId = 10L
+        val types = mapOf(
+            "themes" to true,
+            "plugins" to true,
+            "uploads" to true,
+            "sqls" to true,
+            "roots" to true,
+            "contents" to true
+        )
+
+        val payload = ActivityLogStore.RewindResultPayload(rewindId, restoreId, siteModel)
+        whenever(activityLogRestClient.rewind(siteModel, rewindId, types)).thenReturn(payload)
+
+        activityLogStore.onAction(ActivityLogActionBuilder.newRewindAction(RewindPayload(siteModel, rewindId, types)))
+
+        val expectedChangeEvent = ActivityLogStore.OnRewind(rewindId, restoreId, ActivityLogAction.REWIND)
+        verify(dispatcher).emitChange(eq(expectedChangeEvent))
     }
 
     private suspend fun initRestClient(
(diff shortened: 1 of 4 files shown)


example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
```diff
@@ -117,7 +117,7 @@ class ActivityLogStoreTest {
         )
 
         val rewindId = "rewindId"
-        val payload = RewindPayload(siteModel, rewindId)
+        val payload = RewindPayload(siteModel, rewindId, mutableMapOf())
```
ParaskP7: Minor: Maybe having that as `mapOf()` is better. Wdyt? In general I usually try to have everything as immutable as possible, even for tests, just so I don't give the wrong impression to the reader. For example, in this case I would have thought that the `RewindPayload` defines it as mutable and as such this need to be passed as mutable.

PS: The same apply to the test below.

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
```diff
@@ -218,7 +218,10 @@ class ActivityLogStoreTest {
         val payload = ActivityLogStore.RewindResultPayload(rewindId, restoreId, siteModel)
         whenever(activityLogRestClient.rewind(siteModel, rewindId)).thenReturn(payload)
 
-        activityLogStore.onAction(ActivityLogActionBuilder.newRewindAction(RewindPayload(siteModel, rewindId)))
+        activityLogStore.onAction(ActivityLogActionBuilder.newRewindAction(RewindPayload(
+                siteModel,
+                rewindId,
+                mutableMapOf())))
```
ParaskP7: Super Minor: This line can be reformatted accordingly to the Kotlin style guide.

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
```diff
@@ -248,6 +251,46 @@ class ActivityLogStoreTest {
         verify(activityLogSqlUtils).getActivityByActivityId(rewindId)
     }
 
+    @Test
+    fun onRewindActionWithTypesCallRestClient() = test {
```
ParaskP7, line 257: Question: The test name suggest that this tests a rewind action `with types` but actually an empty list of types are passed. Am I missing anything?

Minor: Usually I would have the `whenever` clause as narrowed down as possible in order to avoid surprises, like below, wdyt? 
```
val rewindId = "rewindId"
val resultPayload = RewindResultPayload(rewindId, null, siteModel)
whenever(activityLogRestClient.rewind(siteModel, rewindId)).thenReturn(resultPayload)
```

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
```diff
@@ -248,6 +251,46 @@ class ActivityLogStoreTest {
...
+        val rewindId = "rewindId"
+        val payload = RewindPayload(siteModel, rewindId, mutableMapOf())
+        val action = ActivityLogActionBuilder.newRewindAction(payload)
+        activityLogStore.onAction(action)
+
+        verify(activityLogRestClient).rewind(siteModel, rewindId)
+    }
+
+    @Test
+    fun emitsRewindResultWhenSendingTypes() = test {
```
ParaskP7, line 281: Question: This test will pass even if no `types` are passed, so the `sending types` suggests something I don't quite understand, right? Can you please explain it a bit to me?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -87,9 +87,14 @@ constructor(
         }
     }
 
-    suspend fun rewind(site: SiteModel, rewindId: String): RewindResultPayload {
+    suspend fun rewind(site: SiteModel, rewindId: String, types: Map<String, Boolean> = mapOf()): RewindResultPayload {
```
ParaskP7: Question: Since a new `types` parameters is added here, maybe the `rewind(...)` function misses now a test within `ActivityLogRestClientTest`?

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
```diff
@@ -172,7 +172,10 @@ class ActivityLogStore
 
     class FetchRewindStatePayload(val site: SiteModel) : Payload<BaseRequest.BaseNetworkError>()
 
-    class RewindPayload(val site: SiteModel, val rewindId: String) : Payload<BaseRequest.BaseNetworkError>()
+    class RewindPayload(
+        val site: SiteModel,
+        val rewindId: String,
+        val types: Map<String, Boolean> = mapOf()) : Payload<BaseRequest.BaseNetworkError>()
```
ParaskP7: Super Minor: This line can be reformatted accordingly to the Kotlin style guide. The surrounding functions are having the correct formatting.

ParaskP7 (commented):
Thank you so much for including me in this PR Annmarie! 🌟

Only minor comments and questions from my side, just for my own educational purposes, so please don't mind me much... 😄

malinajirka (commented):
Thank you @zwarm for the changes!

I skipped the issues Petros commented on. Thank you @ParaskP7!

Have you considered introducing a new data class instead of using a generic map in the payload? With the current approach the client doesn't know what they can pass as parameters and they can even pass parameters which are invalid. AFAIU the list of supported types will be hardcoded anyway (we don't fetch it from the server, right?). I think it might become a bit more readable and easier to use from the client. We could use gson's "toMap" extension function to convert the data class into a map. Wdyt?

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
zwarm: Thanks @ParaskP7 - this test has been changed - not using map anymore and mutable was just an oversight -  nice catch

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
zwarm: Test has been updated. Tks. :)

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
zwarm: Test updated to include types. Tks

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
zwarm: Removed

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
zwarm: Added in 1ef014a

malinajirka (approved):
It seems you unintentionally edited my comment instead of posting it as a new comment. So here is what you wrote

```
> Have you considered introducing a new data class instead of using a generic map in the payload?

Yes, it was option number two - I switched it in. And we don't have to do anything for conversion, 
gson will take care of it when it builds the request because it uses <String, Any> :)
```


Thanks for the changes @zwarm! Works as expected.

PR #1762, opened 2020-11-11
Feature/add p2 flag to sitemodel

### LINKED ISSUE #1757: Add P2 support to SiteModel and connected test with real P2 site
Site endpoint should return `is_wpforteams_site` in `options` object. We need to map it to SiteModel and provide an appropriate getter.

We should test this with a real P2 site.

khaykov:
Fixes #1757 

This PR adds a P2 flag to SiteModel and related models, mapper, and DB.

I also added a connected test with a real P2 site that checks for the flag, so we would know if something unexpected will change on the backend. You will have to get fresh `tests.properties` for the test to work.


develric (approved):
Hey @khaykov thanks for this PR 🙇 ! Code LGTM and I smoke tested it with WPAndroid develop and didn't notice abnormalities 😄. Also tested the db update script and it looks good. :shipit:

PR #1763, opened 2020-11-11
ScanStore: Fetch Scan State

ashiagr:

Jetpack Scan APIs [Internal Ref `pcdRpT-3J`]

`/sites/$site/scan` API is used to
- know if the scan feature is unlocked for a site
- get progress of a current scan
- get a list of threats for a site.

This PR introduces `/sites/$site/scan` API to the `ScanStore` and adds corresponding model classes, `ScanAction`, fetch methods in the `ScanRestClient` and `ScanStore`.

#### To test

- Add a breakpoint on this line 
https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/90f65e413066434adfcda59552a8cc12ee93778a/fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt#L54

- debug temporary test `testFetchScanState` in the `ReleaseStack_ScanTestJetpack` to see fetch results in the `payload.scanStateModel` at the breakpoint above.

You might need to update `tests.properties` if you haven't already to get jetpack complete site details.

#### Note
API response is not persisted anywhere yet. Storing the data to DB and getting from it via `ScanStore` will be supported in an upcoming PR. Tests will also be added at that point.

#### Questions
1. As we use the same API for different actions listed above, I'm in dilemma if there should be separate actions to 

   - get just the "scan feature unlocked status" [with a separate model to just contain information about this status]
   - get progress of a current scan
   - get list of threats for a site. 

    Let me know if you have suggestions.

2. ~Is it fine to map both `most_recent` and `current` to a common object `ScanProgressStatus` (See c54b4e4)? Seems like only one of them appear in the response based on the scan state and they have common fields `is_initial`, `timestamp` and `progress`.~ I've kept them separate in 799a7b4

#### Todo
- ~Map "current" object when state is "scanning"~ Done in c54b4e4
- ~Map "reason"~ Done in 5a7e6b9


fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
```diff
@@ -0,0 +1,37 @@
+package org.wordpress.android.fluxc.model.scan
+
+data class ScanStateModel(
```
ParaskP7, line 5: Idea: As I understand this model is an almost 1-to-1 mapping of the response, which in this case might not do as any favours. Maybe it is better if a separate model is created per `state`. For example, we could have a sealed class that contains all individual models per state. This way, it might be easier later on to deal with the difference in business logic. Wdyt?

Disclaimer: I don't know much about how this feature will progress, it is just that usually when I see `state` or `type` or something that indicates a difference in response I found it better if those are mapped using a sealed class instead of 1-to-1 enums. But, obviously, maybe this is not the case.

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ThreatModel.kt
```diff
@@ -0,0 +1,24 @@
+package org.wordpress.android.fluxc.model.scan
+
+data class ThreatModel(
```
ParaskP7, line 3: Question: Why was the thread model extracted to a separate model, is that on purpose because it will be used with another model in the future? Otherwise, I find it better if all the relevant models are located in a single file, sometimes not necessary in a single class, but having them in a single file helps.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
```diff
@@ -17,4 +33,92 @@ class ScanRestClient(
     requestQueue: RequestQueue,
     accessToken: AccessToken,
     userAgent: UserAgent
-) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent)
+) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
```
ParaskP7, line 35: Question: Shouldn't there be an associated unit test with this rest client as well?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
```diff
@@ -17,4 +33,92 @@ class ScanRestClient(
...
+                            isInitial = it.isInitial
+                    )
+                }
+        )
+    }
+
+    private fun buildErrorPayload(site: SiteModel, errorType: ScanStateErrorType) =
+            FetchedScanStatePayload(ScanStateError(errorType), site)
+
+    private fun <T> genericToError(
```
ParaskP7: Question: This is the same `genericToError(...)` method that exist also within `ActivityLogRestClient`. Is that a duplicate that can be reused or is it needed to be be created as such every time a new such rest client is created?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
```diff
@@ -0,0 +1,49 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.scan
+
+import com.google.gson.annotations.SerializedName
+import org.wordpress.android.fluxc.network.Response
+
+data class ScanStateResponse(
```
ParaskP7, line 7: Question: Could it be possible that we create a table with all the possible response results of the api for this model? Since there are indeed different combinations, it might be beneficial if we actually come up with a table or even add mock `.json` responses somewhere, which will contain all the possible results. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
```diff
@@ -0,0 +1,49 @@
...
+    val state: String,
+    val threats: List<Threat>?,
+    val credentials: List<Credentials>?,
```
ParaskP7: Question: From what I understand by playing a bit with the responses that I received from the api, the `threats` and `credentials` can be an empty list. But, can they actually be null as well? When will that happen? Is it when we get a specific state other than `idle` or `scanning`?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt
```diff
@@ -1,6 +1,7 @@
 package org.wordpress.android.fluxc.persistence
 
+import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-class ScanSqlUtils
+class ScanSqlUtils @Inject constructor()
```
ParaskP7, line 7: Question: How will this class be used? Will it need the `@Inject constructor()` afterwards somehow into injecting extra fields?

ParaskP7 (commented):
Good job Ashita! 🌟 

And again, thank you for including me in this! 🙏 

PS: About your questions, I don't know much about the different actions that you refer to and their usages, but I did add some thoughts on the different response that the api provides us. You will find those on individual comments below. Maybe this will provide some help from my side.

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
ashiagr: I've based this class upon [RewindStatusModel](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/model/activity/RewindStatusModel.kt) class which has a similar structure and responsibility. 

I can try converting it to a sealed class if it helps.

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ThreatModel.kt
ashiagr: This was done on purpose as we have another API `/sites/SITE_ID/scan/threat/THREAT_ID` that might need `ThreatModel`.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ashiagr: Yep, tests are coming soon.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
ashiagr: Yeah I noticed at least `threats` can be `null` for status `unavailable`.
`{"body":{"state":"unavailable","threats":null,"has_cloud":false,"reason":"wpcom_site"},"status":200,"headers":{"Allow":"GET"}}`

I wasn't able to verify if `credentials` can be null.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt
ashiagr: This class will be used to store scan results into db and is injected into `ScanStore`. 
From [docs](https://dagger.dev/dev-guide/):
> Classes that lack @Inject annotations cannot be constructed by Dagger. 

Am I missing something?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
ashiagr: Nice suggestion! I'll surely consider once response structure is finalised and I start writing tests. Thank you!

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
ParaskP7: Oh cool, wasn't aware of the `RewindStatusModel`, maybe it is better to keep them consistent. I am not sure how exactly this is going to be used later on, as such I cannot suggest to either converting it or not. For sure, I would talk it through first!

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ThreatModel.kt
ParaskP7: Great, thanks for the explanation, it makes sense then!

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
ParaskP7: Cool, thanks for letting me know! I think when we get to the table discussed above or a list of all possible `.json` responses by the api, things will get clearer.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ashiagr: Hmm, I'm not sure if rest clients in this case should interact with one another. May be I should extract to a util class? What do you suggest?

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
ashiagr: @malinajirka, @zwarm  your opinion needed 😀

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt
ParaskP7: 🤓 Don't know, it might be that I am missing something! I haven't worked with Dagger for a long time now... I was thinking that having the `@Inject` annotation as such is only necessary for when field are included in the constructor, else it is not, probably I am wrong!

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ParaskP7: Yea, maybe extracting it into a util class might be a good step forward. I am not sure how this `genericToError(...)` is handled on all the rest of the `RestClient` classes, maybe it is a good idea to figure out the common denominator, extract that logic and reuse that everywhere.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt
ashiagr: ok I ran `testFetchScanState() `without `@Inject` annotation in the `ScanSqlUtils` and get this error:

`error: [Dagger/MissingBinding] org.wordpress.android.fluxc.persistence.ScanSqlUtils cannot be provided without an @Inject constructor or an @Provides-annotated method.`

I see few classes with `@Inject` annotation and empty constructor e.g. `WhatsNewSqlUtils`. Let me know if you have any suggestion if we can remove it, Dagger confuses me as well 😅.

ParaskP7: Yeap, yeap, thank you so much for testing it for me! 🤗 I did test it as well just know and it is needed! I really need to brush my knowledge of Dagger one of these days... too much of Koin spoiled me lately! 😝

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
malinajirka: I like the idea of using sealed classes with a more strict structure. However, since the Scan and Rewind actions are quite similar and we'll probably introduce similar classes in WPAndroid (eg ScanProgressChecker) I'd keep them consistent. So I'd either change both to the new structure or keep the current structure.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
malinajirka: > Hmm, I'm not sure if rest clients in this case should interact with one another.

I agree that they shouldn't interact with each other.

> May be I should extract to a util class? What do you suggest?

I quickly looked at the usages and I guess we could extract it to a util class. I usually extract a code when I need to copy-paste the code for the third time (it's easier to find the common denominator). But feel free to extract it now.

malinajirka (commented):
> As we use the same API for different actions listed above, I'm in dilemma if there should be separate actions to
> - get just the "scan feature unlocked status" [with a separate model to just contain information about this status]
> - get progress of a current scan
> - get list of threats for a site.
> 
> Let me know if you have suggestions.

Since the API (AFAIK) doesn't support returning a different response for these different use cases. In other words the API always returns the same response. I'd personally vote for having a single action with the whole model in FluxC.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ParaskP7: The `for the 3rd time` rule it a good one, I like! 🌟

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
ashiagr: > I'd either change both to the new structure or keep the current structure.

I checked and found `RewindStatusModel` is used in several classes:

```
RewindStatusService
ActivityLogRestClient
ActivityLogSqlUtils
ReleaseStack_ActivityLogTestJetpack
RewindProgressCheckerTest
ActivityLogViewModelTest
ActivityLogRestClientTest
```
Refactoring it to new structure + testing that it hasn't introduced new bugs will take some time. I'd suggest we keep current structure for now. 

I really like the idea of using sealed classes for different states and will definitely use it in the future where we do not have to make so many changes to the existing logic.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ashiagr: I've tried to extract `errorType` mapping to a mapper class here: 1803d71.

> I am not sure how this genericToError(...) is handled on all the rest of the RestClient classes, maybe it is a good idea to figure out the common denominator

I gave this class a generic name `NetworkErrorMapper`. We can probably use this class every time we have to map an `errorType`. Currently it is used only for ActivityLog and Scan.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
```diff
@@ -0,0 +1,52 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.scan
```
malinajirka, line 1: I'm always unsure about this. I prefer to use as many non-nullable fields as possible throughout the codebase. However, I'm not sure we can rely on the API that it'll always return the values. I think in other models we have most of the fields nullable in the response and we try to make them non-nullable during the mapping to models (eg. list is nullable in the response, but we store an empty list in the model). Having said that, if we know the API won't return nulls here, I'll all up for keeping them non-nullable even in the response. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
```diff
@@ -20,9 +27,67 @@ class ScanStore @Inject constructor(
...
+                }
+            }
+        }
     }
 
     override fun onRegister() {
         AppLog.d(AppLog.T.API, this.javaClass.name + ": onRegister")
     }
+
+    @SuppressLint("WrongConstant")
```
malinajirka: Could you please double check we need to use this suppress annotation here?

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
```diff
@@ -0,0 +1,41 @@
+package org.wordpress.android.fluxc.model.scan
```
malinajirka, line 1: I noticed that the model for rewind is called "Rewind**Status**Model", but we use "Scan**State**Model" here. Is it intentional? If it's not we might want to consider renaming the scan model, wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import com.google.gson.annotations.SerializedName
+import org.wordpress.android.fluxc.network.Response
+import java.util.Date
+
+data class ScanStateResponse(
+    val state: String,
+    val threats: List<Threat>?,
+    val credentials: List<Credentials>?,
+    val reason: String?,
```
malinajirka: Nitpick: I personally prefer to use `@SerializedName` annotation for all the fields no matter the structure of their names. It allows us to rename the fields without worrying about breaking the api contract.

malinajirka (commented):
Great job, thanks @ashiagr! I've left a few comments, but nothing vital. Let me know what you think.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
ashiagr: Nice catch! Removed in f7caf3a

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
ashiagr: Added in 20144a0

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
ashiagr: I noticed we used terms like "FETCH_REWIND_**STATE**",  "FetchRewind**State**Payload", "storeRewind**State**", "FetchedRewind**State**Payload" and felt like using "State" in the model name also. I don't have a strong opinion and can change it to "Scan**Status**Model". I'll keep rest of the names like  "FETCH_SCAN_**STATE**" similar to rewind module.

Is that ok?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
ashiagr: > if we know the API won't return nulls here, I'll all up for keeping them non-nullable even in the response. 

Response contains nullable fields if the API returns null. I've verified several fields in the response.

E.g. threats is null for "unavailable" state.

`{"body":{"state":"unavailable","threats":null,"has_cloud":false,"reason":"wpcom_site"},"status":200,"headers":{"Allow":"GET"}}`

Also 
`reason` is null for states other than "unavailable" state.
`most_recent` is null for "scanning"  state.
`current` is null for "idle" state.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1763#discussion_r522059589

> I think in other models we have most of the fields nullable in the response and we try to make them non-nullable during the mapping to models (eg. list is nullable in the response, but we store an empty list in the model).

I tried to be consistent with the [RewindStatusModel](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/model/activity/RewindStatusModel.kt) which also contains nullable fields like `credentials`, `reason`.

I can convert them to non-nullable fields for the scan model though.

malinajirka: > Response contains nullable fields if the API returns null. I've verified several fields in the response. 

Just to be clear, I was suggesting we might want to make them `nullable` (not `non-nullable`) in the Response object since we don't know what the API will return. Having said that if we are at least a bit confident that we tested all the possible responses and some of the fields were never null, I guess we can keep them `non-nullable`.

> I tried to be consistent with the RewindStatusModel which also contains nullable fields like credentials, reason.

Yeah, that make sense. Let's keep both consistent. Good call.

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
malinajirka: Ohh I see, thanks for the clarification. Let's keep it as is for now. We might want to consider renaming the RewindStatusModel to RewindStateModel in the future.

malinajirka (approved):

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStateResponse.kt
```diff
@@ -5,47 +5,47 @@ import org.wordpress.android.fluxc.network.Response
...
-            val type: String,
-            val slug: String,
-            val name: String,
-            val version: String,
-            val isPremium: Boolean
+            @SerializedName("type") val type: String,
+            @SerializedName("slug") val slug: String,
+            @SerializedName("name") val name: String,
+            @SerializedName("version") val version: String,
+            @SerializedName("isPremium") val isPremium: Boolean
```
malinajirka, line 35: Could you please just double check this field - does the api really return "isPremium" and not "is_premium"? It seems the api returns "is_" prefix for other boolean fields.

ashiagr: I had the same doubt 😅. Checked it is `isPremium`.

"threats": [
		{
			"id": 35220155,
			"signature": "Vulnerable.WP.Extension",
			"description": "The plugin Calendar (version 1.3.1) has a known vulnerability",
			"first_detected": "2020-11-14T01:33:27.000Z",
			"status": "current",
			"fixable": {
				"fixer": "update",
				"target": "1.3.14"
			},
			"extension": {
				"type": "plugin",
				"slug": "calendar",
				"name": "Calendar",
				"version": "1.3.1",
				**"isPremium": false**
			}
		}
	],

malinajirka: Awesome, you are always several steps ahead 🥇

PR #1765, opened 2020-11-12
NPE in okhttpstack

planarvoid:
With updated OkHttp it's no longer possible to send invalid POST requests with empty body. That's now causing a NullPointerException. This fix adds an empty body instead when it was null before.

The other changes are small improvements based on what Android Studio suggested

This could be tested with the Reader tags in the [linked WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13372).


malinajirka (approved):
LGTM, thanks!

PR #1766, opened 2020-11-12
Add support for rewind downloads endpoint

### LINKED ISSUE #1764: Add support for rewind downloads endpoint
The goal of this ticket is to add support for preparing a backup for download

https://public-api.wordpress.com/wpcom/v2/sites/182675569/rewind/downloads

The request will take the following form:
```
{
  "rewindId": "1603788401.423",
  "types": {
    "themes": false,
    "plugins": false,
    "uploads": true,
    "sqls": false,
    "roots": false,
    "contents": false
  }
}
```
The response will take the following form:
```
{
  "downloadId": 276158,
  "rewindId": "1603788401.423",
  "backupPoint": "2020-10-27T08:46:41+00:00",
  "startedAt": "2020-10-27T11:00:23+00:00",
  "progress": 0
}
```

zwarm:
Fixes #1764 

This PR adds support for `https://public-api.wordpress.com/wpcom/v2/sites/182675569/rewind/downloads` endpoint.
The logic was added to `ActivityLogStore`, as it's an extension of Activity Log functionality.

I was on the fence about creating `DownloadRequestTypes` instead of just reusing `RewindRequestTypes` and am open to a reuse if we can find a proper name. `RewindAndDownloadRequestTypes` just feels like a mouth full. Same goes for Error types - they don't appear to be reused among the separate functions in ActivityLog, so I maintained consistency and created a new set for Download. 

To test
The feature is not being used yet, so there is nothing to test.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
@@ -309,42 +309,6 @@ class ActivityLogRestClientTest {
         }
     }
 
-    private suspend fun initFetchActivity(
-        data: ActivitiesResponse = mock(),
-        error: WPComGsonNetworkError? = null
-    ): Response<ActivitiesResponse> {
-        val response = if (error != null) Response.Error<ActivitiesResponse>(error) else Success(data)
-        whenever(wpComGsonRequestBuilder.syncGetRequest(
-                eq(activityRestClient),
-                urlCaptor.capture(),
-                paramsCaptor.capture(),
-                eq(ActivitiesResponse::class.java),
-                eq(false),
-                any(),
-                eq(false))
-        ).thenReturn(response)
-        whenever(site.siteId).thenReturn(siteId)
-        return response
-    }
-
-    private suspend fun initFetchRewindStatus(
-        data: RewindStatusResponse = mock(),
-        error: WPComGsonNetworkError? = null
-    ):
-            Response<RewindStatusResponse> {
-        val response = if (error != null) Response.Error<RewindStatusResponse>(error) else Success(data)
-        whenever(wpComGsonRequestBuilder.syncGetRequest(
-                eq(activityRestClient),
-                urlCaptor.capture(),
-                paramsCaptor.capture(),
-                eq(RewindStatusResponse::class.java),
-                eq(false),
-                any(),
-                eq(false))).thenReturn(response)
-        whenever(site.siteId).thenReturn(siteId)
-        return response
-    }
-
     @Test
     fun postRewindOperationWithTypes() = test {
         val restoreId = 10L
@@ -429,6 +393,42 @@ class ActivityLogRestClientTest {
         val payload = activityRestClient.download(site, rewindId, types)
 
         assertTrue(payload.isError)
+    }
+
+    private suspend fun initFetchActivity(
+        data: ActivitiesResponse = mock(),
+        error: WPComGsonNetworkError? = null
+    ): Response<ActivitiesResponse> {
+        val response = if (error != null) Response.Error<ActivitiesResponse>(error) else Success(data)
+        whenever(wpComGsonRequestBuilder.syncGetRequest(
+                eq(activityRestClient),
+                urlCaptor.capture(),
+                paramsCaptor.capture(),
+                eq(ActivitiesResponse::class.java),
+                eq(false),
+                any(),
+                eq(false))
+        ).thenReturn(response)
+        whenever(site.siteId).thenReturn(siteId)
+        return response
+    }
+
+    private suspend fun initFetchRewindStatus(
+        data: RewindStatusResponse = mock(),
+        error: WPComGsonNetworkError? = null
+    ):
+            Response<RewindStatusResponse> {
+        val response = if (error != null) Response.Error<RewindStatusResponse>(error) else Success(data)
+        whenever(wpComGsonRequestBuilder.syncGetRequest(
+                eq(activityRestClient),
+                urlCaptor.capture(),
+                paramsCaptor.capture(),
+                eq(RewindStatusResponse::class.java),
+                eq(false),
+                any(),
+                eq(false))).thenReturn(response)
+        whenever(site.siteId).thenReturn(siteId)
+        return response
     }
 
     private suspend fun initPostRewind(
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
@@ -345,41 +345,6 @@ class ActivityLogRestClientTest {
         return response
     }
 
-    private suspend fun initPostRewind(
-        data: RewindResponse = mock(),
-        error: WPComGsonNetworkError? = null
-    ): Response<RewindResponse> {
-        val response = if (error != null) Response.Error<RewindResponse>(error) else Success(data)
-
-        whenever(wpComGsonRequestBuilder.syncPostRequest(
-                eq(activityRestClient),
-                urlCaptor.capture(),
-                eq(null),
-                eq(mapOf()),
-                eq(RewindResponse::class.java)
-        )).thenReturn(response)
-        whenever(site.siteId).thenReturn(siteId)
-        return response
-    }
-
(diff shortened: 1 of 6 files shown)


example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
```diff
@@ -430,4 +431,63 @@ class ActivityLogRestClientTest {
...
+                plugins = true,
+                uploads = true,
+                sqls = true,
+                roots = true,
+                contents = true)
+        initPostDownload(rewindId = rewindId, data = response, requestTypes = types)
+
+        val payload = activityRestClient.download(site, rewindId, types)
+
+        assertEquals(downloadId, payload.downloadId)
```
ParaskP7, line 378: Question: Why is `downloadId` the only one value asserted and not the whole `payload`? Is this done by design just to verify it is not an error one?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -116,6 +120,31 @@ constructor(
         }
     }
 
+    suspend fun download(site: SiteModel, rewindId: String, types: DownloadRequestTypes): DownloadResultPayload {
+        val url = WPCOMV2.sites.site(site.siteId).rewind.downloads.url
+        val request = mapOf("rewindId" to rewindId,
```
ParaskP7: Minor: The `request` variable can become a one-liner.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -116,6 +120,31 @@ constructor(
...
+                "types" to types)
+        val response = wpComGsonRequestBuilder.syncPostRequest(this, url, null, request, DownloadResponse::class.java)
+        return when (response) {
+            is Success -> {
+                DownloadResultPayload(
+                        rewindId,
```
ParaskP7: Question: Why not getting the `rewindId` from the response itself? If that is indeed not needed since I understand the request/response should match ids, then maybe the `rewindId should be removed from the response. But, then again, I am not sure if this response will be used elsewhere where the `rewindId` would be required, if it does, then it makes sense to keep it there.

ParaskP7 (commented):
Good job Annmarie! 🌟

And again, thank you for including me in this! It is so nice for me to be able to see all these new to me aspects of adding such a solution in our codebase. 🙏

Apart from the minors and questions I posted individually, I can suggest the following generic minor comments:
- Store and rest client classes and tests affected by this PR can be reformatted according to Kotlin style guide. 📏  📐 
- There are some occasions of redundant qualifier names that can be fixed (an example is the `ActivityLogStore` within `ActivityLogStoreTest`). 🔍 
- There are some occasions of explicit type arguments that can be fixed (an example is the `DownloadResponse` within the `ActivityLogRestClientTest`). 🔍 
- I would have personally created multiple commits to derive with this solution. I am just saying that so you know what to expect from my side when I start contributing like you all. 😊

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
zwarm: @ParaskP7 - The assumption I made in this case is that if donwloadId is there, then yes it didn't fail.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
zwarm: @ParaskP7 - yes, I know, but my eyes can read it easier this way. I changed it anyway

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
zwarm: @ParaskP7 - Good eye, it can come from the response in the success branch. In commit 03ca555

zwarm:
> * reformatted according to Kotlin style guide

@ParaskP7 - my settings were off, thanks for pointing this out. I'll take care of this in my next PR - which builds off of this PR. 👍

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
ParaskP7: Do you think it is best if these kind of tests do only that and then another set of test could assert the whole payload for verification purposes? Or, do you think it it better that we have one test that asserts the full payload and get over with it? Maybe, we do have another set or tests that do that already, on another layer?

It seems to me that if I were to have written this test I would have created and assert the whole payload instead. I am just curious and trying to understand better, in order I can later on do the very same thing you all doing.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
ParaskP7: No worries, it is a super minor and I was just pointing that for two main reasons:
1) The obvious one that it saves lines (as per the comment).
2) The fact that even in this form, it would need to be reformatted further, and that would take a couple of more lines.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
ParaskP7: Coolio, perfect, thanks much! 🌟

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
```diff
@@ -430,4 +431,63 @@ class ActivityLogRestClientTest {
 
         assertTrue(payload.isError)
     }
+
+    private suspend fun initPostDownload(
```
malinajirka: Nitpick: We might want to consider moving this method to the bottom/top of the class. Wdyt?

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
```diff
@@ -430,4 +431,63 @@ class ActivityLogRestClientTest {
...
+        data: DownloadResponse? = null,
```
malinajirka: Nitpick: Is there any reason why we want to keep the data nullable? If not, we might want to consider making this data non-nullable and replace "null" with "mock()"? Wdyt?

malinajirka (approved):
LGTM, thanks @zwarm! ;)

> I was on the fence about creating DownloadRequestTypes instead of just reusing RewindRequestTypes and am open to a reuse if we can find a proper name. RewindAndDownloadRequestTypes just feels like a mouth full. Same goes for Error types - they don't appear to be reused among the separate functions in ActivityLog, so I maintained consistency and created a new set for Download.

I'd have gone with the same solution as you did.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
zwarm: I moved all "initPost" methods to the bottom of the class - there were existing methods scattered around too.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
zwarm: @malinajirka Because I followed the pattern of the other init functions. I changed them all in the latest commit.

malinajirka: Ohh, I missed that. Thanks for improving the code ;)!

PR #1768, opened 2020-11-12
Woo: Order Product Attribute fixes

AmandaRiu:
This PR adds the following fixes to the `WCOrderModel.LineItem.Attribute` logic:
- Added a new custom deserializer to prevent crashes due to unexpected values being returned by the API. For example, `display_value` containing a json array instead of a string. (see https://github.com/woocommerce/woocommerce-android/issues/3153)
- Capitalize the first letter of each attribute returned from `getAttributesAsString`
- Filter out line items where the `display_key` begins with `_`.
- Added test scenarios for each of the above items.

### To Test
- Run the unit test `WCOrderModelTest`
- Run the tests in `MockedStack_WCOrdersTest`

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/order/WCOrderModelTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/order/WCOrderModelTest.kt
@@ -95,7 +95,7 @@ class WCOrderModelTest {
             assertEquals(1, attributes.size)
 
             assertEquals("size", attributes[0].key)
-            assertEquals("Medium", attributes[0].value)
+            assertEquals("medium", attributes[0].value)
         }
 
         with(renderedLineItems[2]) {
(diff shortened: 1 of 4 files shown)


anitaa1990 (approved):
Looks good! Thanks for taking care of this one @AmandaRiu :shipit:

anitaa1990:
Should this target the release branch though since we're opening a new beta release with this fix? cc @loremattei this fix would require a new beta release for Woo 🙇‍♀️

loremattei:
@anitaa1990 please note that if we target the release branch, this change won't get any testing at all since the release is on Monday and it takes about 24h for a beta to be deployed by Google Play. 

Also, let's ping @AliSoftware as he's currently the release manager for Woo :-)

0nko (approved):
Nice work @AmandaRiu! :shipit:

PR #1769, opened 2020-11-12
Add XPostsStore for Retrieving a site's xpost candidates

mchowning:
Related [WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13184)

This adds the capability for retrieving and persisting a user/sites xposts. 

I decided to persist this data in two tables to avoid duplicating site information when the same site was an xpost candidate for multiple sites. One table simply persists the site information for a given xpost candidate (`XpostSite`) (name, URL, subdomain, blavatar, etc.). I could have used the `SiteModel` for this, but updating that object with information from another endpoint felt a bit risky with little (if any) benefit, so I kept them separate. I then added a second table that serves the purpose of connecting source site (`SiteModel`), to the relevant xpost target sites (`XPostSite`). That table makes it possible to look up all the `XPostSites` available for a given `SiteModel`.

### Testing

This can be tested via the related [WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/13184). In addition, it can be tested in the example app:

1. Log into a P2-themed site in the example app
2. Tap on "Sites"
3. Tap on "Fetch XPosts"
4. Observe that a successful response is displayed in the log at the bottom of the screen.


mchowning:
:wave: @planarvoid ! Could you take a look at the code in this PR to see if what I'm doing makes sense in the context of FluxC's architecture. In particular, I'm adding a new `Store` because xposts seemed somewhat distinct from the other functionality of the `SiteStore` (which seems to already be a pretty oversized `Store`), but I'd be curious to hear your thoughts.

Feel free to leave testing and approval of this PR to @cameronvoell or whichever dev reviews this along with the related WPAndroid, and Gutenberg PRs.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/XPostsRestClient.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.wordpress.android.fluxc.network.UserAgent
+import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import javax.inject.Inject
+import javax.inject.Named
+import javax.inject.Singleton
+
+@Singleton
+class XPostsRestClient @Inject constructor(
```
cameronvoell, line 18: It's not immediately obvious to me why we created a new class here instead of using  [ReactNativeWPComRestClient](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blame/develop/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/reactnative/ReactNativeWPComRestClient.kt) . Any thoughts on that?

mchowning: I think of the `ReactNativeStore` and the associated .com and .org rest clients as being singularly focused on handling dynamic requests. Really, they're a deviation from the pattern in FluxC of statically typing all the requests. To a certain extent giving those statically typed guarantees is the entire purpose of FluxC, and the `ReactNativeStore` abandons that purpose (out of necessity, because javascript). So I feel like it makes sense to keep them separate from all the other statically-typed calls that follow the more typical (and more Android friendly) FluxC patterns.

The XPosts Store/RestClient could have been included somewhere else (like the `SiteStore`). The main reason I didn't do that is because the `SiteStore` is already big (too big imho), and I didn't want to make it bigger.

As far as keeping the `XPostsStore`, but sharing an already existing Rest Client, it looked to me like most of the other stores had their own rest client, so I was mostly just following what I perceived to be the existing pattern.

With that said, I'm definitely open to hearing your thoughts on this and adopting a different approach.

cameronvoell: makes sense, I'm fine leaving it how you have it now. Thanks for the response 👍

cameronvoell (approved):
Looks good to me!

PR #1771, opened 2020-11-15
Adds a RequestQueue provider that performs retries on redirect

antonis:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2718

**Description**
This PR:
* Extends BasicNetwork by adding retries on temporary redirect (307)
* Adds an HTTP Client that does not follow redirects
* Adds a RequestQueue provider that performs retries on redirect

This is needed because the mshot backend service redirects to a loading gif image when the thumbnail is not ready. This occurs when the thumbnail has not been accessed recently for the specific language and viewport size. When this thumbnail is requested it is cached on the server and is readily available according to the server cache policy.

**To test**
This can be tested with https://github.com/wordpress-mobile/WordPress-Android/pull/13386


mkevins (approved):
Nice work Antonis. Tested via: https://github.com/wordpress-mobile/WordPress-Android/pull/13386, LGTM!

chipsnyder (approved):
Tested along with https://github.com/wordpress-mobile/WordPress-Android/pull/13386 Nice work :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/module/ReleaseNetworkModule.java
```diff
@@ -93,6 +103,14 @@ public RequestQueue provideRequestQueue(@Named("regular") OkHttpClient.Builder o
         return newRequestQueue(okHttpClientBuilder, appContext);
     }
 
+    @Singleton
+    @Named("noRedirects")
```
aforcier: I think we should try to follow convention here and call this `no-redirects` instead (like `custom-ssl`).

aforcier:
This looks okay to me and I verified that the connected tests all pass 👍 I just left one small consistency comment.

I think it might've made sense to write a test (live or mocked) for this but it's tricky since we're not actually using this new provider in FluxC. (I might argue that some of the mshots functionality could have been implemented to FluxC instead of WPAndroid, but it seems to rely on existing Glide setup which is already part of WPAndroid and may be messy to extract for not much benefit.)

Also the way this is written, we can't use retry-on-redirect functionality in a `custom-ssl` context (for example, to support self-hosted sites with self-signed SSL). But we haven't needed this before and mshots is a WordPress.com tool - we can refactor (or create a fourth request queue) if we ever need to use this in the future.

maxme:
> I think it might've made sense to write a test (live or mocked) for this but it's tricky since we're not actually using this new provider in FluxC

I agree and I think it's fine to not test the `no-redirects` queue.

I'd like to see a redirect test. A live one can be done with the right `.htaccess` and I think it would help understand how this is used, it would be good as documentation. And a good way to catch this particularity if we ever have to rewrite that layer. @antonis  I can help you setup the live server.

antonis:
@mkevins, @chipsnyder, @aforcier, @maxme 
Thank you all for reviewing this! I really appreciate your help and feedback.

I'll change the annotation to `no-redirects` and try to add tests to make this more consistent.

@maxme I'll appreciate any help with setting up  the live server for testing since I haven't really touched this part of our environment before.

fluxc/src/main/java/org/wordpress/android/fluxc/module/ReleaseNetworkModule.java
antonis: Fixed with fabfd98e9588ee82bb3d8f0d489c68f31d316c99

antonis:
Hello @aforcier, @maxme 👋 
I added a couple of unit tests with 74f889da694b32a2ed24365b3e3d6645f78b85f9
and a connected test with 72f0d7932be52f9a1eddf650aa326cd8156affeb

Regarding the later I used the endpoint just to validate that retries occur on redirects from the server (which [does not happen by default with Volley](https://github.com/google/volley/blob/b080a3488b0f82e0ef70191e8848c631376e80e0/src/main/java/com/android/volley/toolbox/BasicNetwork.java#L209)).

PR #1772, opened 2020-11-16
Issue/1767 download status

### LINKED ISSUE #1767: Add support for checking the status of a backup preparation download
The goal of this ticket is to add support for checking the status of a backup preparation download

GET https://public-api.wordpress.com/wpcom/v2/sites/182675569/rewind/downloads

The response will take the following form:

RUNNING
```
{
  "downloadId": 276158,
  "rewindId": "1603788401.423",
  "backupPoint": "2020-10-27T08:46:41+00:00",
  "startedAt": "2020-10-27T11:00:23+00:00",
  "progress": 49
}
```

FINISHED

```
{
  "downloadId": 276158,
  "rewindId": "1603788401.423",
  "backupPoint": "2020-10-27T08:46:41+00:00",
  "startedAt": "2020-10-27T11:00:23+00:00",
  "downloadCount": 0,
  "validUntil": "2020-10-28T11:00:25+00:00",
  "url": "https://public-api.wordpress.com/wpcom/v2/sites/182675569/rewind/downloads/276158/data?token=6VxkPOSkvwcLQfhPRMIh9TCNaJKygYueLjnzzS5wNl1R05dnc3jKEvmgAuCbmhLp"
}

```

zwarm:
Fixes #1767 

This PR adds support for checking download preparation status.
GET https://public-api.wordpress.com/wpcom/v2/sites/182675569/rewind/downloads

Note: I named everything "download", but after submitting this PR I decided to rename methods/objects to BackupDownload so it has more context. That will be in a separate PR following approval of this one.

This PR follows along with the same patterns used for checking rewind status.

P.S. I also adjusted minor Kotlin inspection finds such as redundant qualifiers.

**To Test**
The feature has not been implemented yet. CI tests passing are okay for this PR.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
@@ -20,7 +20,6 @@ import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.activity.RewindStatusModel
 import org.wordpress.android.fluxc.network.BaseRequest.BaseNetworkError
-import org.wordpress.android.fluxc.network.BaseRequest.GenericErrorType
 import org.wordpress.android.fluxc.network.BaseRequest.GenericErrorType.NETWORK_ERROR
 import org.wordpress.android.fluxc.network.UserAgent
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComGsonNetworkError
@@ -198,7 +197,7 @@ class ActivityLogRestClientTest {
 
     @Test
     fun fetchActivity_dispatchesErrorOnFailure() = test {
-        initFetchActivity(error = WPComGsonNetworkError(BaseNetworkError(GenericErrorType.NETWORK_ERROR)))
+        initFetchActivity(error = WPComGsonNetworkError(BaseNetworkError(NETWORK_ERROR)))
 
         val payload = activityRestClient.fetchActivity(requestPayload, number, offset)
 
@@ -407,7 +406,10 @@ class ActivityLogRestClientTest {
         assertEmittedDownloadStatusError(payload, DownloadStatusErrorType.GENERIC_ERROR)
     }
 
-    private fun assertEmittedDownloadStatusError(payload: FetchedDownloadStatePayload, errorType: DownloadStatusErrorType) {
+    private fun assertEmittedDownloadStatusError(
+        payload: FetchedDownloadStatePayload,
+        errorType: DownloadStatusErrorType
+    ) {
         with(payload) {
             assertEquals(this.site, site)
             assertTrue(this.isError)
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
@@ -30,13 +30,16 @@ import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Re
 import org.wordpress.android.fluxc.network.rest.wpcom.activity.ActivityLogRestClient.ActivitiesResponse
 import org.wordpress.android.fluxc.network.rest.wpcom.activity.ActivityLogRestClient.ActivitiesResponse.Page
 import org.wordpress.android.fluxc.network.rest.wpcom.activity.ActivityLogRestClient.DownloadResponse
+import org.wordpress.android.fluxc.network.rest.wpcom.activity.ActivityLogRestClient.DownloadStatusResponse
 import org.wordpress.android.fluxc.network.rest.wpcom.activity.ActivityLogRestClient.RewindResponse
 import org.wordpress.android.fluxc.network.rest.wpcom.activity.ActivityLogRestClient.RewindStatusResponse
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
 import org.wordpress.android.fluxc.store.ActivityLogStore.ActivityLogErrorType
 import org.wordpress.android.fluxc.store.ActivityLogStore.DownloadRequestTypes
+import org.wordpress.android.fluxc.store.ActivityLogStore.DownloadStatusErrorType
 import org.wordpress.android.fluxc.store.ActivityLogStore.FetchActivityLogPayload
 import org.wordpress.android.fluxc.store.ActivityLogStore.FetchedActivityLogPayload
+import org.wordpress.android.fluxc.store.ActivityLogStore.FetchedDownloadStatePayload
 import org.wordpress.android.fluxc.store.ActivityLogStore.FetchedRewindStatePayload
 import org.wordpress.android.fluxc.store.ActivityLogStore.RewindRequestTypes
 import org.wordpress.android.fluxc.store.ActivityLogStore.RewindStatusErrorType
@@ -395,6 +398,48 @@ class ActivityLogRestClientTest {
         assertTrue(payload.isError)
     }
 
+    @Test
+    fun fetchActivityDownload_dispatchesGenericErrorOnFailure() = test {
+        initFetchDownloadStatus(error = WPComGsonNetworkError(BaseNetworkError(NETWORK_ERROR)))
+
+        val payload = activityRestClient.fetchActivityDownload(site)
+
+        assertEmittedDownloadStatusError(payload, DownloadStatusErrorType.GENERIC_ERROR)
+    }
+
+    private fun assertEmittedDownloadStatusError(payload: FetchedDownloadStatePayload, errorType: DownloadStatusErrorType) {
+        with(payload) {
+            assertEquals(this.site, site)
+            assertTrue(this.isError)
+            assertEquals(errorType, this.error.type)
+        }
+    }
+
+    @Test
+    fun fetchActivityDownload_dispatchesResponseOnSuccess() = test {
+        val progress = 55
+        val downloadResponse = DOWNLOAD_STATUS_RESPONSE.copy(progress = progress)
+        initFetchDownloadStatus(downloadResponse)
+
+        val payload = activityRestClient.fetchActivityDownload(site)
+
+        with(payload) {
+            assertEquals(this.site, site)
+            assertNull(this.error)
+            assertNotNull(this.downloadStatusModelResponse)
+            this.downloadStatusModelResponse?.apply {
+                assertEquals(this.downloadId, DOWNLOAD_STATUS_RESPONSE.downloadId)
+                assertEquals(this.rewindId, DOWNLOAD_STATUS_RESPONSE.rewindId)
+                assertEquals(this.backupPoint, DOWNLOAD_STATUS_RESPONSE.backupPoint)
+                assertEquals(this.startedAt, DOWNLOAD_STATUS_RESPONSE.startedAt)
+                assertEquals(this.downloadCount, DOWNLOAD_STATUS_RESPONSE.downloadCount)
+                assertEquals(this.validUntil, DOWNLOAD_STATUS_RESPONSE.validUntil)
+                assertEquals(this.url, DOWNLOAD_STATUS_RESPONSE.url)
+                assertEquals(this.progress, progress)
+            }
+        }
+    }
+
     private suspend fun initFetchActivity(
         data: ActivitiesResponse = mock(),
         error: WPComGsonNetworkError? = null
@@ -482,6 +527,23 @@ class ActivityLogRestClientTest {
                         "types" to requestTypes)),
                 eq(DownloadResponse::class.java)
         )).thenReturn(response)
+        whenever(site.siteId).thenReturn(siteId)
+        return response
+    }
+
+    private suspend fun initFetchDownloadStatus(
+        data: DownloadStatusResponse = mock(),
+        error: WPComGsonNetworkError? = null
+    ): Response<DownloadStatusResponse> {
+        val response = if (error != null) Response.Error<DownloadStatusResponse>(error) else Success(data)
+        whenever(wpComGsonRequestBuilder.syncGetRequest(
+                eq(activityRestClient),
+                urlCaptor.capture(),
+                paramsCaptor.capture(),
+                eq(DownloadStatusResponse::class.java),
+                eq(false),
+                any(),
+                eq(false))).thenReturn(response)
         whenever(site.siteId).thenReturn(siteId)
         return response
     }
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
@@ -411,9 +411,9 @@ class ActivityLogRestClientTest {
         errorType: DownloadStatusErrorType
     ) {
         with(payload) {
-            assertEquals(this.site, site)
-            assertTrue(this.isError)
-            assertEquals(errorType, this.error.type)
+            assertEquals(site, site)
+            assertTrue(isError)
+            assertEquals(errorType, error.type)
         }
     }
 
@@ -426,8 +426,8 @@ class ActivityLogRestClientTest {
         val payload = activityRestClient.fetchActivityDownload(site)
 
         with(payload) {
-            assertEquals(this.site, site)
-            assertNull(this.error)
+            assertEquals(site, site)
+            assertNull(error)
             assertNotNull(this.downloadStatusModelResponse)
             this.downloadStatusModelResponse?.apply {
                 assertEquals(this.downloadId, DOWNLOAD_STATUS_RESPONSE.downloadId)
--- a/example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
@@ -19,6 +19,7 @@ import org.wordpress.android.fluxc.annotations.action.Action
 import org.wordpress.android.fluxc.generated.ActivityLogActionBuilder
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.activity.ActivityLogModel
+import org.wordpress.android.fluxc.model.activity.DownloadStatusModel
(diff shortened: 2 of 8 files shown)


example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
```diff
@@ -395,6 +397,51 @@ class ActivityLogRestClientTest {
...
+    @Test
+    fun fetchActivityDownload_dispatchesGenericErrorOnFailure() = test {
+        initFetchDownloadStatus(error = WPComGsonNetworkError(BaseNetworkError(NETWORK_ERROR)))
+
+        val payload = activityRestClient.fetchActivityDownload(site)
+
+        assertEmittedDownloadStatusError(payload, DownloadStatusErrorType.GENERIC_ERROR)
+    }
+
+    private fun assertEmittedDownloadStatusError(
```
ParaskP7, line 409: Question: Why did this code extracted to a private function? Will it be reused?

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
```diff
@@ -395,6 +397,51 @@ class ActivityLogRestClientTest {
...
+        payload: FetchedDownloadStatePayload,
+        errorType: DownloadStatusErrorType
+    ) {
+        with(payload) {
+            assertEquals(this.site, site)
+            assertTrue(this.isError)
+            assertEquals(errorType, this.error.type)
```
ParaskP7: Minor: The `this` keyword is not needed here, is it?

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
```diff
@@ -395,6 +397,51 @@ class ActivityLogRestClientTest {
...
+        }
+    }
+
+    @Test
+    fun fetchActivityDownload_dispatchesResponseOnSuccess() = test {
```
ParaskP7, line 421: Minor: I have a few suggestion for improving this test, instead writing all those down for you, how about if I provide the code snippet?
```
    @Test
    fun fetchActivityDownload_dispatchesResponseOnSuccess() = test {
        val downloadResponse = DOWNLOAD_STATUS_RESPONSE.copy(progress = 55)
        initFetchDownloadStatus(downloadResponse)

        val payload = activityRestClient.fetchActivityDownload(site)

        with(payload) {
            assertEquals(site, site)
            assertNull(error)
            assertNotNull(downloadStatusModelResponse)
            downloadStatusModelResponse?.apply {
                assertEquals(downloadId, downloadResponse.downloadId)
                assertEquals(rewindId, downloadResponse.rewindId)
                assertEquals(backupPoint, downloadResponse.backupPoint)
                assertEquals(startedAt, downloadResponse.startedAt)
                assertEquals(downloadCount, downloadResponse.downloadCount)
                assertEquals(validUntil, downloadResponse.validUntil)
                assertEquals(url, downloadResponse.url)
                assertEquals(progress, downloadResponse.progress)
            }
        }
    }
```

Wdyt Annmarie?

Question: For future reference, would you like me to continue doing so or would you instead prefer it written no matter how long?

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
```diff
@@ -365,6 +368,47 @@ class ActivityLogStoreTest {
...
+                )
+        )
+        val action = ActivityLogActionBuilder.newFetchDownloadStateAction(payload)
+        activityLogStore.onAction(action)
+
+        verify(activityLogRestClient).fetchActivityDownload(siteModel)
+    }
+
+    @Test
+    fun storeFetchedDownloadStatusToDb() = test {
```
ParaskP7, line 387: Question: How about testing the error path as well?

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
```diff
@@ -365,6 +368,47 @@ class ActivityLogStoreTest {
...
+        val downloadStatusModel = mock<DownloadStatusModel>()
+        val payload = FetchedDownloadStatePayload(downloadStatusModel, siteModel)
+        whenever(activityLogRestClient.fetchActivityDownload(siteModel)).thenReturn(payload)
+
+        val fetchAction = ActivityLogActionBuilder.newFetchDownloadStateAction(FetchDownloadStatePayload(siteModel))
+        activityLogStore.onAction(fetchAction)
+
+        verify(activityLogSqlUtils).replaceDownloadStatus(siteModel, downloadStatusModel)
+        val expectedChangeEvent = ActivityLogStore.OnDownloadStatusFetched(ActivityLogAction.FETCH_DOWNLOAD_STATE)
+        verify(dispatcher).emitChange(eq(expectedChangeEvent))
```
ParaskP7, line 397: Question: Can this be considered as part of this db test? It seems that it can be a separate test altogether?

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
```diff
@@ -365,6 +368,47 @@ class ActivityLogStoreTest {
...
+    @Test
+    fun returnDownloadStatusFromDb() {
+        val downloadStatusModel = mock<DownloadStatusModel>()
+        whenever(activityLogSqlUtils.getDownloadStatusForSite(siteModel))
+                .thenReturn(downloadStatusModel)
+
+        val downloadStatusFromDb = activityLogStore.getDownloadStatusForSite(siteModel)
+
+        verify(activityLogSqlUtils).getDownloadStatusForSite(siteModel)
+        assertEquals(downloadStatusModel, downloadStatusFromDb)
```
ParaskP7: Question: Does this assertion provide actual value for this test? How does it help?

fluxc/src/main/java/org/wordpress/android/fluxc/model/activity/DownloadStatusModel.kt
```diff
@@ -0,0 +1,14 @@
+package org.wordpress.android.fluxc.model.activity
+
+import java.util.Date
+
+data class DownloadStatusModel(
```
ParaskP7, line 5: Idea: I did leave the very same comment on @ashiagr's PR about domain modelling. I am just going to comment it again so we can spin up a conversation on that again (if that's ok).

My suggestion is that this model can be split into two separate models, using a sealed class. Something like a `DownloadStatusRunningModel` and a `DownloadStatusFinishedModel` model. This way there would be no nullable values per model and the usage of those later on would be much more straightforward. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ActivityLogSqlUtils.kt
```diff
@@ -170,6 +196,21 @@ class ActivityLogSqlUtils
...
+                localSiteId = site.id,
+                remoteSiteId = site.siteId,
+                downloadId = this.downloadId,
+                rewindId = this.rewindId,
+                backupPoint = this.backupPoint.time,
+                startedAt = this.startedAt.time,
+                progress = this.progress,
+                downloadCount = this.downloadCount,
+                validUntil = this.validUntil?.time,
+                url = this.url
```
ParaskP7, line 210: Minor: The `this` keyword is not needed here, is it?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ActivityLogSqlUtils.kt
```diff
@@ -289,4 +330,43 @@ class ActivityLogSqlUtils
...
+                    downloadId,
+                    rewindId,
+                    Date(backupPoint),
+                    Date(startedAt),
+                    progress,
+                    downloadCount,
+                    validUntil?.let {
+                        Date(it)
+                    },
+                    url
```
ParaskP7, line 368: Idea: Maybe the usage of named arguments could help here to verify correctness without the need to jump to and back into their declaration?

ParaskP7 (commented):
Thanks again for including me in this PR Annmarie! 🌟 

As always, I have only questions, minors and ideas as comments, still learning in progress, please bare with me! 📚 

Also, I have a couple of extra generic things that I want to ask:
1) The `ActivityFixture` file can be renamed file to include the `Log` word in it. Actually a lot of names are missing the `Log` word in them. Maybe someone can spend some time and create a separate PR to improve on that. To me, having it as plain `Activity` can be confusing. What is even worse, having some of the class/file names as `Activity`, while others as `ActivityLog` can create even more confusion in the future. Also, the package name might need to change to include 'log' as well. Wdyt?
1) I always suggest and trying hard myself to separate the refactoring from the feature commits in order to make my coding and later on the code review more targeted and easy. How do you all handling that, do you prefer one over the other?

zwarm:
> 1. The `ActivityFixture` file can be renamed file to include the `Log` word in it. Actually a lot of names are missing the `Log` word in them. Maybe someone can spend some time and create a separate PR to improve on that. To me, having it as plain `Activity` can be confusing. What is even worse, having some of the class/file names as `Activity`, while others as `ActivityLog` can create even more confusion in the future. Also, the package name might need to change to include 'log' as well. Wdyt?

This is a good suggestion, please go ahead and create a low priority issue for this. 

> 2. I always suggest and trying hard myself to separate the refactoring from the feature commits in order to make my coding and later on the code review more targeted and easy. How do you all handling that, do you prefer one over the other?

I don't have a preference, so whatever works for you.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
zwarm: It very well may be. This is the generic error and until I get to the UI implementation, I don't know if there will be additional error situations I will need to handle.

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
zwarm: Following existing patterns for rewind

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
zwarm: > Question: For future reference, would you like me to continue doing so or would you instead prefer it written no matter how long?

Thanks for the input, I appreciate the time you took to write this out, but pease don't spend your time rewriting unless there is something horribly wrong. Perhaps we can start sharing these types of things in a monthly learning meetup - kind of like a "writing test best practices" series. Wdyt?

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
zwarm: > Question: How about testing the error path as well? 

All the error paths are unknown at the moment. This is on my future to-do list. thanks

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
zwarm: Removed in latest commit

fluxc/src/main/java/org/wordpress/android/fluxc/model/activity/DownloadStatusModel.kt
zwarm: > My suggestion is that this model can be split into two separate models, using a sealed class

I really like this idea. Using sealed classes with a stricter structure and not mucking around with nulls is a + in my book.  My only hesitation is that this functionality mimics the existing rewind functionality and I wouldn't want to change one without the other. I want to check with @malinajirka / @ashiagr and see if they are planning to switch out to existing functionality, as I want to stay consistent.

Thanks 👍

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
ParaskP7: Great, yes I agree! Although, I usually will spend that time anyway, it is just in this case I was trying to understand this test and some of the inputs/outputs. I find it helps me playing around so please don't worry about me. Also, I am doing my code review from the IDE, so changing and testing code is a matter of seconds for me, at least in this case.

About the monthly learning meetup and various series such as this, I would totally ❤️  if we can arrange something like that! @malinajirka, @ashiagr is this possible, would you like that as well? Or @zwarm did you mean more of a whole `Android` thingy meetup?

fluxc/src/main/java/org/wordpress/android/fluxc/model/activity/DownloadStatusModel.kt
ParaskP7: Thanks for the detailed explanation Annmarie! Yeap, please talk it through with the rest and decide on the way you want to go forward, it is just an idea, and consistency is the key!

ParaskP7:
> This is a good suggestion, please go ahead and create a low priority issue for this.

Perfect, will do that then, thanks Annmarie! When I do create the issue (probably tomorrow) I'll link it in this PR for our reference.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
malinajirka: We had a couple of learning meetups. There was even one about unit test best practices. Unfortunately, it was in-person learnup so there is no recording 😢. I'd love to have more of those!

fluxc/src/main/java/org/wordpress/android/fluxc/model/activity/DownloadStatusModel.kt
malinajirka: As I mentioned on [the "original" convo](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1763#discussion_r522036285) I really like the sealed classes approach. I also commented there that I wouldn't change `ScanStateModel` without changing `RewindStatusModel`.

However, in this particular case I don't feel strongly about the consistency. "DownloadStatusModel" is not similar to "RewindStatusModel" anyway. The Rewind model contains "status" field so it's at least readable what status it is in. However, the download model doesn't contain any such info so it's very difficult to see that it can actually be in two different states.

So to summarize, I personally wouldn't mind even if we refactored just this class. Having said that, feel free to keep it as is. Whatever you think is the best, tbh I'm not sure.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
malinajirka, line 409: Nitpick: We might want to consider moving all the private methods to the bottom of the class. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
```diff
@@ -112,6 +123,13 @@ class ActivityLogStore
         return emitDownloadResult(payload, DOWNLOAD)
     }
 
+    suspend fun fetchActivitiesDownload(
```
malinajirka, line 126: Note: I find this name a bit confusing, I'm not sure I understand why the method name includes "activities". I'd have called it something like "fetchDownloadState". But I understand you kept consistency with "fetchActivitiesRewind". Am I missing something or do you find the name confusing too? (no need to change anything, I was just curious if I'm missing something)

malinajirka (approved):
Great job @zwarm! The code LGTM. I've left just one super minor comment so I'm merging this PR.

malinajirka:
> The ActivityFixture file can be renamed file to include the Log word in it. Actually a lot of names are missing the Log word in them. Maybe someone can spend some time and create a separate PR to improve on that. To me, having it as plain Activity can be confusing. What is even worse, having some of the class/file names as Activity, while others as ActivityLog can create even more confusion in the future. Also, the package name might need to change to include 'log' as well. Wdyt?

@ParaskP7 I agree that the current state can be confusing sometimes. Here's my understanding of the current state. `ActivityLog` refers to the ActivityLog feature (list of events on a site), whereas `Activity` refers to a single item (event). I agree that some classes/files/fields are not following this "rule", so these should be definitely renamed. I'm not sure about the ones which are intentionally called "Activity(ies)", perhaps we could come up with a better name for them (eg. ActivityLogEvent/ActivityLogItem/...). Let's continue with this discussion on the ticket.

ParaskP7:
> @ParaskP7 I agree that the current state can be confusing sometimes. Here's my understanding of the current state. ActivityLog refers to the ActivityLog feature (list of events on a site), whereas Activity refers to a single item (event). I agree that some classes/files/fields are not following this "rule", so these should be definitely renamed. I'm not sure about the ones which are intentionally called "Activity(ies)", perhaps we could come up with a better name for them (eg. ActivityLogEvent/ActivityLogItem/...). Let's continue with this discussion on the ticket.

Thanks for your input @malinajirka! Didn't think about the `ActivityLog` plural and `Activity` as single, it makes sense, yes. Indeed, we should make sure that everything is following this "rule" wherever possible. Still something troubles in me within when reading plain `Activity`, probably I am too used to this keyword being referred to the view layer I guess. Yes, let continue this discussion on the ticket. 🙏

fluxc/src/main/java/org/wordpress/android/fluxc/store/ActivityLogStore.kt
zwarm: Oh yes, I hate the name too and was just staying consistent. I am going to add this change to the rename PR (if it hasn't been closed yet). 🙇

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
zwarm: I'll add this into my rename PR too! Thanks

ParaskP7:
> Perfect, will do that then, thanks Annmarie! When I do create the issue (probably tomorrow) I'll link it in this PR for our reference.

As suggested, I created this [issue](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1775) for our future reference.

PR #1774, opened 2020-11-18
Persist the fetched products

0nko:
This PR adds product persistence to the `fetchProductListSynced` call to make it consistent with the rest of fetch requests, since there is currently no other way to save products in the DB from the store as a separate action.


ThomazFB (approved):
Hey @0nko, thanks for covering this point, I didn't notice this inconsistency with the WCProductStore behavior. I did some tests to see if it would generate any state problem with the `WCProductLeaderboardsMapper` methods, but it's working fine and it also reduces the number of requests over time when we call the `asProductList`, nice job! LGTM :shipit:

PR #1777, opened 2020-11-20
Issue/1776 wc cached order count

### LINKED ISSUE #1776: Modify hasCachedOrdersForSite() to use a count
Right now `WCOrderStore.hasCachedOrdersForSite()` loads the entire list of a site's orders to determine whether there are any orders, which is causing `SQLiteBlobTooBigException` crashes in [WCAndroid](https://github.com/woocommerce/woocommerce-android/issues/572). We can address this by simply using a count, or loading a single order.

nbradbury:
Fixes #1776 - this PR changes `hasCachedOrdersForSite()` to use a simple count query, which is much more efficient than loading all the orders for a site. I also updated `fetchOrders` to use the same count to determine the offset when loading more orders.

This is being submitted because Sentry shows a large number of `SQLiteBlobTooBigExceptions` coming from `hasCachedOrdersForSite()`, and I expect that's due to us loading all the orders just to check if a site has orders. I don't expect this to fix every one of those exceptions but this should resolve quite a few of them.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderSqlUtilsTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderSqlUtilsTest.kt
@@ -98,6 +98,21 @@ class OrderSqlUtilsTest {
     }
 
     @Test
+    fun testGetOrderCountForSite() {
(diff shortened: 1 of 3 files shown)


0nko (approved):
:shipit:

PR #1779, opened 2020-11-23
ScanStore: ScanRestClient unit tests

ashiagr:

This PR adds tests for `ScanRestClient`.

This PR also makes following minor fixes:
- Adds few missing fields to `Credentials` - 946eddb
- Updates nullability of `ThreatModel` fields - cb8c83c

**To test**

That tests run fine and cover important flows/ results for success, errors and different scan states.

**Note**
1. `ThreatModel` fields are not yet finalised involving threat status, types, context and few more. I'll update them in a separate PR along with tests (most probably with threat details API integration).
2. I wasn't able to create a 'PROVISIONING' scan state. Let me know if anyone knows how to test it.


example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
```diff
@@ -0,0 +1,324 @@
...
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Success
+import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import org.wordpress.android.fluxc.network.rest.wpcom.scan.ScanStateResponse.Threat
+import org.wordpress.android.fluxc.store.ScanStore.FetchedScanStatePayload
+import org.wordpress.android.fluxc.store.ScanStore.ScanStateErrorType
+import org.wordpress.android.fluxc.test
+
+@RunWith(MockitoJUnitRunner::class)
+class ScanRestClientTest {
```
ParaskP7, line 40: Minor (🔍 ): I understand that we prefer to write tests using the `'test something'` convention. Also, I saw you used that on `ScanStoreTest` as well. What if we convert those test to this convention as well, instead of the camelcase underscore convention. Also, and we have a brief chat with @malinajirka about that, personally for new tests I will be using the `'given xyz, when xyz, then xyz`. I find using at least the commas help a lot with the test readability.
Minor (🔍 ): I am seeing a lot of test being duplicated, with only one assertion being the diff. I am 💯 in favour of single assertion tests and that is what I also do and live by. However, in this case since the `payload` is being tested, and this can be considered as one assertion, no matter if the internals are being asserted separately, I believe merging those tests would help. Wdyt?
Minor (🔍 ): This test suite can be reformatted. PS: Can you and me @ashiagr check out indent setting cause when I run reformat I get the indentation corrected  on my side. Do you have a different setup? PSS: The same applies to the `ScanStoreTest` suite.
Minor (🔍 ): Probably all of the `this.` references can be removed from this test suite. Can you double check this? Actually you removed most of them like 17' mins ago, YOU ARE GOOD, but some of them as still there. Not sure if you want them there on purpose.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
```diff
@@ -0,0 +1,324 @@
...
+            accessToken,
+            userAgent
+        )
+    }
+
+    @Test
+    fun fetchScanState_buildsCorrectRequestUrl() = test {
+        val successResponseJson = UnitTestUtils.getStringFromResourceFile(
+            this.javaClass,
+            "wp/jetpack/scan/jp-daily-scan-with-threat.json"
```
ParaskP7: Minor (🔍 ): I suggest all the `.json` file paths be extracted in constant and reused when necessary. Wdyt?

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
```diff
@@ -0,0 +1,324 @@
...
+        val responseType = object : TypeToken<ScanStateResponse>() {}.type
+        return Gson().fromJson(json, responseType) as? ScanStateResponse
+    }
+
+    private suspend fun initFetchScanState(
+        data: ScanStateResponse? = null,
+        error: WPComGsonNetworkError? = null
+    ): Response<ScanStateResponse> {
+        val nonNullData = data ?: mock()
+        val response = if (error != null) Response.Error<ScanStateResponse>(error) else Success(nonNullData)
```
ParaskP7: Minor (🔍 ): The `ScanStateResponse` explicit type argument can be safely removed.

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
```diff
@@ -0,0 +1,69 @@
...
+            scanSqlUtils,
+            initCoroutineEngine(),
+            dispatcher
+        )
+    }
+
+    @Test
+    fun `fetch scan state triggers rest client`() = test {
+        val payload = FetchScanStatePayload(siteModel)
+        whenever(scanRestClient.fetchScanState(siteModel)).thenReturn(
```
ParaskP7: Minor (🔍 ): It seems that this can fit in one line.

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
```diff
@@ -0,0 +1,69 @@
...
+    fun `error on fetch scan state returns the error`() = test {
+        val error = ScanStateError(ScanStateErrorType.INVALID_RESPONSE, "error")
+        val payload = FetchedScanStatePayload(error, siteModel)
+        whenever(scanRestClient.fetchScanState(siteModel)).thenReturn(payload)
+
+        val fetchAction = ScanActionBuilder.newFetchScanStateAction(FetchScanStatePayload(siteModel))
+        scanStore.onAction(fetchAction)
+
+        val expectedEventWithError = ScanStore.OnScanStateFetched(payload.error, ScanAction.FETCH_SCAN_STATE)
+        verify(dispatcher).emitChange(eq(expectedEventWithError))
```
ParaskP7: Minor (🔍 ): It seems that the `eq(...)` usage is not required for this verification.

ParaskP7 (commented):
Once more, thanks for including me in this PR @ashiagr ! 🌟 

Such an amazing job with the tests, I love them and do believe that in the long run they will provide us with great value! ❤️ 

You will find only minor (🔍 ) comments from my side, nothing else.

PS: On the json naming convention, I have the below questions (❓ ) and ideas (💡 ). Please feel free to ignore all them if you find me too much (PS: I do find myself a bit too much all the time! 😄 ):
- What if ever response file is named with the `xyz-scan-<state>` convention for consistency purposes? For example, I would have named the `jp-complete-scanning` as `jp-complete-scan-scanning`. Wdyt? Also, for some of those the `<state>` keyword is missing, like the `jp-daily-scan-with-threat`. I would have added that there as well.
- Also, maybe `scan` can go in front of `daily/complete` since this is the parent in the hierarchy? For example, I would have named the `jp-complete-scan-idel` as `jp-scan-complete-idle` and then for the daily I would have something like `jp-scan-daily-idle`. This way, it can be considered as fake folder like hierarchy, which when those json file grow to larger number help.
- On one of the response you added the `backup` keyword, why is that?

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
ashiagr: In 80e1351

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
ashiagr: In 4b5ef66

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ashiagr: It gives me type expected error if I remove it. Can you check in your IDE and confirm?
Screenshot 2020-11-24 at 4 03 34 PM

ParaskP7: The `<>` needs to be removed as well, probably this is why it complains. Can you try that Ashita?

ParaskP7: In my case, the IDE actually suggests that and by running a quick action on it, it removes it itself, what about your IDE? Isn't it shown as grey as well?

image

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ashiagr: > I understand that we prefer to write tests using the 'test something' convention. Also, I saw you used that on ScanStoreTest as well. What if we convert those test to this convention as well, instead of the camelcase underscore convention. Also, and we have a brief chat with @malinajirka about that, personally for new tests I will be using the 'given xyz, when xyz, then xyz. I find using at least the commas help a lot with the test readability.

Yeah we should discuss about the naming convention. Right now different naming conventions are used in the code base and it is difficult to prefer one over another. 

> I am seeing a lot of test being duplicated, with only one assertion being the diff. I am 💯 in favour of single assertion tests and that is what I also do and live by. However, in this case since the payload is being tested, and this can be considered as one assertion, no matter if the internals are being asserted separately, I believe merging those tests would help. Wdyt?

We might get different responses for a state for example it may or may not return credentials based on whether they're set on the server even though both sites might be in IDLE  scan state. Based on different responses, tests are split into different cases. I'll try to merge some of them though.

> This test suite can be reformatted. PS: Can you and me @ashiagr check out indent setting cause when I run reformat I get the indentation corrected on my side. Do you have a different setup? PSS: The same applies to the ScanStoreTest suite.

Actually `ktlint` allows both indent spaces 4 & 8. There has been long discussions around it. E.g. [this](https://shorturl.at/acjo5) and [this](https://shorturl.at/GHNP9) and seems like 4 is more preferable. 

Quoting from the second link:
> 4 is reasonable, especially if ktx is using 4. 4 it is!

But I don't have a strong opinion. I can change it to 8 indent spaces. We can surely compare our setup 😄.

> Probably all of the this. references can be removed from this test suite. Can you double check this? Actually you removed most of them like 17' mins ago, YOU ARE GOOD, but some of them as still there. Not sure if you want them there on purpose.

Yeah I removed them a while ago. I'll remove the rest 👍.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ashiagr: > The <> needs to be removed as well, probably this is why it complains. 

I already tried that. It still complains 🤷‍♀️ with type inference failed error.

Screenshot 2020-11-24 at 4 40 50 PM

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ParaskP7: Thank for all the responses Ashita and I agree with all!

PS: About the indentation/reformatting issue, my setup was actually wrong, after explicitly applying `Set from... -> Kotlin style guide` on my IDE this got fixed. I wonder whether we should explicitly say that as an instruction for new team members and contributors on our `README.md`. This way we will not forget to have this setup.
PSS: The official `Kotlin Style Guide` and also `Detekt` are defaulting to 4, so I am sure this is the way to go.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ashiagr: There was a `kotlin` older version issue, got sorted by comparing our setup, thanks @ParaskP7 🙏.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ashiagr: In 05b382a

ashiagr:
Thanks so much for the review, @ParaskP7 !

I've addressed most of your comments. Some more inline:

> What if ever response file is named with the xyz-scan-<state> convention for consistency purposes? For example, I would have named the jp-complete-scanning as jp-complete-scan-scanning. Wdyt? Also, for some of those the <state> keyword is missing, like the jp-daily-scan-with-threat. I would have added that there as well.

Scan state added to file name in a3f755b

> Also, maybe scan can go in front of daily/complete since this is the parent in the hierarchy? For example, I would have named the jp-complete-scan-idel as jp-scan-complete-idle and then for the daily I would have something like jp-scan-daily-idle. This way, it can be considered as fake folder like hierarchy, which when those json file grow to larger number help.

First part of the filename includes the plan name, so ignoring this comment.

> On one of the response you added the backup keyword, why is that?

First part of the filename includes the plan name. I got the unavailable state response from a daily backup plan and so it includes the "backup" keyword.

Ready for another round 🙇‍♀️.

malinajirka (approved):
Amazing job @ashiagr 👏 

The code looks great and I love the tests 🤩 .

>Updates nullability of ThreatModel fields

I'm still a bit worried about keeping some of the fields as not-nullable in the response. If we had unit tests on the API which would cover these scenarios I'd feel ok. However, afaik we don't have such unit tests. Moreover, if someone decides to stop sending a field it won't be considered as a breaking change (the definition of the endpoint won't change). However, if it happens our app (at least old versions) will start crashing on NPE. I think it'd be at least worth a discussion with someone from the API, wdyt?

ashiagr:
Thanks for reviewing it @malinajirka!

> I'm still a bit worried about keeping some of the fields as not-nullable in the response. If we had unit tests on the API which would cover these scenarios I'd feel ok. However, afaik we don't have such unit tests. 

It's a valid concern and I fully agree. 

Actually fields that are marked not null for `ScanStateResponse.Threat` right now are:

`id` ,`signature` ,`first_detected ` -  [API code ](https://opengrok.a8c.com/source/xref/trunk/wp-content/lib/vaultpress/threat-types.php#32) throws an error and it seems they should always be set.
```
if ( ! isset( $threat->id ) || ! is_numeric( $threat->id ) ) {
   throw new \InvalidArgumentException( 'missing or invalid threat ID' );
}
if ( ! isset( $threat->signature ) || ! $threat->signature ) {
   throw new \InvalidArgumentException( 'missing threat signature name' );
}
if ( ! isset( $threat->firstDetected ) || ! $threat->firstDetected ) {
    throw new \InvalidArgumentException( 'missing threat firstDetected timestamp' );
}
```

`status` - is set to [valid status or blank](https://opengrok.a8c.com/source/xref/trunk/wp-content/lib/vaultpress/threat-types.php#51).

`description` - is set to [valid description or blank](https://opengrok.a8c.com/source/xref/trunk/wp-content/lib/vaultpress/threat-types.php#45).

That said, I haven't done full justice to `ThreatModel`. Several fields need to be finalised and I plan to add tests for it later as mentioned in the PR description (notes).

malinajirka:
Thanks for the reply with all the useful links!

I agree that that the current code of the API and the app are compatible. Hopefully, no-one will change the API code in the future 🤞.

PR #1780, opened 2020-11-23
ScanStore: Store scan state into DB

ashiagr:

This PR stores scan state into DB.
Currently it does not save credentials and threats and focuses mainly on storing scan progress status.

**To test**
    That scan progress status is correctly saved into the DB.

**Notes**

1. Note that out of 4 different scan states (`IDLE`, `SCANNING`, `PROVISIONING`, `UNAVAILABLE`) only `IDLE` and `SCANNING` states return the progress status. 

    `IDLE` state: returns `most_recent` status (`current` is `null`)
    `SCANNING` state: returns `current` status (`most_recent` is `null`)

Since either `most_recent` or `current` is valid for these states, fields like `progress`, `isInitial`, `startDate` are not created separately for `most_recent` and `current`.

2. We might want to reuse `RewindStatusCredentials` for storing `Credentials` for scan functionality as both refer to the same set of credentials. Currently `RewindStatusCredentials` is not used anywhere in _WordPress-Android_, will consider storing credentials returned in scan state response later when we really need it.


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt
```diff
@@ -1,7 +1,155 @@
...
+        return WellSql.select(ScanStateBuilder::class.java)
+            .where()
+            .equals(ScanStateTable.LOCAL_SITE_ID, site.id)
+            .endWhere()
+            .asModel
+            .firstOrNull()
+    }
+
+    private fun ScanStateModel.toBuilder(site: SiteModel): ScanStateBuilder {
+        val startDate = when (state) {
```
malinajirka, line 49: Nitpicks:
1. I think we can drop the brackets to make it a bit more concise.
2. I personally always try to explicitly list all the enum/sealedclass values and avoid using "else/default". When someone introduces a new state it forces them to go over all the switch/when statements and double check if all of them behave as expected. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt
```diff
@@ -1,7 +1,155 @@
...
+    private fun getScanStateBuilder(site: SiteModel): ScanStateBuilder? {
+        return WellSql.select(ScanStateBuilder::class.java)
+            .where()
+            .equals(ScanStateTable.LOCAL_SITE_ID, site.id)
+            .endWhere()
+            .asModel
+            .firstOrNull()
+    }
+
+    private fun ScanStateModel.toBuilder(site: SiteModel): ScanStateBuilder {
```
malinajirka, line 48: I'm wondering if we should write unit tests for the "builder" methods, wdyt?

malinajirka (approved):
Thanks @ashiagr! I've left two minor comments. It LGTM - feel free to merge it.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt
ashiagr: In 40cd5b4

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt
ashiagr: They're covered by connected tests. As discussed, I'll get back to adding unit test too later on after covering other higher priority tasks.

malinajirka:
Thanks for the changes, LGTM! ;)

PR #1784, opened 2020-11-25
Changed download limit from Int to Long

anitaa1990:
This PR partially addresses woocommerce/woocommerce-android#3057 by modifying the `download_limit` field from Int to Long to fix a crash when fetching/updating products.

#### To test
- From `wp-admin`, update the `download_limit` field of a product to `123123124124`.
- Open the example app, click on `Woo` -> `Products` -> `Fetch single product` and enter the `productId` from the previous step.
- Verify that the app does not crash and there is no error in the logs similar to `java.lang.NumberFormatException: Expected an int but was 123123124124 at line 1 column 1188 path $.data.download_limit`

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
@@ -100,7 +100,7 @@ class MockedStack_WCProductsTest : MockedStack_Base() {
             assertEquals(product.getNumVariations(), 2)
             assertEquals(product.getDownloadableFiles().size, 1)
             assertEquals(product.downloadExpiry, 10)
-            assertEquals(product.downloadLimit, 2)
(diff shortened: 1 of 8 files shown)


nbradbury:
@anitaa1990 Changes are good, but I think the download limit needs to be changed to a `Long` in [WCProductVariationModel](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/67d51cb295bf0c39ebf83fa653a793725bf95f90/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductVariationModel.kt#L53) as well.

anitaa1990:
> @anitaa1990 Changes are good, but I think the download limit needs to be changed to a Long in WCProductVariationModel as well.

Thanks for the review @nbradbury! Good call on `WCProductVariationModel`. I've fixed that in  6cd6e1c. @0nko ready for another round 👍

0nko (approved):
Looks good! 👍

PR #1785, opened 2020-11-25
ScanStore: Start new scan

ashiagr:

This PR introduces `/sites/$site/scan/enqueue` API and adds corresponding `ScanAction`, `ScanRestClient` and `ScanStore` methods and tests.

**To test**
That tests run fine and cover important flows/ results for success, error states.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ScanTestJetpack.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ScanTestJetpack.kt
@@ -75,7 +75,7 @@ class ReleaseStack_ScanTestJetpack : ReleaseStack_Base() {
         assertNotNull(scanStateForSite)
         scanStateForSite?.apply {
             assertNotNull(state)
-            assertEquals(state, IDLE)
+            Assert.assertTrue(listOf(IDLE, SCANNING).contains(state))
         }
 
         assertNotNull(fetchedScanStatePayload)
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ScanTestJetpack.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ScanTestJetpack.kt
@@ -5,6 +5,7 @@ import org.greenrobot.eventbus.Subscribe
 import org.junit.Assert
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
 import org.junit.Test
 import org.wordpress.android.fluxc.TestUtils
 import org.wordpress.android.fluxc.annotations.action.Action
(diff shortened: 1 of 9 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ScanTestJetpack.kt
```diff
@@ -148,6 +149,18 @@ class ReleaseStack_ScanTestJetpack : ReleaseStack_Base() {
         }
     }
 
+    @Test
+    fun testStartScan() {
```
malinajirka, line 153: ⚠️  I'm wondering if this tests won't conflict with `testFetchScanState`. `testFetchScanState` expects that the state is `IDLE`. Ideally tests shouldn't affect each other and order shouldn't matter. Unless we are able to clean up after the test (in other words stop the scan), I don't think we should start it. We can include other team members in this discussion as it's an interesting one :). Wdyt?

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
```diff
@@ -254,6 +295,11 @@ class ScanRestClientTest {
         return Gson().fromJson(json, responseType) as? ScanStateResponse
     }
 
+    private fun getStartScanResponseFromJsonString(json: String): ScanStartResponse? {
```
malinajirka: Idea 💡 : I'm wondering if it'd be better to throw an exception instead of returning an error. If we silently return null, it might be more difficult to find why is the test failing. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStartResponse.kt
```diff
@@ -0,0 +1,15 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.scan
+
+data class ScanStartResponse(
```
malinajirka, line 5: We might want to use "@SerializedName" annotation to get rid of the "_" in the names. I'd personally use it for all the fields so we can freely rename them in the future without a risk of breaking the contract.

malinajirka (commented):
Thanks @ashiagr! I've left a few comments but it looks great overall.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanStartResponse.kt
ashiagr: I generated it from the json and forgot to add `@SerializedName` annotations. Added in fd0e7a7.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ashiagr: As discussed, removed `?` which should take care of this situation - in d09c124.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ScanTestJetpack.kt
ashiagr: As discussed, added assertion for possible states that test site can be in 2fa22d1 to handle tests order.

malinajirka:
Thanks for the changes ;) LGTM

malinajirka (approved):

PR #1786, opened 2020-11-26
[Bug] Change product variations order by logic to fix UI scrolling issue

ParaskP7:
Fix requirement for [WooCommerce-Android - #2848](https://github.com/woocommerce/woocommerce-android/issues/2848)
Complimentary PR to [WooCommerce-Android - #3217](https://github.com/woocommerce/woocommerce-android/pull/3217)

This PR is the first step towards solving the aforementioned [WooCommerce-Android](https://github.com/woocommerce/woocommerce-android) issue. Briefly, it changes the order by logic for getting product variations from the database in order to align that with the api request order by logic. The ordering logic for both is now by `date`. Previously, the database was ordering by `menu` instead, which was causing pagination to not work and forcing the UI to always scroll on top for every new loading of additional product variations.


anitaa1990 (approved):
Looks good @ParaskP7! Nice work :shipit:

PR #1787, opened 2020-11-30
Woo: Parse fees in order details

hichamboushaba:
Requirement for https://github.com/woocommerce/woocommerce-android/issues/3141

Adds support for parsing and exposing the fees details of an order.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
@@ -28,7 +28,7 @@ open class WellSqlConfig : DefaultWellConfig {
     annotation class AddOn
 
     override fun getDbVersion(): Int {
-        return 124
+        return 125
     }
 
     override fun getDbName(): String {
@@ -1351,6 +1351,9 @@ open class WellSqlConfig : DefaultWellConfig {
                             "INITIAL BOOLEAN," +
                             "REASON TEXT)"
                     )
+                }
+                125 -> migrate(version) {
+                    db.execSQL("ALTER TABLE WCOrderModel ADD FEE_LINES TEXT")
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


anitaa1990 (approved):
LGTM :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
```diff
@@ -1352,6 +1352,9 @@ open class WellSqlConfig : DefaultWellConfig {
                             "REASON TEXT)"
                     )
                 }
+                125 -> migrate(version) {
```
anitaa1990, line 1355: Oops..sorry I merged the PR too soon! We would need to use `migrateAddOn(ADDON_WOOCOMMERCE, version)` instead of `migrate(version)` since the latter is used for the WordPress apps and right now this will cause a crash in that app (because the `WCOrderModel` table does not exist in the WordPress apps. 

Would you mind opening a PR with a fix?

anitaa1990 (commented):
Sorry @hichamboushaba I think I merged the PR too soon. There is a migration bug that would cause the WordPress app to crash. Would you mind opening a PR to fix that please?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
anitaa1990: Actually opened a quick PR [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1790), and assigned you as a reviewer 👍

mchowning, line 1355: :wave: @hichamboushaba and @anitaa1990 ! I noticed that we skipped adding a migration for version 124 here. Was that intentional?

hichamboushaba: @mchowning thanks for pointing this out, it's a mistake from me 🤦, I'll have another PR to fix it on Monday.

PR #1789, opened 2020-11-30
ScanStore: Introduce ThreatMapper and custom deserialisers

ashiagr:

This is a small preparatory PR that
1. Refactors code to introduce `ThreatMapper` and `ThreatMapperTest` to ease mapping threats based on different types (database, core, file etc) later on.
2. Introduces new fields `context`, `fileName`, `rows`, `diff` to be used later on to map threat types.
3. Introduces custom deserialiser for Threat context object.

   `context` obj can either be:
   - JsonObject - e.g.
   ```
   "context": {
       "3": "echo <<<HTML", 
       "4": "X5O!P%@AP[4\\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*",
       "5": "HTML;",
       "marks": {
         "4": [
           [0, 68]
         ]
       }
     }  
   ```
   - an empty string
   - not present

4. Introduces custom deserialiser for `Fixable` object.
   `fixable` obj can either be:
   - JsonObject - e.g.
   ```
   "fixable": {
       "fixer": "update",
       "target": "4.2.2"
     }
   ```
   - boolean set to false if threat is not fixable
   ```
   "fixable": false
   ```

#### To Test

- `ThreatContextDeserialiser` behaves correctly for different threat context responses.
- `FixableDeserialiser` behaves correctly for different fixable responses. 

(unit tests covered in `ThreatMapperTest`).

Rest of the changes (new fields, fields nullability) can be left for testing in the upcoming PRs which will map `Threat` to different types and use `ThreatResponse` in the `Threat details API `integration.


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/FixableDeserializer.kt
```diff
@@ -0,0 +1,34 @@
...
+         * If a threat is not fixable, fixable is returned as a boolean (false).
+         * It is set to null in this case to avoid json parsing issue.
+         */
+        return if (context != null && json != null && json.isJsonObject) {
+            val fixableType = object : TypeToken<Fixable>() { }.type
+            val result: Fixable?
+            result = try {
+                context.deserialize(json, fixableType)
+            } catch (e: JsonSyntaxException) {
+                null
```
malinajirka, line 27: Is this state expected or just a fallback? 

I'd expect `json.isJsonObject` returns false when the value is a boolean and not a jsonObject. If that's the case, we might want to log an exception in this catch clause. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/ThreatContextDeserializer.kt
```diff
@@ -0,0 +1,136 @@
...
+        threatContextJsonObject: JsonObject,
+        marksJsonObject: JsonObject? = null,
+        key: String
+    ): ContextLine? {
+        var contextLine: ContextLine? = null
+
+        val lineNumber = try {
+            key.toInt()
+        } catch (ex: NumberFormatException) {
+            return null
```
malinajirka, line 80: We might want to consider logging a message here, wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/ThreatContextDeserializer.kt
```diff
@@ -0,0 +1,136 @@
...
+            return highlights
+        }
+    }
+
+    private fun getSelectionRange(rangeArrayJsonElement: JsonElement): Pair<Int, Int>? {
+        return try {
+            val startIndex = rangeArrayJsonElement.asJsonArray.get(0).asInt
+            val endIndex = rangeArrayJsonElement.asJsonArray.get(1).asInt
+            Pair(startIndex, endIndex)
+        } catch (ex: ClassCastException) {
```
malinajirka, line 124: We might want to consider logging a message here, wdyt?

malinajirka (approved):
Thanks @ashiagr, great job! ;) 

It's a pity that we can't use the default deserializers :(. I never enjoy parsing an API which returns different types for the same fields.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/FixableDeserializer.kt
ashiagr: It is just a fallback similar to [this one](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/a097f94c384cc146c9a187144ea73eb1ab64c893/fluxc/src/main/java/org/wordpress/android/fluxc/model/EditorTheme.kt#L106-L113).

ashiagr:
Thanks for the review @malinajirka ! 

As discussed, I've created a nice-to-have [ticket in the project board ](https://github.com/orgs/wordpress-mobile/projects/75#card-50468548) about adding testing version of `AppLog`, of what we tried and the issues we encountered while running tests on doing so.

PR #1790, opened 2020-12-01
Fixed migration script to use woocommerce

anitaa1990:
PR #1787 introduced a migration script to `WellSqlConfig`.This would potentially cause a crash in the WordPress app (if that commit hash is used) since the `WCOrderModel` does not exist on WordPress.  I accidentally merged that PR before realising this. This PR just modifies the migration script to fix this.


hichamboushaba:
Thanks for taking care of this @anitaa1990, I didn't notice the other migrate method 🤦

hichamboushaba (approved):

anitaa1990:
Thanks @hichamboushaba! No worries! I made a [similar mistake in the past](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1648) 😓

PR #1791, opened 2020-12-01
ScanStore: Map ThreatModels based on types

ashiagr:

This PR converts `ThreatModel` to a sealed class and maps threat models based on different types.

Based on [Calypso logic](https://opengrok.a8c.com/source/xref/calypso/client/components/jetpack/threat-item/utils.ts) and [threat types logic](https://opengrok.a8c.com/source/xref/trunk/wp-content/lib/vaultpress/threat-types.php#70)
- File based threats (`threat.context != null`) will provide code `context` to be displayed.
- Core modification threats (`threat.diff != null)` will contain a git `diff` string.
- Database threats (`threat.rows != null || 'Suspicious.Links' == threat.signature`) will contain `rows` information.
- VulnerableExtension threats (`threat.extension != null`)  will contain `extension` information.

If none of these, threat is mapped as a `GenericThreat`.

#### To test

That unit tests in `ThreatsMapperTest` run fine for different types of threats. 

#### Note

- JSON data samples are taken from [here](https://opengrok.a8c.com/source/xref/trunk/bin/tests/isolated/suites/JetpackScan/data/) with some of the fields replaced with ones sent in the API response (e.g. `firstDetected` is sent as `first_detected`).


fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/threat/ThreatMapper.kt
```diff
@@ -1,38 +1,110 @@
...
+        return when {
+            response.fileName != null && response.diff != null -> {
+                CoreFileModificationThreatModel(
+                    id = id,
+                    signature = signature,
+                    description = description,
+                    status = status,
+                    firstDetected = firstDetected,
+                    fixable = fixable,
+                    fixedOn = response.fixedOn,
```
ashiagr: These fields are repeated in each threat model. Wondering if we should club them into `BaseArgs` as done in the [API code](https://opengrok.a8c.com/source/xref/trunk/wp-content/lib/vaultpress/threat-types.php?r=216961#68) 🤔.

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/threat/ThreatModel.kt
```diff
@@ -2,41 +2,180 @@ package org.wordpress.android.fluxc.model.scan.threat
...
-    val fixable: Fixable? = null,
-    val extension: Extension? = null,
-    val firstDetected: Date,
-    val fixedOn: Date? = null,
-    var context: ThreatContext? = null,
-    val fileName: String? = null,
-    val diff: String? = null,
-    val rows: String? = null
+sealed class ThreatModel(
+    open var id: Long,
```
malinajirka: Could you please elaborate what was the reason for changing vals->vars? I think that in an ideal scenario all our models would be immutable. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/threat/ThreatModel.kt
```diff
@@ -2,41 +2,180 @@ package org.wordpress.android.fluxc.model.scan.threat
...
-    val status: String?, // TODO: ashiagr - possible values: fixed / ignored / current
-    val fixable: Fixable? = null,
-    val extension: Extension? = null,
-    val firstDetected: Date,
-    val fixedOn: Date? = null,
-    var context: ThreatContext? = null,
-    val fileName: String? = null,
-    val diff: String? = null,
-    val rows: String? = null
+sealed class ThreatModel(
```
malinajirka: I think we could move all the fields from this constructor into the class itself and mark them as `abstract`. It'd simplify the sub-classes. Wdyt?

Something like
```
sealed class ThreatModel {
    abstract val id: Long
    abstract val signature: String
    abstract val description: String
    abstract val status: ThreatStatus
    abstract val firstDetected: Date
    abstract val fixable: Fixable?
    abstract val fixedOn: Date?

    data class GenericThreatModel(
        override val id: Long,
        override val signature: String,
        override val description: String,
        override val status: ThreatStatus,
        override val firstDetected: Date,
        override val fixable: Fixable? = null,
        override val fixedOn: Date? = null
    ) : ThreatModel()
.....
}
```

or we might consider taking it even one step further and write something like

```
data class BasicThreatModel(
    val id: Long,
    val signature: String,
    val description: String,
    val status: ThreatStatus,
    val firstDetected: Date,
    val fixable: Fixable
)

sealed class ThreatModel {
    abstract val basicThreatModel: BasicThreatModel

    data class GenericThreatModel(
        override val basicThreatModel: BasicThreatModel
    ) : ThreatModel()

    data class CoreFileModificationThreatModel(
        override val basicThreatModel: BasicThreatModel,
        val fileName: String,
        val diff: String
    ) : ThreatModel()

    data class VulnerableExtensionThreatModel(
        override val basicThreatModel: BasicThreatModel,
        val extension: Extension
    ) : ThreatModel()

    data class DatabaseThreatModel(
        override val basicThreatModel: BasicThreatModel,
        val rows: List<Row>? = null
    ) : ThreatModel() {
        data class Row(
            val id: Int,
            val rowNumber: Int,
            val description: String? = null,
            val code: String? = null,
            val url: String? = null
        )
    }
....
}
```

I used `BasicThreatModel` name just for the showcase purposes, I'm not sure if it's a good name.

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/threat/ThreatMapper.kt
malinajirka: See comment on the ThreatModel class - > summary: I think we could use this approach even on the level of the sealed class. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/RowsDeserializer.kt
```diff
@@ -0,0 +1,93 @@
...
+        var row: Row? = null
+
+        val rowNumber = try {
+            key.toInt()
+        } catch (ex: NumberFormatException) {
+            return null
+        }
+
+        val contentsForKey = rowJsonObject.get(key)
+        val contentsForKeyAsJsonObject = if (contentsForKey?.isJsonObject == true) {
```
malinajirka: Nitpick 🔍 : We might want to consider simplifying this statement:
```
val contentsForKeyAsJsonObject = contentsForKey?.takeIf { it.isJsonObject }
```
Wdyt?

We could take it even a few steps further and end up with something like this 
```
private fun getRow(key: String, rowJsonObject: JsonObject): Row? =
            rowJsonObject.get(key)?.takeIf { it.isJsonObject }?.asJsonObject?.let { contents ->
                try {
                    Row(
                            rowNumber = key.toInt(),
                            id = contents.get(ID)?.asInt ?: 0,
                            description = contents.get(DESCRIPTION)?.asString,
                            code = contents.get(CODE)?.asString,
                            url = contents.get(URL)?.asString
                    )
                } catch (ex: ClassCastException) {
                    null
                } catch (ex: IllegalStateException) {
                    null
                } catch (ex: NumberFormatException) {
                    null
                }
            }
```

Please feel free to ignore this comment. Use the version you like better. I just wanted to propose an alternative.

malinajirka (commented):
Great job @ashiagr! I've left a couple of comments - nitpicks and minor suggestions. Let me know what you think.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/RowsDeserializer.kt
ashiagr: I like this alternative, still not used to using `takeIf`, thanks for suggesting it 👍.
In 8c56ab8

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/threat/ThreatModel.kt
ashiagr: Reverted in 77f1cf4.
It was unintentional, I didn't realise I was doing it.

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/threat/ThreatModel.kt
ashiagr: Yes, this is what was missing - thanks for explaining it 🙏.
In 77f1cf4.

malinajirka (approved):
Thank you for all the changes ;)! LGTM

PR #1792, opened 2020-12-01
Fetch threat details

ashiagr:

This PR introduces `/sites/$site/scan/threat/$threat_id` API and adds corresponding `ThreatAction`, fetch methods in the `ThreatRestClient` and `ThreatStore`.

#### To test
That tests run fine and cover important flows/ results for success, error states.

**Question**

Would it be helpful to include a connected test as well for threat details? We'll need  threat id for a threat on the real site and might need to first run fetch scan state or integrate fetch threats API only for tests to get one.

**Note**

This PR covers fetching threat details only from the network. Storing threat details into DB, if needed, will be covered in a separate PR.


malinajirka (approved):
Thanks @ashiagr! LGTM 

> Would it be helpful to include a connected test as well for threat details? We'll need threat id for a threat on the real site and might need to first run fetch scan state or integrate fetch threats API only for tests to get one.

Could we hardcode a valid threat id in the test.properties? We sometimes do this - eg. for images [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/e7af20209b24c9eb358c229d7b7a943baaf7893e/example/tests.properties-example#L25). If that's the only thing blocking us from having a connected test for this flow, I'd consider adding it. Wdyt?

ashiagr:
Thanks for the review @malinajirka !

> Could we hardcode a valid threat id in the test.properties? We sometimes do this - eg. for images here. If that's the only thing blocking us from having a connected test for this flow, I'd consider adding it. Wdyt?

Wow, thanks for showing this example scenario! I'll hardcode it and add a test.

ashiagr:
Connected test added, ready for another look @malinajirka!
You may need to pull latest tests.properties.

malinajirka:
Thanks for adding the test ;)!

When I run the test locally it fails with
```
java.lang.IllegalArgumentException: Required value was null.
	at org.wordpress.android.fluxc.release.ReleaseStack_ThreatTestJetpack.testFetchThreat(ReleaseStack_ThreatTestJetpack.kt:68)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
```

Screenshot 2020-12-10 at 08 40 18

Is it possible that some other tests fixed the threat so it's not valid anymore?

ashiagr:
I had the correct one in my local `tests.properties` when I tested it. Looks like there was some confusion when I copied to mobile secrets. I've updated it again. It might take some time for mobile secrets to reflect it. Sorry about the confusion 😞.

malinajirka:
No worries at all ;).

I've tested it with the new ID, but I still receive the same error. Any ideas?
Screenshot 2020-12-11 at 12 16 40 
Screenshot 2020-12-11 at 12 16 30

Screenshot 2020-12-11 at 12 17 06

ashiagr:
I think I know the reason now. It's funny but good that I know now (partially) 😄.

The site has a scan daily plan and it auto fixes the (fixable) threat each day. The threat that I uploaded to the site was a fixable one that got fixed the previous day:

Screenshot 2020-12-11 at 10 14 18 PM

Ideally it should not have re-shown in the scan results the next time, but a new threat is created with another id this time which created a confusion. 

Screenshot 2020-12-11 at 10 14 34 PM

I confirmed different threat ids on the wordpress API console after scanning twice:

Screenshot 2020-12-11 at 10 14 43 PM

So I think first time also correct id was set to `tests.properties` but by the time you tested, it got changed. 

Should we drop the idea of including connected test for it? Or may be defer the decision?

malinajirka:
Good investigation :). Make sense. 

If we can easily create a non-fixable threat which would remain on the site forever (with the same id), I'd give it a try. If it's complicated or not doable I'd drop the idea of the connected test. Wdyt?

ashiagr:
> If we can easily create a non-fixable threat which would remain on the site forever (with the same id), I'd give it a try. 

I've added to backlog here: 
https://github.com/wordpress-mobile/WordPress-Android/issues/13326

I'll see how we can generate one and add a test then.

Can we close this PR?

malinajirka:
This PR contains even other changes. I'd personally remove the connected test and merged the PR afterwards. Wdyt?

PR #1795, opened 2020-12-06
Woo/shipping label product ids fix

anitaa1990:
The GET shipping labels API endpoint now provides the `product_ids` field in the [response json](https://github.com/Automattic/woocommerce-services/pull/2125). This is needed in order to fetch the product details (images, name, SKU etc) associated with the product ids of a shipping label. This PR adds the `product_ids` field to the local db for use in the Woo app.

#### To test
- Open the example app and click on `Woo` -> `Shipping Labels` -> `Select site` -> `Fetch shipping labels`. Enter a valid order id and verify that the shipping labels for the order is fetched correctly and stored in the local db correctly. Check that the `WCShippingLabelModel` table in `Database Inspector` includes the `product_ids` field and it is not empty.
- Run `WCShippingLabelSqlUtilsTest` and `WCShippingLabelStoreTest`.
- The `product_ids` field is not available by default in all the API responses. For shipping labels created before the 1.24.1 version was released, the `product_ids` field will not be available. Since I had a few shipping labels created before then, I was able to test this scenario. This scenario is added to the `WCShippingLabelStoreTest` but if anyone would like to test this in the example app, please reach out to me so that I can provide you access to my test site :)


anitaa1990:
@hichamboushaba @mchowning, I have updated the migration script to the correct version in this PR. So this should fix the issue that @mchowning mentioned in [this PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1787#discussion_r536414946). Thanks for bringing that up @mchowning 🙇

mchowning:
> I have updated the migration script to the correct version in this PR.

Thanks so much @anitaa1990 !

I see that the db version is fixed to now be the correct number (1 + the latest migration block), so the latest migration block will be run. ❤️  

We still have the `onUpgrade` function skipping a migration for version 124. I did a quick migration test, and it does look like the migration of older version still works fine even with us skipping a migration for 124 (which makes sense because our migrations [run _every_ migration between `oldVersion` and (`newVersion`-1)](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/c00483eba00bace075a280aaa6cd31d1cea41cf6/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt#L47)). 👍 

I do feel like it would be helpful to include either a standalone comment or a comment in an empty `124 -> { ...` block to briefly explain that 124 was accidentally skipped. I think that could help avoid some confusion and investigation in the future if someone else notices the skipped version, but wdyt?

hichamboushaba:
>  run every migration between oldVersion and (newVersion-1)

@mchowning I think it actually runs between oldVersion and newVersion, as we are using the operator `..`, and it's what causes my confusion about using the newVersion there. I think we should replace `..` by `until` to exclude the newVersion from the migration.

And I agree that a comment can be useful, and why not even replace 125 by 124 there, as the change wasn't released yet 🤔.

anitaa1990:
> why not even replace 125 by 124 there, as the change wasn't released yet

Good point! I didn't notice that we had skipped 124 but since it's not yet release we can make that change. I updated it in 29f7b3f

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/shippinglabels/WCShippingLabelModel.kt
```diff
@@ -88,6 +89,22 @@ class WCShippingLabelModel(@PrimaryKey @Column private var id: Int = 0) : Identi
...
 
+    /**
+     * Returns the list of products the shipping labels were purchased for
+     *
+     * For instance: "[60, 61, 62]" would be split into a list
+     * [60, 61, 62]
+     */
+    fun getProductIdsList(): List<Long> {
+        return productIds
+                .trim() // remove extra spaces between commas
```
hichamboushaba: @anitaa1990 I think the comment on this line is not totally accurate, as it's applied before the split, it will delete the white space before `[` and after `]`, not between commas, unless I'm missing something.

WDYT? and is it necessary? I'm asking as there is a `trim` applied to individual ids later 🙂

anitaa1990: There are some cases where the API has sent data that includes spaces before and after the `[` and `]` so I'm not sure it's wise to remove that 😅 But I agree the comment is a bit misleading. I updated it in `1ebc5d8`

hichamboushaba (approved):
LGTM! and thanks for fixing the issue with the migration version 🙏 
:shipit:

PR #1797, opened 2020-12-07
ScanStore: Fix ScanState table fields

ashiagr:
Issue: https://github.com/wordpress-mobile/WordPress-Android/issues/13326

This is a small PR that 
- Fixes `ScanState` table fields (type, name) created in the `WellSqlConfig.onUpgrade` (called on app upgrade when db version is upgraded) - 4607c9a 
- Implements `Response` in `ScanStartResponse`

It can be tested via PR: https://github.com/wordpress-mobile/WordPress-Android/pull/13555


fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
```diff
@@ -38,9 +38,9 @@ data class ScanStateModel(
 
     data class ScanProgressStatus(
         val startDate: Date,
-        val duration: Int = 0,
-        val progress: Int,
-        val error: Boolean = false,
-        val isInitial: Boolean
+        val duration: Int? = null,
```
malinajirka: I don't mind this change, bust just to make sure we are one the same page regarding the nullability. Feel free to ignore this comment if you changed the fields to nullable fields for a different reason.

I think, it's ok to have non-nullable fields in our domain models when we have an ok default/fallback value. This approach is used for example with the "hasCloud" field - it's nullable in the response model (ScanStateResponse), but non-nullable in the domain model (ScanStateModel). 

Sometime we want to keep them nullable even in the domain model as we don't have an ok default/fallback value 
For example: 
- port (port 0 or -1 doesn't make sense and it's better to use null when it's empty) 
- if we had lastScan field (we could set it to Date(0), but it wouldn't make sense from the user point of view so it's better to keep it as null).

So if we have an ok default/fallback value for nullable fields I'd personally consider making them non-nullable in the domain model. Does it make sense, wdyt?

malinajirka (commented):
Thanks @ashiagr. I've left a comment for you to consider. Other than that it LGTM.

fluxc/src/main/java/org/wordpress/android/fluxc/model/scan/ScanStateModel.kt
ashiagr: In 6f3cad6

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt
```diff
@@ -86,11 +86,11 @@ class ScanSqlUtils @Inject constructor() {
         @Column var remoteSiteId: Long,
         @Column var state: String,
         @Column var startDate: Long,
-        @Column var duration: Int,
```
malinajirka, line 89: I'm wondering if we need to make the startDate nullable after the update of `ScanProgressStatus` ? Wdyt?

ashiagr: It was set to zero [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/6f3cad651052eb27c4b62b6d7ad16701b030ca6d/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/ScanSqlUtils.kt#L50-L52), changed to null explicitly in 89a46c1.

malinajirka (approved):
LGTM, thanks!

PR #1798, opened 2020-12-08
Fix the broken migration script

0nko:
This PR fixes the broken migration script by adding the missing comma.


ParaskP7 (approved):
Thanks for the quick fix @0nko ! 🌟

PR #1799, opened 2020-12-10
Woo: Add support for deleting image from product variation

AmandaRiu:
This PR adds support for deleting the image from a product variation. This is done by sending only `{ id: 0 }` as the value for the variation image. It’s important that *only* the `id` is sent, because if `src` get’s set the API will expect a new image is being *added* and throw an error. [This functionality was recently added to the API in Woo 4.7](https://github.com/woocommerce/woocommerce/pull/27299) so if a store is running an older copy of Woo then the error `woocommerce_variation_invalid_image_id` will be returned from the API. 

This PR adds the following: 
- If the image `id=0`, then the `toJson()` method will only include the `id` property 
- New error type `woocommerce_variation_invalid_image_id` added
- Two new tests added to `MockedStack_WCProductsTest`

### To Test
- Run the tests in `MockedStack_WCProductsTest` and verify success.


nbradbury (approved):
:shipit:

Issue #1801, opened 2020-12-15
Restore testRedeemingCardWithWrongCountryCode test when API issues is resolved


aforcier:
#1802 suppresses a test broken by API-side changes - those are being looked into, and we should remove the @Ignore annotation when that's complete.

cc @khaykov @malinajirka

aforcier:
Restored in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1805.

PR #1802, opened 2020-12-15
Temporarily disable a failing test - testRedeemingCardWithWrongCountryCode

malinajirka:
This PR temporarily disables a test which is failing due to a bug on the API.

We plan to revert this change when the bug on the API is fixed - https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1801


aforcier:
:shipit:

PR #1803, opened 2020-12-16
Create Store Attribute base endpoints handling

ThomazFB:
Summary
==========
Partially deliver implementation for the [Woo Issue 3204](https://github.com/woocommerce/woocommerce-android/issues/3204), there will be further upcoming PRs in order to deliver data persistence, models and mappers. This PR adds support to retrieve, create, update and delete Attribute data in the site. Also adds to the example app the capability to test and log the Attributes output from a given site.

*Tests coverage on upcoming PRs*

Screenshots & How to Test
==========
| Select the Woo section | Scroll down until the Attribute option becomes visible and click on it | Fetch, create, update and remove attributes
| screenshot-1608543978049 | screenshot-1608543983921 | screenshot-1608535524554 |

To verify if the changes are working under the Woo site, go to Products > Attributes on your Woo site and make sure everything is as expected.

Screen Shot 2020-12-21 at 4 25 56 AM


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/JetpackRequestBuilderExt.kt
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Extends JetpackTunnelGsonRequestBuilder to make available a new JSON-formatted PUT and DELETE requests,
+ * triggers it and awaits results synchronously.
+ * @param restClient rest client that handles the request
+ * @param url the request URL
+ * @param body the content body, which will be converted to JSON using [Gson][com.google.gson.Gson]
+ * @param clazz the class defining the expected response
+ */
+
+suspend fun <T : Any> JetpackTunnelGsonRequestBuilder.syncPutRequest(
```
AmandaRiu: Curious why you didn't just add these to the `JetpackTunnelGsonRequestBuilder` directly? It doesn't seem like much is gained by moving it into an Extension file.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductAttributesStore.kt
```diff
@@ -0,0 +1,83 @@
...
+        name: String,
+        type: String = "select",
+        orderBy: String = "menu_order",
+        hasArchives: Boolean = false
+    ): WooResult<AttributeApiResponse> =
+            coroutineEngine.withDefaultContext(AppLog.T.API, this, "createStoreAttributes") {
+                restClient.postNewAttribute(
+                        site, mapOf(
+                        "name" to name,
+                        "slug" to name,
```
AmandaRiu: The user has the ability to create a custom slug in `wp-admin`, curious why we aren't allowing that option on the backend?

AmandaRiu (changes requested):
@ThomazFB Really nice work on this. I especially love the output in the Example app - very detailed and nicely formatted for viewing. I did find some issues with the test in the Example app though. For example, I can update an attribute and get an error, but the example app treats it like it’s successful. The Update, Create and Delete options all act like the action was successful even when there is an error response from the API. To reproduce:
1. Click **UPDATE ATTRIBUTE**
2. Leave the first dialog blank
3. Leave the second dialog blank. Flipper shows the response is a 404, but the UI shows “Attribute Updated” and doesn’t display an error. 

Also left a couple code comments for your review.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/JetpackRequestBuilderExt.kt
ThomazFB: Well, that's something I'm curious about your opinion. The main reason I did that was related to the module distribution on FluxC, since the Woo codebase is on a different module I understood that adding functions directly to the class would insert code that is never used or required by any part inside the FluxC module, instead, it would be only used on the Woo plugin module part, so I decided to make it as an extension operation exclusive to Woo. But I'm not sure if we want to keep this kind of consistency around the entire FluxC lib. Do you think it would be better to just move everything to the FluxC module side?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductAttributesStore.kt
ThomazFB: Oops, nice catch! That was a mistake of mine, it's fixed on https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1803/commits/b26e1033957220e7f407093bb2cb1fabe8294827

ThomazFB:
Hey @AmandaRiu, thank you so much for your review, about the error handling and logging, I've improved it right here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1803/commits/da0a3543aceb44ec88e6de4dc62362e234e21419. Let me know what you think!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/JetpackRequestBuilderExt.kt
AmandaRiu: yeah I understand your point, but these methods would be used eventually and I can see a scenario where the logic ends up getting duplicated.

AmandaRiu (approved):
Looks like all the requested changes have been made. :shipit:

PR #1804, opened 2020-12-16
Use binary configure

jkmassel:
This PR replaces the one-off encryption scheme in this project with the binary `configure` gradle plugin which allows us to avoid setting up Ruby in order to run `configure` using `fastlane`. 

**To Test CI Integration**
- Ensure automated tests pass

**To Test `apply`**
- Check out this PR locally
- Run `rm -rf .configure-files vendor && git reset --hard` to reset the repo to its default post-checkout state
- Run `./gradlew applyConfiguration`
- Note that the files are correctly decrypted in the `.configure-files` directory.

**To Test `update`**
- Check out this PR locally
- Run `rm -rf .configure-files vendor && git reset --hard` to reset the repo to its default post-checkout state
- Run `./gradlew updateConfiguration`
- Note that the `.configure` file has a new pinned hash corresponding to the latest mobile secrets commit
- Note that the `.enc` files have changed


loremattei:
This is awesome @jkmassel!

CI is green, so I think it works, but when I tried `.gradlew applyConfiguration` locally, it finished without errors, but I couldn't find any decrypted file in the `.configure-files` folder. 

Also, I took a look at the plugin and, if I got it correctly, the `update` task is not available yet. I wonder if we should provide that as well before merging this PR, because otherwise devs may have a hard time if they need to update the secrets.

jkmassel:
@loremattei – this is ready for another look! I've added support for `update` and updated the testing instructions for both scenarios, so hopefully it'll be more successful this time.

We can commit the changes to the secrets in a subsequent PR, I think – for now it's useful to have something to test the `update` command against.

oguzkocer (approved):
Testing instructions worked great for me. I love how clean the implementation is now. Thank you for working on this @jkmassel! 🙇

PR #1806, opened 2020-12-17
ScanStore: Store threats data in db

ashiagr:
Parent https://github.com/wordpress-mobile/WordPress-Android/issues/13326

This PR 
- Removes `ThreatStore`, `ThreatAction`, `ThreatRestClient` as threats can be managed within the `ScanStore` due to how APIs are structured (threats returned as part of Scan data) and removing them helps in avoiding complexity between `ThreatStore` & `ScanStore` relationship - 2682769
- Stores/ retrieves threat data into/ from db - 1f2b8aa, 432ed80
- Based on different threat types, special objects like `ThreatContext`, `Extension`, `List<Rows>` are stored as json strings -  17dfbc7, 52e3142

To test
`ScanStoreTest`, `ThreatSqlUtilsTest` cover important flows and run properly.

Note: There's scope of optimisation: we can probably pass response threat model to the `ThreatSqlUtils` to avoid json conversion while saving special objects. Lmkwyt.


malinajirka (approved):
Thanks @ashiagr! Looks great, it feels definitely better to keep it within the ScanStore  - good call!

> Note: There's scope of optimisation: we can probably pass response threat model to the ThreatSqlUtils to avoid json conversion while saving special objects. Lmkwyt.

I personally wouldn't worry about it. We don't expect thousands of threats anyway - more of 1-3 threats. I'd keep it as is unless we encounter performance issues while testing. Thanks for thinking about it though.

AliSoftware:
@malinajirka Hello there!

It seems that you created a `1.7.0-beta-1` release and tag after merging this PR. The issue is that there was already an official `1.7.0` at that time, so that new beta should have been a `1.8.0-beta-1` instead.

It happens that I just released a hotfix 1.7.1 on top of 1.7.0 – to bring back [a fix we patched back in 1.6.26.1 but forgot to merge back to develop](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1781), so 1.7.1 now includes this old patch on top of 1.7.0.

It's already end of my (long) day so I won't have time to fix that myself before my AFK, but I'll let you in the hands of @jkmassel to decide what to do with that `1.7.0-beta-1`.

Personally I recommend either one of those options to fix this:

 - Either create a new tag name in the place of that `1.7.0-beta-1` (aka, rename 1.7.0-beta-1) to name it `1.8.0-beta-1` instead (the name it should have had all along), so that it has the appropriate semver naming. Then consider bringing the changes from the `1.7.1` hotfix back on top of that newly-named `1.8.0-beta-1` to then create `1.8.0-beta-2` on top of that. Then update the client apps to point to `1.8.0-beta-2`
 - Or, completely forget the existence of `1.7.0-beta-1` (delete the release and tag, even?), and directly create a brand new `1.8.0-beta-1` that would contain _both_ the patch from this PR and the patch that landed in `1.7.1` (and got since merged back to develop, and should [soon land back to `trunk` too](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1814)), Then update the client apps to point to that new `1.8.0-beta-1`.

In either case it might be worth considering deleting the 1.7.0-beta-1 GitHub pre-release, or at least edit its description to mention it's not appropriate release to use, and to use 1.8.0-beta-N instead.

ashiagr:
👋 @AliSoftware, 

`1.7.0-beta-1` tag was created by me. It somehow displayed incorrect creator name. Apologies for the troubles caused because of it 😞.

Thank you @jkmassel for creating `1.8.0-beta.1` tag for us.

I'll delete `1.7.0-beta-1` tag once client app starts pointing to the new `1.8.0-beta-N` tag.

PR #1807, opened 2020-12-17
Issue/13268 add activity type endpoint

malinajirka:
Parent issue https://github.com/wordpress-mobile/WordPress-Android/issues/13268

This PR adds an endpoint for "fetch activity types" action which is used to get available activity types for ActivityLog filter.

I had to use a custom deserializer since the API respones uses values as keys.

For example
```
{
  "groups": {
    "post": {
      "name": "Posts and Pages",
      "count": 69
    },
```
`post` in this response is used as a key, but it's actually a value (id of the activity type).

Note: It seems the server currently ignores the "after" and "before" dates. We use a different format than Calypso. However, we use this format on all the APIs. I plan to fix this in a separate PR when we decide on the best approach.

To test:
Run the connected test
Test the feature in WPAndroid PR - tbd
Let the CI handle the rest.


example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
```diff
@@ -414,6 +419,71 @@ class ActivityLogRestClientTest {
...
+                        ),
+                        15
+                )
+        )
+        val siteId = site.siteId
+
+        val payload = activityRestClient.fetchActivityTypes(siteId, null, null)
+
+        assertEquals(
+                payload.activityTypeModels!![0],
```
ashiagr: Nitpick: IDE warning:
`Unnecessary non-null assertion (!!) on a non-null receiver of type List<ActivityTypeModel>`

fluxc/src/main/java/org/wordpress/android/fluxc/model/activity/ActivityTypeModel.kt
```diff
@@ -0,0 +1,7 @@
+package org.wordpress.android.fluxc.model.activity
+
+data class ActivityTypeModel(
+    val key: String,
+    val name: String,
+    val count: Int?
```
ashiagr: ❓Should we make domain `ActivityTypeModel.count ` non-nullable?

Response `ActivityTypeModel.count `when `null` is mapped to `0` in domain model:
`ActivityTypeModel(requireNotNull(it.key), requireNotNull(it.name), it.count ?: 0)`

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -390,4 +452,19 @@ class ActivityLogRestClient(
         val validUntil: Date?,
         val url: String?
     )
+
+    data class ActivityTypesResponse(
+        @JsonAdapter(ActivityTypesDeserializer::class) val groups: Groups?,
+        val totalItems: Int?
+    ) {
```
ashiagr: We might want to implement `Response` interface for response models according to [WordPress-FluxC-Android Documentation: Response Models](https://github.com/wordpress-mobile/WordPress-FluxC-Android#response-models).

Feel free to ignore if you intentionally didn't implement to remain consistent with the rest of the class.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityTypesDeserializer.kt
```diff
@@ -0,0 +1,77 @@
...
+        return groups.get(key)?.takeIf { it.isJsonObject }?.asJsonObject?.let { item ->
+            try {
+                ActivityType(
+                        key = key,
+                        name = item.get(NAME)?.asString,
+                        count = item.get(COUNT)?.asInt
+                )
+            } catch (ex: ClassCastException) {
+                null
+            } catch (ex: NumberFormatException) {
```
ashiagr: ❓ Do we need to catch `NumberFormatException`?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
```diff
@@ -180,6 +186,34 @@ class ActivityLogRestClient(
...
+        val response = wpComGsonRequestBuilder.syncGetRequest(
+                this,
+                url,
+                params,
+                ActivityTypesResponse::class.java
+        )
+        return when (response) {
+            is Success -> {
+                buildActivityTypesPayload(response.data, remoteSiteId)
+            }
```
ashiagr: Nitpick: Short branches can be put on the same line as the condition, without braces.

```suggestion
            is Success -> buildActivityTypesPayload(response.data, remoteSiteId)        
```

ashiagr (commented):
Nice work @malinajirka! Let's consider some minor nitpicks in the code and we should be good to merge it.

I haven't included comments for line-wraps, fields explicit type declarations as it is more of a personal preference. We can probably consider practising them more often in our future work or automate the process.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
malinajirka: Fixed in 8d419ce2d4b05ba8c61c0cba6c6dc133d3bc0093

fluxc/src/main/java/org/wordpress/android/fluxc/model/activity/ActivityTypeModel.kt
malinajirka: Nice catch, fixed in 8d419ce2d4b05ba8c61c0cba6c6dc133d3bc0093

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
malinajirka: Fixed in 8d419ce2d4b05ba8c61c0cba6c6dc133d3bc0093

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityTypesDeserializer.kt
malinajirka: I thought "asInt" throws this exception, nice catch;). Fixed in 8d419ce2d4b05ba8c61c0cba6c6dc133d3bc0093

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt
malinajirka: Fixed in 8d419ce2d4b05ba8c61c0cba6c6dc133d3bc0093

ashiagr (approved):
LGTM 👍.

Feel free to merge after fixing failed CircleCI tests.

malinajirka:
Thanks @ashiagr! I've reverted a change where I started passing the date in "GMT+0". I think I'll need to test the timezone thoroughly before this is released. It seems the DateRange picker component doesn't work with timezones correctly + I'll need to verify if the API takes the timezone into account or it just ignores it.

ashiagr (approved):
> I think I'll need to test the timezone thoroughly before this is released. 

Agree, timezones are confusing. We'll figure out what works for us. Approving this one.

PR #1782, opened 2020-11-23
Merge back hotfix 1.6.26.1 -> develop

AliSoftware:
Hotfix 1.6.26.1 was cut from tag 1.6.26 and recently merged back to trunk, time to merge it back to develop too to follow git-flow [ref: p77Llu-9z9-p2]

PR #1808, opened 2020-12-18
Merge hotfix/1.6.27.1 into develop

AliSoftware:
Somehow we accidentally closed https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1782 instead of merging it, which means that the fix that we landed in 1.6.26.1 never made it into `develop` and never got included in the later versions of FluxC.

This required a hotfix for 1.6.27.1 – that WC 5.6's production build depends on – so that we can then do a WC 5.6.1 hotfix immediately after.

---

Note: 1.7.0-beta-1 (the currently latest tag) will benefit from getting this 1.6.26.1 hotfix merged in too, but since the 1.7.0-beta-1 tag currently points to the same commit as `develop`, I plan to do a `1.7.0-beta-2` tag pointing on `develop` just right after this PR is merged into `develop` (\cc @loremattei for confidence check)


AliSoftware:
I messed up bad when created the tag, this 1.6.27.1 contains way more than it's supposed to, my bad! Closing this PR now to avoid propagating the mess.

PR #1809, opened 2020-12-18
Merge hotfix/1.6.27.1 into trunk

AliSoftware:
Somehow we accidentally closed https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1782 instead of merging it, which means that the fix that we landed in 1.6.26.1 never made it into `develop` and never got included in the later versions of FluxC.

This required a hotfix for 1.6.27.1 – as WC 5.6's production build depends on 1.6.27 – so that we can then do a WC 5.6.1 hotfix immediately after with this FluxC hotfix


AliSoftware:
I messed up when created the tag, this 1.6.27.1 contains way more than it's supposed to, my bad! Closing this PR now to avoid propagating the mess.

PR #1816, opened 2020-12-22
Fix date range timezones

malinajirka:
Parent issue https://github.com/wordpress-mobile/WordPress-Android/issues/13268

This PR adjusts after/before dates based on the current timezone.

When the user selects a date range eg. 10-12th of January, they expect to see data for 10-12th in their timezone. Therefore the app needs to adjust the date range based on the current timezone of the device.

For example
2020-01-**10**T**00**:00:00+00:00 becomes 2020-01-**09**T**21**:00:00+00:00 when the device's timezone is GMT+3. The user selected "10th of January" which is "9th of January 21pm" in GMT+0 (server timezone).

To test:
Test in WPAndroid https://github.com/wordpress-mobile/WordPress-Android/pull/13655


ashiagr (approved):
Works as described 👍.

Thank you so much for fixing tricky timezone issues and also adding tests which made reviewing easier!

PR #1818, opened 2020-12-22
Update xpost test method name to be camel case

mchowning:
Updates the xpost test method to use camel case like the rest of the project.


cameronvoell (approved):
LGTM!

PR #1819, opened 2020-12-24
ScanStore: Get threat details by id

ashiagr:
This PR gets threat model from the Db for the given threat id.

EDIT: It also fixes threat mapping when a fixable threat with unknown `FixType` (like "rollback") was returned as a non fixable threat (`fixable = null)`. It is now returned with a non-null fixable property with `FixType` set to `UNKNOWN`.

**To test** 
That newly introduced tests in `ThreatSqlUtilsTest`, `ThreatMapperTest` run fine.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/model/scan/threat/ThreatsMapperTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/model/scan/threat/ThreatsMapperTest.kt
@@ -162,7 +162,6 @@ class ThreatsMapperTest {
 
         val model = mapper.map(threat.copy(fixable = Fixable(fixer = "rollback", file = null, target = null)))
 
-        assertNotNull(model.baseThreatModel.fixable)
         assertEquals(model.baseThreatModel.fixable?.fixer, FixType.UNKNOWN)
     }
 
--- a/example/src/test/java/org/wordpress/android/fluxc/model/scan/threat/ThreatsMapperTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/model/scan/threat/ThreatsMapperTest.kt
@@ -16,6 +16,7 @@ import org.wordpress.android.fluxc.model.scan.threat.ThreatModel.ThreatStatus
 import org.wordpress.android.fluxc.model.scan.threat.ThreatModel.VulnerableExtensionThreatModel
 import org.wordpress.android.fluxc.model.scan.threat.ThreatModel.VulnerableExtensionThreatModel.Extension.ExtensionType
 import org.wordpress.android.fluxc.network.rest.wpcom.scan.threat.Threat
+import org.wordpress.android.fluxc.network.rest.wpcom.scan.threat.Threat.Fixable
 import kotlin.test.assertEquals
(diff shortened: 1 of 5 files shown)


example/src/test/java/org/wordpress/android/fluxc/model/scan/threat/ThreatsMapperTest.kt
```diff
@@ -154,6 +155,17 @@ class ThreatsMapperTest {
...
     }
 
+    @Test
+    fun `maps fixable threat correctly for unknown fix type`() {
+        val threatJsonString = UnitTestUtils.getStringFromResourceFile(javaClass, THREAT_FIXABLE_JSON)
+        val threat = getThreatFromJsonString(threatJsonString)
+
+        val model = mapper.map(threat.copy(fixable = Fixable(fixer = "rollback", file = null, target = null)))
+
+        assertNotNull(model.baseThreatModel.fixable)
```
ParaskP7: Minor (🔍): This `NotNull` assertion is covered by the `Equals` assertion below, isn't it? Meaning that if the `fixable` is null it will fail on the second assertion as well. As such, I would consider removing this assertion based on the fact that it can be considered a duplication. Wdyt?

PS: The same applies to the test above.
PSS: The same applies for the newly created test within `ThreatSqlUtilsTest.kt` and the `assertThat(threat).isNotNull` line.

example/src/test/java/org/wordpress/android/fluxc/persistence/ThreatSqlUtilsTest.kt
```diff
@@ -152,4 +152,17 @@ class ThreatSqlUtilsTest {
...
+    fun `threat model gets retrieved for the given threat id`() {
+        val threatId = dummyBaseThreatModel.id
+        val dummyThreat = GenericThreatModel(baseThreatModel = dummyBaseThreatModel)
+        threatSqlUtils.replaceThreatsForSite(site, listOf(dummyThreat))
+
+        val threat = threatSqlUtils.getThreatByThreatId(threatId)
+
+        assertThat(threat).isNotNull
+        assertThat(threat).isInstanceOf(GenericThreatModel::class.java)
+        assertEquals(threatId, requireNotNull(threat).baseThreatModel.id)
```
ParaskP7: Minor (🔍): Consider removing the extra `requireNotNull(...)` wrapper check as this check is covered by the above `isInstanceOf(...)` check and can be considered duplicate.

ParaskP7 (approved):
👋 @ashiagr !

I have tested the solution (by running the tests) and can verify correctness. As such, I am approving this PR. I have left a couple of minor (🔍) test related comments for you to check. However, feel free to merge the PR anyway. I am not merging the PR myself since I want to give you the opportunity to first check my comments.

ashiagr:
>  I have left a couple of minor (🔍) test related comments for you to check.

Thanks @ParaskP7! Both of them make sense. I've addressed them in 75fc544. 
Ready for another look.

ParaskP7:
Thanks @ashiagr ! 🌟 

It looks great, I am now going to merge this PR.

🙏

PR #1823, opened 2021-01-07
getAudioTypesOnly added to MimeTypes

### LINKED ISSUE #1822: getAudioTypesOnly is missing from MimeTypes
## Expected behavior
When an instance of `MimeTypes` is created there should be a method that allows us to get all audio-based MIME types. 

## Actual behavior
Currently this functionality to get MIME types based on the media type only supports videos and images.

### Steps to reproduce the behavior

 Create a new instance of `MimeTypes` like `new MimeTypes()` and realize there's no getter for the audio types.

jd-alexander:
Fixes #1822
## Findings
`getAudioTypesOnly` was missing from MimeTypes. This functionality is now required by the Audio Block in Mobile Gutenberg because the Gutenberg Bridge's integration with the native media pickers need to filter media content by the audio MIME types only. 

## Solution
To add this behavior to `MimeTypes`

## Testing
A unit test is included in this solution that verifies that the getter works appropriately. For manual tests, you can create an instance of `MimeTypes` and notice `getAudioTypesOnly`

## Reviewing 

## Submitter Checklist

- [x] I have considered adding accessibility improvements for my changes.


develric (approved):
Hey @jd-alexander , thanks for this PR. LGTM 👍 !

PR #1824, opened 2021-01-07
ScanStore: Fix/ Ignore Threat

ashiagr:
Parent wordpress-mobile/WordPress-Android#13326

This PR adds following endpoints [Internal Ref: pcdRpT-3J (Jetpack Scan API Audit)]

1. **Fix Threats** POST `/sites/$site/alerts/fix?threat_ids[1]=<id1>&threat_ids[2]=<id2>`
2. **Ignore Threat** POST `/sites/$site/alerts/$threat_id?ignore=true`
3. **Fetch Fix Threats Status** GET `/sites/$site/alerts/fix?threat_ids[1]=<id1>&threat_ids[2]=<id2>`

and adds corresponding `ScanAction`, models and methods in the `ScanRestClient` and `ScanThreatStore`.

To test

That newly added tests in `ScanRestClientTest`, `ScanStoreTests` run fine.

Notes

1. "Fix Threats" POST request is asynchronous. To check on the status of the fix, GET request is provided separately. 
2. To fix a single threat, we can pass corresponding threat id in the list to the "Fix Threats" API.
3. "Fix Threat Status" response is returned directly and not stored into the db. We'll consider adding to db if a need arises during implementation in the client app.
4. Connected tests not included yet (added to back log) as real threat ids from the test site might become invalid during internal testing. Will figure out a way those threats are not updated and then include tests.
5. For **Ignore Threat**, currently just checking if the request has succeeded and not worrying about the data in the response as it contains site's rewind site state data which doesn't seem to be useful, documented in calypso code [here](https://opengrok.a8c.com/source/xref/calypso/client/state/data-layer/wpcom/sites/alerts/ignore.js?r=bbad57e4#41) and commented in the Scan API Audit p2 post here (pcdRpT-3J#comment-201).

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
@@ -35,9 +35,11 @@ import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Re
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Success
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
 import org.wordpress.android.fluxc.network.rest.wpcom.scan.threat.FixThreatsResponse
+import org.wordpress.android.fluxc.network.rest.wpcom.scan.threat.FixThreatsStatusResponse
 import org.wordpress.android.fluxc.network.rest.wpcom.scan.threat.Threat
 import org.wordpress.android.fluxc.store.ScanStore.FetchedScanStatePayload
 import org.wordpress.android.fluxc.store.ScanStore.FixThreatsErrorType
+import org.wordpress.android.fluxc.store.ScanStore.FixThreatsStatusErrorType
 import org.wordpress.android.fluxc.store.ScanStore.IgnoreThreatErrorType
 import org.wordpress.android.fluxc.store.ScanStore.ScanStartErrorType
 import org.wordpress.android.fluxc.store.ScanStore.ScanStateErrorType
@@ -360,6 +362,31 @@ class ScanRestClientTest {
         }
     }
 
+    @Test
+    fun `fetch fix threats status dispatches response on success`() = test {
+        initFetchFixThreatsStatus(FixThreatsStatusResponse(ok = true, fixThreatsStatus = listOf()))
+
+        val payload = scanRestClient.fetchFixThreatsStatus(siteId, listOf(threatId))
+
+        with(payload) {
+            assertEquals(remoteSiteId, this@ScanRestClientTest.siteId)
+            assertNull(error)
+        }
+    }
+
+    @Test
+    fun `fetch fix threats status dispatches api error on failure from api`() = test {
+        initFetchFixThreatsStatus(FixThreatsStatusResponse(ok = false, fixThreatsStatus = null))
+
+        val payload = scanRestClient.fetchFixThreatsStatus(siteId, listOf(threatId))
+
+        with(payload) {
+            assertEquals(remoteSiteId, this@ScanRestClientTest.siteId)
+            assertTrue(isError)
+            assertEquals(FixThreatsStatusErrorType.API_ERROR, error.type)
+        }
+    }
+
     private fun assertEmittedScanStateError(payload: FetchedScanStatePayload, errorType: ScanStateErrorType) {
         with(payload) {
             assertEquals(site, this@ScanRestClientTest.site)
@@ -443,7 +470,7 @@ class ScanRestClientTest {
     private suspend fun initIgnoreThreat(
         error: WPComGsonNetworkError? = null
     ): Response<Any> {
-        val response = if (error != null) Response.Error(error) else Success(data = Any())
+        val response = if (error != null) Response.Error(error) else Success(Any())
         whenever(
             wpComGsonRequestBuilder.syncPostRequest(
                 eq(scanRestClient),
@@ -451,6 +478,26 @@ class ScanRestClientTest {
                 anyOrNull(),
                 anyOrNull(),
                 eq(Any::class.java)
+            )
+        ).thenReturn(response)
+        return response
+    }
+
+    private suspend fun initFetchFixThreatsStatus(
+        data: FixThreatsStatusResponse? = null,
+        error: WPComGsonNetworkError? = null
+    ): Response<FixThreatsStatusResponse> {
+        val nonNullData = data ?: mock()
+        val response = if (error != null) Response.Error(error) else Success(nonNullData)
+        whenever(
+            wpComGsonRequestBuilder.syncGetRequest(
+                eq(scanRestClient),
+                urlCaptor.capture(),
+                anyOrNull(),
+                eq(FixThreatsStatusResponse::class.java),
+                eq(false),
+                any(),
+                eq(false)
             )
         ).thenReturn(response)
         return response
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
@@ -34,8 +34,11 @@ import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Success
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import org.wordpress.android.fluxc.network.rest.wpcom.scan.threat.FixThreatsResponse
 import org.wordpress.android.fluxc.network.rest.wpcom.scan.threat.Threat
 import org.wordpress.android.fluxc.store.ScanStore.FetchedScanStatePayload
+import org.wordpress.android.fluxc.store.ScanStore.FixThreatsErrorType
+import org.wordpress.android.fluxc.store.ScanStore.IgnoreThreatErrorType
 import org.wordpress.android.fluxc.store.ScanStore.ScanStartErrorType
 import org.wordpress.android.fluxc.store.ScanStore.ScanStateErrorType
 import org.wordpress.android.fluxc.test
@@ -53,6 +56,7 @@ class ScanRestClientTest {
     private lateinit var urlCaptor: KArgumentCaptor<String>
     private lateinit var scanRestClient: ScanRestClient
     private val siteId: Long = 12
+    private val threatId: Long = 1
 
     @Before
     fun setUp() {
@@ -304,6 +308,58 @@ class ScanRestClientTest {
         }
     }
 
+    @Test
+    fun `fix threats dispatches response on success`() = test {
+        val fixThreatsResponse = FixThreatsResponse(ok = true)
+        initFixThreats(fixThreatsResponse)
+
+        val payload = scanRestClient.fixThreats(siteId, listOf(threatId))
+
+        with(payload) {
+            assertEquals(remoteSiteId, this@ScanRestClientTest.siteId)
+            assertNull(error)
+        }
+    }
+
+    @Test
+    fun `fix threats dispatches api error on failure from api`() = test {
+        val fixThreatsResponse = FixThreatsResponse(ok = false)
+        initFixThreats(fixThreatsResponse)
+
+        val payload = scanRestClient.fixThreats(siteId, listOf(threatId))
+
+        with(payload) {
+            assertEquals(siteId, this@ScanRestClientTest.siteId)
+            assertTrue(isError)
+            assertEquals(FixThreatsErrorType.API_ERROR, error.type)
+        }
+    }
+
+    @Test
+    fun `ignore threat dispatches response on success`() = test {
+        initIgnoreThreat()
+
+        val payload = scanRestClient.ignoreThreat(siteId, threatId)
+
+        with(payload) {
+            assertEquals(remoteSiteId, this@ScanRestClientTest.siteId)
+            assertNull(error)
+        }
+    }
+
+    @Test
+    fun `ignore threats dispatches generic error on failure`() = test {
+        initIgnoreThreat(error = WPComGsonNetworkError(BaseNetworkError(NETWORK_ERROR)))
+
+        val payload = scanRestClient.ignoreThreat(siteId, threatId)
+
+        with(payload) {
+            assertEquals(remoteSiteId, this@ScanRestClientTest.siteId)
+            assertTrue(isError)
+            assertEquals(IgnoreThreatErrorType.GENERIC_ERROR, error.type)
+        }
+    }
+
     private fun assertEmittedScanStateError(payload: FetchedScanStatePayload, errorType: ScanStateErrorType) {
         with(payload) {
             assertEquals(site, this@ScanRestClientTest.site)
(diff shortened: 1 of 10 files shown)


example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
```diff
@@ -304,6 +310,83 @@ class ScanRestClientTest {
...
 
+    @Test
+    fun `fix threats dispatches response on success`() = test {
+        val fixThreatsResponse = FixThreatsResponse(ok = true)
+        initFixThreats(fixThreatsResponse)
+
+        val payload = scanRestClient.fixThreats(siteId, listOf(threatId))
+
+        with(payload) {
+            assertEquals(remoteSiteId, this@ScanRestClientTest.siteId)
```
ParaskP7: Minor (🔍): The `this@ScanRestClientTest` is optional here since `siteId` is different to `remoteSiteId`. Consider removing it unless you might want to rename `remoteSiteId` to `siteId` by the end of this PR.

PS: The same applies on all the newly added test.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
```diff
@@ -304,6 +310,83 @@ class ScanRestClientTest {
...
+            assertNull(error)
+        }
+    }
+
+    @Test
+    fun `fix threats dispatches api error on failure from api`() = test {
```
ParaskP7, line 327: 1) Question (❓): Is `ok = false` considered an API error?
2) Question (❓): Is the actual API error test missing, just like it exist for the `ignore threat` scenario?

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
```diff
@@ -128,4 +143,91 @@ class ScanStoreTest {
...
+        whenever(scanRestClient.fixThreats(siteId, threatIds)).thenReturn(
+            FixThreatsResultPayload(siteId)
+        )
+
+        val action = ScanActionBuilder.newFixThreatsAction(payload)
+        scanStore.onAction(action)
+
+        verify(scanRestClient).fixThreats(siteId, threatIds)
+    }
+
```
ParaskP7, line 162: 1) Question (❓): Is the success test missing here?
2) Question (❓): `DISCLAIMER: This is a food for thought question.` Since what is emitted is being asserted (via verification), for both, the success and error cases, maybe verifying that the rest client is being triggered is obsolete and these kind of test be omitted altogether in favor of having only the emitted related tests, which always include the triggering of the rest client. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/FixThreatsStatusDeserializer.kt
```diff
@@ -0,0 +1,51 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.scan.threat
+
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
+import com.google.gson.JsonElement
+import com.google.gson.JsonObject
+import org.wordpress.android.fluxc.network.rest.wpcom.scan.threat.FixThreatsStatusResponse.FixThreatStatus
+import java.lang.reflect.Type
+
+class FixThreatsStatusDeserializer : JsonDeserializer<List<FixThreatStatus>?> {
```
ParaskP7, line 10: 1) Praise (❤️): Really like the KDoc, which clearly shows the input and output of the deserializer.
2) Question (❓): Wouldn't an accompanying test or set of test for the deserializer, explicitly or implicitly, whichever is easier, make sure that everything works as expected and any future changes don't break functionality? For example, by transforming the input json to the output response.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/FixThreatsStatusResponse.kt
```diff
@@ -0,0 +1,17 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.scan.threat
+
+import com.google.gson.annotations.JsonAdapter
+import com.google.gson.annotations.SerializedName
+import org.wordpress.android.fluxc.network.Response
+
+data class FixThreatsStatusResponse(
+    @SerializedName("ok") val ok: Boolean?,
+    @SerializedName("threats") @JsonAdapter(FixThreatsStatusDeserializer::class)
+    val fixThreatsStatus: List<FixThreatStatus>?
```
ParaskP7: Minor (🔍): I am seeing both styles being used in our codebase. IMHO the below might be more readable, wdyt?

```
    @JsonAdapter(FixThreatsStatusDeserializer::class)
    @SerializedName("threats") val fixThreatsStatus: List<FixThreatStatus>?
```

With the above style the only difference is the extra line on top of the usual serialized field, without changing anything else.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
```diff
@@ -82,8 +95,86 @@ class ScanRestClient(
...
+                    FixThreatsStatusErrorType.INVALID_RESPONSE,
+                    FixThreatsStatusErrorType.AUTHORIZATION_REQUIRED
+                )
+                val error = FixThreatsStatusError(errorType, response.error.message)
+                FetchFixThreatsStatusResultPayload(remoteSiteId = remoteSiteId, error = error)
+            }
+        }
+    }
+
+    // Build Payloads
```
ParaskP7: Minor (🔍): Probably this is a left-over comment that can be removed. Please ignore if I am wrong.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
```diff
@@ -145,6 +236,28 @@ class ScanRestClient(
...
 
-    private fun buildErrorPayload(site: SiteModel, errorType: ScanStateErrorType) =
+    private fun buildFixThreatsStatusPayload(
+        response: FixThreatsStatusResponse,
+        remoteSiteId: Long
+    ): FetchFixThreatsStatusResultPayload {
+        var error: FixThreatsStatusErrorType? = null
+        val fixThreatStatusModels = response.fixThreatsStatus?.mapNotNull {
+            if (it.id == null) {
+                error = FixThreatsStatusErrorType.MISSING_THREAT_ID
```
ParaskP7, line 245: Question (❓): If we have 10 threats being fixed and on of them errors, then a generic error is shown and the rest of the threads are ignored, I am understanding this right, is this the intended behavior?

fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
```diff
@@ -86,6 +105,50 @@ class ScanStore @Inject constructor(
...
+    private fun emitIgnoreThreatResult(payload: IgnoreThreatResultPayload): OnIgnoreThreatStarted {
+        return if (payload.error != null) {
+            OnIgnoreThreatStarted(payload.error, IGNORE_THREAT)
+        } else {
+            OnIgnoreThreatStarted(IGNORE_THREAT)
+        }
+    }
+
+    suspend fun fetchFixThreatsStatus(payload: FetchFixThreatsStatusPayload): OnFixThreatsStatusFetched {
+        val resultPayload = scanRestClient.fetchFixThreatsStatus(
```
ParaskP7: Minor (🔍): This can become a one-liner, just how it is done on the above function.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
```diff
@@ -124,6 +213,39 @@ class ScanStore @Inject constructor(
         }
     }
 
+    class FixThreatsPayload(val remoteSiteId: Long, val threatIds: List<Long>) : Payload<BaseNetworkError>()
```
ParaskP7, line 213: Question (❓): Where should the `removeSiteId` be used in favor of `sideId`. For example, if I were to implement this I would have used `sideId` instead, but probably I would be wrong. Would you mind explaining the difference?

ParaskP7 (approved):
👋 @ashiagr !

I have tested the solution and it works as expected. Kudos on splitting the commits, it made it easy to review. 🌟 

I will be approving this PR since all my comments are questions with some minors, which most of the more important minors are test related. As such I am giving you some time to review my comment, but, feel free to merge the PR if you want.

PS: If it was me I would probably split this PR into 3 separate PRs (`fix threats`, `ignore threat` and `fetch fix threats status`) and then have them being reviewed individually. However, I am not sure if that would be too much in this case. Would you mind sharing your opinion so I become more aligned with you and the team in general.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ashiagr: Removed in fc38dc5

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
ashiagr: You're right :). We can remove such tests. Thank you!
Removed in a4b146a.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/FixThreatsStatusResponse.kt
ashiagr: In 1c38192

fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
ashiagr: Site can have a local as well as a remote site id. I'm not very sure how local site ids are generated but since we're not saving site's threat statuses into db so considered using a remote site id here.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ashiagr: Yep, this is the intended behaviour. It replicates similar behaviour in fetch activity logs [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ae96bc1d4958c87cbced799df33741138a0457ba/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt#L260). 

On second thought, we can consider sending threat status models with valid status along with the error. I can revisit this based on implementation in the client app.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
ashiagr: In 3c0934b

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/FixThreatsStatusDeserializer.kt
ashiagr: > Wouldn't an accompanying test or set of test for the deserializer

Implicitly these deserialisers are hit in the connected tests where real data is used. But in this case, I have deferred adding them as mentioned in the description.  I've added a task in the back log. I can also add explicit tests for deserialisers at that time.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ashiagr: >  Is ok = false considered an API error?

Good point. I tried to get an error by passing an invalid `threat_id` and got this in the response

```
{
	"ok": true,
	"threats": {
		"1": {
			"error": "not_found"
		}
	}
}
```
It returned ` ok = true` in the response even for the `not found` error.
I'm not sure if `error` (instead of `threats`) will be returned along with `ok` when it is set to `false`, so didn't include it in the response or add any test for it.

As a general convention, I see `ok` when set to `false` is accompanied with an `error` and reported as an `API Error`
Example: In activity log [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClient.kt#L120). 
 
But I am not 100% sure here, I'll add a note in the back log to check this further.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ashiagr: Removed in 174aaae

ashiagr:
Thanks you for the review @ParaskP7 🙏! Really appreciate your time.

I've addressed your review comments. Let me know if anything is not clear.

> PS: If it was me I would probably split this PR into 3 separate PRs (fix threats, ignore threat and fetch fix threats status) and then have them being reviewed individually. However, I am not sure if that would be too much in this case. Would you mind sharing your opinion so I become more aligned with you and the team in general. 

Since these endpoints were closely related with similar kind of tests, and boilerplate code, I combined them into single PR.  
I tried to split them into separate commits, but I realise now it must have been too much for a single PR. Apologies if it caused you troubles in the review.  

Ready for another look 🙇‍♀️.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClientTest.kt
ParaskP7: Oh cool, thanks so much for this clarification Ashita, it makes sense then! Although, the path to the error and what is being done there might never be trigger or test, which is what threw me away in the first place.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/threat/FixThreatsStatusDeserializer.kt
ParaskP7: Great, thank you for the clarification Ashita, makes sense!

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/scan/ScanRestClient.kt
ParaskP7: Thanks for verifying that for me Ashita! 🌟 For me, and for now, it is okay to just be aware of that and keep this thought while progress further with this feature.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt
ParaskP7: Makes sense, thanks Ashita, this helps! 🌟

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
ParaskP7: Thanks for removing those tests @ashiagr ! 🌟 

However, what I was trying to say is that because of these tests the `success` path is kind of hidden, thus by removing those we need to add the equivalent amount of `failure` tests to cover the `success` path. Wdyt?

ParaskP7:
> I've addressed your review comments. Let me know if anything is not clear.

Thank you for addressing the comments @ashiagr ! I have only one comment left to discuss with you, the I resolved the rest, please let me know what you think. 🙏 

> Since these endpoints were closely related with similar kind of tests, and boilerplate code, I combined them into single PR.
I tried to split them into separate commits, but I realise now it must have been too much for a single PR. Apologies if it caused you troubles in the review.

No worries, it didn't cause me any trouble. I was just saying what I would have done in order to try to find out the right balance as a whole (or mainly for me 😅). The pros of doing a split would have been quicker feedback from more individuals, one per PRs. But the cons would have been that you might ended up with some merge conflicts depending on the opening strategy of those multiple PRs, so ⚖️ !

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
ashiagr: I see, as discussed converted triggers tests to emitted related tests here: a0a4e71

ParaskP7: Thank you so much @ashiagr ! I am now going to approve the PR, feel free to merge it when the completes.

ParaskP7 (approved):

PR #1834, opened 2021-01-12
Jetpack Section: Backup Screen - Introduce Rewindable Only Flag

ParaskP7:
Parent [WordPress-Android#13629](https://github.com/wordpress-mobile/WordPress-Android/issues/13629)

This PR introduces the `rewindableOnly` flag to the `getActivityLogForSite` function. When this flag is passed as `true`, then the only activities that are being retrieved from the database are the rewindable ones. Else, if this flag is passed as `false`, then all the activities are being retrieved (as previously).

This solution is required to build the `Backups` screen. 

To test:
- Trigger the `ReleaseStack_ActivityLogTestJetpack` connected test suite and verify that all tests pass.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
@@ -150,9 +150,9 @@ class ReleaseStack_ActivityLogTestJetpack : ReleaseStack_Base() {
 
         this.mCountDownLatch = CountDownLatch(1)
 
-        val firstActivity = activityLogModel(1, true)
-        val secondActivity = activityLogModel(2, true)
-        val thirdActivity = activityLogModel(3, true)
+        val firstActivity = activityLogModel(index = 1, rewindable = true)
+        val secondActivity = activityLogModel(index = 2, rewindable = true)
+        val thirdActivity = activityLogModel(index = 3, rewindable = true)
         val activityModels = listOf(firstActivity, secondActivity, thirdActivity)
 
         activityLogSqlUtils.insertOrUpdateActivities(site, activityModels)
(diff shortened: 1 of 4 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
```diff
@@ -128,9 +150,9 @@ class ReleaseStack_ActivityLogTestJetpack : ReleaseStack_Base() {
 
         this.mCountDownLatch = CountDownLatch(1)
 
-        val firstActivity = activityLogModel(1)
-        val secondActivity = activityLogModel(2)
-        val thirdActivity = activityLogModel(3)
+        val firstActivity = activityLogModel(1, true)
```
malinajirka: 🔍 Nitpick: We usually add explicit parameter names to boolean parameters.
```suggestion
        val firstActivity = activityLogModel(1, rewindable  = true)
```

malinajirka (approved):
LGTM, great job! ;) Feel free to merge it without changes.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
ParaskP7: This makes sense, thanks for pointing that out @malinajirka ! 🌟 

I am also on the fence here, not sure which is better. Having that as such for tests and using the IDE to see the parameter name sometimes is enough, but this is very subjective. For production code I would also be more strict with that. See below:

image

In that case, and since the `index` parameter is something that would benefit from the explicit parameter names I would convert them both or non, just to have that consistency in place. How do you feel about that?

malinajirka: > Having that as such for tests and using the IDE to see the parameter name sometimes is enough

Some devs have these IDE parameters disabled. I'm not sure if they can be enabled if you are using eg. vim. Moreover, they are not visible on GitHub and a lot of us do most of the reviews here. So imo we shouldn't rely on this behavior of the default AS's editor. Wdyt?

> In that case, and since the index parameter is something that would benefit from the explicit parameter names I would convert them both or non, just to have that consistency in place. How do you feel about that?

I agree, I'd change both.

ParaskP7: > Some devs have these IDE parameters disabled. I'm not sure if they can be enabled if you are using eg. vim. Moreover, they are not visible on GitHub and a lot of us do most of the reviews here. So imo we shouldn't rely on this behavior of the default AS's editor. Wdyt?

True that, I was basing a lot of what I said on my development setup. 👍 

> I agree, I'd change both.

Great, will do this change asap and let you know! 🙏

ParaskP7: 👋 @malinajirka !

This is now done: a474430b889c9b225be57d6f3ff61ae4c8ffd3d6

malinajirka:
Thanks, LGTM! ;)

PR #1835, opened 2021-01-13
Remove "Create a new Page" and "Customize your site" from Quick Start

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/WordPress-Android/issues/13713

Can be tested with: https://github.com/wordpress-mobile/WordPress-Android/pull/13756

📓 I left the references to the tour steps in place for now because I wasn't sure if we would want to incorporate them somewhere else or not. If we don't think they're worth keeping around, I can open a more extensive clean-out PR against develop.

### To test:
- Navigate to the Create WordPress.com site
    - Choose Site > ➕ > Create WordPress.com site
- Complete the site creation flow
- Choose "Yes, Help Me" on the "Want a little help..." dialog
- **Expect** To no longer see the options for "Create a new Page" or "Customize your site"

## Screenshots
Before | After
--- | ---
<kbd><a href="https://user-images.githubusercontent.com/3384451/103805173-39468d80-5021-11eb-98cf-5c13127b30b9.jpg"></a></kbd> |  <kbd><a href="https://user-images.githubusercontent.com/3384451/104388919-4f17ef00-5508-11eb-9a78-da86a3f53ecd.jpg"></a></kbd>

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/QuickStartStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/QuickStartStore.kt
@@ -7,6 +7,7 @@ import org.wordpress.android.fluxc.annotations.action.Action
 import org.wordpress.android.fluxc.persistence.QuickStartSqlUtils
 import org.wordpress.android.fluxc.store.QuickStartStore.QuickStartTaskType.CUSTOMIZE
 import org.wordpress.android.fluxc.store.QuickStartStore.QuickStartTaskType.GROW
+import org.wordpress.android.fluxc.store.QuickStartStore.QuickStartTaskType.UNKNOWN
 import org.wordpress.android.util.AppLog
 import javax.inject.Inject
 import javax.inject.Singleton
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/QuickStartStore.kt
```diff
@@ -23,9 +24,11 @@ constructor(private val quickStartSqlUtils: QuickStartSqlUtils, dispatcher: Disp
...
         UPLOAD_SITE_ICON("upload_site_icon", CUSTOMIZE, 2),
         CHOOSE_THEME("choose_theme", CUSTOMIZE, 3),
-        CUSTOMIZE_SITE("customize_site", CUSTOMIZE, 4),
-        CREATE_NEW_PAGE("create_new_page", CUSTOMIZE, 5),
-        VIEW_SITE("view_site", CUSTOMIZE, 6),
+        // Disabeling until a long term plan is made https://github.com/wordpress-mobile/WordPress-Android/issues/13713
+        CUSTOMIZE_SITE("customize_site", UNKNOWN, 0),
+        // Disabeling until a long term plan is made https://github.com/wordpress-mobile/WordPress-Android/issues/13713
+        CREATE_NEW_PAGE("create_new_page", UNKNOWN, 1),
+        VIEW_SITE("view_site", CUSTOMIZE, 4),
```
mkevins, line 31: Not requesting a change, just curious why the order for `VIEW_SITE` was changed (i.e. could it have remained 6, I just want to make sure I'm not missing anything).

mkevins (approved):
Tested via steps described here: https://github.com/wordpress-mobile/WordPress-Android/pull/13756#issue-553837877. Code looks good. I only had one question, but it's not a blocker. Nice work! :+1:

fluxc/src/main/java/org/wordpress/android/fluxc/store/QuickStartStore.kt
chipsnyder: It probably could have remained 6. I'm not sure what would happen in that case. Before the change in the `CUSTOMIZE` group, `CUSTOMIZE_SITE` and `CREATE_NEW_PAGE` were 4 and 5. But since I removed those from that group I didn't want to leave a gap in the ordering.

mkevins: Thanks for the explanation :+1:

PR #1836, opened 2021-01-13
Refactor ScanStateErrors

malinajirka:
This PR
1. Replace MISSING_THREAT_ID, MISSING_THREAT_SIGNATURE and MISSING_THREAT_FIRST_DETECTED with INVALID_RESPONSE. `Error message` is used to help us differentiate between these error cases.
2. Extracts "map Threat to Threat Model" into its own method so it can be easily re-used for "Scan History" endpoint which is coming in the next PR.

Note: Review by commits might be easier to follow.
Note2: I just realized the branch name is misleading, sorry. 

To Test:
CI will run unit tests and that should be enough


ashiagr (approved):
LGTM 👍. Thanks for this refactoring 🙏.

PR #1841, opened 2021-01-14
Issue/13326 add scan history endpoint 2

malinajirka:
Parent issue https://github.com/wordpress-mobile/WordPress-Android/issues/13326

This PR introduced "scan/history" endpoint which will be used on Scan History screen in WPAndroid.

- adds fetchScanHistory action: Action which fetches scan history from the server and stores it into ThreatModelTable. We decided to store items from both /scan and /scan/history into the same table. All ThreatModels returned by /scan endpoint have "ThreatStatus == current" and all ThreatModels returned by /scan/history endpoint have "ThreatStatus == fixed|ignored".
- adds getScanHistoryForSite method: Used to retrieve ThreatModels for scan history screen from the local db.
- splits replaceThreatsForSite into two methods (remove and insert): I decided to split this method into two methods for two reasons. Firstly, replace evoked to me that both operations happen within a single transaction, but they were not. Secondly, I wanted to add "statuses" parameter which felt quite confusing when used with "replace" method and required me to introduce a business logic into the db layer. The business logic is now placed [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/31052c17b0137711197ce78b48b22d8c0a9538ce/fluxc/src/main/java/org/wordpress/android/fluxc/store/ScanStore.kt#L189).

To test:
Running connected tests should be enough.


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ScanTestJetpack.kt
```diff
@@ -159,6 +159,18 @@ class ReleaseStack_ScanTestJetpack : ReleaseStack_Base() {
         assertNull(scanStartResultPayload.error)
     }
 
+    @Test
+    fun fetchScanHistory() {
+        val site = authenticate(CompleteJetpackSite)
```
ashiagr: Idea (💡): How about we use `ScanDailyJetpackSite` here, site creds already exist in the `test.properties`? `CompleteJetpackSite` currently returns zero scan history records and might not cover any error while storing threats to db.

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
```diff
@@ -50,7 +60,18 @@ class ScanStoreTest {
     @Mock private lateinit var threatSqlUtils: ThreatSqlUtils
     @Mock private lateinit var dispatcher: Dispatcher
     @Mock private lateinit var siteModel: SiteModel
-    @Mock private lateinit var threat: ThreatModel
+    @Mock private lateinit var buildConfigWrapper: BuildConfigWrapper
+    private val threat: ThreatModel = GenericThreatModel(
```
ashiagr: Idea (💡): How about we rename `threat` to `threatInCurrentStatus`? Since we now include tests that return it based on the threat status, it might be easier to understand why this threat is being returned at the test level.

ashiagr (approved):
Thanks @malinajirka! Everything looks great, good job 👍. Just left two ideas, feel free to merge after considering them.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ScanTestJetpack.kt
malinajirka: Nice catch! I meant to do that and I forgot about it. Fixed in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1841/commits/ce8026c42dde60f1a8528c0aee51534216a8f955

malinajirka: I also updated the test so it verifies that the data get saved into the db in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1841/commits/e3e37358f5be4ee1f88c5b539524943990c2ea2b

example/src/test/java/org/wordpress/android/fluxc/store/ScanStoreTest.kt
malinajirka: Good idea! I introduced threatInCurrentStatus and threatInFixedStatus fields in e6a47153b6c91c262c8d0e683ddcebbc83bf38a4

malinajirka:
Thanks for the review @ashiagr! It's ready for another round ;)!

ashiagr (approved):
Thanks for the changes @malinajirka! All looks good to me 👍.
Feel free to merge it after CI is done.

PR #1847, opened 2021-01-20
Fix duplicated Top Performers products

ThomazFB:
Summary
==========
Fix for the https://github.com/woocommerce/woocommerce-android/issues/3460. Sometimes the Top Performers view at the `My Store` section shows the same Top Performer product twice.

How to Test
==========
It still not clear when the site returns duplicated product information from the Top Performers endpoints, so there's no reliable way to reproduce that, so this step by step is just a suggestion about how normally this issue have been reproduced
1. Go to the Woo app at the `My Store` section and make sure the `Today` tab is selected (preferably make sure that no order was created)
2. Go to the Woo site and make a single order with a single product
3. Go back to the Woo app and do a scroll refresh
4. Verify if the same product shows up twice

Screenshot
==========
screenshot-1610706461726

Update release notes:


anitaa1990:
Hey @ThomazFB 👋 I was able to reproduce this issue by installing a fresh version of the app (without any cache) and logging into a store. It looks like the API response returned is correct but the app stores the same product twice in `WCTopPerformerProductModel`. But when refreshing the issue no longer exists since we delete existing records before adding new ones.

#### API respons returns only 1 product
Screenshot 2021-01-21 at 10 29 08 AM

#### Local Db stores the same product twice
Screenshot 2021-01-21 at 10 30 12 AM


I'm not sure why the app is storing the same product twice, when there is no data in `WCTopPerformerProductModel` 🤔

ThomazFB:
Awesome finding, @anitaa1990! Thank you so much for that, I'll investigate using this scenario described by you and see what's causing this duplicated insert. 😄

ThomazFB:
Hey @anitaa1990, the PR is ready for a second run. This was a very tricky problem to detect but the issue was concentrated at the `WCProductLeaderboardsMapper.asProductList` method, this method does a merge between local and remote products to make sure we request only the products that don't exist inside already our local database. 

The main issue here was that when we asked for the remote products the `WCProductStore` already insert these new fetched Products into the local database, so this was causing the local list to include repeated products when we go fetch the database products right after that. Then the Mapper duplicates the Top Performers and inserts the same item twice. To solve this I only needed to invert the fetching order, first the local ones and after that the missing products that will be requested to the Site.

Again, thank you for detecting a reliable way to reproduce this issue as this was the main start point for me to pinpoint the central problem, anyway I'm leaving the `distinct` usage inside the `WCLeaderboardsStore` as the last filter before returning the data to the UI so we can be defensive from Site plugin bugs, do you think that makes sense?

anitaa1990 (commented):
Thanks for the fix @ThomazFB! That is indeed a tricky case. Would you mind adding a comment to the code where the inversion takes place so that we don't accidentally change the order again and re-introduce the same bug again 😄

ThomazFB:
Hey @anitaa1990! Thanks for your review, I've added the comments enforcing why we should keep the operation order right here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1847/commits/2d0b017470c79e89b826ac3b06667219d2fbb79b, let me know what you think!

anitaa1990 (approved):
Looks good :shipit:

PR #1849, opened 2021-01-22
Added methods to fetch product count + virtual product count for an order

anitaa1990:
This PR adds new methods to `ProductSqlUtils` to fetch the products count and the virtual products count for an order. We seem to be getting a few `SqliteBlobTooBigException` crashes when fetching `getProductsByRemoteIds` method in `WCProductStore`.

There are 2 places in the Woo app where this method is called:
- When trying to find if there are any virtual products for the order.
- When trying to find the total product count available locally for the order. 

Both of these scenarios require only the product count and not the entire product model list. So this PR adds methods to fetch just the product count for an order.

### To test
Run `ProductSqlUtilsTest.kt`.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
@@ -105,7 +105,7 @@ object ProductSqlUtils {
                 .count().toInt()
     }
 
-    fun getVirtualProductCountForOrder(site: SiteModel, remoteProductIds: List<Long>): Int {
+    fun getVirtualProductCountByRemoteIds(site: SiteModel, remoteProductIds: List<Long>): Int {
         return WellSql.select(WCProductModel::class.java)
                 .where().beginGroup()
                 .isIn(WCProductModelTable.REMOTE_PRODUCT_ID, remoteProductIds)
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
@@ -96,6 +96,25 @@ object ProductSqlUtils {
                 .asModel
     }
 
+    fun getProductCountByRemoteIds(site: SiteModel, remoteProductIds: List<Long>): Int {
+        return WellSql.select(WCProductModel::class.java)
+                .where().beginGroup()
+                .isIn(WCProductModelTable.REMOTE_PRODUCT_ID, remoteProductIds)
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -96,6 +96,25 @@ object ProductSqlUtils {
...
+    fun getProductCountByRemoteIds(site: SiteModel, remoteProductIds: List<Long>): Int {
+        return WellSql.select(WCProductModel::class.java)
+                .where().beginGroup()
+                .isIn(WCProductModelTable.REMOTE_PRODUCT_ID, remoteProductIds)
+                .equals(WCProductModelTable.LOCAL_SITE_ID, site.id)
+                .endGroup().endWhere()
+                .count().toInt()
+    }
+
+    fun getVirtualProductCountForOrder(site: SiteModel, remoteProductIds: List<Long>): Int {
```
0nko: This is a method that counts the virtual products in general, not just for an order, isn't it?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -681,6 +681,20 @@ class WCProductStore @Inject constructor(
...
+     * if it exists in the database
+     */
+    fun getProductCountByRemoteIds(site: SiteModel, remoteProductIds: List<Long>): Int =
+            ProductSqlUtils.getProductCountByRemoteIds(site, remoteProductIds)
+
+    /**
+     * returns the count of virtual products for the given [SiteModel] and [remoteProductIds]
+     * if it exists in the database
+     */
+    fun hasVirtualProductsOnly(site: SiteModel, remoteProductIds: List<Long>): Int =
```
0nko: The name of the methods indicates the result is a boolean. Did you mean to check if the count is 0?

0nko (commented):
Looks good, I just left a couple of questions in the code.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
anitaa1990: I had initially added logic here to return a boolean flag but thought it would be better to have that logic on the Woo side instead. Changed the method name in 50f7cc9

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
anitaa1990: True! It returns a list of products for the remoteProductIds. I modified the name of the method in 50f7cc9 👍

0nko (approved):
:shipit:

PR #1850, opened 2021-01-22
Added logic to parse products response as json object as well as json array

anitaa1990:
There are a bunch of crashes in the Woo app when parsing the products response. This is related to woocommerce/woocommerce-android#2374. There are a couple of different crashes happening when parsing:
- product categories: the app is expecting a json array but the response returned is null.
- cross sell ids: the app is expecting a json array but the response is either null or a json object.
- upsell ids: the app is expecting a json array but the response is either null or a json object.
- grouped products: the app is expecting a json array but the response is either null or a json object.
- variations: he app is expecting a json array but the response is either null or a json object.

This could be because there could be some plugins on the merchant's store that might be messing with the API response. This PR handles the above scenarios by ensuring that we can parse the response as a `jsonObject`, `jsonArray` or `jsonNull`.

### To test
- Run `MockedStack_WCProductsTest`.
- Smoke test the GET products feature in the example app to ensure that everything is working correctly.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
@@ -257,13 +257,19 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
         try {
             if (jsonString.isNotEmpty()) {
                 val jsonElement = Gson().fromJson(jsonString, JsonElement::class.java)
-                if (jsonElement.isJsonArray) {
-                    jsonElement.asJsonArray.forEach { jsonArray ->
-                        jsonArray.asLong.let { productIds.add(it) }
+                when {
+                    jsonElement.isJsonNull -> {
+                        return emptyList()
                     }
-                } else if (jsonElement.isJsonObject) {
-                    jsonElement.asJsonObject.entrySet().forEach {
-                        productIds.add(it.value.asLong)
+                    jsonElement.isJsonArray -> {
+                        jsonElement.asJsonArray.forEach { jsonArray ->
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -239,24 +239,38 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
...
-                Gson().fromJson(jsonString, JsonElement::class.java).asJsonArray.forEach { jsonElement ->
-                    jsonElement.asLong.let { productIds.add(it) }
+                val jsonElement = Gson().fromJson(jsonString, JsonElement::class.java)
+                when {
+                    jsonElement.isJsonNull -> {
+                        return emptyList()
+                    }
+                    jsonElement.isJsonArray -> {
+                        jsonElement.asJsonArray.forEach { jsonArray ->
+                            jsonArray.asLong.let { productIds.add(it) }
```
0nko, line 266: Could there possible be a case when the array has different types, like Strings?

0nko (commented):
Nice work! 👍 Just one question in the code.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
anitaa1990: All the crashes I've seen so far have indicated that the response is something like:
```
{
 "1": 11,
"2": 12,
"3": 13
}
```
So I've left it as a `Long` for now 👍

0nko (approved):
:shipit:

PR #1852, opened 2021-01-22
Only add search parameters when a searchQuery exists

### LINKED ISSUE #3470: Product and Order data loading issues when passing the search query parameter into a request
**Describe the bug**
Some merchants using our app are reporting of ocasional problems loading data from the site, this was caused by an issues between our app and the site where we request data passing the `search` parameter into the URL even when there's no search query to inform, causing some very specific sites to return.

**To Reproduce**
Sites affected by this aren't able to load Orders and Products properly so it's pretty easy to reproduce the problem,  the hard part is to be able to come up with a Woo Store site with the exact same configuration issue since it still not clear why only some sites behave like this. The main suspicion is plugin related problems.

ThomazFB:
Summary
==========
Fixes https://github.com/woocommerce/woocommerce-android/issues/3470. Some merchants using our app are reporting of ocasional problems loading data from the site, this was caused by an issues between our app and the site where we request data passing the `search` parameter into the URL even when there's no search query to inform, causing some very specific sites to return 

How to Test
==========
Sites affected by this aren't able to load Orders and Products properly so it's pretty easy to reproduce the problem, the hard part is to be able to come up with a Woo Store site with the exact same configuration issue since it still not clear why only some sites behave like this. The main suspicion is plugin related problems.

So in order to be able to properly verify the changes done here, the ideal test scenario would be to make sure using Charles Proxy that the Woo app requests doesn't include the search parameter for each of these operations:

1. Fetch Product list
2. Fetch Top Performers Products stats
3. Fetch Product tags
4. Fetch Order list


nbradbury (approved):
Looks great! :shipit:

PR #1853, opened 2021-01-25
Jetpack Section: Backup Download - Empty Fetching Backup Download

ParaskP7:
Parent [WordPress-Android#13329](https://github.com/wordpress-mobile/WordPress-Android/issues/13329)

This PR enhances the `fetchBackupDownloadState(...)` API call to take into account the empty response scenario for when the call succeeds. As such, this PR will make sure that the app will not crash when a `fetchBackupDownloadState(...)` call succeeds with an empty array. This scenario will happen when there is no backup download happening but the client queries for it.

This solution is required to build the `Backup Download` progress row for the `Activity Log` and `Backup` screens.

To test:
- Trigger the `fetchEmptyActivityBackupDownload_dispatchesResponseOnSuccess` test within the `ActivityLogRestClientTest` suite.

PS: As always, I suggest doing a commit-by-commit review as this will make it much easier for you to review the overall solution.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/activity/ActivityLogRestClientTest.kt
@@ -475,6 +475,19 @@ class ActivityLogRestClientTest {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


malinajirka (approved):
LGTM, thanks!

PR #1855, opened 2021-01-28
Adds `type` and `is_beta` parameters in FETCH_BLOCK_LAYOUTS action

antonis:
**Description**
This PR adds `type` and `is_beta` parameters in fetch block layouts call both for WPCom and self hosted site.
This make it so that the api will be able to return type ( mobile or web ) specific page layouts as well as beta versions.

**To test**
This can be tested with https://github.com/wordpress-mobile/WordPress-Android/pull/13905
or by running the `testFetchBlockLayouts()` and `testFetchBlockLayoutsBeta()` tests.

**Context**
This PR clones the [iOS implementation](https://github.com/wordpress-mobile/WordPress-iOS/pull/15678)

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestWPCom.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestWPCom.java
@@ -215,6 +215,15 @@ public class ReleaseStack_SiteTestWPCom extends ReleaseStack_Base {
 
     @Test
     public void testFetchBlockLayouts() throws InterruptedException {
+        testFetchBlockLayouts(false);
+    }
+
+    @Test
+    public void testFetchBlockLayoutsBeta() throws InterruptedException {
+        testFetchBlockLayouts(true);
+    }
+
+    private void testFetchBlockLayouts(boolean isBeta) throws InterruptedException {
         authenticateAndFetchSites(BuildConfig.TEST_WPCOM_USERNAME_TEST1,
                 BuildConfig.TEST_WPCOM_PASSWORD_TEST1);
         SiteModel firstSite = mSiteStore.getSites().get(0);
@@ -229,7 +238,7 @@ public class ReleaseStack_SiteTestWPCom extends ReleaseStack_Base {
         mNextEvent = TestEvents.BLOCK_LAYOUTS_FETCHED;
         mDispatcher.dispatch(SiteActionBuilder.newFetchBlockLayoutsAction(
                 new SiteStore.FetchBlockLayoutsPayload(firstSite, supportedBlocks,
-                        828.0f, 2.0f)));
+                        828.0f, 2.0f, isBeta)));
         mCountDownLatch = new CountDownLatch(1);
         assertTrue(mCountDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
     }
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestXMLRPC.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestXMLRPC.java
@@ -236,6 +236,15 @@ public class ReleaseStack_SiteTestXMLRPC extends ReleaseStack_Base {
 
     @Test
     public void testFetchBlockLayouts() throws InterruptedException {
+        testFetchBlockLayouts(false);
+    }
+
+    @Test
+    public void testFetchBlockLayoutsBeta() throws InterruptedException {
+        testFetchBlockLayouts(true);
+    }
+
+    private void testFetchBlockLayouts(boolean isBeta) throws InterruptedException {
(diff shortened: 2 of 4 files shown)


antonis:
Hey @enejb 👋 
I have some trouble testing this on Android since I get an empty result set when I add the `type` parameter.
Since on iOS it still works as expected it should be something obvious I'm doing wrong on the Android side. Please let me know if you see something weird in this PR or the server logs.

enejb:
I checked the endpoints and they do return the data as expected.
I wonder if some of the assumptions about the data are currently incorrect. 

For example now there are some layouts that have multiple categories.

antonis:
Hey @enejb 👋 

> I checked the endpoints and they do return the data as expected.
> I wonder if some of the assumptions about the data are currently incorrect.
> 
> For example now there are some layouts that have multiple categories.

Thank you for your reply. I checked the multiple categories assumption but it does not cause this.

I tried to replicate it with the `common-block-layouts` used in self hosted sites. For simplicity I also removed the `supported_blocks` parameter. What remains is:
`https://public-api.wordpress.com/wpcom/v2/common-block-layouts/?preview_width=536.0&scale=1.0&_locale=en_US&is_beta=true&type=mobile`
This url returns no result as I observed:
`{"layouts":[],"categories":[]}`

If I remove the `type` parameter the results are returned
`https://public-api.wordpress.com/wpcom/v2/common-block-layouts/?preview_width=536.0&scale=1.0&_locale=en_US&is_beta=true`

I tried to remove other parameters to validate my hypothesis and noticed that removing just the `_locale` parameter (and keeping the `type` did the trick:
`https://public-api.wordpress.com/wpcom/v2/common-block-layouts/?preview_width=536.0&scale=1.0&is_beta=true&type=mobile`

The `locale` parameter is added in [all Android calls by default](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/a097f94c384cc146c9a187144ea73eb1ab64c893/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java#L70) unless specified otherwise. This parameter has a [leading underscore](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/a097f94c384cc146c9a187144ea73eb1ab64c893/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java#L138) for `v2` calls.
Removing the `_US` from the `_locale` parameter value also does the trick:
`https://public-api.wordpress.com/wpcom/v2/common-block-layouts/?preview_width=536.0&scale=1.0&_locale=en&is_beta=true&type=mobile`

I [pushed a fix](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1855/commits/83f386795f1c7662c0c6d5de464c1834e1b67dc7) removing `_locale` parameter completely. This resolves the issue but is probably not correct since I see that the server side (`D55810-code`) considers the locale value.
An alternative client side fix would involve removing the country from the passed value (e.g. instead of `en_US` to send just `en`) but this would be an exception to all other Android calls.
We can also align with the Home Page Picker implementation and [use a `language` parameter](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/36f5116482a0af7427e3643ac6a0b6860254ba42/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/theme/ThemeRestClient.java#L163) instead of `_locale`. The later would require changes both in the clients (Android, iOS) and the server.
Wdyt?

cc @chipsnyder

enejb:
Thanks for the investigation. 

I think it is definitely something we need to fix on the back-end, I will work on this fix today. So that we can merge this PR early next week :)

enejb:
Once D56236-code is merged we should be able to revert your last commit and ship this :)

antonis:
Hey @enejb 👋 
Thank you for fixing this on the server side 🙇 

> Once D56236-code is merged we should be able to revert your last commit and ship this :)

I reverted the last commit and I'll get into testing this asap.

enejb:
I have shipped D56236-code but when I am testing this PR locally via https://github.com/wordpress-mobile/WordPress-Android/pull/13905

I still end up with a blank screen? 
Screen Shot 2021-02-01 at 11 20 47 AM

https://public-api.wordpress.com/wpcom/v2/common-block-layouts/?preview_width=536.0&scale=1.0&_locale=en_US&is_beta=true&type=mobile now does return results as expected. 

Any ideas? Maybe I am doing something wrong when I am testing this?

antonis:
Hello @enejb 👋 

Thank you for reviewing this again 🙇 

> I have shipped D56236-code but when I am testing this PR locally via [wordpress-mobile/WordPress-Android#13905](https://github.com/wordpress-mobile/WordPress-Android/pull/13905)

I retested this with the build from the above PR without a change and on debug mode and it works for me. Tried both a WP.com site and a Jurassic Ninja one 🤷 

> Any ideas? Maybe I am doing something wrong when I am testing this?

I'm not sure. There is a caching mechanism on mobile but the stored results are used only when the api call fails. If the problem persists we can check this later when we are both online.

enejb (approved):
@antonis I tested this using my phone by downloading the apk that was build and it worked all as expected. Lets ship this! :)

PR #1857, opened 2021-01-29
Woo store address

0nko:
This PR adds additional columns to the Woo settings table for the store address, which is required for shipping labels.

**To test:**
1. Run the example app & login
2. Go to Woo
3. Tap on the Log Woo Settings button
4. Notice the address, city, postal code, state and country are printed in the log


anitaa1990 (approved):
Looks good :shipit:

PR #1820, opened 2020-12-29
Add Attributes Persistence

ThomazFB:
Summary
==========
Finish implementation for the [Woo Issue 3204](https://github.com/woocommerce/woocommerce-android/issues/3204). This PR adds support to persist all retrieve, create, update and delete Attribute data operations coming from the site. Also adding unit test coverage

Screenshots & How to Test
==========
| Select the Woo section | Scroll down until the Attribute option becomes visible and click on it | Fetch, create, update and remove attributes
| ------------- | ------------- | ------------- |
| screenshot-1608543978049 | screenshot-1608543983921 | screenshot-1608535524554 |

To verify if the changes are working under the Woo site, go to Products > Attributes on your Woo site and make sure everything is as expected.

Screen Shot 2020-12-21 at 4 25 56 AM

PR #1858, opened 2021-02-02
Sync operations between Attributes and Variable Products

ThomazFB:
Summary
==========
Finish implementation for the [Woo Issue 3512](https://github.com/woocommerce/woocommerce-android/issues/3512). This PR adds support to edit/add/remove/update all attributes from *a given product and variation*, note that this is different from editing the Global Attributes from the entire Woo Site, which was introduced by [PR 1803](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1803) and [PR 1803](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1820). This PR also addresses a lot of noted bugs within the Attributes and Terms operations

Screenshots & How to Test
==========
| Select the Woo section | Select the Products Section | Attach or Detach Attributes from a Product
| Screenshot_1612377245 | Screenshot_1612377248 |Screenshot_1612376305 |

To verify if the changes are working under the Woo site, go to Products > All Products > Select the updated Product on the list and go to the Attributes section on your Woo site and make sure everything is as expected.

image

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductAttributeFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductAttributeFragment.kt
@@ -19,8 +19,8 @@ import org.wordpress.android.fluxc.example.ui.StoreSelectorDialog
 import org.wordpress.android.fluxc.example.utils.showSingleLineDialog
 import org.wordpress.android.fluxc.example.utils.toggleSiteDependentButtons
 import org.wordpress.android.fluxc.model.SiteModel
-import org.wordpress.android.fluxc.model.product.attributes.WCGlobalAttributeModel
-import org.wordpress.android.fluxc.model.product.attributes.terms.WCAttributeTermModel
+import org.wordpress.android.fluxc.model.attribute.WCGlobalAttributeModel
+import org.wordpress.android.fluxc.model.attribute.terms.WCAttributeTermModel
 import org.wordpress.android.fluxc.store.WCGlobalAttributeStore
 import javax.inject.Inject
 
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductAttributeFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductAttributeFragment.kt
@@ -56,6 +56,7 @@ class WooProductAttributeFragment : Fragment(), StoreSelectorDialog.Listener {
         delete_product_attributes.setOnClickListener(::onDeleteAttributeButtonClicked)
         update_product_attributes.setOnClickListener(::onUpdateAttributeButtonClicked)
         fetch_product_single_attribute.setOnClickListener(::onFetchAttributeButtonClicked)
+        fetch_term_for_attribute.setOnClickListener(::onFetchAttributeTermsButtonClicked)
         create_term_for_attribute.setOnClickListener(::onCreateAttributeTermButtonClicked)
     }
 
@@ -173,6 +174,33 @@ class WooProductAttributeFragment : Fragment(), StoreSelectorDialog.Listener {
         }
     }
 
+    private fun onFetchAttributeTermsButtonClicked(view: View) {
+        try {
+            showSingleLineDialog(
+                    activity,
+                    "Enter the attribute ID you want to fetch terms:"
+            ) { editText ->
+                coroutineScope.launch {
+                    takeAsyncRequestWithValidSite {
+                        wcAttributesStore.fetchAttributeTerms(
+                                it,
+                                editText.text.toString().toLongOrNull() ?: 0
+                        )
+                    }?.apply {
+                        model?.forEach {
+                            logSingleAttributeTermResponse(it) }
+                                ?.let { prependToLog("========== Attribute Terms Fetched =========") }
+                                ?: takeIf { isError }?.let {
+                                    prependToLog("Failed to fetch Terms. Error: ${error.message}")
+                                }
+                    } ?: prependToLog("Failed to fetch Terms. Error: Unknown")
+                }
+            }
+        } catch (ex: Exception) {
+            prependToLog("Couldn't create Attributes. Error: ${ex.message}")
+        }
+    }
+
     private fun onCreateAttributeTermButtonClicked(view: View) {
         try {
             showSingleLineDialog(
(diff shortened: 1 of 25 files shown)


ThomazFB:
Hey @nbradbury, this PR is ready for review! Just a heads up that this PR changes some Class names to reflect the naming convention from the site and make it easier to differ a Product Attribute from a Global Attribute, so this may affect your current code at the Woo site at least on the Class naming declaration level. 

The main part to be aware of is that all `WCProductAttribute` naming was changed to `WCGlobalAttribute`, meanwhile, a new type of model called `WCProductAttributeModel` was created.

nbradbury:
I should've noticed this in my previous review, but I see now that when we fetch the list of attributes, we fetch the terms for each attribute in a separate request. For a site that has a lot of attributes, this results in a lot of network requests (as illustrated by the screenshot below). This approach will likely need some re-thinking.

Screen Shot 2021-02-03 at 1 57 32 PM

ThomazFB:
Hey @nbradbury, thanks for this feedback! I've tried in early development to avoid this problem, but unfortunately I didn't found a way to fetch a full Terms list unbound from a specific Attribute as we do for Products and Global Attributes. As you can see in the [Woo Rest API for Attribute Terms](https://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-attribute-terms), the current endpoint to fetch terms is specific for a given Attribute ID set on the URL and I don't see a good alternative or available endpoint for a more general data request, so I'm not sure which options we have right now to avoid this.

ThomazFB:
I wonder maybe if iOS found a better way to work with this. Hey @Ecarrion, quick question! On the iOS side, when fetching all attributes terms, do the app request every terms list for each attribute on a separate request as described [here](https://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-attribute-terms)? Or there's a known endpoint to the app to do a batch fetching of all exiting terms from the store unbound from any attribute?

Thank you for your time!

nbradbury:
I'm not able to add a product attribute, and I'm definitely entering a valid productId, attributeId, and termId. The problem appears to be in `WCProductModel.addAttribute`, because the attribute passed to that method never gets added. So an empty attribute list is passed when the product is updated.

Screen Shot 2021-02-03 at 3 09 42 PM

ThomazFB:
Hey @nbradbury, thanks for catching this one, the issue was happening only when adding Attributes to an empty product attributes list, it's fixed now right here: 21021ca089f916b7d2740d4bee7c0b1fe2071827

Ecarrion:
> I wonder maybe if iOS found a better way to work with this. Hey @Ecarrion, quick question! On the iOS side, when fetching all attributes terms, do the app request every terms list for each attribute on a separate request as described [here](https://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-attribute-terms) or there's a known endpoint to the app to do a batch fetching of all exiting terms from the store unbound from any attribute?
> 
> Thank you for your time!

Hey @ThomazFB AFAIK that's the only endpoint we have to fetch the terms of a global attribute. 
I haven't started implemented that part yet but why do you need to fetch all terms beforehand? Is it possible to fetch the specific ones as soon as you enter the attribute options screen?

nbradbury:
> Hey @nbradbury, thanks for catching this one, the issue was happening only when adding Attributes to an empty product attributes list

'Fraid it's still not working for me. I can see that it's now correctly sending the added attribute:

request

But the response comes back with an empty attribute list:

response

ThomazFB:
> Hey @ThomazFB AFAIK that's the only endpoint we have to fetch the terms of a global attribute.
I haven't started implemented that part yet but why do you need to fetch all terms beforehand? Is it possible to fetch the specific ones as soon as you enter the attribute options screen?

Thanks for your input @Ecarrion! I decided to fetch all Global attribute terms beforehand as a very early decision trying to enforce data consistency and avoid storing the Attributes data incomplete. But there's no reason to do that if we end up making a lot of requests all at once, so thanks for your suggestion!

@nbradbury do you think it does make sense to change the Global Attribute flow to only fetch the terms when it's needed? This way we would lazily acquire the data and avoid mashing the Store with requests, but it would be necessary for the UI to request this data since the FluxC can't predict when it would be necessary to fetch it.

nbradbury:
> do you think it does make sense to change the Global Attribute flow to only fetch the terms when it's needed?

Yes, definitely. But probably in a separate PR.

nbradbury:
@ThomazFB I think I may have found why adding an attribute isn't working for me. Looking at the request body, I see that `options` is being sent twice, first as an empty array and then with the term.

options

But maybe you'd want me to take this over since you're supposed to be AFK?

nbradbury:
After deleting an attribute, I see this crash:

```
 android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
        at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:7753)
        at android.view.ViewRootImpl.requestLayout(ViewRootImpl.java:1225)
        at android.view.View.requestLayout(View.java:23093)
        at android.view.View.requestLayout(View.java:23093)
        at android.view.View.requestLayout(View.java:23093)
        at android.view.View.requestLayout(View.java:23093)
        at android.view.View.requestLayout(View.java:23093)
        at android.widget.TextView.checkForRelayout(TextView.java:8902)
        at android.widget.TextView.setText(TextView.java:5730)
        at android.widget.TextView.setText(TextView.java:5571)
        at android.widget.TextView.setText(TextView.java:5528)
        at org.wordpress.android.fluxc.example.MainExampleActivity.prependToLog(MainExampleActivity.kt:32)
        at org.wordpress.android.fluxc.example.FragmentExtensions.prependToLog(FragmentExt.kt:10)
        at org.wordpress.android.fluxc.example.ui.products.WooProductsFragment.logProduct(WooProductsFragment.kt:795)
```

ThomazFB:
Hey @nbradbury, thanks for all your inputs and feedback! I've submitted some fixes this morning, so it's worth verifying if all these problems persist.

f842ade078d02224f27bf26929d83f013b034936 address the JSON body issue, the serialization should be working just fine now.
5d294cbf439395d9079128e4b113fbad9a8a2663 and 4ead26021a767b45b4d9ed9d6feef153a9e2520b modify the Terms fetching logic to be done in a lazy fashion, I've also added a `withTerms` parameter on the `fetchAttribute` method so we can choose if we want to fetch a given attribute with or without the Terms
fdc6c17a3f7641cd57bb05af36c8159699d89d20 and 8c597337ce700ea9302562f0ef01c437f8dfde04 should address these main thread errors for the Example app.

Let me know how everything goes now!

PS: Sorry, I've only seen your suggestions about doing the terms fetching in a separate PR just now, if you want I can split these changes into another branch.

nbradbury (approved):
Everything appears to be working now! :shipit:

PR #1859, opened 2021-02-02
Introduce new FormattableContentType for Scan

malinajirka:
This PR implements mapping for "Scan" FormattableContentType. This type is already supported by the server. With the introduction of the Scan screen, we can introduce support for this type into the apps.

To test:
Test in [WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/pull/13941/commits)


ParaskP7 (approved):
👋 @malinajirka  !

I have reviewed and tested this PR as per the instruction, good job! 🌟 

I have only one suggestion (💡) for you: What if we added a new test within the `FormattableContentMapperTest` with the `Scan` type to cover this case? Maybe by just updating the `notificationBodyResponse` field within that test and adding the new `Scan` type? Btw, I feel that these json fields within this test should be moved to `.json` files, but this is a refactor for another time (I guess).

PS: Feel free to merge the PR yourself when you are ready.

malinajirka:
Thanks for the review! ;)

> What if we added a new test within the FormattableContentMapperTest with the Scan type to cover this case? Maybe by just updating the notificationBodyResponse field within that test and adding the new Scan type?

Added in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1859/commits/f61edee8267bcef09fc171d39a2dde36d77f32a4

> Btw, I feel that these json fields within this test should be moved to .json files, but this is a refactor for another time (I guess).

Good call, fixed in 74b61b5

ParaskP7 (approved):
Excellent job @malinajirka !

Big 👍 from me. ❤️

PR #1861, opened 2021-02-03
WCS account settings endpoint

hichamboushaba:
This PR is a requirement for Shipping Labels project, and needed right now for the ticket https://github.com/woocommerce/woocommerce-android/issues/3019.
It adds fetching the endpoint: `/wc/v1/connect/account/settings` to get the last used package id.

I also added an `id` field to the `PredefinedPackage` class, to expose the box id.

The endpoint returns also an object `storeOptions` which contains the different units used by the store, but as the same information si already returned by the settings/general and settings/products endpoints, I didn't expose it.

#### Testing

1. Run the example app & login
2. Go to Woo
3. Click on "Shipping Labels"
4. Click on "Get account settings"
5. Confirm the account settings are printed in the log.


0nko (approved):
:shipit:

PR #1863, opened 2021-02-04
Jetpack Section: Update Activity Log Page Size

ParaskP7:
Parent [WordPress-Android#13629](https://github.com/wordpress-mobile/WordPress-Android/issues/13629)

This PR updates activity log page size from `20` to `100`. This is done since the API enhancement is not ready yet. As such, and
in order to avoid some of the scrolling issues on the 'Backups' screen the activity log page size has been increase to 100 (x5 increase).

This PR will not be fixing the scrolling issues on the 'Backups' screen. It will just make them harder to appear. To avoid any
scrolling issue and API enhancement in required.

To test:
- Trigger the `ReleaseStack_ActivityLogTestJetpack` connected test suite and verify that all tests pass.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
@@ -164,7 +164,7 @@ class ActivityLogStoreTest {
         val rowsAffected = 1
         val activityModels = listOf<ActivityLogModel>(mock())
 
-        val action = initRestClient(activityModels, rowsAffected, totalItems = 100)
+        val action = initRestClient(activityModels, rowsAffected, totalItems = 500)
         whenever(activityLogSqlUtils.insertOrUpdateActivities(any(), any())).thenReturn(rowsAffected)
 
         activityLogStore.onAction(action)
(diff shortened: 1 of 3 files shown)


ParaskP7:
👋 @leandroalonso !

With this PR, Android will start using `100` as the activity log page size. I don't know if you need a similar change. If not, maybe it is worth considering doing on iOS as well. Wdyt?

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
```diff
@@ -164,7 +164,7 @@ class ActivityLogStoreTest {
         val rowsAffected = 1
         val activityModels = listOf<ActivityLogModel>(mock())
 
-        val action = initRestClient(activityModels, rowsAffected, totalItems = 100)
+        val action = initRestClient(activityModels, rowsAffected, totalItems = 500)
```
zwarm, line 167: @ParaskP7 - ❓ Just making sure `500` was intentional and not a typo. Other than that all is as expected. 👍

zwarm (approved):
Other than that one double check question before merging, all LGTM. :)

example/src/test/java/org/wordpress/android/fluxc/store/ActivityLogStoreTest.kt
ParaskP7: 👋 @zwarm !

Yes, `500` was intentional, since I increased the page size `x5` (from previous `20`), I increased this value `x5` as well (from previous `100`).

PR #1868, opened 2021-02-11
Dynamic cards store

planarvoid:
This PR moves the logic to keep the pinned/removed/hidden dynamic cards on per site level to the FluxC. This could be tested with the related WPAndroid PR. 
In this PR I'm creating a new table for the card state (removed/pinned) and the hidden cards are kept in a list so they'll restart on app restart. 

Test with this WPAndroid PR - https://github.com/wordpress-mobile/WordPress-Android/pull/14036


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/DynamicCardSqlUtils.kt
```diff
@@ -0,0 +1,100 @@
...
+        cardType: DynamicCardType
+    ) {
+        WellSql.insert(
+                DynamicCardBuilder(
+                        siteId = siteId,
+                        dynamicCardType = cardType.toString(),
+                        state = PINNED.toString()
+                )
+        ).execute()
+    }
```
renanferrari: Maybe we should check if there's already a pinned card here and then just update it, as I believe this would solve [the issue reported here](https://github.com/wordpress-mobile/WordPress-Android/pull/14036#pullrequestreview-590653912).

renanferrari (commented):
Thanks for the changes! The code looks pretty good, I just left one comment below that is related to the issue reported on the related WPAndroid PR.

planarvoid:
Good catch @renanferrari! I've implemented a slightly simpler version where I check whether there is a previously pinned card and if it exists, I unpin it. I've also added a similar check to the `remove` call so it can't happen that a card is both pinned and removed. Let me know if it makes sense! (fix is in 3f3c63d14)

renanferrari (approved):
Thanks for the changes! The solution makes sense for me and everything seems to be working as expected now.

LGTM :shipit: 

I'm not so sure what's the correct FluxC process anymore, so I'm gonna let you merge these PRs.

PR #1872, opened 2021-02-12
Shipping Labels Settings: expose more details, and allow updating

hichamboushaba:
This is needed for https://github.com/woocommerce/woocommerce-android/issues/3536, and the rest of SL tasks, it adds the following:
1. Exposes more details from the settings endpoint: store owner details and other settings.
2. Add the ability to update the account settings.

#### Testing
1. Open the example app and login.
2. Click on Woo, then Shipping Labels.
3. Click on "Update account settings"
4. Confirm that a dialog with different settings is displayed.
5. Confirm that the values in the dialog match the current settings.
6. Update some settings.
7. Click on Save.
8. Confirm that `Settings Updated` is printed.
9. Confirm in wp-admin that the settings were updated.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/shippinglabels/AccountSettingsApiResponse.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/shippinglabels/AccountSettingsApiResponse.kt
@@ -10,11 +10,18 @@ data class AccountSettingsApiResponse(
     @SerializedName("userMeta") val userMeta: UserMeta
 ) {
     data class FormData(
-        @SerializedName("selected_payment_method_id") val selectedPaymentId: Int?
+        @SerializedName("enabled") val isCreatingLabelsEnabled: Boolean,
+        @SerializedName("selected_payment_method_id") val selectedPaymentId: Int?,
+        @SerializedName("paper_size") val paperSize: String,
+        @SerializedName("email_receipts") val isPaymentReceiptEnabled: Boolean
     )
 
     data class FormMeta(
         @SerializedName("can_manage_payments") val canManagePayments: Boolean,
+        @SerializedName("master_user_name") val storeOwnerName: String,
+        @SerializedName("master_user_login") val storeOwnerUserName: String,
+        @SerializedName("master_user_wpcom_login") val storeOwnerWpcomUserName: String,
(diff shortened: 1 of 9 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/shippinglabels/AccountSettingsApiResponse.kt
```diff
@@ -10,11 +10,18 @@ data class AccountSettingsApiResponse(
...
+        @SerializedName("paper_size") val paperSize: String,
+        @SerializedName("email_receipts") val isPaymentReceiptEnabled: Boolean
     )
 
     data class FormMeta(
         @SerializedName("can_manage_payments") val canManagePayments: Boolean,
+        @SerializedName("master_user_name") val storeOwnerName: String,
+        @SerializedName("master_user_login") val storeOwnerUserName: String,
+        @SerializedName("master_user_wpcom_login") val storeOwnerWpcomUserName: String,
+        @SerializedName("master_user_email") val storeOwnerWpcomEmail: String,
```
hichamboushaba, line 24: Even that this says `master_user_email`, it's actually the WordPress.com account email (https://github.com/Automattic/woocommerce-services/blob/c97384bc78254bd484485f3f781917af605788c2/classes/class-wc-connect-account-settings.php#L45), that's why I chose to name it `storeOwnerWpcomEmail`

nbradbury:
@hichamboushaba Code looks good, but I'm lost in wp-admin. Where do I go to "Confirm in wp-admin that the settings were updated?"

hichamboushaba:
Thanks @nbradbury for the review, assuming WC Shipping is installed in your store, the settings are in WooCommerce -> Settings -> Shipping -> WooCommerce Shipping (https://d.pr/i/gHMctV)

Just a remark regarding the naming, the switch in wp-admin says "Shipping Labels" which makes it look like it controls the whole shipping labels feature, but it's actually for just creating new labels, users can still reprint previously created labels even if it's disabled, and that's why I called it `isCreatingLabelsEnabled` in our model.

nbradbury:
> the settings are in WooCommerce -> Settings -> Shipping -> WooCommerce Shipping

Thanks for that, I knew I'd been there before but couldn't find it earlier.

nbradbury (approved):
:shipit:

PR #1871, opened 2021-02-12
Add check for empty order product attribute key

anitaa1990:
This PR is a hotfix that just copies everything from this PR #1778. It looks like the changes from that PR has been removed during a merge conflict. 

This PR just fixes the following crash when an unexpected empty key is discovered in the order product attribute json array:
```
ava.util.NoSuchElementException: Char sequence is empty.
    at kotlin.text.StringsKt___StringsKt.first(_Strings.kt:71)
    at kotlin.text.StringsKt.first
    at org.wordpress.android.fluxc.model.WCOrderModel$LineItem.getAttributesAsString(WCOrderModel.kt:115)
    at com.woocommerce.android.model.OrderKt.toAppModel(Order.kt:184)
    at com.woocommerce.android.ui.orders.details.OrderDetailRepository.getOrder(OrderDetailRepository.kt:242)
    at com.woocommerce.android.ui.orders.details.OrderDetailViewModel.start(OrderDetailViewModel.kt:108)
    at com.woocommerce.android.ui.orders.details.OrderDetailViewModel.<init>(OrderDetailViewModel.kt:103)
    at com.woocommerce.android.ui.orders.details.OrderDetailViewModel_AssistedFactory.create(OrderDetailViewModel_AssistedFactory.java:49)
```

cc @AliSoftware

PR #1870, opened 2021-02-12
Fixed crash when trying to parse variations field in WCProductModel

anitaa1990:
Fixes woocommerce/woocommerce-android#3553 by parsing just the size of the `variations` instead of trying to parse the values of the `variations` array.

#### To test
- Open the example app and click on `Woo` - > `Products` -> `Fetch Single product`.
- Enter a productId for a variable product and ensure that the variations count is displayed correctly.
- Enter a productId for a simple product and ensure that the product is fetched correctly.

This PR is a hotfix and should be merged to the release branch so that we can release another beta. cc @AliSoftware

PR #1874, opened 2021-02-12
Merge hotfix 1.10.1 into develop

AliSoftware:
We just cut a hotfix from the `1.10.0` tag (used by the `release/6.0` branch of WCiOS currently in beta) to fix 2 issues #1870 and #1871 and create hotfix `1.10.1`.

So now it's time to merge the `1.10.1` hotfix back into `develop`.

### DIFF
--- a/example/src/test/resources/wc/lineitems.json
+++ b/example/src/test/resources/wc/lineitems.json
@@ -108,6 +108,10 @@
       },
       {
         "display_value": []
+      },
+      {
+        "display_key": "",
+        "display_value": "empty key"
       }
     ],
     "sku":"woo-vneck-tee",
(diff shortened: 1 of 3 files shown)


jkmassel (approved):

jkmassel:
Approved – I see @anitaa1990 is on the reviewer list, so I'll let her merge if 👍

AliSoftware:
tbh she already approved the other PR about the merge to `trunk` (and whose diff is *exactly* the same in practice…) and I since we already have a FluxC `1.11.0-beta-1` (used by WPAndroid only), I wouldn't want this hotfix to risk not making it in a beta-2 in case someone needed to cut a `1.11.0-beta-2` (while working on WPAndroid or WCAndroid's `develop`) before this one gets merged, so I'd say let's merge ASAP.

(would probably not be a big deal if it was merged afterwards as long as the hotfix end up being included in the final release of `1.11.0`, but I just want to be sure we don't risk having that hotfix skipped like it happened during `1.6.27.x` in December, especially because this `1.10.1` hotfix re-fixes a regression that December's `1.6.27.x` hotfix was all about back then and failed to make its way to `develop`… so fool me once…)

PR #1875, opened 2021-02-12
Woo: Shipping rates endpoint

0nko:
This PR adds the support for retrieving shipping rates.

**To test:**
1. Go to Woo -> Shipping labels -> Select site
2. Tap on Get shipping rates
3. Wait for the data to be loaded
4. Enter package size parameters
5. Notice a response is printed in the log

**Unit tests:**

Run **WCShippingLabelStoreTest** tests.

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -148,5 +148,5 @@ dependencies {
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlinCoroutinesVersion"
 
     // Multidex
-    implementation 'com.android.support:multidex:1.0.3'
+    implementation 'androidx.multidex:multidex:2.0.1'
 }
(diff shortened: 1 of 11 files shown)


nbradbury:
@0nko Code looks good, but I'm not sure what this dialog is asking, and why it's asking for it?

package

0nko:
Thanks for the review, @nbradbury!

> Code looks good, but I'm not sure what this dialog is asking, and why it's asking for it?

Oh, that's just a name/ID used by the endpoint, which gets returned along with the shipping rates. For example, you could have 2 shipping packages and each will get a separate quote. You can then use this ID to identify which is which.

nbradbury (approved):
Looks good! :shipit:

PR #1878, opened 2021-02-18
Improve attributes method operations

ThomazFB:
Summary
==========
This PR changes how we modify any Product attribute, instead of requesting any updates from separate methods, now we take advantage of the `updateProduct` call from the `WCProductStore` so the update works with any product changes. As a small additional tweak, this PR also adds the option to request all cached attributes directly from the Database instead of fetching them from the site.

Screenshots & How to Test
==========

## Update Product
| Select the Woo section | Select the Products Section | Select Update Product | Attach or Detach Attributes from a Product
| ------------- | ------------- | ------------- | ------------- |
| Screenshot_1612377245 | Screenshot_1612377248 | Screenshot_1613603879 | Screenshot_1613611606 |

⚠️  Don't forget to inform a valid product ID at the top of the Update Product view

To verify if the changes are working under the Woo site, go to Products > All Products > Select the updated Product on the list and go to the Attributes section on your Woo site and make sure everything is as expected.

image

## Fetch Attributes from Database
 
| Select the Woo section | Scroll down until the Attribute option becomes visible and click on it | Select Fetch attributes from database
|Screenshot_1612377245 | Screenshot_1613614095 | Screenshot_1613613983 |

⚠️  Make sure data comes directly as it is from database ignoring any Site changes

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt
@@ -17,6 +17,10 @@ import androidx.fragment.app.Fragment
 import dagger.android.support.AndroidSupportInjection
 import kotlinx.android.parcel.Parcelize
 import kotlinx.android.synthetic.main.fragment_woo_update_product.*
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
 import org.wordpress.android.fluxc.Dispatcher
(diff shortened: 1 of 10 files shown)


nbradbury:
I'm wondering how we'll detach an attribute term in WCAndroid since it requires a term id, which isn't exposed in the product attribute list. I guess we simply fetch all terms and get the term id by matching the name?

term

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGlobalAttributeStore.kt
```diff
@@ -48,6 +48,10 @@ class WCGlobalAttributeStore @Inject constructor(
                         ?: WooResult(WooError(GENERIC_ERROR, UNKNOWN))
             }
 
+    fun fetchCachedStoreAttributes(
```
nbradbury: This works fine, but in FluxC "fetch" usually means fetching from the backend. How about something like `loadCachedStoreAttributes` instead?

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt
```diff
@@ -524,6 +538,80 @@ class WooUpdateProductFragment : Fragment() {
...
+                                }
+                    }
+                }
+            }
+        } catch (ex: Exception) {
+            prependToLog("Couldn't create Attribute Term. Error: ${ex.message}")
+        }
+    }
+
+    private suspend fun handleProductAttributesSync(
```
nbradbury, line 606: This would benefit from a comment header explaining what this `fun` is doing (I should've mentioned this on the last PR, sorry).

nbradbury:
@ThomazFB Code looks good, just had a couple of nit (Nick?) picks.

ThomazFB:
Hey @nbradbury! Thanks for the awesome "Nick" picks 😆! Regarding the Term ID, I've improved the `WCGlobalAttributeModel` in order to make it easier for us to do changes to the options List<String> as we talked about yesterday on slack and changed the Example app to make sure we're adding Terms through a String just fine. Here's the commit: 00f386a751256310f9f6d73897f535cd82b87858. Let me know what you think about this solution.

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooUpdateProductFragment.kt
ThomazFB: Sure! Here's the change: a363e64279c19cd9d8c0e1d02b51f20e205cb97c

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGlobalAttributeStore.kt
ThomazFB: Agreed! Here's the change: c513e288aa6158c67408f0dfef8ec838755627dd

ThomazFB:
The PR is ready for another round!

nbradbury (approved):
Looks good! :shipit:

PR #1879, opened 2021-02-18
Pass the correct id when updating Woo Product Settings

hichamboushaba:
Currently when changing the product settings, the new changes are not reflected on the app. The cause is that we were passing a wrong id to the SQL request. This PR fixes this.

This was reported in the issue https://github.com/woocommerce/woocommerce-android/issues/3575

#### Testing
1. Open the example app.
2. Click on Woo, then "Fetch WC product settings"
3. Check that the current settings are logged.
4. Go to wp-admin -> WooCommerce -> Settings -> Products
5. Change the measurement units.
6. Go back to the example app, and click on "Fetch WC product settings"
7. Confirm that the new settings are printed.


nbradbury (approved):
Fix is good! :shipit:

PR #1883, opened 2021-02-19
Shipping Labels account settings enhancements

hichamboushaba:
This just adds two more things to the SL account settings:
1. Exposes the field `can_edit_settings`, even if it's always true (as explained here: p91TBi-4uy-p2), to keep better compatibility, we'll use it for controlling the display of settings.
2. Adds the paper size type A4.


0nko (approved):
Looks good 👍

PR #1888, opened 2021-02-24
Return comment filter with response and add a limit option for local comment requests.

khaykov:
This is a counterpart to [this](https://github.com/wordpress-mobile/WordPress-Android/pull/14140) wpandroid PR.

This PR adds a `CommentStatus` to payload returned when requesting all comments. `CommentStatus` can be null when we don't need to match specific action to a specific CommentStatus.

This PR also adds a logic for detection of comments that were removed from server or moved to a different status.

In addition this PR adds a limit option when requesting site comments.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
@@ -80,7 +80,7 @@ public class CommentSqlUtils {
             return 0;
         }
 
-        if (!Arrays.asList(statuses).contains(CommentStatus.ALL)) {
+        if (Arrays.asList(statuses).contains(CommentStatus.ALL)) {
             return removeComments(site);
         }
 
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
@@ -77,13 +77,11 @@ public class CommentSqlUtils {
                       .execute();
     }
 
-    public static int removeDeletedComments(SiteModel site, List<CommentModel> comments, int maxEntriesInResponse,
-                                            int requestOffset, CommentStatus... statuses) {
+    public static int removeCommentGaps(SiteModel site, List<CommentModel> comments, int maxEntriesInResponse,
+                                        int requestOffset, CommentStatus... statuses) {
         if (site == null || comments == null || comments.isEmpty()) {
             return 0;
         }
-
-        ArrayList<Long> remoteIds = new ArrayList<>();
 
         Collections.sort(comments, new Comparator<CommentModel>() {
             @Override
@@ -94,13 +92,13 @@ public class CommentSqlUtils {
             }
         });
 
+        ArrayList<Long> remoteIds = new ArrayList<>();
         for (CommentModel comment : comments) {
             remoteIds.add(comment.getRemoteCommentId());
         }
 
-        long topTimeStamp = comments.get(0).getPublishedTimestamp();
-        long bottomTimeStamp = comments.get(comments.size() - 1).getPublishedTimestamp();
-
+        long startOfRange = comments.get(0).getPublishedTimestamp();
+        long endOfRange = comments.get(comments.size() - 1).getPublishedTimestamp();
 
         ArrayList<CommentStatus> targetStatuses = new ArrayList<>();
         if (Arrays.asList(statuses).contains(CommentStatus.ALL)) {
@@ -119,7 +117,7 @@ public class CommentSqlUtils {
                                            .equals(CommentModelTable.LOCAL_SITE_ID, site.getId())
                                            .isIn(CommentModelTable.STATUS, targetStatuses)
                                            .isNotIn(CommentModelTable.REMOTE_COMMENT_ID, remoteIds)
-                                           .greaterThen(CommentModelTable.PUBLISHED_TIMESTAMP, topTimeStamp)
+                                           .greaterThen(CommentModelTable.PUBLISHED_TIMESTAMP, startOfRange)
                                            .endWhere()
                                            .execute();
         }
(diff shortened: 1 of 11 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
```diff
@@ -182,6 +185,22 @@ public CommentStore(Dispatcher dispatcher, CommentRestClient commentRestClient,
...
+     * @param statuses Array of status or CommentStatus.ALL to get all of them.
+     * @param limit Maximum number of comments to return. 0 is unlimited.
+     */
+    @SuppressLint("WrongConstant")
+    public List<CommentModel> getCommentsForSite(SiteModel site, boolean orderByDateAscending, int limit,
+                                                 CommentStatus... statuses) {
+        @Order int order = orderByDateAscending ? SelectQuery.ORDER_ASCENDING : SelectQuery.ORDER_DESCENDING;
+        return CommentSqlUtils.getCommentsForSite(site, order, limit, statuses);
+    }
+
```
develric: Not sure if I'm missing anything, wdyt if we merge this 2 getCommentsForSite function as we did for `CommentSqlUtils.getCommentsForSite` already? I was also thinking we maybe could even remove the CommentSqlUtils`public static List<CommentModel> getCommentsForSite(SiteModel site, @Order int order, CommentStatus... statuses)` but I guess it's better to leave that function call in case it's used by other apps, so we keep the binary compatibility.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
```diff
@@ -90,25 +107,35 @@ public static CommentModel getCommentByLocalCommentId(int localId) {
...
+
+        if (limit > 0) {
+            query.limit(limit);
+        }
+
         ConditionClauseBuilder<SelectQuery<CommentModel>> selectQueryBuilder =
-                WellSql.select(CommentModel.class)
-                        .where().beginGroup()
-                        .equals(CommentModelTable.LOCAL_SITE_ID, site.getId());
+                query.where().beginGroup()
```
develric, line 191: I think it was already there, but since I thought to it better share 😄 , should't we have an endGroup() here also ? Or not having the Group at all maybe 🤔

develric (commented):
Had a first test round here using the companion WP Android PR and I think it's looking good 👍 . As said in the companion PR I would like to have another round of test/check on that other PR but leaving here just a couple of notes for your review 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
khaykov: My understanding is that we need a group in case the statuses include `CommentStatus.ALL` so we could add `selectQueryBuilder.isIn(CommentModelTable.STATUS, Arrays.asList(statuses));` to the query.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
```diff
@@ -75,21 +77,73 @@ public static int removeComments(SiteModel site) {
...
+        if (site == null || comments == null || comments.isEmpty()) {
             return 0;
         }
 
+        Collections.sort(comments, new Comparator<CommentModel>() {
+            @Override
+            public int compare(CommentModel o1, CommentModel o2) {
+                long x = o2.getPublishedTimestamp();
+                long y = o1.getPublishedTimestamp();
+                return (y < y) ? -1 : ((x == y) ? 0 : 1);
```
develric: >(y < y)

Is this correct? 🤔  I would expect something more like `(x<y)` or the reverse

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
```diff
@@ -75,21 +77,73 @@ public static int removeComments(SiteModel site) {
...
+        int numOfDeletedComments = 0;
+
+        // try to trim comments from the top
+        if (requestOffset == 0) {
+            numOfDeletedComments += WellSql.delete(CommentModel.class)
+                                           .where()
+                                           .equals(CommentModelTable.LOCAL_SITE_ID, site.getId())
+                                           .isIn(CommentModelTable.STATUS, targetStatuses)
+                                           .isNotIn(CommentModelTable.REMOTE_COMMENT_ID, remoteIds)
+                                           .greaterThen(CommentModelTable.PUBLISHED_TIMESTAMP, startOfRange)
```
develric: Not sure, but should it be `greaterThenOrEqual` or `lessThenOrEqual` in all these 3 delete queries?

PS: it has nothing to do with this PR but it's my poor english or should it be `Than`; again not related to this PR, just curious about what should it be 😄

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
develric: ahah....forget about it, makes sense 😄

develric (commented):
Hey @khaykov , I'm still testing the WPAndroid companion app but I thought to share a couple of things below in case they could be relevant. Thanks for your feedback on it 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
khaykov: Yep, you are right - it's supposed to be (x<y) 🤦

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
khaykov: `OrEqual` statement will include comments that we just got from server into the query (there is `isNotIn` check, but still).

And yes, I also think it should be `Than` :) It's a 3rd party library, so things like this happen :D

develric (approved):
Thanks for the changes @khaykov 🙇 ! Used the WPAndroid companion PR to test and works well for me 👍  (leaving few more details on test covered in the WPAndroid PR)

PR #1892, opened 2021-02-25
Add order shipping line fields

### LINKED ISSUE #1885: Woo: Add all properties into ShippingLine in WCOrderModel
At the moment, `getShippingLineList()` inside WOrderModel.kt returns a list of `ShippingLine` object that [only contains "method_title" as its property.](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/96c59a96a69f31634198fa4a481b90a0c405ab39/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt#L75)

The `ShippingLine` object needs to be modified to include the other properties available inside a shipping line item, for example "total_cost" and "total". 

Screen Shot 2021-02-20 at 16 11 58

At the moment my need for that is to support shipping refund in the app.

hafizrahman:
This will fix #1885 and will initially be used for enabling shipping refund in Woo Mobile.

**To test:**
1. Run the test on `example/src/test/java/org/wordpress/android/fluxc/wc/order/WCOrderModelTest.kt`
2. Make sure everything passes, especially `testGetShippingLines` and `testGetShippingLinesAttributes`


0nko (approved):
Nice work, @hafizrahman! 🥇

PR #1903, opened 2021-02-28
Issue-3605. Fetch customers

kidinov:
#3605 3 and 4

[Commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1903/commits/8037521f75e5c6c46e6720c57933b6ede01e8b88)

Description. An API call, caching, models, and DTOs for /customers/ endpoint

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/customer/CustomerRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/customer/CustomerRestClient.kt
@@ -1,0 +1,49 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.customer
+
+import android.content.Context
+import com.android.volley.RequestQueue
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.generated.endpoint.WOOCOMMERCE
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.UserAgent
+import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
+import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackError
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackSuccess
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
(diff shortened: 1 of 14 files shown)


0nko:
I noticed your commits are very large with hundreds of lines of code. It's a bit harder to follow the changes, especially since commit messages can't really encompass everything and end up being too generic. 

I'd suggest splitting up the changes into smaller chunks that is logically tied together. Going even further, we usually try to keep our PRs under 500 LoC. I know it's hard sometimes (I struggle with it myself) but anything larger makes it harder to review, so just something to keep in mind.

kidinov:
@0nko Sorry. I have this habit because currently in the company I work for, merges are not allowed because every commit should have a ticket number in the message (to generate release notes, better integrations with ticket systems, CI knows more, etc). Also, every commit should solve a single task (ideally) and if it's too big it means that the task was way too complex (probably it's the mistake I made here then)

One thing is not clear for me here though. If I would have many commits per PR, then it can be harder for the reviewer to understand what exactly changes current PR brings - I can not target my branch to a parent branch because of the restrictions I have (I work in the fork). With 1 PR = 1 commit I just give a link to the commit and you can check the actual changes the PR brings.

0nko:
I see. Well, we don't have any restriction like that :). 

An example of a good PR description and commit messages: https://github.com/woocommerce/woocommerce-android/pull/3559

It describes the main purpose of the PR and list the steps to test the new functionality (if applicable).

> If I would have many commits per PR, then it can be harder for the reviewer to understand what exactly changes current PR brings

Interesting, I'd argue the opposite :). I could quickly glance at the list of commits and get the gist of what sort of changes have been made. Then I can look at the "Files changed" tab and see all the code changes (just like the single commit).

>  I can not target my branch to a parent branch because of the restrictions I have (I work in the fork).

I know. We normally target the parent branch, but here I usually wait for the parent PR to be merged before I dive into the next one.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/customer/CustomerRestClient.kt
```diff
@@ -34,11 +41,73 @@ class CustomerRestClient(
...
+        role: String? = null,
+        context: String? = null,
+        remoteCustomerIds: List<Long>? = null,
+        excludedCustomerIds: List<Long>? = null
+    ): WooPayload<Array<CustomerApiResponse>> {
+        val url = WOOCOMMERCE.customers.pathV3
+
+        val orderBy = when (sortType) {
+            NAME_ASC, NAME_DESC -> "name"
+            DATE_ASC, DATE_DESC -> "name"
```
0nko: Should this be "date"?

0nko (commented):
The code looks good, I just found one possible mistake.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/customer/CustomerRestClient.kt
kidinov: Should be "include". [Fixed](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1903/commits/081bc947a806d19c732f8bc1aad6a05b4224d12a) =)

```

orderby | string | Sort collection by object attribute. Options: id, include, name and registered_date. Default is name.
```

0nko (approved):
:shipit:

PR #1907, opened 2021-03-04
Issue-3605. Example of customer creation

kidinov:
Part number 9 of [#3605](https://github.com/woocommerce/woocommerce-android/issues/3605), contains an example of the usage of FluxC functionality of customer creation

**Testing:**
1. Open the example
2. Sign in
3. Click "WOO" button
4. Scroll down and click "CUSTOMERS" button
5. Select site
6. Click "CREATE CUSTOMER"
7. Type data you need
8. Click "CREATE" button
9. Check the logs

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/FragmentExt.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/FragmentExt.kt
@@ -11,8 +11,8 @@ fun Fragment.prependToLog(s: String) {
 }
 
 fun Fragment.replaceFragment(fragment: Fragment) {
-    fragmentManager?.beginTransaction()
-            ?.replace(R.id.fragment_container, fragment)
-            ?.addToBackStack(null)
-            ?.commit()
+    parentFragmentManager.beginTransaction()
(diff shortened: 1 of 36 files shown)


0nko:
I've tried to test the customer list fetching, but it doesn't seem to be working (or maybe I'm using it wrong). When I don't add any filter or search string, I'd expect it to fetch all customers, but it says there is nothing and just shows a blank screen. 

image

When I do add something as a search string (a part of the name of an existing customer), again, it shows nothing.

kidinov:
@0nko do you have customers on this site? The log shows "Customers list is empty". If you don't have customers maybe try to create one via the flow described in this PR and check if it appears here in the list.

What returns to you for your site:
```
curl https://example.com/wp-json/wc/v3/customers \
    -u consumer_key:consumer_secret
```
?

0nko:
Hmm, weird. I'm getting an empty list, as well. But when I open the Customers page in wp-admin I can see 13 customers:

image

0nko (approved):
Looks good! 👍

Issue #1908, opened 2021-03-04
Wrong data in "localOrderId" column of WCShippingLabel


hichamboushaba:
The field `localOrderId` in the table `WCShippingLabel` references the remote order id, and not the local one.
It should be renamed.

PR #1909, opened 2021-03-04
Woo Shipping Labels: purchase labels

hichamboushaba:
This adds support for purchasing shipping labels.

To test:
1. Create an order that satisfies shipping label creation in your store.
2. Open the example app.
3. Click on Woo -> Shipping Labels.
4. Click on purchase a shipping label.
5. Enter the order and package details.
6. Confirm that the shipping label details are printed in the console.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/shippinglabels/WooShippingLabelFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/shippinglabels/WooShippingLabelFragment.kt
@@ -235,8 +235,10 @@ class WooShippingLabelFragment : Fragment() {
                         return@launch
                     }
                     if (origin == null || destination == null) {
-                        prependToLog("Invalid origin or destination address:\n" +
-                                "Origin:\n$origin\nDestination:\n$destination")
+                        prependToLog(
+                                "Invalid origin or destination address:\n" +
+                                        "Origin:\n$origin\nDestination:\n$destination"
+                        )
                         return@launch
                     }
                     var name: String
@@ -261,25 +263,30 @@ class WooShippingLabelFragment : Fragment() {
 
                                         val box: ShippingLabelPackage?
                                         if (height == null || width == null || length == null || weight == null) {
-                                            prependToLog("Invalid package parameters:\n" +
-                                                "Height: $height\nWidth: $width\nLength: $length\nWeight: $weight")
+                                            prependToLog(
+                                                    "Invalid package parameters:\n" +
+                                                            "Height: $height\n" +
+                                                            "Width: $width\n" +
+                                                            "Length: $length\n" +
+                                                            "Weight: $weight"
+                                            )
                                         } else {
                                             box = ShippingLabelPackage(
-                                                name,
-                                                "medium_flat_box_top",
-                                                height!!,
-                                                length!!,
-                                                width!!,
-                                                weight!!
+                                                    name,
+                                                    "medium_flat_box_top",
+                                                    height!!,
+                                                    length!!,
+                                                    width!!,
+                                                    weight!!
                                             )
 
                                             coroutineScope.launch {
                                                 val result = wcShippingLabelStore.getShippingRates(
-                                                    site,
-                                                    order.remoteOrderId,
-                                                    origin,
-                                                    destination,
-                                                    listOf(box)
+                                                        site,
+                                                        order.remoteOrderId,
+                                                        origin,
+                                                        destination,
+                                                        listOf(box)
                                                 )
 
                                                 result.error?.let {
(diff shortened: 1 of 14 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/shippinglabels/WCShippingLabelMapper.kt
```diff
@@ -32,4 +36,36 @@ class WCShippingLabelMapper
...
+                currency = labelItem.currency ?: ""
+                productNames = labelItem.productNames.toString()
+                productIds = labelItem.productIds.toString()
+                refund = labelItem.refund.toString()
+                dateCreated = labelItem.dateCreated?.toString() ?: ""
+
+                localOrderId = orderId
+                formData = gson.toJson(FormData(origin = origin, destination = destination))
+
+                localSiteId = site.id
```
hichamboushaba, line 67: Some details are missing from the actual API response, there is a task to rework the API to include them.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/shippinglabels/WCShippingLabelMapper.kt
```diff
@@ -32,4 +36,36 @@ class WCShippingLabelMapper
...
+                packageName = labelItem.packageName ?: ""
+                rate = labelItem.rate?.toFloat() ?: labelItem.refundableAmount?.toFloat() ?: 0F
+                refundableAmount = labelItem.refundableAmount?.toFloat() ?: 0F
+                currency = labelItem.currency ?: ""
+                productNames = labelItem.productNames.toString()
+                productIds = labelItem.productIds.toString()
+                refund = labelItem.refund.toString()
+                dateCreated = labelItem.dateCreated?.toString() ?: ""
+
+                localOrderId = orderId
```
hichamboushaba, line 64: the `localOrderId` is actually referencing the remote order ID, so what we are doing here is correct, we should rename it in a separate PR.
I've filed a separate issue for this #1908

example/src/main/java/org/wordpress/android/fluxc/example/ui/shippinglabels/WooShippingLabelFragment.kt
```diff
@@ -221,89 +225,68 @@ class WooShippingLabelFragment : Fragment() {
         get_shipping_rates.setOnClickListener {
             selectedSite?.let { site ->
                 coroutineScope.launch {
-                    prependToLog("Loading shipping data...")
+                    val orderId = showSingleLineDialog(requireActivity(), "Enter order id:", isNumeric = true)
+                            ?.toLong() ?: return@launch
```
hichamboushaba, line 229: I added this one, as some orders may not satisfy the shipping labels creation conditions, so it's better to use a specific order for the tests.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/shippinglabels/WCShippingLabelPackageData.kt
```diff
@@ -0,0 +1,17 @@
+package org.wordpress.android.fluxc.model.shippinglabels
+
+import com.google.gson.annotations.SerializedName
+
+data class WCShippingLabelPackageData(
```
hichamboushaba, line 5: I didn't find a better name, if you have one, please let me know 😉

0nko (commented):
When I was testing the purchase I got the following exception after entering the package dimensions:

```
java.util.NoSuchElementException: List is empty.
        at kotlin.collections.CollectionsKt___CollectionsKt.first(_Collections.kt:212)
        at org.wordpress.android.fluxc.example.ui.shippinglabels.WooShippingLabelFragment$onViewCreated$8$$special$$inlined$let$lambda$1.invokeSuspend(WooShippingLabelFragment.kt:360)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7656)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
```
         
It looks like there are no rates available for the given addresses, which is strange, since I use the same order for testing without any changes and it does give me rates in the Woo app.

hichamboushaba:
Maybe it doesn't have any rates for the package "medium_flat_box_top" that we hardcode, did you test just fetching the rates and see what it gives?

I'll update the logic to log an error if it's empty instead of crashing.

hichamboushaba:
@0nko I added a check to see if the list of rates is not empty before trying to access it in 9f8dfd3, I'm also caching the purchased labels after purchase in 7da18c5

Ready for another round

0nko (approved):
Nice work, @hichamboushaba! :shipit:

PR #1911, opened 2021-03-05
Site plan support for MimeTypes

### LINKED ISSUE #1884: Add WordPress Site Plan MIME types
The `MimeTypes` needs support for allowed MIME types based on a site’s plan ([Dot Com](https://en.support.wordpress.com/accepted-filetypes/)) or configuration ( [Dot Org](https://href.li/?https://codex.wordpress.org/Uploading_Files)) as a means of filtering unavailable MIME types from the media pickers. 

## Expected behavior
`MimeTypes` should only return the MIME types that are supported by a site's plan. 

## Actual behavior
`MimeTypes` returns all MIME types that are defined as acceptable types.

jd-alexander:
Fixes #1884 

WP Android https://github.com/wordpress-mobile/WordPress-Android/pull/14353

## Solution
An enum was introduced that represents the different plans and they were wired to the MIME types that are currently supported within the WP Android app. 

The main source for this logic change is located on the WordPress.com website https://wordpress.com/support/accepted-filetypes/ . There are a few deviations from this nonetheless, for video support, which is not restricted to any site plan, but available for all users. The main changes are as follows: 

1. Audio files are only available on paid plans. 
2. Key and zip files are only available on paid plans while all other document types are available across the board.

## Testing
1. From a testing standpoint, several unit tests were created to verify that this functionality works as expected. 
2. A WP Android PR, implementing these API changes will be created to validate that they work as expected. 

## Reviewing 

## Submitter Checklist

- [x] I have considered adding accessibility improvements for my changes.


fluxc/src/main/java/org/wordpress/android/fluxc/utils/MimeTypes.kt
```diff
@@ -80,7 +91,7 @@ class MimeTypes {
      * .key - missing - "application/keynote"
      * .zip (Archive File Format)
      */
-    private val documentTypes = listOf(
+    private val wpComPaidAndSelfHostedDocumentTypes = listOf(
```
planarvoid: If I understand correctly, this list contains all the items from the `wpComFreeDocumentTypes` and some additional ones, right? Would it maybe make sense to change it to something like: 
```
private val wpComPaidAndSelfHostedDocumentTypes = wpComFreeDocumentTypes  + listOf(MimeType(APPLICATION, Subtype.KEYNOTE, listOf("key")),MimeType(APPLICATION, Subtype.ZIP, listOf("zip"))
```
this way it would be obvioius that the `free` is just a subset of the `paid` types. 
One thing to consider would be to have:
- wpComPaidAndSelfHostedDocumentTypes - just for the document types that are additional in the paid plans
- wpComFreeDocumentTypes - all the types allowed in the free plans
- documentTypes - `wpComPaidAndSelfHostedDocumentTypes + wpComFreeDocumentTypes`

I'd also update the comment above to explain either of these approaches.

planarvoid (commented):
The changes are looking good 👍  thanks! I have one comment that might improve the readibility. Let me know what you think

fluxc/src/main/java/org/wordpress/android/fluxc/utils/MimeTypes.kt
jd-alexander: Thanks for these thoughts. I was wondering if this would be a worthwhile approach to reduce duplication so the explanation here is quite helpful and I will make these changes.

jd-alexander: I did these changes here 99f384f

planarvoid (approved):
Thanks for the changes, it's looking good 👍

Issue #1912, opened 2021-03-05
WC: Update just a product's attributes


nbradbury:
For WCAndroid, we need a way to update only a product's attributes. This could use the same call we use to update the product, except passing only its attributes, eg:

```
product {
   attributes: [
       {
           id: 3,
           name: "Size"
           options: [
               "S",
               "M",
               "L"
           ]
       }
   ] 
}
```
Note we want to save **all** attribute & term changes - both local and global - in one call. This can be done by simply including only the `attributes` section when updating a product.

So we need a method to accept a site model, remoteProductId, and a list of attributes (local and/or global) with their terms.

ThomazFB:
Closing this issue since the fix is already merged

PR #1913, opened 2021-03-06
Feature/add dedicated product attributes call

ThomazFB:
Summary
==========
Address issue #1912. This PR adds support to edit/add/remove/update all attributes from *a given product and variation* in a way that we only update the product from a given attribute list instead of posting the entire product and everything included on it


example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
```diff
@@ -660,10 +674,168 @@ class WooProductsFragment : Fragment() {
...
+        try {
+            showSingleLineDialog(
+                    activity,
+                    "Enter the product ID you want to attach attributes:"
+            ) { productIdEditText ->
+                showSingleLineDialog(
+                        activity,
+                        "Enter the attribute ID you want to attach:"
+                ) { attributeIdEditText ->
+                    showSingleLineDialog(
```
nbradbury: I'm confused why we're asking for a term ID here, since the goal is to update all the product's attributes (both global and local) in a single call, and local terms don't have IDs. Am I misunderstanding?

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
```diff
@@ -660,10 +674,168 @@ class WooProductsFragment : Fragment() {
...
+                        }
+                    }
+                }
+            }
+        } catch (ex: Exception) {
+            prependToLog("Couldn't create Attribute Term. Error: ${ex.message}")
+        }
+    }
+
+    private suspend fun requestSelectedProductData(
```
nbradbury: This would benefit from a comment header stating its purpose, and it could be named something more descriptive.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -715,6 +716,57 @@ class ProductRestClient(
...
     }
 
+    /**
+     * Makes a PUT request to
+     * `/wp-json/wc/v3/products/[WCProductModel.remoteProductId]/variations/[WCProductVariationModel.remoteVariationId]`
+     * to replace a variation's attributes with [WCProductVariationModel.attributes]
+     *
+     * Returns a WooPayload with the Api response as result
+     *
+     * @param [site] The site to fetch product reviews for
```
nbradbury: Typo: this says "product reviews."

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -715,6 +716,57 @@ class ProductRestClient(
...
+                }
+    }
+
+    /**
+     * Makes a PUT request to `/wp-json/wc/v3/products/[WCProductModel.remoteProductId]`
+     * to replace a product's attributes with [WCProductModel.attributes]
+     *
+     * Returns a WooPayload with the Api response as result
+     *
+     * @param [site] The site to fetch product reviews for
```
nbradbury: Same typo ("reviews").

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
ThomazFB: Thanks for the tip! Fixed here: 976da948eee813f4c9e7144f36a8eb3fafe4788c

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
ThomazFB: Thanks for the tip! Fixed here: 976da948eee813f4c9e7144f36a8eb3fafe4788c

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
ThomazFB: Agreed! Added here: 545b2f7ab8596a0bf112dfe2c3ebd8934dd13c37

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
ThomazFB: Sorry, this was a lack of attention of mine. We agreed about informing Terms as Strings instead of IDs to match the app user flow. I've fixed this here: 3373f6c67162e2c546e4b65a11f16301ee6c8848

ThomazFB:
Hey @nbradbury, thank you for your time reviewing these changes, the PR is now ready for a second round!

nbradbury:
@ThomazFB This is much better, but I think it's overlooking what we really want to do here. As discussed in that [Slack thread with Ernesto](https://a8c.slack.com/archives/C70PAM3RA/p1614887568402600), we want to save **all** attribute & term changes - both local and global - in one call. This can be done by simply including only the `attributes` section when updating a product.

So we need a method to accept a site model, remoteProductId, and a list of attributes (local and/or global) with their terms. I'm sorry I wasn't clearer about that in the issue (I've just updated it).

ThomazFB:
Oh, sorry about that @nbradbury, I misunderstood some points of your request, I'll change that and return here. I have just one question, do you think it's necessary to have a UI flow on the example app for this new method operation? I'm asking this because I'm not sure how we could properly differentiate this call from the existing ones, and if it does makes sense to do that.

nbradbury:
> do you think it's necessary to have a UI flow on the example app for this new method operation?

Probably not, especially if there's a test in WCAndroid to verify that it works.

ThomazFB:
@nbradbury the PR is ready for a new round, let me know if these changes deliver what you need!

nbradbury:
@ThomazFB This looks like exactly what we need - thanks!  Later today I'll update my WCAndroid branch to use this FluxC hash, then test it from there.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -837,10 +846,11 @@ class WCProductStore @Inject constructor(
 
     suspend fun submitProductAttributeChanges(
         site: SiteModel,
-        product: WCProductModel
+        productId: Long,
+        attributes: List<WCProductAttributeModel>
```
nbradbury: Should this be  a `List<WCProduct.ProductAttribute`, because that's what's stored in the product model?

ThomazFB: Definitely! Since you mentioned that, I've decided to fully delete the `WCProductAttributeModel` and only use the `WCProduct.ProductAttribute` since it's the model we're using on the Woo app, so it makes more sense to use only one model to represent a Local Product Attribute. I've commited th change right here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1913/commits/7b521af03047362f0890d3b6d9192159e493d8e6

ThomazFB:
@nbradbury forgot to ping you, but your changes request is done! The PR is ready for another round 😄

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
nbradbury: > I've decided to fully delete the `WCProductAttributeModel`

I think that's a good call! To go along with this, should we delete the `WCProductAttributeModel` table, or will that still be used?

ThomazFB: Oh, I think I made a confusion on that part, originally during the early stage of development, the WCGlobalAttributeModel was called as WCProductAttributeModel, but after I noticed that they should be called as Global, I renamed the corresponding table too, so the Table that you mentioned have a ALTER TABLE command changing it name to Global too.

So, as it is, there's no table for WCProductAttributeModel, and that's why the model doesn't have any WellSql annotation inside of it, because it was stored as a JSON inside the attributes field of the WCProductModel, the existing Table is made just for the Global Attributes, so I think we're safe on this.

nbradbury: Oh, I should've noticed that - thanks for clarifying.

nbradbury:
@ThomazFB Code is looking very good, but it looks like the latest changes to `develop` need to be pulled in. PR #1882 added a `WCOrderStore.OnOrderSummariesFetched` class which needs to be accessible from WCAndroid.

ThomazFB:
Hey @nbradbury, thanks for pointing that, I updated the branch with `develop`, I think it'll work now.

nbradbury (approved):
Looks great! :shipit:

PR #1914, opened 2021-03-08
Merge release/1.12.0 -> trunk

AliSoftware:
Just made a stable [release 1.12.0](https://github.com/wordpress-mobile/WordPress-FluxC-Android/releases/tag/1.12.0) in preparation for WooCommerce 6.2 code freeze + upcoming WP code freeze (\cc @JavonDavis).

---

\cc @anitaa1990 just FYI (in case you want to rehearse the logic of "which commit to tag?" that we followed last time we did that step during release shadowing 😉 ):

At the time of WC code-freeze:

 - WordPress's `develop` was pointing to FluxC `1.12.0-beta-4`
 - WooCommerce's `develop` was pointing to FluxC 7df2303c62b8299e7650a03396ff84637c178f76, which is 4 merge-commits behind `1.12.0-beta-4`, and those 4 merge commits added to FluxC since seems to all only be related to changes in gutenberg-mobile and WordPress, so including them in WooCommerce should be harmless.
 - At the time of this writing, FluxC's `develop` doesn't have any new commit since `1.12.0-beta-4` (i.e. `1.12.0-beta-4` == `develop`). So we don't have to wonder if we should also include more recent changes post-last-beta here.

That's why I'm making the new version that we will use for WC code-freeze be cut from `1.12.0-beta-4` (and not 7df2303c62b8299e7650a03396ff84637c178f76) so that a single stable version will be able to be used both for WC and WP code freezes.


antonis (approved):
Thank you for handling this release @AliSoftware 🙇  
The code changes on the `wordpress` side LGTM 🎉

jkmassel (approved):

PR #1916, opened 2021-03-08
Issue/1915 hackweek db downgrade

### LINKED ISSUE #1915: Reset database when downgraded
For part of my "hack week" project, I plan to reset the SQLite database when its version is downgraded. Currently a downgrade crashes the app.

Note that this will only be enabled in `DEBUG` builds.

nbradbury:
Closes #1915 - In the past the example app would crash after we switched to a branch that used a previous db version, requiring us to uninstall, re-install, and login again.

The solution required simply overriding `onDowngrade()`, and use that to drop & recreate all the tables. This is only enabled for debug builds, and then only when the following has been added to `gradle.properties`:

```
wp.ENABLE_DATABASE_DOWNGRADE = true
```

So now when a downgrade is detected, you'll see a toast alerting that the db has been reset. To test, I recommend changing [getDbVersion()](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ddcd93d1a5a31bae57e326f01c2519937e8da349/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt#L31) in `WellSqlConfig` to this:

```
override fun getDbVersion(): Int {
    return 999
}
```

Run that once, then each time afterwards reduce the version by one. That will trigger the downgrade.

hack


antonis (approved):
Thank you for fixing this @nbradbury 🙇 
I tested this both with the **example** app and **WordPress-Android** and it works as expected 👍 
The code changes LGTM 🎉

nbradbury:
Thanks for the review, @antonis! Since you've approved this PR, can you merge it? Or should I do that now?

antonis:
> Thanks for the review, @antonis! Since you've approved this PR, can you merge it? Or should I do that now?

Done. Thanks for fixing this @nbradbury, it improves our daily workflow 👍

PR #1918, opened 2021-03-09
Issue/1917 hack week change log font size

### LINKED ISSUE #1917: Enable developers to customize the font size in the example app's log
OG's like me have a tough time reading the 12sp text in the log. We should add a value to `gradle.properties` that enables setting a custom size.

nbradbury:
Closes #1917 - the hard-coded 12sp font size for the example app's log is too tiny for people like me to see clearly, so this small PR adds the ability to set a custom font size in the example app's `build.gradle`.

**Note:** I'd be happy to close this PR and simply change the log font size to something more reasonable (like, say, 14sp or 16sp) in the layout file, but wasn't sure if most of us were fine with 12sp.

### DIFF
--- a/example/gradle.properties-example
+++ b/example/gradle.properties-example
@@ -1,5 +1,7 @@
 # OAuth credentials for example app
 wp.OAUTH.APP.ID = wp
 wp.OAUTH.APP.SECRET = wp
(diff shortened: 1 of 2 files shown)


nbradbury:
I've decided to close this and simply increase the hard-coded size separately.

PR #1920, opened 2021-03-09
Woo: Customers endpoint

0nko:
This PR contains a complete set of FluxC implementations of the Customers API endpoint, which is the first step of the order creation project:

- REST client
- Data persistence
- Store
- Unit tests
- Example app UI (Woo -> Customers)

This is a feature branch, all code has already been reviewed.


0nko:
The PR is ready for a merge.

nbradbury:
The customer search isn't working for me, and I tried on two different stores. As you can see, my site has customers:

web-customers

But when I search for customers, I always get a blank screen:

empty

And looking at the response, it's always empty:

response

I've tried text search, email search, id search, and leaving all the search fields blank, but no luck.

0nko:
The customers are really users of the `Customer` type. You probably don't have any. The endpoint has a parameter for type with a default value of `customer`. The `wp-admin`, on the other hand, uses the `all` value, which includes users without one specified. I could change the example to use `all`, as well. WDYT?

nbradbury:
> The customers are really users of the `Customer` type. You probably don't have any.

Oh, okay, that explains it. I changed one of my user's roles to "Customer" and now they correctly appear in the customer search. However, I would expect that feature to show me the same data as the Woo Customers page in `wc-admin`. Would changing the type to `all` show me that data, or does it just return Users rather than Customers?

**Update:** I guess my real question is, when does someone who purchases from your store become a user with the Customer role?

nbradbury (approved):
Looks good! :shipit:

PR #1923, opened 2021-03-10
Issue/1922 hack week cleanup example toolbar

nbradbury:
Closes #1922 - as part of my hack week project, this PR adds a very basic toolbar with a back button to the example app.

Screenshot_1615463846

Issue #1919, opened 2021-03-09
Clean up the example app's screens


nbradbury:
Our example app's screens are mostly a jarring assortment of mis-aligned buttons and other views. I know this app is just for us, but we deserve better :)

As part of my hack week, I'm going to clean it up a bit.

![example](https://user-images.githubusercontent.com/3903757/110535414-3eda5580-80ee-11eb-8973-789a6e5be3cc.png)

nbradbury:
Fixed via #1924

PR #1924, opened 2021-03-11
Issue/1919 hack week cleanup example woo

nbradbury:
As part of #1919, this PR cleans up all of the WooCommerce fragments. Before and after shots of the main Woo screen below (ignore the toolbar, that comes in #1923).

woo1


0nko (approved):
:shipit:

PR #1925, opened 2021-03-12
Fix breaking tests: Use unique content for WPCom post tests

aforcier:
This fixes the recent breakage in the live network connected tests.

The tests started breaking consistently a few days ago. I tracked down the cause to a change in the .com `posts/new` API (see D58079-code). That API change updates the endpoint to attempt to match any new post request to an existing post. If a match is found, the existing post ID is returned, rather than a new one being created.

Since our tests generally use the same dummy content in subsequent runs (and we don't clean up posts afterwards), this change caused a few of our tests to fail. For example, expecting the created post status to be `private`, but instead getting `draft` because that happened to be the status of the post created by an earlier test.

This change just modifies every instance where we're setting the post content in `ReleaseStack_PostTestWPCom` to append the current timestamp, ensuring the post is always considered unique by the API.

The specific tests that were failing were:

```
testClearTagsFromPost
testAutoSaveDoesNotUpdateModifiedDate
testChangePublishedPostToScheduled
testFetchPostStatus
testAddLocationToRemotePost
```

Anyone can review this PR, not necessarily the folks I'm pinging 🙂

### To test:

Run the WPCom post tests, ensure all tests are passing.


malinajirka (approved):
LGTM, thanks!

PR #1926, opened 2021-03-16
Fix first day of week

### LINKED ISSUE #3472: Display Weekly Stats Mon thru Sun (instead of Sun thru Sat)
**Describe the bug**
When viewing stats in the Woo Android app, the week is shown as Sun through Saturday (unlike the browser or iOS app) 
Report here: 3677458-zen

**To Reproduce**
1. Go to WP-Admin
2. Click on Stats
3. Select the "Store" tab and the "Weeks" view
4. Notice that the week is displayed Monday through Sunday (this is not the case in the mobile Android app, it's Sunday through Saturday instead, dividing the weekend.)

**Screenshots**
image

**Expected behavior**
The Android version of the app would display the weekly stats in the same way that the browser and iOS versions do (Monday through Sunday) 

- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).

**Mobile Environment**
Please include:
- Device: multiple
- Android version: 11
- WooCommerce Android version: 5.7

</details>

malinajirka:
Fixes https://github.com/woocommerce/woocommerce-android/issues/3472

Note - I recommend reviewing this PR by commits.

DateUtils class is a bit tricky to use, since AFAICT it doesn't take into account user's Locale (or timezone) in most methods. 

This PR is trying to fix getter for first/last day of week. FirstDayOfWeek is dependant on Locale - eg. in US a week starts with Sunday but in France with Monday. 

https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/90557ea25977c6311f41fb2360e0681295d98e4e - Adds Locale as parameter so we can provide Locale from tests. Fixes "getQuantityInWeeks" so it takes into account locale dependant firstDayOfWeek.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/23134262d1c8ad657db91fb1566da7bd2deb90cb - Add tests for the previous commit.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/be0f34cf647e0a1f0ac2a327d0e6074acb345e39 - Add a test for one more edge case.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/2e86b824f73453dfc87415a20117cbb07d267821 - add tests for getConstantForLastDayOfWeek(..) introduced in the first commit.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/3bedec22f391d8cf81f6a99dd43460484effaf1f - Removes non-deterministic test which relies on VM's settings. This test never passed on my machine but it passes on CI. It doesn't feel like unit test and it's dependant on local environment. I wasn't able to easily come up with a fix, so I decided to replace it with simpler tests in https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/4b5875bb61375e772e6ee52b9b96bab41361506a.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/8d8ced271bcdd8fde1d982d3bfb58804edc81c1a - Makes getFirstDayOfWeek and getLastDayOfWeek methods Locale aware.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/fe118a57e83ff01b30ee4a7bfaa2fecd2f881c27 - Passes Calendar as parameter to make the methods unit testable. Ideally, we'd refactor this whole class from static (object) singleton into an injectable class. We'd also use injectable dateTimeProvider/calendarProvider instead of using static methods. However, this felt like a huge change and not worth the effort at the moment.

To test:
- Run the tests
- [Test WCAndroid app](https://github.com/woocommerce/woocommerce-android/pull/3706)

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/utils/DateUtilsTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/utils/DateUtilsTest.kt
@@ -108,6 +108,13 @@ class DateUtilsTest {
         val endDateCalendar3 = DateUtils.getEndDateCalendar(endDate3, Locale.US)
         val quantity3 = DateUtils.getQuantityInWeeks(startDateCalendar3, endDateCalendar3)
         assertEquals(1, quantity3)
+
+        val startDate4 = DateUtils.getDateFromString("2019-01-14")
+        val endDate4 = DateUtils.getDateFromString("2019-01-20")
+        val startDateCalendar4 = DateUtils.getStartDateCalendar(startDate4, Locale.US)
+        val endDateCalendar4 = DateUtils.getEndDateCalendar(endDate4, Locale.US)
+        val quantity4 = DateUtils.getQuantityInWeeks(startDateCalendar4, endDateCalendar4)
+        assertEquals(2, quantity4)
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


anitaa1990 (approved):
LGTM :shipit: Thanks for adding all the testing scenarios Jirka 🥇

PR #1927, opened 2021-03-17
Replace isButton with isMobileButton in the Note response

planarvoid:
The note response from the API has changed. It should now be sending the `isMobileButton` field


renanferrari (approved):
LGTM 👍

PR #1929, opened 2021-03-18
Remove the failing API call

0nko:
This PR fixes the failing Customers API test by removing an API call that returns the 501 (Not implemented) HTTP error.


nbradbury (approved):
:shipit:

PR #1931, opened 2021-03-24
[Woo] Support decimal value in stock quantity property for Simple and Variation Products.

hafizrahman:
---

### ⚠️ 👋  This PR contains database changes. To avoid complication, this should not be merged until [WooCommerce Android PR #3737](https://github.com/woocommerce/woocommerce-android/pull/3737) is ready as well. 

---

This PR is **part 1 of 3** to [fix this issue](https://github.com/woocommerce/woocommerce-android/issues/3611#issuecomment-805665832).
Required by: woocommerce/woocommerce-android#3737

The overarching goal is to make sure Woo Mobile app can display Simple and Variation products with decimal value in their stock quantity attribute. 

## What this PR does:
1. **[Persistence]** Update models to use Double instead of Int, so decimal values can be stored and displayed on the Woo Mobile side.
2. **[Networking]** Update API responses to use Double instead of Int, so that decimal values from API can be accepted in Woo Mobile.
2. **[Networking]** Because core API does not support decimal values when adding/updating a product, make sure the decimal value in models is converted back to integer before sending to the backend.

## Related tasks to be done in [WooCommerce Android PR 3737](https://github.com/woocommerce/woocommerce-android/pull/3737)
- [ ] The Woo Mobile app has to be adjusted to be able to receive and display decimal values in stock quantity.
- [ ] The Woo Mobile app has to display stock quantity attribute to be read-only, unless it is a rounded decimal (e.g: `10.0`).

---

To test:
1. Run tests in `example/src/test/java/org/wordpress/android/fluxc/wc/product/ProductSqlUtilsTest.kt` and make sure all is OK, especially `testInsertOrUpdateProductWithDecimalQuantity()`

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -1166,7 +1166,10 @@ class ProductRestClient(
         // only allowed to change the following params if manageStock is enabled
         if (updatedProductModel.manageStock) {
             if (storedWCProductModel.stockQuantity != updatedProductModel.stockQuantity) {
-                body["stock_quantity"] = updatedProductModel.stockQuantity
+                // Conversion/rounding down because core API only accepts Int value for stock quantity.
+                // On the app side, make sure it only allows whole decimal quantity when updating, so that
+                // there's no undesirable conversion effect.
+                body["stock_quantity"] = updatedProductModel.stockQuantity.toInt()
             }
             if (storedWCProductModel.backorders != updatedProductModel.backorders) {
                 body["backorders"] = updatedProductModel.backorders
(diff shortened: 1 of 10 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -1166,7 +1166,10 @@ class ProductRestClient(
         // only allowed to change the following params if manageStock is enabled
         if (updatedProductModel.manageStock) {
             if (storedWCProductModel.stockQuantity != updatedProductModel.stockQuantity) {
-                body["stock_quantity"] = updatedProductModel.stockQuantity
+                // Conversion/rounding down because core API only accepts Int value for stock quantity.
+                // On the app side, make sure it only allows whole decimal quantity when updating, so that
+                // there's no undesirable conversion effect.
```
anitaa1990, line 1171: Good call with adding the comments here <3

anitaa1990 (approved):
Nice work on this @hafizrahman 🥇 The code changes look good and I tested the following scenarios without any issues:
- [x] Fetch and update a single product that has a stock quantity in decimal value/without decimal value.
- [x] Fetch and update a single product variation that has a stock quantity in decimal value/without decimal value.
- [x] Tested db migration by updating the example app from a previous version to the version in this PR.

Just a nit from my end that it's worth mentioning other testing steps in the PR description so anyone can test the above mentioned scenarios from the example app along with running the unit tests.

I'll refrain from merging this PR till the Woo PR can be reviewed and merged :shipit:

anitaa1990:
Merging this PR since the Woo PR is also ready to merge 👍 Nice work @hafizrahman!

PR #1933, opened 2021-03-25
Make attributeList a mutable field instead of a lazy single instance

### LINKED ISSUE #1932: addVariant and removeVariant at WCProductVariationModel not working properly
All operations of editing/adding/removing any `ProductVariantOption` get ignored after the first operation is called. This issue is happening because `attributeList` is currently a single instance `val` created lazily that doesn't get updated after the first call.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/69edb50bf272316d08ac7f07847d070a0d0fae02/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductVariationModel.kt#L94

ThomazFB:
Fix issue #1932 by making the `attributeList` property inside `WCProductVariationModel` a fieldless read-only value instead of a locked single instance.

This issue is currently blocking some scenarios for the Variation Attributes tasks.


nbradbury (approved):
:shipit:

PR #1934, opened 2021-03-25
Shipping Labels: update purchase logic to use the API

hichamboushaba:
This implements the changes discussed in this post: p91TBi-4ND-p2, the goal is to use the async flow of the purchase API, and poll the label status endpoint afterwards until the purchase is complete.
It introduces a new class called `Poller` to make reusing the logic easier for other endpoints.

#### Testing
The [change](https://github.com/Automattic/woocommerce-services/pull/2390) to WCShip is not released yet, so you need to use the following steps:
1. Upload the following version of WCShip to your store [woocommerce-services.zip](https://github.com/wordpress-mobile/WordPress-FluxC-Android/files/6210751/woocommerce-services.zip), after releasing 1.25.11 which will contain the change, it will auto update to it.
2. Open the example app.
3. Click on Woo, then "Shipping Labels"
4. Make a SL purchase using the button "purchase shipping label"
5. Confirm that the purchase is done correctly.

And it would be nice too to review the different test cases in the unit tests.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/Poller.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/Poller.kt
@@ -22,7 +22,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
 class Poller(private val delayInMs: Long, private val maxRetries: Int) {
     suspend fun <T : WooPayload<*>> poll(
         request: suspend () -> T,
-        predicate: (T) -> Boolean,
+        predicate: (T) -> Boolean
     ): T {
         return pollInternal(request)
                 .filter { predicate.invoke(it) }
@@ -30,7 +30,7 @@ class Poller(private val delayInMs: Long, private val maxRetries: Int) {
     }
 
     private suspend fun <T : WooPayload<*>> pollInternal(
-        request: suspend () -> T,
+        request: suspend () -> T
     ): Flow<T> {
         var retries = 0
         return flow {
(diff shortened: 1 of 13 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/Poller.kt
```diff
@@ -0,0 +1,52 @@
...
+ * and automatic retrying on errors, the logic is as follows:
+ * 1. Invoke the request.
+ * 2. If the request fails, retry for the number of times specified by maxRetries after the specified delay
+ * 3. If the request succeeds, check the predicate agains the result
+ * 4. If it's true, then return the result.
+ * 5. If it's false, repeat from 1.
+ *
+ * @param delayInMs: the delay between requests in milliseconds
+ * @param maxRetries: the maximum number of retries when an error occurs
+ */
```
0nko, line 21: Great documentation!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/Poller.kt
```diff
@@ -0,0 +1,52 @@
...
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.flow
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
+
+/**
+ * A utility class that handles polling, it allows polling a request, with a defined delay between the requests
+ * and automatic retrying on errors, the logic is as follows:
+ * 1. Invoke the request.
+ * 2. If the request fails, retry for the number of times specified by maxRetries after the specified delay
+ * 3. If the request succeeds, check the predicate agains the result
```
0nko: Typo: "agains"

0nko (approved):
I love how polling works! 👏 The SL purchase is working as described. Really great job, Hicham!

There's just one typo, otherwise it's good to go 👍.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/Poller.kt
hichamboushaba: Fixed in 83cdece

PR #1936, opened 2021-03-26
Filter Jetpack sites for Jetpack app

ashiagr:
### Description
- This PR adds filtering for Jetpack sites if the request to fetch sites comes from the Jetpack app.

### How to test
- WordPress-Android PR:  https://github.com/wordpress-mobile/WordPress-Android/pull/14356

### Merge Instructions

~~This PR targets `feature/jetpack-app-basics-fluxc`. We'll decide next week if we want to merge these changes to `develop`.~~

- Make sure below PRs are reviewed and ready to be merged with respective `develop` branch:
   - https://github.com/wordpress-mobile/WordPress-Android/pull/14397
   - https://github.com/woocommerce/woocommerce-android/pull/3805
- Remove `Not Ready for Merge` label
- Merge this PR with `develop`


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -146,12 +149,14 @@ public SiteRestClient(Context appContext, Dispatcher dispatcher, RequestQueue re
                           AccessToken accessToken, UserAgent userAgent) {
         super(appContext, dispatcher, requestQueue, accessToken, userAgent);
         mAppSecrets = appSecrets;
+        mIsJetPackApp = appContext.getPackageName().contains(JETPACK);
```
ParaskP7: 👋 @ashiagr !

Good idea of using `getPackageName` to figure out whether this is a `jetpack` app or not. However, I am being a bit skeptical about the following:
1) Maybe it is better to have this `context.getPackageName()` flow from the client side that is using the `FluxC` library and as such to pass instead an `isJetpack` boolean as a constructor argument that the clients will set themselves instead of binding the `FluxC` library with a client specific logic. Wdyt?
2) Although `context.getPackageName()` is an easy win, I wonder if this ends up being a bit risky to use. This will only work if the package name contains `jetpack` and the fact that we have a hard coded constant that we need to match to it. This means that if later on, for some unknown reason, the package name changes, due to maybe some Google Play or product constraints, then it might become harder to identify why things are breaking. One should know that this is how it works. Also, it might end up failing because the hardcoded constant is defined incorrectly (like with a typo). Wdyt?

Question (❓): Why not using `BuildConfig` with an `IS_JETPACK` boolean res value instead? Where there any concerns about this vs. the `getPackageName()` solution?

ashiagr:
👋 @ParaskP7 

As per our discussion on `isJetpackApp` handling, instead of getting app info from `context.getPackageName()`, I've added filters param to the `fetchSites()` method. I now pass `jetpack` filter from the client app if Jetpack product flavor is used.

P.S.

1. There were three site filters in the [Calypso code](https://opengrok.a8c.com/source/xref/trunk/wp-content/lib/site-filter.php?r=211256&fi=filter_blog#10), I've supported all of them and added connected tests in f940f79. 
2. Since `SiteStore` and `SiteRestClient` classes don't seem to be tests friendly (also no tests exists for them), I didn't include unit tests. 
3. You might need to update `.mobile-secrets` locally as I updated it.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
ashiagr: We resolved it over Zoom call, and agreed to pass filters from the client app. Even though `SiteRestClient` had access to app context, we decided not to use it following the "separation of concerns" principle for our Rest library and Android code.

instaflux/src/main/java/org/wordpress/android/fluxc/instaflux/MainInstafluxActivity.java
```diff
@@ -178,7 +181,7 @@ public void onAccountChanged(OnAccountChanged event) {
         } else {
             if (!mSiteStore.hasSite() && event.causeOfChange == AccountAction.FETCH_ACCOUNT) {
                 AppLog.d(AppLog.T.API, "Account data fetched - fetching sites");
-                mDispatcher.dispatch(SiteActionBuilder.newFetchSitesAction());
+                mDispatcher.dispatch(SiteActionBuilder.newFetchSitesAction(new FetchSitesPayload(new ArrayList())));
```
ParaskP7: 1) Warning (⚠️): Consider using `ArrayList<>` instead of a raw parameterized class `ArrayList`. PS: Same apply everywhere.
2) Idea (💡): I suggest using an empty default constructor for `FetchSitesPayload`. This way all usages of `new FetchSitesPayload(new ArrayList())` (in Java) and `FetchSitesPayload(emptyList())` (in Kotlin), will become simpler, like `new FetchSitesPayload()` (in Java) and `FetchSitesPayload(emptyList())` (in Kotlin) respectively. This way all the clients will benefit from a simpler api, with the base minimum amount of changes to it.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
```diff
@@ -184,6 +186,15 @@ public void onErrorResponse(@NonNull WPComGsonNetworkError error) {
         add(request);
     }
 
+    @NotNull private Map<String, String> getFetchSitesParams(List<SiteFilter> filters) {
```
ParaskP7: 1) Warning (⚠️): The `import org.jetbrains.annotations.NotNull;` seems to be the wrong non null annotation. Please replace that with the `import androidx.annotation.NonNull;` and remove the wrong import.
2) Minor (🔍): Consider moving the `@NonNull` annotation to a separate line above the method.
3) Minor (🔍): Consider adding a a`@NonNull` annotation for the `filters` method parameter and remove the `filters != null` check. Note that this change will propagate to making the `filters` parameter on `fetchSites(...)` method `@NonNull` as well. I would also suggest to make the `filters` field and constructor arguments of `FetchSitesPayload` a `@NonNull`. This way the api will become much stricter and as such more straightforward to use.
4) Suggestion (💡): I wonder whether it is time to start extracting constants from hardcoded values both, here (to start with) and everywhere else (later on). For example, I am seeing that the `fields` hardcoded values are used twice, and I am afraid that whoever might potentially change that will forget to change the other one. This is why I personally like seeing every hardcoded value, even if that is being used once, to become a constant from the get going. Feel free to ignore my suggestion if you feel differently.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java
```diff
@@ -136,6 +140,79 @@ public void testWPComJetpackMultisiteSiteFetch() throws InterruptedException {
         assertFalse(mSiteStore.hasSitesAccessedViaWPComRest());
     }
 
+    @Test
```
ParaskP7, line 143: 1) Praise (❤️): I love the test you added here, kudos!
2) Suggestion (💡): Maybe this test suite can benefit from an additional test with at least two filters, which will make sure that the `comma` functionality works, both locally and on backend wise. Wdyt?

ParaskP7 (changes requested):
👋 @ashiagr !

I have reviewed and tested this PR as per the instructions, everything works as expected, good job! 🌟 

I have left a couple of warning (⚠️) comments, some minor (🔍) comments and a few suggestions (💡) for you to consider.

I am still skeptical whether this change needs to target the `jetpack` feature branch or directly the `develop`. While reviewing this PR I got to thinking that `Woo` is going to be using this new enhanced api as well, through `newFetchSitesAction(...)`. As such, when that eventually enters `develop`, the `Woo` team would need to update their usages of that api. This makes me think that it is better if that is merged to `develop` sooner that later so that we don't forget about this going forward and so that we don't create too many requirements before merging later on. I also suggest that after this is merged to `develop`, we ourselves go and created two PRs, one on the `WordPress` and one on the `Woo` side to merge this change with the apps' `develop` as soon as possible. This way, all apps will be able to reuse this functionality from there after (`WordPress`, `Jetpack` and `Woo`) and we will stop thinking about that anymore, progressing with the next task in line. Also, since the `tests.properties` where updates to account for this change, new tests where written to support this, this makes me feel that merging to `develop` now is better.

Having said that, if you and @zwarm feel strongly about keeping the feature branch on `FluxC` and do the merge later on, I am okay with that path too. Even thought this path will make us start maintain and update yet another feature branch, on a separate repo, it hopeful shouldn't be that difficult. 🤞

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
ashiagr: 1., 2., 3.: Done in 78a96b7. 
4: Done in 8edb84d. No strong opinion on this.

instaflux/src/main/java/org/wordpress/android/fluxc/instaflux/MainInstafluxActivity.java
ashiagr: 2: Done in 8edb84d
1: Got fixed with 2.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java
ashiagr: 2: Suggested test added in eb41793.

instaflux/src/main/java/org/wordpress/android/fluxc/instaflux/MainInstafluxActivity.java
ParaskP7: 👋 @ashiagr !

Thanks for your changes. 🙏 

About `2:` probably you meant to use this d33ee3ecc9768b4993cd74fe6a3aaac4041375ff commit hash, right? In anyway, I have verified the changes and LGTM. 🌟 

PS: From your changes, I understood that I wasn't clear enough with my suggestion, apologies for that. What I meant is using an `extra` empty constructor, while keeping the other one as well, so that to make the `filters` field private and force a constructor only usage. However, feel free to keep it as is, the end result is similar, the only diff is the exposure of the field, that's all.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java
ParaskP7: 👋 @ashiagr !

Thanks for your changes, LGTM! 🙏 

PS: No strong opinion on this too, just a preference on my side I guess. 😄

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_SiteTestJetpack.java
ParaskP7: 👋 @ashiagr !

Thanks for your changes. 🙏

About `2:` probably you meant to use this 458b96317b43d1b5c352f317c2c940aa2a2d9bbe commit hash, right? In anyway, I have verified the changes and LGTM. 🌟

PS: This improvement you did within eb41793afef99b7401e7849c76ee0d5adbde5ce1 is great! ❤️

ParaskP7 (approved):
👋 @ashiagr !

I have reviewed and tested this PR as per the instructions, everything works as expected, good job! 🌟 

PS: Thanks for the updates! ❤️

PR #1943, opened 2021-04-07
Add System status endpoint client

0nko:
This PR adds the `wc/v3/system_status` endpoint client and it only requests the list of active plugins on the site. It's an in-place replacement for the `WooPluginRestClient` and it supports the loading of the installed & active plugins for users with the _Store manager_ role (previously, shipping label creation wasn't available to them).

I kept the original client because it contains other information and allows to also load inactive plugins, which could be useful in the future. The system status endpoint doesn't have any `slug` property, so I just used the name instead in the database. 

Since the Store method signatures haven't changed, all that's needed in the Woo app is to update the FluxC version.

**To test:**
1. Log in
2. Tap on the Woo button
3. Tap on the Shipping labels button
4. Select site with WCShip extension
5. Tap on Get shipping plugin info button
6. Verify the correct plugin's information is printed

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/module/ReleaseWCNetworkModule.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/module/ReleaseWCNetworkModule.kt
@@ -23,6 +23,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.product.ProductRestClie
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.product.attributes.ProductAttributeRestClient
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.refunds.RefundRestClient
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.shippinglabels.ShippingLabelRestClient
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.system.WooSystemRestClient
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.taxes.WCTaxRestClient
 import javax.inject.Named
 import javax.inject.Singleton
@@ -146,6 +147,17 @@ class ReleaseWCNetworkModule {
         token: AccessToken,
         userAgent: UserAgent
     ) = WooPluginRestClient(dispatcher, wpComBuilder, appContext, requestQueue, token, userAgent)
+
+    @Singleton
+    @Provides
+    fun provideWooSystemRestClient(
+        appContext: Context,
+        requestBuilder: JetpackTunnelGsonRequestBuilder,
+        dispatcher: Dispatcher,
+        @Named("regular") requestQueue: RequestQueue,
+        token: AccessToken,
+        userAgent: UserAgent
+    ) = WooSystemRestClient(dispatcher, requestBuilder, appContext, requestQueue, token, userAgent)
 
     @Singleton
     @Provides
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
@@ -1,0 +1,59 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.system
+
+import android.content.Context
+import com.android.volley.RequestQueue
+import com.google.gson.annotations.SerializedName
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.generated.endpoint.WOOCOMMERCE
+import org.wordpress.android.fluxc.generated.endpoint.WPCOMREST
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.UserAgent
+import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Error
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Success
+import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackError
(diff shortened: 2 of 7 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
```diff
@@ -0,0 +1,57 @@
...
+    userAgent: UserAgent
+) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
+    suspend fun fetchActivePlugins(site: SiteModel): WooPayload<ActivePluginsResponse> {
+        val url = WOOCOMMERCE.system_status.pathV3
+
+        val response = jetpackTunnelGsonRequestBuilder.syncGetRequest(
+                this,
+                site,
+                url,
+                mapOf("_fields" to "active_plugins"),
```
hichamboushaba: I didn't take time to review this yet, I just wanted to suggest something, WDYT about including `inactive_plugins` as well, to keep the same feature as before, by returning all installed plugins, with their correct status?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/module/ReleaseWCNetworkModule.kt
```diff
@@ -147,6 +148,17 @@ class ReleaseWCNetworkModule {
         userAgent: UserAgent
     ) = WooPluginRestClient(dispatcher, wpComBuilder, appContext, requestQueue, token, userAgent)
 
+    @Singleton
```
wzieba, line 151: What's the reason to treat `WooSystemRestClient` as a singleton?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
+import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackError
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackSuccess
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.toWooError
+import javax.inject.Singleton
+
+@Singleton
```
wzieba, line 19: `@Singleton` is already applied in provide method above, maybe we could remove it from here?
```suggestion
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/module/ReleaseWCNetworkModule.kt
0nko: All the other REST clients are singletons so I used the same approach. I think the reason is to use a single request queue, but I'm not 100% sure.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
0nko: Sure 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
```diff
@@ -0,0 +1,60 @@
...
+class WooSystemRestClient
+constructor(
+    dispatcher: Dispatcher,
+    private val jetpackTunnelGsonRequestBuilder: JetpackTunnelGsonRequestBuilder,
+    appContext: Context?,
+    requestQueue: RequestQueue,
+    accessToken: AccessToken,
+    userAgent: UserAgent
+) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
+    suspend fun fetchActivePlugins(site: SiteModel): WooPayload<ActivePluginsResponse> {
```
hichamboushaba: nit: now that we are including inactive plugins, I think this should be renamed to something like: `fetchInstalledPlugins`

hichamboushaba (approved):
Nice 👍, I tested it, and it works as expected, I just left one comment regarding the naming of the function `fetchActivePlugins`, and there is a failing lint check.
Feel free to merge once both are addressed.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
0nko: Oh right, thanks!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
0nko: `[ERROR] /home/circleci/project/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt:17: Network clients must be annotated with @Singleton [RegexpMultiline]` 

I'm reverting the commit.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/module/ReleaseWCNetworkModule.kt
wzieba: Gotcha. I see other `*RestClients` are injected in the same way, I'm not sure if it's desired behavior though. I don't see that `WooSystemRestClient` keeps any state and `RequestQueue` is injected so making `WooSystemRestClient` a singleton or not shouldn't create any queue.

My concerns are about a memory leak - here we create a class that will live through the whole app lifecycle but we might not use it. WDYT?

Nitpick: we could also use constructor injection to reduce some boilerplate, WDYT?

0nko: Whoops, I just noticed your comment after merging. I'll take a look at the refactoring and create another PR.

wzieba: @0nko sure no problem! Now I see [your comment below](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1943#discussion_r609928937) with 

> Network clients must be annotated with @Singleton

which is I believe custom lint check. So there must be a reason to keep network clients as singletons.

PR #1944, opened 2021-04-08
Shipping label creation eligibility

hichamboushaba:
This PR integrates the shipping [label creation eligibility endpoint](https://github.com/Automattic/woocommerce-services/pull/2334), as discussed on Slack, I decided to cache the result, to follow the same flow as other order details, and to avoid abrupt UI changes when loading the order details.

The PR includes a second change fc14073, which does the following:
1. Renames `localOrderId` in the shipping label model to `remoteOrderId` as it's what it contains.
2. Removes the columns holding the account settings, as they are not used by the app, and they are not linked to a single shipping label, but to the account. The same settings are fetched using https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/a20c110c79a8918e49563ec28585b6d59e07f1fc/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCShippingLabelStore.kt#L296
And if we need to cache them, we should do it in a separate table anyway.

This change can be in a separate PR, as it's unrelated, but I decided to combine both of them to handle all the DB changes in one single migration.

#### Testing
1. Open the example app.
2. Click on Woo, then "Shipping Labels"
3. Click on "Check Label Creation Eligibility"
4. Enter the order id, and the other parameters.
5. Confirm that the eligibility is printed in the logs.
6. Retry with the same parameters and confirm that the eligibility is fetched from the DB.
7. Retry with different parameters and confirm that it's fetched from the API.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/shippinglabels/WooShippingLabelFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/shippinglabels/WooShippingLabelFragment.kt
@@ -462,7 +462,7 @@ class WooShippingLabelFragment : Fragment() {
                         val label = it.first()
                         prependToLog(
                                 "Purchased a shipping label with the following details:\n" +
-                                        "Order Id: ${label.localOrderId}\n" +
+                                        "Order Id: ${label.remoteOrderId}\n" +
                                         "Products: ${label.productNames}\n" +
                                         "Label Id: ${label.remoteShippingLabelId}\n" +
                                         "Price: ${label.rate}"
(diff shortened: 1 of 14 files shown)


0nko (approved):
Tested different scenarios and it worked as described. Nice work! 💯

0nko:
Is there something that needs to be done before merging this?

hichamboushaba:
> Is there something that needs to be done before merging this?

I had the `Not Ready For Merge` just to be sure that the WCAndroid PR is reviewed at the same time, I'm merging it now.
Thanks @0nko

PR #1946, opened 2021-04-09
DI boilerplate cleanup

0nko:
This PR removes the unnecessary `@Provides` methods and replaces them with constructor injection. No functionality should be affected.

**To test:**
- Smoke test the app
- Run unit tests
- Run connected tests


example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedNetworkModule.java
```diff
@@ -55,14 +28,23 @@ public ResponseMockingInterceptor provideResponseMockingInterceptor() {
...
+    @Named("custom-ssl")
+    public OkHttpClient provideCustomOkHttpClientInstance(OkHttpClient.Builder builder) {
+        return builder.build();
+    }
+
+    @Singleton
+    @Provides
+    @Named("no-redirects")
+    public OkHttpClient provideNoRedirectOkHttpClientInstance(OkHttpClient.Builder builder) {
+        return builder.build();
```
wzieba: I don't know if you want to do this in scope of this PR but, according to `OkHttp` docs, 

> OkHttpClients Should Be Shared

https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#okhttpclients-should-be-shared

WDYT about merging those 3 instances into one? (here and in `release` modules)

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/vertical/VerticalRestClient.kt
```diff
@@ -14,17 +14,18 @@ import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
...
 
 @Singleton
-class VerticalRestClient
-constructor(
+class VerticalRestClient @Inject constructor(
     dispatcher: Dispatcher,
     private val wpComGsonRequestBuilder: WPComGsonRequestBuilder,
     appContext: Context?,
-    requestQueue: RequestQueue,
+    @Named("regular") requestQueue: RequestQueue,
```
wzieba, line 28: Might be out of scope of this PR but maybe you'd be interested in this improvement:

`@Named` seems to me as quite typo error prone solution. It's also quite difficult to navigate in IDE, I usually have to look for `@Named("regular")` string in project to find usages of it.

I'd like to suggest a custom `@Qualifier` annotation, like this:

```kotlin
@Qualifier
@Retention(AnnotationRetention.RUNTIME)
annotation class RequestQueue(val type: RequestQueueType) {
}

enum class RequestQueueType {
    REGULAR, NO_REDIRECTS, CUSTOM_SSL
}
```

and then use it like

```kotlin
@RequestQueue(REGULAR) requestQueue: RequestQueue,
```

wzieba (approved):
Constructor injection ftw! Thank you, that's good improvement - I like the numbers of line changed.

I've added two suggestions. Those might be out of scope of this PR but you I'd appreciate your opinion on those. Thanks!

example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedNetworkModule.java
0nko: Each time a new client type was created, the previous settings that are not overwritten would need to be reset. I'm worried this could break something and would require more rigorous testing, so I'd leave it for another PR if that's ok.

wzieba: You're right, good point with previous settings of a client. I'm totally ok to leave it for another PR - I'll willingly handle it in the scope of the Slow Orders project.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/vertical/VerticalRestClient.kt
0nko: I think it'll add a lot of unrelated changes, let's keep it for a different PR.

example/src/androidTest/java/org/wordpress/android/fluxc/module/MockedNetworkModule.java
wzieba: I've created an issue: https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1949

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/vertical/VerticalRestClient.kt
wzieba: > I think it'll add a lot of unrelated changes, let's keep it for a different PR.

I agree. I've created issue so we won't forget about it later: https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1950

wzieba:
I've merged `develop` and added one change in 710a872e9b5f8e67b2fd1887da5df72037a74206 commit connected with change introduced in #1962

PR #1947, opened 2021-04-11
Parse preferred blog info and bio from like endpoints

develric:
This PR adds the parsing of preferred blog info and bio from like endpoints. It is a companion PR to the WPAndroid PR [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14437). You can use that PR and description to test this.

Requires D59220 endpoint changes to be shipped first (now deployed).


khaykov (approved):
Looks good 👍 Confirmed that it works well with the client side PR. Feel free to merge whenever you are ready.

PR #1948, opened 2021-04-12
Take into account shipping lines in API responses related to refund.

hafizrahman:
This PR is required to enable [Woo Mobile PR #3825](https://github.com/woocommerce/woocommerce-android/pull/3825).

Update: This PR is already approved, but let's postpone merging until the above PR is ready.

## What this PR does.

When a refund involves shipping lines, the API responses for both creating and fetching such refund treats the shipping lines separately from refunded products/items. 

An example response for fetching a single Refund where a shipping line is refunded looks like this:
```
{
  "data": [
  {
    "id": 181,
    "date_created": "2021-04-12T14:37:56",
    "date_created_gmt": "2021-04-12T07:37:56",
    "amount": "15.00",
    "reason": "",
    "refunded_by": 1032737,
    "meta_data": [],
    "line_items": [],
    "shipping_lines": [
      {
        "id": 87,
        "method_title": "Shipping",
        "method_id": "",
        "instance_id": "0",
        "total": "-1.00",
        "total_tax": "0.00",
        "taxes": [],
        "meta_data": [
          {
            "id": 651,
            "key": "_refunded_item_id",
            "value": "72",
            "display_key": "_refunded_item_id",
            "display_value": "72"
          }
        ]
      }
    ]
}
```

Note these two separate arrays within the main `data` object:
- "line_items": for refunded products, this one's empty because in the example we're not refunding a product
- "shipping_lines": for refunded shipping lines

With this PR, we're taking into account the values returned in `shipping_lines` by putting them the WCRefundModel. My use case right now is so that I can grab the refunded item id data from the app's DB, so I can:
- Display information about already refunded shipping lines
- Exclude already refunded shipping lines being refunded again.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/refunds/RefundFixtures.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/refunds/RefundFixtures.kt
@@ -8,5 +8,6 @@ val REFUND_RESPONSE = RefundResponse(
         "10.00",
         "No reason",
         false,
+        emptyList(),
         emptyList()
 )
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/refunds/RefundMapper.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/refunds/RefundMapper.kt
@@ -31,7 +31,8 @@ class RefundMapper
                             it.sku,
                             it.price?.toBigDecimalOrNull()
                     )
-                }
+                },
+                response.shippingLineItems
         )
     }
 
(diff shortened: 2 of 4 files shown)


0nko (approved):
Looking good! 👍

0nko:
I just noticed that the target branch is `develop`. Is that correct? The reference PR in the description is targeting `feature/add-shipping-refund`.

hafizrahman:
> I just noticed that the target branch is develop. Is that correct? The reference PR in the description is targeting feature/add-shipping-refund.

Good catch and thank you for the review! I didn't realize that before. It should be OK to merge this right to develop, but let's wait until [Woo Android #3825](https://github.com/woocommerce/woocommerce-android/pull/3825) is ready as well.

PR #1952, opened 2021-04-12
Woo Shipping Labels: expose box_weight in the package data model

hichamboushaba:
To implement https://github.com/woocommerce/woocommerce-android/issues/3829, we need to include the box_weight in the calculation, this PR exposes it.


0nko (approved):
:shipit:

PR #1955, opened 2021-04-14
Merge Release/1.14.1 into develop 

JavonDavis:


mokagio (approved):
🚀 

We need this in so that we can then make https://github.com/wordpress-mobile/WordPress-Android/pull/14432 use the change 👍

PR #1956, opened 2021-04-14
Create and delete variation Endpoints

ThomazFB:
Summary
==========
This PR defines endpoints methods to delete or generate an empty variation from a given Product. 

- `generateEmptyVariation` method request the creation of a variation with no value defined and attributes set to `Any` 
- `deleteVariation` completely deletes a variation from the Product.

How to Test
==========

| Enter the Product section inside Woo | Click on Update Product | Interact with generate and delete variations, check the site to verify if changes were applied |
| Screenshot_20210416_031747 | Screenshot_20210416_031605 | Screenshot_20210416_031531 |


nbradbury:
@ThomazFB Should the generate & delete variation buttons be hidden for non-variable products? I guess the same goes for the attribute buttons, too.

ThomazFB:
Hey @nbradbury, thanks for your review! Your suggestion makes total sense, but just to follow how the view handles fields for dedicated product types, I've decided to leave the button disabled instead of hiding it, WDYT about this solution? 

Here's the change: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1956/commits/6c341d366a7f50baa891669c2978fa0828d22658

ThomazFB:
> Sorry to be nitpicky, but this could be a simple if statement. Also, we can avoid the hardcoded "variable" by comparing to CoreProductType

No apologies needed at all @nbradbury, I've been taking bad practices under the FluxC example app, and your suggestion is definitely needed, thanks for pointing that! 🙇  Here's the fix: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1956/commits/ecc11029b2be2346b2a7d53f6264fe92cca88532

ThomazFB:
> This fails because CoreProductType.VARIABLE.toString() returns uppercase "VARIABLR".

Oh gosh 🤦 , sorry, Nick. Went too confident on the small fix and didn't notice the `toString()` instead of `value`. I should've tested this simple fix to avoid my mistake, thanks for your patience. It's fixed here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1956/commits/3035bc6a68a52416b259829ece6cd1c0668ea53c

nbradbury (approved):
:shipit:

PR #1957, opened 2021-04-14
issue/1953-woo-update-product-attributes

### LINKED ISSUE #1953: Woo: Don't update product attributes when they haven't changed
[This WCAndroid issue](https://github.com/woocommerce/woocommerce-android/issues/3835) involves not passing product attributes when updating a product if the attributes haven't changed. This turns out to be an issue in FluxC rather than WCAndroid.

nbradbury:
Fixes #1953 - previously updating a product would always pass the list of attributes, even when they haven't changed. This was because we did a naive comparison of the attributes as a string, which would lead FluxC to think attributes had changed when only the order of elements had changed.

This PR addresses this by comparing the attributes as models rather than strings.

This also moves the `ProductAttribute.toJson()` function from WCAndroid, since it makes more sense to have that in FluxC. Last but not least, it corrects a bug in the WCAndroid version of that function that would output an attribute's options as a simple comma-separated string rather than a JSON array, resulting in a crash when an array was expected.

To test, use Flipper with [this draft WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/3840) and:

- Go to product detail
- Make a simple change
- Update the product
- Verify via Flipper that attributes aren't sent

Then:

- Make a change to an attribute
- Update the product
- Verify via Flipper that attributes _are_ sent

PS: This really should've been a single commit, but I had to add several commits in order to test the changes in WCAndroid.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
@@ -201,9 +201,6 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
      * Returns true if this product has the same attributes as the passed product
      */
     fun hasSameAttributes(otherProduct: WCProductModel): Boolean {
-        AppLog.e(T.API, "this attributes = ${this.attributes}")
-        AppLog.e(T.API, "other attributes = ${otherProduct.attributes}")
-
         // do a quick string comparison first so we can avoid parsing the attributes when possible
         if (this.attributes == otherProduct.attributes) {
             return true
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -147,6 +155,20 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
...
+            val jsonOptions = JsonArray().also {
+                for (option in options) {
+                    it.add(option)
+                }
+            }
+            return JsonObject().also { json ->
+                json.addProperty("id", id)
+                json.addProperty("name", name)
+                json.addProperty("visible", visible)
+                json.add("options", jsonOptions)
```
anitaa1990, line 170: Just want to note that we need to include the `variation` property here as well. This was added as part of the fix in woocommerce/woocommerce-android#3834

nbradbury: Thanks, @anitaa1990! Added in 7a36f1d.

ThomazFB (approved):
Thanks for fixing this, Nick. It was a weird mistake on my part to use string comparison to verify such a complex data change. I've tested this change with my current FluxC branch attached to the Woo variation/attributes codebase, and this `toJson` new method definition now also avoids us to remove the `use for variation` mark by accident, so it's solving two problems in one shot! Awesome job! LGTM :shipit:

PR #1962, opened 2021-04-15
Share single instance of `OkHttpClient`

### LINKED ISSUE #1949: Share single instance of the `OkHttpClient`
We should create new `OkHttpClient`s variations based on existing, base `OkHttpClient` using `OkHttpClient#newBuilder`. In current approach with creating `OkHttpClient`s based on `OkHttpClient.Builder` , we don't share common thread and connection pool which results with limited http calls performance and wasting resources.

See [code snippet for e.g. `ReleaseOkHttpClientModule`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/e7af20209b24c9eb358c229d7b7a943baaf7893e/fluxc/src/main/java/org/wordpress/android/fluxc/module/ReleaseOkHttpClientModule.java) and [`OkHttp` docs](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#okhttpclients-should-be-shared) for more details.

wzieba:
Resolves: #1949 

### Context

This PR is about improving `http` communication performance by sharing the same instance of `OkHttpClient` for different variants of `OkHttpClient`s we use in `FluxC`: `regular`, `no-redirects` and `custom-ssl`

To read more about how it'll improve performance, please [see `OkHttp` docs.](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#okhttpclients-should-be-shared)

### Testing

I don't have any test cases as this change affects all `http` communication for both `WPAndroid` and `WCAndroid` so I can only think of smoke testing those apps.

I've tested `WCAndroid` and all works fine. We shouldn't expect any changes.


example/src/androidTest/java/org/wordpress/android/fluxc/ditests/OkHttpReleaseInjectionTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    @Test
+    fun assertNoRedirectsOkHttpClientBuilderHasExpectedSettings() {
+        noRedirectsOkHttpClient.apply {
+            assertThat(cookieJar).isEqualTo(customCookieJar)
+            assertThat(followRedirects).isFalse
+            assertThat(hostnameVerifier).isEqualTo(DEFAULT_HOSTNAME_VERIFIER)
+
+            assertThat(connectTimeoutMillis).isEqualTo(BaseRequest.DEFAULT_REQUEST_TIMEOUT)
+            assertThat(readTimeoutMillis).isEqualTo(BaseRequest.UPLOAD_REQUEST_READ_TIMEOUT)
+            assertThat(writeTimeoutMillis).isEqualTo(BaseRequest.DEFAULT_REQUEST_TIMEOUT)
```
wzieba, line 54: Comparing to current implementation, this PR updates `no-redirects` client to have the same timeouts as [`custom-ssl`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5f2fb2f652ec711d17c5b7392c54491b1259f370/fluxc/src/main/java/org/wordpress/android/fluxc/module/ReleaseOkHttpClientModule.java#L66-L71) and [`regular`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5f2fb2f652ec711d17c5b7392c54491b1259f370/fluxc/src/main/java/org/wordpress/android/fluxc/module/ReleaseOkHttpClientModule.java#L77-L82).

[In PR that introduced `no-redirects`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1771) and in usages I haven't spotted any reason to have different timeouts on this client.

In the end though, this won't change anything in current usage of this client. We use it only in [`ReleaseNetworkModule#provideNoRedirectsRequestQueue`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5f2fb2f652ec711d17c5b7392c54491b1259f370/fluxc/src/main/java/org/wordpress/android/fluxc/module/ReleaseNetworkModule.java#L113) which wraps client in [`OkHttpStack`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5f2fb2f652ec711d17c5b7392c54491b1259f370/fluxc/src/main/java/org/wordpress/android/fluxc/network/OkHttpStack.java#L32) which [eventually sets timeouts of Volley's request](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5f2fb2f652ec711d17c5b7392c54491b1259f370/fluxc/src/main/java/org/wordpress/android/fluxc/network/OkHttpStack.java#L92-L94).

Motivation for this change is to achieve consistency between `no-redirects` and other clients in future usages.

wzieba:
hi @antonis @planarvoid and @aforcier 👋  

I've requested a review from you as Github suggested me to do so - it seems you've been recently making changes to those files so you might be familiar with them. Is it ok for you to review this PR?

aforcier:
@wzieba I'll plan on looking at this PR next week (but no worries if someone else is able to get to it before I do 🙂)

example/src/androidTest/java/org/wordpress/android/fluxc/ditests/OkHttpReleaseInjectionTest.kt
antonis: > [In PR that introduced `no-redirects`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1771) and in usages I haven't spotted any reason to have different timeouts on this client.

That's true. Thank you for aligning this 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/network/OkHttpStack.java
```diff
@@ -89,9 +88,11 @@ public HttpResponse executeRequest(Request<?> request, Map<String, String> addit
...
         int timeoutMs = request.getTimeoutMs();
 
-        mClientBuilder.connectTimeout(timeoutMs, TimeUnit.MILLISECONDS);
-        mClientBuilder.readTimeout(timeoutMs, TimeUnit.MILLISECONDS);
-        mClientBuilder.writeTimeout(timeoutMs, TimeUnit.MILLISECONDS);
+        final OkHttpClient timeoutAwareClient = mOkHttpClient.newBuilder()
+                                                             .connectTimeout(timeoutMs, TimeUnit.MILLISECONDS)
+                                                             .readTimeout(timeoutMs, TimeUnit.MILLISECONDS)
+                                                             .writeTimeout(timeoutMs, TimeUnit.MILLISECONDS)
+                                                             .build();
```
antonis, line 95: My understanding is that a new instance `timeoutAwareClient` of `OkHttpClient` is created in each `executeRequest`. Is that correct? (probably I'm missing something in how this works)

fluxc/src/main/java/org/wordpress/android/fluxc/network/OkHttpStack.java
```diff
@@ -113,13 +114,8 @@ public HttpResponse executeRequest(Request<?> request, Map<String, String> addit
...
 
 
-        for (Interceptor interceptor : mClientBuilder.interceptors()) {
-            mClientBuilder.addNetworkInterceptor(interceptor);
-        }
-
-        OkHttpClient client = mClientBuilder.build();
         okhttp3.Request okHttpRequest = okHttpRequestBuilder.build();
-        Call okHttpCall = client.newCall(okHttpRequest);
+        Call okHttpCall = timeoutAwareClient.newCall(okHttpRequest);
```
antonis, line 118: Can we use the `mOkHttpClient` here?

antonis (commented):
Great work @wzieba 👍 
I focused my testing effort on `WordPress-Android` and did not notice any issues with the introduced performance optimization.
The added test cases look great and overall the code changes LGTM 🎉 
I added a couple of comments that are questions and not necessarily change request.

fluxc/src/main/java/org/wordpress/android/fluxc/network/OkHttpStack.java
wzieba: Yes, that's true. This is also [the current behavior](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5f2fb2f652ec711d17c5b7392c54491b1259f370/fluxc/src/main/java/org/wordpress/android/fluxc/network/OkHttpStack.java#L121) of `OkHttpStack` but instead of creating a new `OkHttpClient` based on `Builder`, we do this based on existing client. This way we share common connection and thread pools between all calls.

fluxc/src/main/java/org/wordpress/android/fluxc/network/OkHttpStack.java
wzieba: Unfortunately we can't - the point of `timeoutAwareClient` is to append `connect`, `read` and `write` timeouts from the `com.android.volley.Request`.

I was thinking if it makes sense as we have our own custom timeouts set for `OkHttpClient` but digging more into this I realized that it eventually makes sense. Citing `Volley`'s documentation on `Request#getTimeoutMs`

>    /**
     * Returns the socket timeout in milliseconds per retry attempt. (This value can be changed per
     * retry attempt if a backoff is specified via backoffTimeout()). If there are no retry attempts
     * remaining, this will cause delivery of a {@link TimeoutError} error.
     */


It seems that `com.android.volley.Request` manipulates timeouts to optimize queue so it's probably worth to set those custom timeouts for every request. Also, [the original implementation](https://gist.github.com/LOG-TAG/3ad1c191b3ca7eab3ea6834386e30eb9#file-volleyokhttp3stackinterceptors-java-L96-L98) which we took also have timeouts applied from a `com.android.volley.Request` to the `OkHttpClient`.

wzieba:
Thank you for the review @antonis ! I've answered your questions and concerns, ready for second round.

fluxc/src/main/java/org/wordpress/android/fluxc/network/OkHttpStack.java
antonis: Thank you for the explanation @wzieba! That makes sense :)

antonis (approved):
Hey @wzieba 👋 
thank you for the explanation 🙇 .
Everything LGTM 🎉 

> I've answered your questions and concerns, ready for second round.

My concern was mainly around the PR description mentioning a shared instance of `OkHttpClient` and wanted to make sure that the new instance there was intentional.
Great job 👍

wzieba:
@antonis thank you! 🙏

PR #1963, opened 2021-04-15
Woo: Expose Shipping Label expiry date

### LINKED ISSUE #3845: Shipping Labels: should I be allowed to reprint shipping labels that are more than 120 days old?
**To prepare:**

* Be on a site with the WooCommerce Shipping & Tax plugin installed that has a completed order with a shipping label created more than 120 days old (or several months up to a year).

**To test:** 

1. Open an order that contains a very old shipping label.
1. Tap "Reprint shipping label".

**Result:** it's possible to reprint shipping labels that are more than 120 days old. 

**Questions:** Should I be allowed to reprint shipping labels that are more than 120 days old? Will the number of days a shipping label is permitted to be reprinted vary depending on the shipping carrier?

old-label
<sup>Tested on a VANKYO S20 tablet running Android 9 h/t @allendav thank you for the report & screenshot!</sup>

<br><br>In comparison, the reprint option is disabled for the same shipping label in WP Admin.

screen-shot-2021-04-14-at-2 18 29-pm
<sup>(internal reference: p5T066-28r-p2#comment-7925 h/t @allendav)</sup>

hichamboushaba:
This is needed to fix https://github.com/woocommerce/woocommerce-android/issues/3845, we need to disable reprinting expired shipping labels, and the check depends on the `expiry_date` field.
The PR adds also the following change: change the type of dates in `WCShippingLabelModel` to long, as storing them as TEXT doesn't add any value. 

#### Testing
1. Open the example app.
2. Click on Woo.
3. Click on Shipping Labels.
4. Click on Fetch Shipping Labels.
5. Confirm that the fetching works.

Please don't merge until the corresponding woo PR is reviewed.


nbradbury (approved):
:shipit:

PR #1966, opened 2021-04-17
Fix/attribute list null value

ThomazFB:
Summary
==========
This PR address a test bug happening at `ReleaseStack_WCProductTest#testUpdateDownloadableFiles`, the reason for the issue is that `attributeList` can return null from the JSON serialization when the `attributes` field is also null.

How to test
==========
Execute the Release stack tests and verify they are all green ✅


nbradbury (approved):
Fix looks good! :shipit:

PR #1969, opened 2021-04-19
ExPlat: Add name property to all Variations and remove unnecessary payload

renanferrari:
This PR updates the ExPlat implementation with two general improvements:

1. 2db91c8 → Adds a `name` property to all `Variation`s, including `Control`, to make it easier to log its value (we can now just call `Variation.name` with no need to check its type first).
2. a9e9ec7 → Removes unnecessary `FetchAssignmentsPayload` so that less boilerplate is required when calling `ExperimentStore::fetchAssignments`.

### To test

I'd trust the existing unit tests with this one, so just make sure they all pass.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/experiments/AssignmentsModel.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/experiments/AssignmentsModel.kt
@@ -30,13 +30,14 @@ data class Assignments(
     }
 }
 
-sealed class Variation {
-    object Control : Variation()
-    data class Treatment(val name: String) : Variation()
-
+sealed class Variation(open val name: String) {
+    object Control : Variation(CONTROL)
+    data class Treatment(override val name: String = TREATMENT) : Variation(name)
     companion object {
+        private const val CONTROL = "control"
+        private const val TREATMENT = "treatment"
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/experiments/AssignmentsModel.kt
```diff
@@ -30,13 +30,14 @@ data class Assignments(
     }
 }
 
-sealed class Variation {
-    object Control : Variation()
-    data class Treatment(val name: String) : Variation()
-
+sealed class Variation(open val name: String) {
+    object Control : Variation(CONTROL)
+    data class Treatment(override val name: String = TREATMENT) : Variation(name)
```
planarvoid: does this need the default value?

planarvoid (approved):
Changes are looking good and the tests are passing 👍  I have one NIT question but feel free to merge this

fluxc/src/main/java/org/wordpress/android/fluxc/model/experiments/AssignmentsModel.kt
renanferrari: Not really, good catch! Fixed in b23fd38.

PR #1970, opened 2021-04-19
ExPlat: Add support to experiment_names parameter

renanferrari:
This PR updates the ExPlat implementation so it can support the new `experiment_names` parameter which was added to the "fetch assignments" endpoint, as defined in https://github.com/Automattic/experimentation-platform/issues/606.

This depends on #1969.

In practice, this mainly impacts the usage of `ExperimentStore::fetchAssignments`, which now requires a list of `String` with the experiment names as its second parameter:

```kotlin
val platform = Platform.WORDPRESS_ANDROID

// Before
val result = experimentStore.fetchAssignments(platform)
result.assignments.variations // Contained variations for all active experiments

// Now
val result = experimentStore.fetchAssignments(platform, listOf("experiment_one", "experiment_two"))
result.assignments.variations // Contains variations only for experiment_one and experiment_two (if they are active)
```

### To test

Make sure all unit tests pass and/or use the following steps on the example app:

1. On the next screen, select a platform to query.
1. Under "Experiment names", add the name of all experiments which you want to request, separated by a comma (keep in mind only active experiments will be returned, so check Abacus if you want to see which ones to request).
1. Optionally, you can add or generate an Anonymous ID to be included in the request as well.
1. Tap **Fetch Assignments**.
1. Check the results printed by the in-app console and verify they match what was requested.

Repeat providing single/multiple experiment names and also no experiment name at all (which should return 0 assignments).

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ExperimentsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ExperimentsFragment.kt
@@ -60,8 +60,6 @@ class ExperimentsFragment : Fragment() {
                     "anonymousId=$anonymousId")
             GlobalScope.launch(Dispatchers.Default) {
                 val result = experimentStore.fetchAssignments(platform, experimentNames, anonymousId)
-                result.assignments.variations // Contained variations for experiments which were currently running
-                result.assignments.variations // Contains variations only for the requests experiments
                 withContext(Dispatchers.Main) {
                     onAssignmentsFetched(result)
                 }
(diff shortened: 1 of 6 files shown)


example/src/main/java/org/wordpress/android/fluxc/example/ExperimentsFragment.kt
```diff
@@ -54,10 +54,14 @@ class ExperimentsFragment : Fragment() {
...
+            val experimentNames = experiment_names_edit_text.text.split(',').map { it.trim() }
             val anonymousId = anon_id_edit_text.text.toString()
-            prependToLog("Fetching assignments with: platform=$platform, anonymousId=$anonymousId")
+            prependToLog("Fetching assignments with: platform=$platform, experimentNames=$experimentNames, " +
+                    "anonymousId=$anonymousId")
             GlobalScope.launch(Dispatchers.Default) {
-                val result = experimentStore.fetchAssignments(platform, anonymousId)
+                val result = experimentStore.fetchAssignments(platform, experimentNames, anonymousId)
+                result.assignments.variations // Contained variations for experiments which were currently running
+                result.assignments.variations // Contains variations only for the requests experiments
```
planarvoid: I'm not sure I understand these two lines. They are exactly the same and don't do anything with the value. What am I missing? Is it for debugging purposes?

renanferrari: Oops, those were not supposed to be there. Fixed in ab44de3.

planarvoid (approved):
thanks for the fix 👍

PR #1972, opened 2021-04-22
ExPlat: Fix request parameter separator to prevent adding spaces

renanferrari:
This PR fixes a bug that went undetected on #1970, where only the first experiment was being returned when the `experiment_names` parameter was separated with a comma + space, instead of just a comma.

The change is pretty straightforward as we just updated the separator parameter in the `joinToString` function.

### To test

Make sure all unit tests pass and/or use the following steps on the example app:

1. On the next screen, select `wpandroid`.
1. Under "Experiment names", add the following experiment names, separated by a comma (you can add as many spaces as you like there):
    - `explat_test_android_mock_experiment_1`
    - `explat_test_android_mock_experiment_2`
1. Tap **Fetch Assignments**.
1. Check the results printed by the in-app console and verify they contain assignments for both experiments.


planarvoid (approved):
LGTM 👍

PR #1974, opened 2021-04-28
Generate Variations only with Valid Attributes

ThomazFB:
Summary
==========
This PR address a small improvement needed by https://github.com/woocommerce/woocommerce-android/pull/3912 changes in order to avoid unnecessary data to be sent to the Store when updating a Product Variation.

How to Test
==========
1. Go to a Variable Product at `Update Product` section
2. Click at the `Generate Variation` button
3. Inspect the request and verify that only the Attributes enabled in the site as `use for variations` was sent inside JSON body


nbradbury (approved):
Looks good! :shipit:

PR #1975, opened 2021-04-28
Added method to fetch the currency symbol for a store + unit test

anitaa1990:
This tiny PR just adds a separate method to `WooCommerceStore` to fetch the currency symbol, without the currency value, for a given store. This is needed for fixing woocommerce/woocommerce-android#3651.

### To test
- Run `MockedStack_WCBaseStoreTest.kt`.


hichamboushaba (approved):
LGTM 👍 :shipit:

PR #1976, opened 2021-04-28
Merge `DebugOkHttpClientModule` and `ReleaseOkHttpClientModule`

### LINKED ISSUE #1961: Merge `DebugOkHttpClientModule` and `ReleaseOkHttpClientModule`
### Issue

It seems that both modules provide almost the same clients with one difference: `DebugOkHttpClientModule` allows to append `okhttp3.Interceptor` to the clients. See 
- [`DebugOkHttpClientModule`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5f2fb2f652ec711d17c5b7392c54491b1259f370/fluxc/src/main/java/org/wordpress/android/fluxc/module/DebugOkHttpClientModule.java)
- [`ReleaseOkHttpClientModule`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5f2fb2f652ec711d17c5b7392c54491b1259f370/fluxc/src/main/java/org/wordpress/android/fluxc/module/ReleaseOkHttpClientModule.java)

This causes two issues:

- Duplicated code between `debug` and `release` variants
- Clients don't have a possibility to append `okhttp3.Interceptor` to `release` variant

### Solution

We should merge those two modules into one and allow clients to choose what interceptors they want to provide to which variant.

wzieba:
Closes: #1961 

### Context

This PR merges `DebugOkHttpClientModule` and `ReleaseOkHttpClientModule` into one: `OkHttpClientModule`

Reasons for this change
- now, both `WCAndroid` and `WPAndroid` are able to add interceptors for both `debug` and `release` variants. Previously, interceptors were added **only** in `DebugOkHttpClientModule`
- we reduce boilerplate code - `Debug` and `Release` modules were almost the same

### Backward compatibility

To make project build, just change any appearance of `ReleaseOkHttpClientModule` or `DebugOkHttpClientModule` to `OkHttpClientModule`.

About adding interceptors:

#### `WPAndroid`
No changes required as `InterceptorModule` [is added only into `debug` component.](https://github.com/wordpress-mobile/WordPress-Android/search?q=InterceptorModule) 

#### `WCAndroid`
Same like for `WPAndroid` - no changes required, [`InterceptorModule` only in `debug` component](https://github.com/woocommerce/woocommerce-android/search?q=InterceptorModule).


instaflux/build.gradle
```diff
@@ -24,6 +24,8 @@ android {
         targetSdkVersion 29
         versionCode 1
         versionName "1.0"
+
+        multiDexEnabled true
```
wzieba, line 28: When executing project-wide gradle tasks I was having a crash during the build about extending methods counts, this line fixes it.

example/src/androidTest/java/org/wordpress/android/fluxc/ditests/TestInterceptorModule.kt
```diff
@@ -0,0 +1,30 @@
...
+@Module
+class TestInterceptorModule {
+    @Provides @IntoSet @Named("interceptors")
+    fun provideDummyInterceptor(): Interceptor = DummyInterceptor
+
+    @Provides @IntoSet @Named("network-interceptors")
+    fun provideDummyNetworkInterceptor(): Interceptor = DummyNetworkInterceptor
+}
+
+object DummyInterceptor : Interceptor {
```
wzieba, line 20: It's an `object` as I want to make it a singleton to later [compare it in tests](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1976/files#diff-674396939919b0383826b4b2e3b46105b00270472fa33123f337837210c56037R75-R76) and make sure we inject this very object.

malinajirka:
Thanks @wzieba! I've reviewed the code and it looks good to me. 
~However, I tried to build WPAndroid `develop` referencing this commit and it fails with the following exception:~

Screenshot 2021-04-29 at 11 43 34

~Any ideas what I might be missing? Does it work for you?~

Sorry, I found the mistake - it works as expected 🦆 .

malinajirka (approved):
Thanks @wzieba! I've tested it in WPAndroid and WooAndroid and it works as expected.

However, before we merge it, we should create PRs in both repos and merge them soon after we merge this PR to make sure `develops` are still compatible. Also don't forget that WPAndroid doesn't use "commit-hashes" but manually created tags - let me know if you need any help.

Tip: It's usually better to create PRs in the app repos at the same time as PR in FluxC so the reviewer doesn't need to build the apps locally and can just use the CI apk. 

P.S. Some of the FluxC integration tests are failing for me locally, however, I've verified they are failing even on `develop` so it's not related to this PR.

wzieba:
Thank you @malinajirka for the review!

> It's usually better to create PRs in the app repos at the same time as PR in FluxC so the reviewer doesn't need to build the apps locally and can just use the CI apk.

You're right thank you for the tip - it'll save time for the reviewer and in the end of the day changes must be merged anyway so that's a good time saver for everyone 👍 

I'll mark it as a draft until I create PRs to the clients' repositories.

PR #1977, opened 2021-04-28
Improve Jetpack tunnel query parameters syntax

hichamboushaba:
In order to implement https://github.com/woocommerce/woocommerce-android/issues/2069, to quote the original issue:

> Currently when requests are sent to the remote site, the URL parameters for the endpoint are encoded in the path parameter to the WordPress.com endpoint for the Jetpack proxy. There are also cases where the parameters for the remote site are set as actual parameters on the proxy endpoint. The proxy endpoint will pass these parameters along in most cases, but we've seen some strange behavior because of it.
> 
> Because of this, let's change how we send parameters to the remote site. Instead of appending them to the path property, bundle them all in a JSON object, URL-encode it, and use the query parameter on the rest-api endpoint. This works for POST requests as well as the body can contain JSON that isn't URL-encoded.

This PR makes this change, by passing path params in the `query` object.
And regarding the DELETE endpoint, its parameters are added to the `body` object, because the backend ignores the `query` parameter for endpoints other than GET (p1619607553382000-slack-CGPNUU63E), and to align with iOS's implementation.

### Testing
Open the example app, and try playing with different scenarios.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequest.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequest.kt
@@ -7,7 +7,6 @@ import org.wordpress.android.fluxc.generated.endpoint.WPCOMREST
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComErrorListener
 import java.lang.reflect.Type
-import java.net.URLEncoder
 
 /**
  * A request making a WP-API call to a Jetpack site via the WordPress.com /jetpack-blogs/$site/rest-api/ tunnel.
@@ -20,11 +19,12 @@ import java.net.URLEncoder
  *
  * Example request:
  * https://public-api.wordpress.com/rest/v1.1/jetpack-blogs/$siteId/rest-api/
- * ?path=%2Fwp%2Fv2%2Fposts%2F%26_method%3Dget%26status%3Ddraft&json=true
+ * ?path=%2Fwp%2Fv2%2Fposts%2F%26json%3Dtrue%26_method%3Dget&query=%7B%22status%22%3A%22draft%22%7D
  *
  * Broken down, the GET parameters are:
- * path=/wp/v2/posts/&_method=get&status=draft
+ * path=/wp/v2/posts/&_method=get
  * json=true
+ * query={"status":"draft"}
  *
  * The path parameter is sent HTML-encoded so that it's discernible from the other arguments by WordPress.com.
  * In this example, this would become a GET request to {JSON endpoint root}/wp/v2/posts/?status=draft.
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequest.kt
```diff
@@ -216,7 +217,7 @@ object JetpackTunnelGsonRequest {
         listener: (T?) -> Unit,
         errorListener: WPComErrorListener
     ): WPComGsonRequest<JetpackTunnelResponse<T>>? {
-        val wrappedBody = createTunnelBody(method = "delete", params = params, path = wpApiEndpoint)
+        val wrappedBody = createTunnelBody(method = "delete", body = params, path = wpApiEndpoint)
```
wzieba, line 220: Can we also change the name `params` in the method signature? I think it'll be less confusing in further usage of this method.

It's only suggestion as I understand that for method consumers those value can still be "parameters" and it might not be important how they're passed later - is this a reason for leaving `params` name?

Also I think now docs needs to be updated as
> * @param params the parameters to append to the request URL

is no longer valid if I'm not mistaken.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequest.kt
```diff
@@ -240,36 +241,28 @@ object JetpackTunnelGsonRequest {
     private fun createTunnelParams(params: Map<String, String>, path: String): MutableMap<String, String> {
         val finalParams = mutableMapOf<String, String>()
         with(finalParams) {
-            put("path", buildRestApiPath(path, params, "get"))
+            put("path", "$path&_method=get")
             put("json", "true")
```
wzieba, line 245: also nit, but for consistency with `put("json", "true")`, can we split to

```suggestion
            put("path", "$path")
            put("_method", "get")
            put("json", "true")
```

?

wzieba (commented):
Leaving some minor comments, I don't approve yet as I'm now smoke-testing PR with this change applied.

wzieba (approved):
All looks good 👍 I've tested all of the basic operations, especially previews of order or products data. Also log in flow, just in case.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequest.kt
hichamboushaba: > is this a reason for leaving params name?

Yes, this is the reason, as the consumers don't have to think about how the tunneling works.
But you are right about the comment, I'll update it, thanks 🙏

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequest.kt
hichamboushaba: I don't think this will work, as the `_method` has to be part of the parameter `path`, and not a separate parameter, URL encoding will encode the `&` symbol, which makes sure that it's part of the parameter `path`.

nbradbury (approved):
Looks good here! :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequest.kt
wzieba: Thanks, I think got it!

PR #1979, opened 2021-04-29
Delete clutter likes data

develric:
This PR adds a purge logic for like data to avoid to pollute user storage space. Logic is pretty simple:

- Added a timestamp that marks the time when the data was fetched
- Added a threshold of 7 days from the fetch timestamp after which the data is considered to be purgeable
- When successfully fetched the first page of a post/comment like data, we check if data with timestamp before the threshold exists for any post/comment different from the currently fetching one. If so, all the data for those post/comment likes are purged

This has a companion PR to the WPAndroid PR [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14555) that can be used for testing. To ease testing you should be able to:

- Decrease the `TIMESTAMP_THRESHOLD` to something like 60 * 1000L
- Activate the db inspector in AS with live updates and use the following query (replace `type like 'post-like'` with `type like 'comment-like'` for comments)
```
select REMOTE_SITE_ID as site, REMOTE_ITEM_ID as item, MIN(TIMESTAMP_FETCHED) as timestamp from LIKEMODEL where type like 'post-like'   group by REMOTE_SITE_ID, REMOTE_ITEM_ID order by timestamp
```

- Browse to fetch post/comment likes within the 60 seconds and observe each post/comment to which you browse adds only one row from the query live update
- Continue browsing for post/comment likes and check the expected rows are deleted

PS: since I was adding some unit testing for post and comments sql likes methods, I also added a bit of unit testing to cover the `before` and `exclude` pagination API logic that we identified as tricky to test otherwise.


example/src/test/java/org/wordpress/android/fluxc/comment/CommentSqlUtilsTest.kt
```diff
@@ -275,6 +284,92 @@ class CommentSqlUtilsTest {
         Assertions.assertThat(cleanedComments.find { it.remoteCommentId == 48L }).isNull()
     }
 
+    @Test
+    fun `insertOrUpdateCommentLikes insert a new like`() {
+        val siteId = 100L
+        val commentId = 1000L
+
+        val localLike = createLike(100L, 1000L)
```
khaykov: I think we can use `siteId` and `commentId` here :)

example/src/test/java/org/wordpress/android/fluxc/comment/CommentSqlUtilsTest.kt
```diff
@@ -275,6 +284,92 @@ class CommentSqlUtilsTest {
...
+        CommentSqlUtils.insertOrUpdateCommentLikes(siteId, commentId, localLikeChanged)
+
+        commentLikes = CommentSqlUtils.getCommentLikesByCommentId(siteId, commentId)
+        Assertions.assertThat(commentLikes).hasSize(1)
+        Assertions.assertThat(commentLikes[0].isEqual(localLike)).isFalse
+        Assertions.assertThat(commentLikes[0].isEqual(localLikeChanged)).isTrue
+    }
+
+    @Test
+    fun `purgePostLikes deletes currently fetched data`() {
```
khaykov: I think it should be `purgeCommentLikes deletes currently fetched data`?

example/src/test/java/org/wordpress/android/fluxc/comment/CommentSqlUtilsTest.kt
```diff
@@ -293,4 +388,23 @@ class CommentSqlUtilsTest {
         commentModels.reverse() // we usually receive comments starting from more recent
         return commentModels
     }
+
+    private fun createLike(siteId: Long, postId: Long, timeStamp: Long = Date().time) = LikeModel().apply {
```
khaykov: `postId` should probably be a `commentId`?

example/src/test/java/org/wordpress/android/fluxc/comment/CommentSqlUtilsTest.kt
```diff
@@ -275,6 +284,92 @@ class CommentSqlUtilsTest {
...
+        postLikes = CommentSqlUtils.getCommentLikesByCommentId(siteId, commentId)
+        Assertions.assertThat(postLikes).isEmpty()
+    }
+
+    @Test
+    fun `purgeCommentLikes delete data older than threshold`() {
+        val siteId = 100L
+        val commentId = 1000L
+
+        val sitePostList = listOf(
```
khaykov: This should probably be `siteCommentsList` :)

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/CommentSqlUtils.java
```diff
@@ -224,16 +229,49 @@ public static int getCommentsCountForSite(SiteModel site, CommentStatus... statu
...
-        return WellSql.delete(LikeModel.class)
-                      .where()
-                      .beginGroup()
-                      .equals(LikeModelTable.TYPE, LikeType.COMMENT_LIKE.getTypeName())
-                      .equals(LikeModelTable.REMOTE_SITE_ID, siteId)
-                      .equals(LikeModelTable.REMOTE_ITEM_ID, remoteCommentId)
-                      .endGroup()
-                      .endWhere()
-                      .execute();
+    public static int purgeCommentLikes(long siteId, long remoteCommentId) {
```
khaykov: For clarity purposes maybe we can rename this method to something like `deleteCommentLikesAndPurgeExpiredLikes`? A bit long but I think it gives a better understanding of what its doing :)

Same with Posts counterpart.

Another option will be to split them, but I'll leave this up to you :)

khaykov (approved):
Thanks for all the tests and guiding me through the details in Slack, @develric ! It looks good to me, and I confirmed that the likes are getting purged 👍 

I noticed couple of minor copy-paste issue, and have one other comment about a method name :)

example/src/test/java/org/wordpress/android/fluxc/comment/CommentSqlUtilsTest.kt
develric: yep, definitely 😄

example/src/test/java/org/wordpress/android/fluxc/comment/CommentSqlUtilsTest.kt
develric: cannot understand why are you supposing I copy pasted 🤔 ... 😄

khaykov (approved):
Thanks for the changes, @develric ! Looks good to me :)

PR #1980, opened 2021-04-29
Add capture payment endpoint

malinajirka:
This PR adds support  for `/payments/orders/<id>/capture_terminal_payment` endpoint which will be used to capture card-present payments. The development of this endpoint is not finished even on the server-side so this PR doesn't contain tests and error paths (except of generic "WooError"). I've added `TODO cardreader` - I plan to "fix or turn into tickets" all these todos before the code goes into production.

To test:
Test this in WCAndroid - https://github.com/woocommerce/woocommerce-android/pull/3943


nbradbury (approved):
Looks good! :shipit:

PR #1981, opened 2021-04-30
Fix a rare crash when fetching the system status endpoint

hichamboushaba:
This is needed to fix the issue https://github.com/woocommerce/woocommerce-android/issues/3925, in some stores, the `inactive_plugin` field is missing from the response, so I changed the response fields to be nullable all, to avoid any issues with them missing.

### Testing
1. Open the example app.
2. Click on Woo.
3. Click on Shipping Labels.
4. Click on Get shipping plugin info.
5. Confirm that the functionality is not impacted.


nbradbury (approved):
Fix is good! :shipit:

PR #1984, opened 2021-05-02
Added endpoint to fetch user info for a store

### LINKED ISSUE #1978: Woo: add option to fetch user role for a store
As part of woocommerce/woocommerce-android#3846, we would need to aadd support to fetch the current user's role from the `/wp/v2/users/me` endpoint.

anitaa1990:
Fixes #1978 

This PR adds the relevant endpoint to fetch the current user's role for a store. The Woo API's don't support roles other than Admin or Store manager. So we use this API to get the current user's role and display an error message if their role does not have the relevant access.

### Changes
- Added a new endpoint `/wp/v2/users/me`.
- Added new store, rest client, mapper classes to fetch the role of the current user.
- Added unit tests for `WCUserStore`.
- Added new button to example app to test the new endpoint.

### Notes
- I added an enum class `WCUserRole` that holds the list of default roles currently supported in Woo. 
- Custom roles for aa store can be created using external plugins. We currently don't support any roles other than Admin and Shop manager so I added an `OTHER` as well to the `UserRole` enum.
- I also found that a single user can have more than 1 user role defined. So added support for that as well.

### Screenshots
. 

### Testing
- Login to a test store as an admin, shop manager, customer, subscriber etc from the example app.
- Click on `Woo` -> `Get User Role` and verify that the current user role is displayed correctly.
- Install the [User Role Editor](https://user-images.githubusercontent.com/22608780/116804544-18a8c300-ab3d-11eb-880e-a60381940a82.png) plugin on your test site.
- Add multiple user roles for a single user. You can also add custom user roles.
- Open the app and follow Step 2 and verify that the current user roles are displayed correctly. If a user has a custom user role, notice that `OTHER` is displayed.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/WooCommerceFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/WooCommerceFragment.kt
@@ -96,7 +96,7 @@ class WooCommerceFragment : Fragment() {
                         prependToLog("${it.type}: ${it.message}")
                     }
                     result.model?.let {
-                        prependToLog("Current user is: ${it.value}")
+                        prependToLog("Current user is: ${it.joinToString(", ")}")
                     }
                 }
             }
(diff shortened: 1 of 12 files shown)


nbradbury (approved):
Nice work, this all works great! :shipit:

PR #1990, opened 2021-05-04
Support `_fields` when fetching list of orders by ids

### LINKED ISSUE #1354: Woo: use _fields when fetching product(s)
Our `OrderRestClient` uses the `_fields` param to limit the size of the response. We should do the same when fetching a single product or product list.

Note: Wait for #1355 to be merged before tackling this.

wzieba:
Closes #1354 

This PR is a complementary to #1157 

It adds `_fields` to the call that doesn't have it yet: `OrderRestClient#fetchOrdersByIds`.

I also added some improvements to readability of the fields we request and added one: `date_modified_gmt` which we were missing.

### Result

On below screens you can see 4 calls.
- 2 first requests "ids" of the orders. There are 2 calls as on test project I have 16 orders and we batch this request by 15 orders per request
- next 2 requests are requests that fetch "rich" `Orders`

You can see that after we reduced size of the requests roughly by half.

> Note: `Duration` is not a good measurement, if you take a look at first two calls, they weren't influence by changes in this PR but duration vary ~20%.

#### Before
before

#### After
after

### How to test

There are basically two equivalent ways, please choose the one you find more effortless:

#### Using `Flipper`
1. Run the `WCAndroid` app (build from this PR), go to `Orders`. 
2. Open `Flipper` app, select `databases` plugin
3. Select `tracks.db` and run `DELETE FROM WCOrderModel` in `SQL` section
4. Refresh the orders
3. Smoke test orders list and details: make sure all expected fields are present

or

#### Reinstalling the app
1. Remove `WCAndroid` from Android device (or just remove app's data)
2. Install the app (build from this PR)
3. Log in, go to `orders`
3. Smoke test orders list and details: make sure all expected fields are present

You can use app build from this PR: https://github.com/woocommerce/woocommerce-android/pull/3962

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCOrderTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCOrderTest.kt
@@ -2,6 +2,7 @@ package org.wordpress.android.fluxc.release
 
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
+import org.greenrobot.eventbus.ThreadMode.MAIN
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
 import org.junit.Assert.fail
@@ -9,6 +10,7 @@ import org.junit.Test
 import org.wordpress.android.fluxc.TestUtils
 import org.wordpress.android.fluxc.action.WCOrderAction
 import org.wordpress.android.fluxc.generated.WCOrderActionBuilder
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
 import org.wordpress.android.fluxc.model.WCOrderModel
 import org.wordpress.android.fluxc.model.WCOrderNoteModel
 import org.wordpress.android.fluxc.model.order.OrderIdentifier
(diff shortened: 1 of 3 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCOrderTest.kt
```diff
@@ -102,6 +106,23 @@ class ReleaseStack_WCOrderTest : ReleaseStack_WCBase() {
         )
     }
 
+    @Test
+    fun testFetchOrdersById() {
```
wzieba, line 110: This test is not strictly connected to the subject of the PR. It doesn't test `_fields` but when I was browsing the code to understand how it works, I've realized we don't have a test for `fetching orders by ids` so I've added it.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderApiResponse.kt
```diff
@@ -39,38 +39,31 @@ class OrderApiResponse : Response {
         val total: String? = null
     }
 
-    val id: Long? = null
-    val number: String? = null
-    val status: String? = null
+    val billing: Billing? = null
```
wzieba, line 42: Changes in this file are just "cosmetic" - I've reordered attributes into alphabetical order. It's much easier this way to compare what we use in `_fields` and what we have from `json` response from the API.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -810,4 +805,39 @@ class OrderRestClient @Inject constructor(
 
         return providers
     }
+
+    companion object {
+        private val ORDER_FIELDS = arrayOf(
```
wzieba, line 810: I've changed plain `String` to the alphabetically-sorted array to make it easier to read and possibly change in the future.

This way I've discovered that we don't request `date_modified_gmt` but we should - so I added it. It's the only change to previous definition of `ORDER_FIELDS`

nbradbury (approved):
Looks good! Once I merge this, if you change the FluxC hash in the related WCAndroid PR I'll merge it, too. :shipit:

PR #1992, opened 2021-05-06
Handle custom errors for capturePayment endpoint

malinajirka:
This PR adds support for transforming [errors specific to capturePayment endpoint](https://github.com/Automattic/woocommerce-payments/pull/1689#discussion_r623880571) and returning them back to the client. Unfortunately, `WooResult` and `WooError` do not support custom errors so I decided to fallback to the default FluxC `Payload`. I've tried to extend `WooResult` in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1983 but decided to close the PR since the introduced complexity wasn't imo worth it.

The client apps don't need to be updated, so feel free to merge this PR.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pay/PayRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pay/PayRestClient.kt
@@ -5,8 +5,18 @@ import com.android.volley.RequestQueue
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.generated.endpoint.WOOCOMMERCE
 import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.model.pay.WCCapturePaymentError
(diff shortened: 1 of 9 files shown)


nbradbury (approved):
This looks good to me. It would be nice at some point to enable custom errors in WooResult and WooError since we'll likely run into this again, but no worries for this PR.

@malinajirka I'll approve this, and the second pair of eyes on this can merge it.

hichamboushaba (approved):
I share Nick's opinion, this can be merged now, and planning more generic and better support for custom errors in WooError.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/pay/PayRestClient.kt
```diff
@@ -66,11 +75,38 @@ class PayRestClient @Inject constructor(
...
+                        site,
+                        paymentId,
+                        orderId
+                )
             }
         }
     }
+
+    private fun mapToCapturePaymentError(error: WPComGsonNetworkError?, message: String): WCCapturePaymentError {
+        val type = when {
```
kidinov, line 99: It looks like that we could benefit here from the use of [exhaustive](https://proandroiddev.com/til-when-is-when-exhaustive-31d69f630a8b) approach.

PR #1998, opened 2021-05-10
Revert "Use Plugin DSL"

wzieba:
Reverts wordpress-mobile/WordPress-FluxC-Android#1973

### Context

We're experiencing an issue with Jitpack builds. For some reason, Jitpack doesn't seem to find either `install` nor `publishToMavenLocal` tasks, even with `maven-publish`, `com.github.dcendents.android-maven`, or `maven` plugins (I've did some tests)

When Jitpack doesn't find those tasks, it forcefully adds plugins (see below). It worked like this before too:
image 

[Look at build of 1.16.2 version](https://jitpack.io/com/github/wordpress-mobile/WordPress-FluxC-Android/1.16.2/build.log)

### Issue

The problem is that Jitpack adds those plugins using `buildscript { ... }` and this force addition breaks build when Plugin DSL is applied:

image e.g. here: https://jitpack.io/com/github/wordpress-mobile/WordPress-FluxC-Android/8428a0d0ab/build.log

I'd like to revert this PR to unblock everyone working on FluxC.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,16 +1,22 @@
+buildscript {
+    ext.kotlinVersion = '1.4.10'
+    ext.kotlinCoroutinesVersion = '1.3.9'
 
-plugins {
-    // Those declarations are just a workaround for a false-positive Kotlin Gradle Plugin warning
-    // https://youtrack.jetbrains.com/issue/KT-46200
-    id "com.android.application" apply false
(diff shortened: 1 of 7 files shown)


antonis (approved):
Thank you for the quick fix @wzieba! I really appreciate this 🙇

PR #2002, opened 2021-05-12
Woo: Adds option to fetch user name + role of the current user 

anitaa1990:
Adds option to fetch user name + role of the current user after logging into the Woo app. 

### Background
The Woo app currently supports only administrator and shop manager role. This is part of the changes in woocommerce/woocommerce-android#3846. So once a user is logged in, we fetch the role of the current user and display an error message if the role does not match admin or shop manager. 

The latest design for that error screen, specifies that we display the user's first and last name along with their role.

### Changes
This PR adds a new model `WCUserModel` that is fetched by the client when calling the `/wp/v2/users/me` API endpoint. Previously, we used to return just the role of the user. This PR returns the `WCUserModel` instead since it includes the user name along with the list of roles.

## Update
- I updated this PR to add a new table to the local db. This just stores the `WCUserModel` as a table and caches the result. 

### Testing
- Run `WCUserStoreTest` and ensure that all tests pass.
- Login to a test store as an admin, shop manager, customer, subscriber etc from the example app.
  - Click on Woo -> Get User Role and verify that the current user role is displayed correctly.
  - Install the User Role Editor plugin on your test site.
  - Add multiple user roles for a single user. You can also add custom user roles.
  - Open the app and follow Step 2 and verify that the current user roles are displayed correctly. If a user has a custom user role, notice that OTHER is displayed.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/user/WCUserRole.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/user/WCUserRole.kt
@@ -1,4 +1,4 @@
-package org.wordpress.android.fluxc.network.rest.wpcom.wc.user
+package org.wordpress.android.fluxc.model.user
 
 enum class WCUserRole(val value: String = "") {
     OWNER("owner"),
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/user/WCUserRole.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/user/WCUserRole.kt
@@ -22,5 +22,5 @@ enum class WCUserRole(val value: String = "") {
         fun fromValue(value: String) = valueMap[value] ?: OTHER
     }
 
-    fun isSupported() = this == ADMINISTRATOR
+    fun isSupported() = this == ADMINISTRATOR || this == SHOP_MANAGER
 }
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/WooCommerceFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/WooCommerceFragment.kt
@@ -96,7 +96,7 @@ class WooCommerceFragment : Fragment() {
                         prependToLog("${it.type}: ${it.message}")
                     }
                     result.model?.let {
-                        prependToLog("Current user is: ${it.joinToString(", ")}")
+                        prependToLog("Current user is: ${it.roles}")
                     }
                 }
             }
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/WooCommerceFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/WooCommerceFragment.kt
@@ -31,8 +31,8 @@ import org.wordpress.android.fluxc.example.ui.stats.WooRevenueStatsFragment
 import org.wordpress.android.fluxc.example.ui.stats.WooStatsFragment
 import org.wordpress.android.fluxc.example.ui.taxes.WooTaxFragment
 import org.wordpress.android.fluxc.generated.WCCoreActionBuilder
-import org.wordpress.android.fluxc.network.rest.wpcom.wc.user.WCUserStore
 import org.wordpress.android.fluxc.store.WCDataStore
+import org.wordpress.android.fluxc.store.WCUserStore
 import org.wordpress.android.fluxc.store.WooCommerceStore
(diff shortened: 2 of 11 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/user/WCUserRole.kt
```diff
@@ -22,5 +22,5 @@ enum class WCUserRole(val value: String = "") {
         fun fromValue(value: String) = valueMap[value] ?: OTHER
     }
 
-    fun isSupported() = this == ADMINISTRATOR
+    fun isSupported() = this == ADMINISTRATOR || this == SHOP_MANAGER
```
hichamboushaba, line 25: Nice, WDYT about adding a similar method to `WCUserModel` that checks that at least one of the roles is supported?

hichamboushaba (approved):
Nice 🥇, I left one suggestion in the comments, but nothing blocking from merging right away

hichamboushaba:
Something else that came to my mind, we will check the role at each app start if I'm not mistaken? did you think about what would happen if the API call fails?

anitaa1990:
Thanks for the review @hichamboushaba!

I added a new commit that will change the column name from `Role` to `Roles` since we can have multiple roles. 

> we will check the role at each app start if I'm not mistaken? did you think about what would happen if the API call fails?

Yup, we will check the role during each app startup. So what I had planned was to fetch the user role info when the app is completely restarted and when the user logs in to the app. This will take place in the background and the info will be stored in the local db. 

So there are 2 scenarios that can happen here:
**If the user is currently eligible to view the app but their role has changed and during that time, the API call fails:**
The user will still be able to view the app. All the API calls will fail anyway. Not ideal, but it's better than blocking them from using the app, when we don't know for sure, whether they are eligible or not. 

**If the user is currently NOT eligible to view the app but their role has changed and during that time, the API call fails:**
We will display the error screen to the user, even though they are eligible to view the app. In this screen, there is a retry option so if they **are** eligible, we will just let them use the app.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/user/WCUserRole.kt
anitaa1990: Do you mean something like this?
```
fun isRoleSupported() = getUserRoles().none { !it.isSupported() }
```

If yes, then sure! I was going to add that to Woo but it could work to have that here 👍

hichamboushaba:
Thanks for the clarifications Anitaa 👍.
One last question, do you think this class WCUserModel can be useful for the other apps (WPAndroid and Jetpack)? if this is the case, we should maybe move it to the common module.

anitaa1990:
> One last question, do you think this class WCUserModel can be useful for the other apps (WPAndroid and Jetpack)? if this is the case, we should maybe move it to the common module.

I thought about it for a long time but some of the implementation is very specific for Woo and so decided to go along with a new table just for Woo. I'll ping the WP team to let them know there is an implementation already available, so if they did want to implement it, it's available already 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/user/WCUserRole.kt
anitaa1990: Fixed in edd3c27

hichamboushaba: Thanks @anitaa1990 for making the change, and sorry I didn't notice your previous comment.
I think the change as implemented will block valid users from accessing (unless I'm misunderstanding something): If a user has multiple roles, and one of them is supported, they should be able to access the app, I just tested by setting my account to have the roles "shop manager" and "customer", and I tested the app, and it's working without issues.

I think we need to change the logic to something like this:
```kotlin
fun isRoleSupported() = getUserRoles().any { it.isSupported() }
```
WDYT?

anitaa1990: 😨 That's a good catch! Thanks! I fixed it in e0e87d9 and updated the unit tests to test with different roles :)

hichamboushaba: Thanks @anitaa1990 🙏

hichamboushaba (approved):
LGTM, nice work :shipit:

PR #2005, opened 2021-05-17
Adding error subtype for malformed media argument.

develric:
This PR adds a MediaErrorSubType persisted in the MediaUploadModel that can be used to better discriminate in the client app an error subcategory and decide to present a localisable and hopefully clearer message to the user.

This PR has a companion in [this WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14670).

Notes for the reviewer:
- This PR adds the kotlin-reflect lib dependency (same version of kotlin dependency). This allow to use reflection to list all the children of a sealed class that seems pretty convenient; but happy to listen to other possible implementation of it if we want to avoid reflection.
- Currently I added subtypes only for `MalformedMediaArgSubType`


develric:
Hi @planarvoid, as per our discussion, given both of us were not happy with introducing reflection, I refactored to avoid it. Hope it's clean enough, let me know wdyt 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/store/media/MediaErrorSubType.kt
```diff
@@ -0,0 +1,63 @@
...
+sealed class MediaErrorSubType(val categoryName: String, val subTypeName: String) {
+    // This enum describes the categories of MediaErrorType that are mapped into one or multiple
+    // MediaErrorSubType. UNDEFINED category collects all MediaErrorType(s) that are not mapped yet.
+    // To map another MediaErrorType, add an item to the enum and define a MediaErrorSubType element
+    enum class MediaErrorSubtypeCategory(val errors: List<MediaErrorSubType>) {
+        UNDEFINED_SUBTYPE(listOf(
+                UndefinedSubType
+        )),
+        MALFORMED_MEDIA_ARG_SUBTYPE(listOf(
+                MalformedMediaArgSubType(MEDIA_WAS_NULL),
```
planarvoid: you could do something like `MalformedMediaArgSubType.Type.values().map { MalformedMediaArgSubType(it) }` here to make it a little less error prone when adding new types 👍

fluxc/src/main/java/org/wordpress/android/fluxc/store/media/MediaErrorSubType.kt
```diff
@@ -0,0 +1,63 @@
+package org.wordpress.android.fluxc.store.media
+
+import org.wordpress.android.fluxc.store.media.MediaErrorSubType.MalformedMediaArgSubType.Type.DIRECTORY_PATH_SUPPLIED_FILE_NEEDED
+import org.wordpress.android.fluxc.store.media.MediaErrorSubType.MalformedMediaArgSubType.Type.MEDIA_FILE_NOT_FOUND_LOCALLY
+import org.wordpress.android.fluxc.store.media.MediaErrorSubType.MalformedMediaArgSubType.Type.MEDIA_WAS_NULL
+import org.wordpress.android.fluxc.store.media.MediaErrorSubType.MalformedMediaArgSubType.Type.NOT_VALID_LOCAL_FILE_PATH
+import org.wordpress.android.fluxc.store.media.MediaErrorSubType.MalformedMediaArgSubType.Type.NO_ERROR
+import org.wordpress.android.fluxc.store.media.MediaErrorSubType.MalformedMediaArgSubType.Type.UNSUPPORTED_MIME_TYPE
+
+sealed class MediaErrorSubType(val categoryName: String, val subTypeName: String) {
```
planarvoid: what do you think about replacing the categoryName type here with `MediaErrorSubtypeCategory`? I've tried it and it makes it a little bit neater 👍 . In the serialize method you'd call `${category.name}:$subTypeName`

planarvoid (commented):
It's looking much better! Thanks for the changes! I have 2 more comments to make the code a little bit cleaner 👍

develric:
Thanks for the review @planarvoid 👍 , I changed as per your feedbacks. To avoid circular dependency/reference that could result in a null `MediaErrorSubType.category` introduced a `private val categories` in the companion object (and used the when approach inside the flatMap; thanks for suggesting this, I like that we protect from missing to add new mapped categories here in this way 👍 ).

Should be ready for another pass 🙇

planarvoid (approved):
Thanks for the changes! It's looking great! I'm not merging this PR because I have to review the WPAndroid part first.

PR #2006, opened 2021-05-18
Shipping Labels: support customs form

hichamboushaba:
Needed for https://github.com/woocommerce/woocommerce-android/issues/3526, this PR adds the ability to pass customs information along the requests to fetch rates and purchase shipping labels.

Please don't merge until https://github.com/woocommerce/woocommerce-android/pull/4060 is reviewed too to avoid breaking the app.

#### Testing
1. Make sure your store has been set up to purchase shipping labels (check p91TBi-3xD-p2 for steps) 
2. Create an order with an international shipping address. 
3. Open the example app, and click on Woo.
4. Click on Shipping Labels.
5. Click on Puchase shipping label, and enter the order id.
6. Enter the package details.
7. Confirm that the purchase is successful.

For choosing the right dimensions, please pick something that you know is a valid value in the web client.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/shippinglabels/ShippingLabelRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/shippinglabels/ShippingLabelRestClient.kt
@@ -253,7 +253,7 @@ class ShippingLabelRestClient @Inject constructor(
             "destination" to destination.toMap(),
             "packages" to packages.map { labelPackage ->
                 val customs = customsData?.first { it.id == labelPackage.id }
-                labelPackage.toMap() + customs.toMap()
+                labelPackage.toMap() + (customs?.toMap() ?: emptyMap())
             }
         )
 
@@ -290,7 +290,7 @@ class ShippingLabelRestClient @Inject constructor(
                 "destination" to destination,
                 "packages" to packagesData.map { labelPackage ->
                     val customs = customsData?.first { it.id == labelPackage.id }
-                    labelPackage.toMap() + customs.toMap()
+                    labelPackage.toMap() + (customs?.toMap() ?: emptyMap())
                 }
         )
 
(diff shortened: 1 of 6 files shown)


hafizrahman:
Haven't checked the code yet, but I followed the test instructions and everything worked and the purchase was successful. 🎉 

Small np:

> Enter the package details.

In this step, the example app gave me `medium_flat_box_top` as the default box id to select, then I had to use `1` value for the dimensions inputs for things to work. I tried some random value previously and the purchase didn't go through. I understand that they should match the selected box's dimensions, so for future testing instructions, it might be worth to mention what value to use for height/width/length/weight.

hichamboushaba:
Thanks for the review @hafizrahman 🙏 
> so for future testing instructions, it might be worth to mention what value to use for height/width/length/weight.

I can't really give some defined values as they depend on the dimensions you have on your store, but I should've mentioned that you need to pick values that work in the web client 🙂.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/shippinglabels/ShippingLabelRestClient.kt
```diff
@@ -242,14 +243,18 @@ class ShippingLabelRestClient @Inject constructor(
...
+        packages: List<ShippingLabelPackage>,
+        customsData: List<WCShippingPackageCustoms>?
     ): WooPayload<ShippingRatesApiResponse> {
         val url = WOOCOMMERCE.connect.label.order(orderId).rates.pathV1
 
         val params = mapOf(
             "origin" to origin.toMap(),
             "destination" to destination.toMap(),
-            "packages" to packages.map { it.toMap() }
+            "packages" to packages.map { labelPackage ->
```
hafizrahman, line 254: Is there a documentation page about how the `packages` parameter should be formatted in the API request?

I wanted to check that they're formatted properly but not sure what the correct format should be when customs data is included.

hichamboushaba: No, there is no documentation page, I'm planning to write a P2 summarizing all my findings regarding customs implementation after finishing the task, and it will include the API changes I found out.
If you are curious enough, you can check the web client's source code here: https://github.com/Automattic/woocommerce-services/blob/864d0f0b14a038605366447b58422b258246498e/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js#L192

hafizrahman: Thanks! Did a spot-check and parameters and values look identical in the js version, looking good.

hafizrahman (approved):
This LGTM. Not merging yet as initially requested.

PR #2010, opened 2021-05-31
Blogging reminders repository

planarvoid:
This PR introduces the database layer and store for blogging reminders. After talking to @develric I've added Room (which was previously discussed and approved multiple times) in order to have a reactive database and store. The reason why I've used it in this PR is that the DB layer here is very small and we have a followup hackweek where we can polish the changes. If it doesn't make sense, I can still replace it with WellSql but doing manually a repository pattern is very error-prone. 

The changes in this PR should be quite straightforward. Let me know what you think! Especially about the database module I've introduced. I think it is quite neat to have WPAndroid database, potentially Woo database and shared database. It should make the development easier.


example/src/main/java/org/wordpress/android/fluxc/example/EditorThemeFragment.kt
```diff
@@ -25,6 +26,7 @@ class EditorThemeFragment : Fragment() {
     @Inject internal lateinit var siteStore: SiteStore
     @Inject internal lateinit var editorThemeStore: EditorThemeStore
     @Inject internal lateinit var dispatcher: Dispatcher
+        @Inject lateinit var database: WPAndroidDatabase
```
renanferrari: I believe this was added by mistake?

fluxc/src/main/java/org/wordpress/android/fluxc/model/BloggingRemindersModel.kt
```diff
@@ -0,0 +1,7 @@
+package org.wordpress.android.fluxc.model
+
+data class BloggingRemindersModel(val siteId: Int, val enabledDays: Set<Day> = setOf()) {
+    enum class Day {
+        MONDAY, TUESDAY, WEDNESDAY, THURSDAY, FRIDAY, SATURDAY, SUNDAY
+    }
```
renanferrari, line 6: It's unfortunate that we can't use `DayOfWeek` here. We should at least try to keep some parity between their numeric values though (monday = 1, tuesday = 2, etc). Adding a `val value = ordinal + 1` should do.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/BloggingRemindersDao.kt
```diff
@@ -0,0 +1,34 @@
...
+    data class BloggingReminders(
+        @PrimaryKey
+        var localSiteId: Int,
+        var monday: Boolean = false,
+        var tuesday: Boolean = false,
+        var wednesday: Boolean = false,
+        var thursday: Boolean = false,
+        var friday: Boolean = false,
+        var saturday: Boolean = false,
+        var sunday: Boolean = false
```
renanferrari, line 32: Good call deciding on using this data structure 👍

renanferrari (commented):
Thanks for the changes! They make sense to me and the code is looking good.

I left just a few comments below.

example/src/main/java/org/wordpress/android/fluxc/example/EditorThemeFragment.kt
planarvoid: Ah no it wasn't but I think we can remove it. I was trying to setup the dagger build and in order to do that I had to have at least one place using the database. I'll write an actual integration test instead 👍

fluxc/src/main/java/org/wordpress/android/fluxc/model/BloggingRemindersModel.kt
planarvoid: I was thinking about this when I saw your PR. Would it be possible to use the same desugaring you did in WPAndroid in FluxC? I'd love to be consistent here and merge the models.

example/src/main/java/org/wordpress/android/fluxc/example/EditorThemeFragment.kt
renanferrari: Thank you!

fluxc/src/main/java/org/wordpress/android/fluxc/model/BloggingRemindersModel.kt
renanferrari: I'd love for us to use desugaring here as well. The only reason I didn't suggest that is that I'm just not sure how it would affect other projects that use this library. If you don't see any problems with that, I'd definitely go for it!

PR #2016, opened 2021-06-03
Woo Shipping Labels: add support for commercial invoices

hichamboushaba:
This PR is needed for https://github.com/woocommerce/woocommerce-android/issues/4119, it just adds a column to save the `commercial_invoice_url`, and updates the example app to test retrieving it.

#### Test
1. Open the example app.
2. Click on Woo.
3. Click on Shipping Labels.
4. Click on Print Commercial Invoice
5. Enter the order and label ids.
6. Confirm that the default PDF reader is opened with the invoice file.

Please note that you need an international shipping label using one of the carriers that require the commercial invoice (for example DHL)


nbradbury (approved):
Looks and works just fine :shipit:

PR #2018, opened 2021-06-07
SiteStore refactoring - feature PR

planarvoid:
This PR contains all the changes to the SiteStore and the related classes. Everything should be reviewed and tested


fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
```diff
@@ -0,0 +1,69 @@
...
+     * @param errorListener the error listener
+     */
+    fun buildGetRequest(
+        url: String,
+        method: XMLRPC,
+        params: List<Any>,
+        listener: (Array<Any>) -> Unit,
+        errorListener: (BaseNetworkError) -> Unit
+    ): XMLRPCRequest {
+        return XMLRPCRequest(url, method, params, listener, errorListener)
```
develric: The url param in the java XMLRPCRequest is not marked as @NonNull (also it is marked as @nonnull in super BaseRequest, but ironically we check in the BaseRequest constructor for null cases 🤔); maybe we should think about using a nullable  . Wdyt? Eventually same goes with the other params (method excluded), and syncGetRequest equivalent

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
```diff
@@ -0,0 +1,69 @@
...
+     */
+    suspend fun syncGetRequest(
+        restClient: BaseXMLRPCClient,
+        url: String,
+        method: XMLRPC,
+        params: List<String>,
+        enableCaching: Boolean = false,
+        cacheTimeToLive: Int = BaseRequest.DEFAULT_CACHE_LIFETIME,
+        forced: Boolean = false
+    ) = suspendCancellableCoroutine<Response> { cont ->
```
develric: Like using the cancellable version 👍

develric (commented):
Nice addition @planarvoid 👍 . Code looks mostly good to me, I just left a comment about nullable fields for your review 🙇 . Let me know.

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
planarvoid: I think I'd rather be strict here and keep it non-null. I think people have to consciously use this builder it will only be used in Kotlin so we should be safe here.

planarvoid:
Thanks for the review 👍 . I've replied to all your comments + I've also applied changed that came from trying to use the builder in the site XMLRPC request builder.

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
develric: Good point, makes sense to me 👍

develric (approved):
Thanks for the follow up @planarvoid , LGTM 👍 . As we discussed I approve but not merge yet, waiting the other related PRs (including testing etc..) land into this one 🙇 .

PR #2019, opened 2021-06-07
Convert SiteStore to Kotlin

planarvoid:
In order to do further improvements this PR converts SiteStore to Kotlin. There should be no behavioural changes except that the payload objects are now immutable. I'll test this change with the following PRs.


fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.java
```diff
@@ -1,2340 +0,0 @@
...
-    /**
-     * Obtains the site with the given (local) id and returns it as a {@link SiteModel}.
-     */
-    public SiteModel getSiteByLocalId(int id) {
-        List<SiteModel> result = SiteSqlUtils.getSitesWith(SiteModelTable.ID, id).getAsModel();
-        if (result.size() > 0) {
-            return result.get(0);
-        }
-        return null;
-    }
```
develric, line 1301: This is not something introduced in this PR of course, just thought it is a good example when using nullability annotations could avoid issues at compile time + advantage of kotlin null-safety 👍 .

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -0,0 +1,1846 @@
...
+        @JvmField val message: String = ""
+    ) : OnChangedError
+
+    data class SiteEditorsError internal constructor(
+        @JvmField val type: SiteEditorsErrorType?,
+        @JvmField val message: String
+    ) :
+            OnChangedError {
+        constructor(type: SiteEditorsErrorType?) : this(type, "") {}
+    }
```
develric: np: some row alignment improvement possible here 👍

develric (commented):
Hey @planarvoid 👋 , thanks for this PR 🙇 ! Left only a np inline comment plus I noticed this: in `SiteStore` we have the `DesignateMobileEditorForAllSitesResponsePayload` with editors as not nullable field. In `SiteRestClient` we actually call `new DesignateMobileEditorForAllSitesResponsePayload(null);` (I guess it was working before given we are in an error scenario where editors are not iterated on). Maybe we can consider to make editors nullable or pass an empty list, wdyt? 

As a side note to share: if the above makes sense, this seems a case where the compiler does not catch a possible issue calling kotlin from java (hope I didn't miss any other similar case while reviewing 🙃). I tried to set @NotNull annotation for editors but annotating all kotlin not nullable methods/fields could be overwhelming maybe + actually it seems we set just a warning level in the linter for this sort of issues 🤷‍♂️

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
planarvoid: fixed in d6b4737

planarvoid:
thanks for the review @develric 

> in SiteStore we have the DesignateMobileEditorForAllSitesResponsePayload with editors as not nullable field. In SiteRestClient we actually call new DesignateMobileEditorForAllSitesResponsePayload(null); (I guess it was working before given we are in an error scenario where editors are not iterated on). Maybe we can consider to make editors nullable or pass an empty list, wdyt?

You're right. I've changed the field to nullable.

> As a side note to share: if the above makes sense, this seems a case where the compiler does not catch a possible issue calling kotlin from java (hope I didn't miss any other similar case while reviewing 🙃). I tried to set @NotNull annotation for editors but annotating all kotlin not nullable methods/fields could be overwhelming maybe + actually it seems we set just a warning level in the linter for this sort of issues 🤷‍♂️

I think these issues will be discovered in the other PRs. I've actually noticed the `DesignateMobileEditorForAllSitesResponsePayload` and fixed it once converting the site rest client.

develric (approved):
Thanks for the changes @planarvoid  (and in general for tackling this 👍 ). I think this is ok to go :shipit:

PR #2020, opened 2021-06-08
Extract site sql utils from site store

planarvoid:
In this PR I'm moving sql calls from Site Store to Site sql utils in order to make the store unit testeable without roboelectric.


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.kt
```diff
@@ -0,0 +1,518 @@
...
+import org.wordpress.android.fluxc.model.layouts.transform
+import org.wordpress.android.fluxc.network.rest.wpcom.site.GutenbergLayout
+import org.wordpress.android.fluxc.network.rest.wpcom.site.GutenbergLayoutCategory
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.AppLog.T.DB
+import org.wordpress.android.util.UrlUtils
+import java.util.ArrayList
+import javax.inject.Inject
+
+class SiteSqlUtils
```
develric, line 34: AFAIU there is no state in this class, wonder if we should mark it with @Singleton or @Reusable, wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java
```diff
@@ -1,431 +0,0 @@
...
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-
-public class SiteSqlUtils {
-    public static class DuplicateSiteException extends Exception {
-        private static final long serialVersionUID = -224883903136726226L;
-    }
-
-    public static SelectQuery<SiteModel> getSitesWith(String field, Object value) {
```
develric, line 45: If I'm not wrong it seems we have removed this method (with Object) from the .kt equivalent. Maybe the SiteSqlUtils makes sense to be used only from the store but in principle we are not granting it so wonder if we should keep this method in case it's used somewhere else (like in the login lib? 😊)

develric (commented):
Hi @planarvoid, this LGTM. Only couple of comments before we procede if you think it's worth 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.kt
planarvoid: I'm wondering if there are performance benefits from something being a singleton. Reusable sounds good though 🤷

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.java
planarvoid: I might be wrong but I'm almost sure the SQL utils are never called directly from anywhere else but FluxC. At least they shouldn't be called or injected. There is a chance someone is doing it wrong and I think we'll find out during testing.

planarvoid:
hey @develric , I've made SiteSqlUtils a Singleton to be consistent 👍

develric (approved):
Thanks for the follow up 🙇 ! :shipit:

PR #2021, opened 2021-06-09
Convert site rest client to kotlin

planarvoid:
This PR contains simple conversion of Site rest client to Kotlin in order to make further improvements


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -0,0 +1,1076 @@
...
+    class ExportSiteResponsePayload : Payload<BaseNetworkError>()
+    data class IsWPComResponsePayload(
+        val url: String,
+        val isWPCom: Boolean = false
+    ) : Payload<BaseNetworkError>()
+
+    data class FetchWPComSiteResponsePayload(
+        val checkedUrl: String,
+        val site: SiteModel? = null
+    ) : Payload<SiteError>()
```
develric, line 124: I guess we are enforcing the url/checkedUrl not being nullable on purpose here, but just wanted to confirm we checked for it 🙇 (looking at WPAndroid it seems not directly used, and same goes for Woo 🤔 )

develric (commented):
Hey @planarvoid 👋 , just wanted to check with you on the below comment, but other than that I'm happy to approve this 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
planarvoid: I think the rest client objects are used only from the site store because only the store calls the rest client methods. I think this means we should be safe.

planarvoid:
thanks for the review @develric , I think we should be safe here 👍

develric (approved):
Thanks for the clarification 🙇 ! This can go ahead in the queue for me 😄 !

PR #2022, opened 2021-06-09
Convert site xmlrpc rest client to kotlin

planarvoid:
In order to do improvements to the SiteStore we need to convert the site XMLRPC rest client to Kotlin. In this PR I'd trust all the tests to pass.


develric (approved):
This is looking good @planarvoid 👍 !

PR #2023, opened 2021-06-10
Convert site store methods to coroutines

planarvoid:
This PR converts the first methods to fetch site, fetch sites, create new site and fetch site formats to coroutines. I've changed both the WPCom and the XMLRPC rest clients whenever possible. This PR could be tested with the related WPAndroid changes - https://github.com/wordpress-mobile/WordPress-Android/pull/14812


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequestBuilder.kt
```diff
@@ -91,13 +92,17 @@ class WPComGsonRequestBuilder
         url: String,
         params: Map<String, String>?,
         body: Map<String, Any>?,
-        clazz: Class<T>
+        clazz: Class<T>,
+        retryPolicy: RetryPolicy? = null
```
develric, line 97: maybe we could add a comment in the java doc about the rationale of this param, wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -1133,11 +1146,15 @@ class SiteStore
         when (actionType) {
             FETCH_PROFILE_XML_RPC -> fetchProfileXmlRpc(action.payload as SiteModel)
             FETCHED_PROFILE_XML_RPC -> updateSiteProfile(action.payload as SiteModel)
-            FETCH_SITE -> fetchSite(action.payload as SiteModel)
-            FETCH_SITES -> fetchSites(action.payload as FetchSitesPayload)
-            FETCHED_SITES -> handleFetchedSitesWPComRest(action.payload as SitesModel)
-            FETCH_SITES_XML_RPC -> fetchSitesXmlRpc(action.payload as RefreshSitesXMLRPCPayload)
-            FETCHED_SITES_XML_RPC -> updateSites(action.payload as SitesModel)
+            FETCH_SITE -> coroutineEngine.launch(T.MAIN, this, "Fetch site") {
+                fetchSite(action.payload as SiteModel)
```
develric: could it make sense to move the `emitChange()` call here as well?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -116,38 +118,31 @@ class SiteRestClient @Inject constructor(
     data class IsWPComResponsePayload(
         val url: String,
         val isWPCom: Boolean = false
-    ) : Payload<BaseNetworkError>()
+    ) : Payload<BaseNetworkError?>()
```
develric: Possibly it's just obvious 😄 , but why do we need it nullable here?

develric (commented):
Great job @planarvoid , going to check the WPAndroid companion PR, but wanted to left just a couple of comments to double check with you 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
planarvoid: we don't, good catch. I think it's an overlook that was caused by splitting the PR into several small ones and merging between them :-). 
Fixed in 5de8dde

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
planarvoid: makes sense 👍
Changed in 1a48168

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequestBuilder.kt
planarvoid: Added in 9876e32

planarvoid:
Thanks for the review @develric , I've implemented all your suggestions 👍

develric (approved):
Thanks for the changes 🙇 ! Approving this one 👍 !

PR #2025, opened 2021-06-10
Woo: fix NPE if shipping_lines is missing from refunds response

hichamboushaba:
This is needed to fix the crash https://github.com/woocommerce/woocommerce-android/issues/4109, which is caused by a null pointer exception at the `shipping_lines` item.
I took the change to make `items` also nullable, the mapper will always convert null items to empty lists.

#### Testing
I wasn't able to reproduce the original issue, as `shipping_lines` is always present in the API response, which means it's probably a plugin that's messing with the response for the affected users.
So just confirming that the feature is not impacted is enough: load refunds of a refunded order in the example app.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/refunds/RefundMapper.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/refunds/RefundMapper.kt
@@ -12,26 +12,26 @@ import javax.inject.Inject
 class RefundMapper @Inject constructor() {
     fun map(response: RefundResponse): WCRefundModel {
         return WCRefundModel(
-                response.refundId,
-                response.dateCreated?.let { fromFormattedDate(it) } ?: Date(),
-                response.amount?.toBigDecimalOrNull() ?: BigDecimal.ZERO,
-                response.reason,
-                response.refundedPayment ?: false,
-                response.items.map {
+                id = response.refundId,
+                dateCreated = response.dateCreated?.let { fromFormattedDate(it) } ?: Date(),
+                amount = response.amount?.toBigDecimalOrNull() ?: BigDecimal.ZERO,
+                reason = response.reason,
+                automaticGatewayRefund = response.refundedPayment ?: false,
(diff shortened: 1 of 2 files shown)


hafizrahman:
> I wasn't able to reproduce the original issue, as shipping_lines is always present in the API response, which means it's probably a plugin that's messing with the response for the affected users.

Oh, that is interesting. Initially I had just assumed that the API response will be consistent with always returning `shipping_lines` even if empty (ref: p91TBi-4V7-p2 ). 

Checked core's code and [it also seems to be set to always return `shipping_lines`](https://github.com/woocommerce/woocommerce/blob/19bdc1c35b06fd1ad41f94bd91d5047474a59c3c/includes/rest-api/Controllers/Version2/class-wc-rest-order-refunds-v2-controller.php#L176):

```
		return array(
			'id'               => $object->get_id(),
			'date_created'     => $data['date_created'],
			'date_created_gmt' => $data['date_created_gmt'],
			'amount'           => $data['amount'],
			'reason'           => $data['reason'],
			'refunded_by'      => $data['refunded_by'],
			'refunded_payment' => $data['refunded_payment'],
			'meta_data'        => $data['meta_data'],
			'line_items'       => $data['line_items'],
			'shipping_lines'   => $data['shipping_lines'],
			'tax_lines'        => $data['tax_lines'],
			'fee_lines'        => $data['fee_lines'],
		);
```

Hadn't considered that the response can be altered elsewhere!

Will test this.

hichamboushaba:
> Initially I had just assumed that the API response will be consistent with always returning `shipping_lines` even if empty (ref: p91TBi-4V7-p2 ).

Yes, that's something that we have to deal with sadly, as the installed plugins may interfere with the APIs responses, so even if Core's code returns empty array always, another plugin may change this. And this is way it was recommended to have all fields in the network layer nullable (discussion: p1606396350412600/1606394991.409000-slack-C02QANACA).

hafizrahman (approved):
Code LGTM and things work OK in my tests as well. I also couldn't replicate the original issue, but the added code should be able to handle it nicely.

PR #2028, opened 2021-06-14
Woo: add a common class that represents object's meta data

hichamboushaba:
This PR creates a common class WCMetaData that represents `meta_data` of different WC classes, this is to match how WC Core does it: https://github.com/woocommerce/woocommerce/blob/b19500728b4b292562afb65eb3a0c0f50d5859de/includes/class-wc-meta-data.php

The class holds optional `display_key` and `display_value` which is returned by the API to format values of some meta_data related to WC orders.

Please don't merge until the Woo app is updated to align with the changes.

##### Order
1. Apply the patch from below
2. Create an order that uses a variable product
3. Open the example app and fetch this order
4. Confirm that you can see the list of attributes in the console

##### Refund
1. Refund an order.
2. Open the example app and fetch this refund.
3. Confirm that the data printed to the console has a list of `WCMetaData` objects which has the key `_refunded_item_id`

#### Patch for the first test
```Index: example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
IDEA additional info:
<+>UTF-8
--- example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt	(revision e1fc3eb87caa0a824d512dc5c902a6ed9e0d09a2)
+++ example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt	(date 1623681112781)
@@ -401,6 +401,7 @@
                                     })
                             )?.let {
                                 prependToLog("Single order fetched successfully!")
+                                prependToLog(it.getLineItemList()[0].getAttributesAsString())
                             } ?: prependToLog("WARNING: Fetched order not found in the local database!")
                         }
                     }
```


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -113,15 +111,16 @@ data class WCOrderModel(@PrimaryKey @Column private var id: Int = 0) : Identifia
         val price: String? = null // The per-item price
 
         @SerializedName("meta_data")
-        private val attributes: JsonArray? = null
+        val metaData: List<WCMetaData>? = null
 
         class Attribute(val key: String?, val value: String?)
```
hichamboushaba, line 116: I kept the `Attribute` class as a special type of meta data

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/refunds/WCRefundModel.kt
```diff
@@ -48,6 +51,6 @@ data class WCRefundModel(
         @SerializedName("method_title")
         val methodTitle: String? = null,
         @SerializedName("meta_data")
-        val metaData: JsonArray? = null
+        val metaData: List<WCMetaData>? = null
```
hichamboushaba, line 54: We need to update the Woo app to align with this API change

anitaa1990 (approved):
LGTM :shipit:

PR #2029, opened 2021-06-14
Woo: fetch and expose order's meta data

hichamboushaba:
This PR fetches and exposes the `meta_data` of the WC Orders, so that the client apps can read them.
This is based on the changes done in #2028 to have a common class representing the meta data, and it's needed to pre-fill the shipping address's phone number (issue https://github.com/woocommerce/woocommerce-android/issues/4195)

#### Testing
1. Create an order in your store.
2. Open the order details in the web client.
3. Go to the section "Custom Fields" and add some custom data.
5. Apply the patch from below and run the example app.
6. Go to Woo, then Orders.
7. Click on "Fetch single order"
8. Enter the order id.
9. Notice that the meta data of the order is printed, and that it includes the field you added.

#### Patch for testing
```
Index: example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
IDEA additional info:
<+>UTF-8
--- example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt	(revision ef4072a5a419c779716589c5a23b621dff278644)
+++ example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt	(date 1623694920378)
@@ -401,6 +401,7 @@
                                     })
                             )?.let {
                                 prependToLog("Single order fetched successfully!")
+                                prependToLog(it.getMetaDataList().toString())
                             } ?: prependToLog("WARNING: Fetched order not found in the local database!")
                         }
                     }

```

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
@@ -30,7 +30,7 @@ open class WellSqlConfig : DefaultWellConfig {
     annotation class AddOn
 
     override fun getDbVersion(): Int {
-        return 153
+        return 154
     }
 
     override fun getDbName(): String {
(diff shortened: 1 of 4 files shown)


wzieba:
@hichamboushaba do you know how I can select a site when using `Fetch single order` in FluxC sample app?

hafizrahman:
@wzieba You can't afaik, in that area the example app picks whatever first site in the SiteModel table that has WooCommerce activated.

wzieba:
Thanks @hafizrahman ! Yeah, I think you're right, there's no way to select a site. I think it would be a very valuable feature - I usually have like 5 test sites activated 😄 . I've created an issue for that, I think it might be a good HACK Week project. #2033 

For now, I'll try to determine what site am I using @hichamboushaba in the requests and we'll be back here with results.

hichamboushaba:
Thanks @wzieba, I didn't notice this part of the example was done differently, we should update it to allow picking a site. I will try to find some time to update it in a separate PR.

wzieba (approved):
Looks and works good!

PR #2031, opened 2021-06-16
Plan offers store refactor

develric:
This PR moves the Plan Offers persistence from Well Sql to the WP Android room database (this DB is already available from the Blogging Reminders project, I thought it was fine to use this same db for Plan Offers). Companion PR [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14940).

In my understanding this should not affect Woo app, but we can probably ask Woo folks to check this PR as well given we are asking them to check the #2018

In comparison to the previous implementation I tried to leverage Relations between involved tables (with onDelete CASCADE from the parent table) and transactions in the way it made sense for me. 

The PlanOffersDao was introduced but I also kept the PlanOffersSqlUtils as an extra abstraction layer in cases where we could need to use multiple DAOs or similar. 

NOTE: I'm adding a 3 new tables but didn't add any migration (relying on the `fallbackToDestructiveMigration`); happy to change this in case we need to consider different options.

Let me know wdyt 🙇


planarvoid (approved):
The changes are looking great 👍 , I've tested the feature from the WPAndroid branch and works it well 👍

PR #2032, opened 2021-06-17
Add API endpoint support for adding custom packages/activating predefined packages.

hafizrahman:
### Description:
As WooCommerce Shipping & Tax now [has API support for creating/activating packages with POST](https://github.com/Automattic/woocommerce-services/pull/2318), this PR updates FluxC to be able to make use of the endpoint.

The immediate need of it is for WooCommerce Android to support package creation / activation in the app. Related issues: 
- [Support creating custom packages](https://github.com/woocommerce/woocommerce-android/issues/3993)
- [Support activating predefined packages](https://github.com/woocommerce/woocommerce-android/issues/3992)

In general there are two types of packages:
1. **Custom package**: Here users create and enter their own package specification.
2. **Predefined packages**: This is supplied by core and is a list of packages provided by the supported carriers. Here the endpoint enables activation of such packages.

This PR adds support for both creating a custom package, and activating a predefined package, as well as add buttons to test them in the example app.

This PR also includes support for creating multiple and mixed package types at once, although tests for these cases do not currently exist in the example app, only within unit tests &mdash; I have only tested them manually.

<hr />

### Testing Instruction:
1. Run example app
2. Tap **Woo**, then tap **Shipping Labels**
2. Tap **Select Site**, and pick a site that already has **WooCommerce Shipping & Tax** plugin enabled and activated
3. Scroll down and find the two buttons: **Create A Custom Package** and **Activate a Random Predefined Package**
4. Tap the **Create A Custom Package** button and complete the creation steps. If all goes well, it will say `Custom package created: true` in the log area
6. Check in WooCommerce Shipping settings area in wp-admin (`wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings`) and confirm that the newly created custom package is listed under the **Packaging** section
7. Go back to example app and tap the  **Activate a Random Predefined Package** button.
8. The button will try to activate a predefined package randomly. The log will say `Activating [package_id] from [carrier_name]`. <br> **8.a:** If it goes through correctly, the log will then say `Predefined package activated: true`.<br>  **8.b:** There's a possibility this does not go through if the selected package is already activated. The log will say `GENERIC_ERROR: At least one of the new predefined packages has the same name as existing packages.` If this happens, retry the button again until it picks something that goes through.
9. Check in WooCommerce Shipping settings area in wp-admin again, and confirm that the newly activated package is listed under Packaging (compare the name with the `[package_id]` listed in the log).

Aside from the testing steps above, run unit tests on `WCShippingLabelStoreTest.kt` and ensure everything passes.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCShippingLabelStore.kt
```diff
@@ -254,14 +254,59 @@ class WCShippingLabelStore @Inject constructor(
...
+                        PredefinedOption(
+                                title = option.value.title,
+                                carrier = carrier.key,
+                                predefinedPackages = predefinedPackages
+                        )
+                )
+            }
+        }
+        return predefinedOptions
+    }
```
hichamboushaba: You don't have to change anything here, just wanted to share with you another way to make the code more functional and without using mutable lists at all:

```kotlin
return result.formSchema.predefinedSchema.entries.map { carrier ->
            carrier.value.map { option ->
                val predefinedPackages = option.value.definitions.map { definition ->
                    PredefinedPackage(
                            id = definition.id,
                            title = definition.name,
                            isLetter = definition.isLetter,
                            dimensions = definition.outerDimensions,
                            boxWeight = definition.boxWeight ?: 0f
                    )
                }

                PredefinedOption(
                        title = option.value.title,
                        carrier = carrier.key,
                        predefinedPackages = predefinedPackages
                )
            } // At this step, we end up with a list of lists of PredefinedOptions
        }.flatMap { it } // Flat it
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCShippingLabelStore.kt
```diff
@@ -453,4 +498,23 @@ class WCShippingLabelStore @Inject constructor(
...
         }
     }
+
+    suspend fun createPackages(
+        site: SiteModel,
+        customPackages: List<CustomPackage> = emptyList(),
+        predefinedPackages: List<PredefinedOption> = emptyList()): WooResult<Boolean> {
+        // We need at least one of the lists to not be empty to continue with API call.
+        if (customPackages.isEmpty() && predefinedPackages.isEmpty()) {
+            return WooResult(WooError(GENERIC_ERROR, UNKNOWN))
```
hichamboushaba: A `WooResult` here will make the consumers unaware of the cause, generally when it's an issue with arguments that we don't want to allow, we would throw an `IllegalArgumentException` with a clear message of what's happening, this way the consumer would fail early, and any hidden bug will be caught during the development phase.

example/src/main/java/org/wordpress/android/fluxc/example/ui/shippinglabels/WooShippingLabelFragment.kt
```diff
@@ -327,6 +329,102 @@ class WooShippingLabelFragment : Fragment() {
...
+                                    site = site,
+                                    customPackages = listOf(customPackage),
+                                    predefinedPackages = emptyList()
+                            )
+                        }
+                        result.error?.let {
+                            prependToLog("${it.type}: ${it.message}")
+                        }
+                        result.model?.let {
+                            prependToLog("$it")
```
hichamboushaba: It would be good to print something meaningful here, to reflect that the package creation succeeded or failed.

example/src/main/java/org/wordpress/android/fluxc/example/ui/shippinglabels/WooShippingLabelFragment.kt
```diff
@@ -327,6 +329,102 @@ class WooShippingLabelFragment : Fragment() {
...
+                                customPackages = emptyList(),
+                                predefinedPackages = listOf(randomParam)
+                        )
+                    }
+                    result.error?.let {
+                        prependToLog("${it.type}: ${it.message}")
+                        return@launch
+                    }
+                    result.model?.let {
+                        prependToLog("$it")
```
hichamboushaba: Same as above.

example/src/test/java/org/wordpress/android/fluxc/wc/shippinglabels/WCShippingLabelStoreTest.kt
```diff
@@ -644,6 +697,162 @@ class WCShippingLabelStoreTest {
...
+
+        val expectedResult = WooResult<Boolean>(wooError)
+        val errorRequestResult = store.createPackages(
+                site = site,
+                customPackages = emptyList(),
+                predefinedPackages = sampleListOfTwoIdenticalPredefinedPackages
+        )
+        assertEquals(errorRequestResult, expectedResult)
+        assertEquals(expectedResult.error.message, "The new predefined package names are not unique.")
+    }
```
hichamboushaba: These tests are overcomplicated, and not useful IMO 🙂.
The system under test (SUT) doesn't know and shouldn't care if passing an id of an existing package would fail or success, it only cares about the response it gets from restClient, and then converts this response to the upper layers, so the test should only care about success or failure, and this success or failure is just mocked.
Which means that all those tests are similar and test the same thing as having one test case like this:
```kotlin
@Test
    fun `returns error when the API call fails`() = test {
        val errorMessage = "error message"
        val response = WooPayload<Boolean>(WooError(GENERIC_ERROR, UNKNOWN, errorMessage))

        whenever(restClient.createPackages(site = any(), customPackages = any(), predefinedOptions = any()))
                .thenReturn(response)

        val expectedResult = WooResult<Boolean>(response.error)
        val errorRequestResult = store.createPackages(
                site = site,
                customPackages = emptyList(),
                predefinedPackages = sampleListOfOnePredefinedPackage
        )
        assertEquals(errorRequestResult, expectedResult)
        assertEquals(expectedResult.error.message, errorMessage)
    }
```
The thing to always remember, when we are testing a SUT, we test its behavior, not the behavior of other parts, so as long it doesn't care about passing duplicate packages or existing packages, we don't test it.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/shippinglabels/ShippingLabelRestClient.kt
```diff
@@ -336,6 +338,88 @@ class ShippingLabelRestClient @Inject constructor(
...
+            // Get all predefined options having the same carrier.
+            val predefinedOptionsForThisCarrier = predefinedOptions.filter { it.carrier == carrier }
+            // Get all package id(s) included in all options.
+            predefinedOptionsForThisCarrier.forEach { option ->
+                option.predefinedPackages.forEach {
+                    packageIds.add(it.id)
+                }
+            }
+            predefinedParam[carrier] = packageIds
+        }
```
hichamboushaba: This can be optimized using Kotlin's functions: 
```kotlin
        val predefinedParam = predefinedOptions
                .groupBy { it.carrier }
                .map { (carrier, options) ->
                    carrier to options.flatMap { it.predefinedPackages }.map { it.id }
                }
                .asMap()
```
The current implementation and while it works iterates over the array multiple times which is not optimal.

The `asMap` is an [alias import](https://kotlinlang.org/docs/packages.html#imports) of `kotlin.collections.toMap` to avoid collision with the `toMap` we have in utils, it's done using the following code:
```kotlin
import kotlin.collections.toMap as asMap
```

hichamboushaba (changes requested):
Great work @hafizrahman 👏 
I left some comments, but nothing major, most of them are to share knowledge :)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCShippingLabelStore.kt
hafizrahman: Oh good idea, I was wondering how to do this properly too. I'll work on that.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCShippingLabelStore.kt
hafizrahman: Ah thanks! Initially the mutable list solution was there because I wanted to avoid the list of lists situation. Thanks for introducing to `flatMap`, something new in my toolbox! I'll use this version, but with use `flatten()` instead as a replacement suggested by AS.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/shippinglabels/ShippingLabelRestClient.kt
hafizrahman: Awesome! `groupBy` is another new tool that precisely does what I needed to do. `flatMap` tidies up the iterations in the original code really nicely. Didn't know `kotlin.collections.toMap` exist either. Thanks for all these :)

hichamboushaba: Really happy to be of help, that's the ultimate goal of PRs, to share knowledge, and we all should learn from it 🙂

example/src/test/java/org/wordpress/android/fluxc/wc/shippinglabels/WCShippingLabelStoreTest.kt
hafizrahman: Ah good point. Initially I added these tests by replicating the available tests in `woocommerce-services`, but I understand now that they make sense there as they're testing the various cases the API has to care about, while here we are only testing the store layer.

As there are multiple possible error message returned by the API depending on what's happening, do you think it still makes sense to have multiple tests for each of those messages?

hichamboushaba: > As there are multiple possible error message returned by the API depending on what's happening, do you think it still makes sense to have multiple tests for each of those messages?

IMO no, as long as we are not manipulating the error message in any way, this line `assertEquals(expectedResult.error.message, errorMessage)` is enough for testing it.

hichamboushaba (approved):
Nice work @hafizrahman, :shipit:

PR #2035, opened 2021-06-24
Making SiteSTore open to allow mocking in androidTest.

develric:
This PR makes the SiteStore open to allow for connected tests mocking in WPAndroid specifically (companion PR [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14922)). 
The original SiteStore.java was not final as well, so in the end we are bringing back the original condition in my understanding.

I was looking for some annotation similar to VisibleForTesting, but could not find something suitable at hand (we can maybe evaluate the introduction of [all-open plugin](https://kotlinlang.org/docs/all-open-plugin.html) in fluxc). Placed a comment in the code to document the why the class was marked open.


planarvoid (approved):
LGTM 👍

PR #2037, opened 2021-06-25
Change custom's item quantity to Int

hichamboushaba:
To align with the web client, and to avoid issues with the API, we need to change the type of item's quantity in the Custom's data to Integer.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/shippinglabels/WCShippingPackageCustoms.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/shippinglabels/WCShippingPackageCustoms.kt
@@ -27,7 +27,7 @@ data class WCShippingPackageCustoms(
 data class WCCustomsItem(
     @SerializedName("product_id") val productId: Long,
     val description: String,
-    val quantity: Float,
+    val quantity: Int,
     val value: BigDecimal,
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/shippinglabels/WCShippingPackageCustoms.kt
```diff
@@ -27,7 +27,7 @@ data class WCShippingPackageCustoms(
 data class WCCustomsItem(
     @SerializedName("product_id") val productId: Long,
     val description: String,
-    val quantity: Float,
+    val quantity: Int,
```
kidinov, line 30: Just to get the context of the change. What API returns float or int? Does the data come from `LineItem`?

hichamboushaba: The API doesn't return anything, we pass this in the request, and yes, we get it from Order's LineItem, but this specific API accepts only integers, as it's proxying the request to a third party API for purchasing the labels.

A side note, please don't merge it until the accompanying PR https://github.com/woocommerce/woocommerce-android/pull/4267 is reviewed too.

kidinov (approved):

PR #2038, opened 2021-06-25
Check blogging reminders presence

planarvoid:
This PR adds a method to check whether any blogging reminders item has been already stored in order to find out whether to show the prologue screen from settings.


renanferrari (approved):
Thanks for the changes!

LGTM 👍

PR #2039, opened 2021-06-28
[Gutenberg] Editor settings w/ gallery flag

mkevins:
#### Related PRs

`gutenberg`: https://github.com/WordPress/gutenberg/pull/33014
`gutenberg-mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3678
`WordPress-Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/14955

### Description

This PR implements a store for fetching and caching the editor settings as a string of raw JSON. The intent here is to continue to rely on FluxC to handle the API endpoint mapping, authentication, and caching layers, but defer parsing to the endpoint consumers in JavaScript. Some basic validation of the response is still present, here, since Gson is "baked into" the method that performs the fetch. This allows us to ensure that only valid JSON is stored in the table.

#### A Shared Endpoint

Since the Global styles rely on the same endpoint (editor settings), this PR also adds an `EditorThemeAction` `UPDATE_EDITOR_THEME`, alongside the existing `FETCH_EDITOR_THEME`. It extracts a method in the `EditorThemeStore` for updating the table with a fetch response, which can be used by the fetch action and update action alike. This decouples the fetch from the local database update, allowing for more flexible integration with other middlewares.

We then dispatch the update action for the `EditorThemeStore` after a successful response from the endpoint shared with the raw editor settings store, allowing us to consolidate the request(s) into a single fetch.

Note: The actual endpoint used for the fetching of editor settings is still being discussed, and it may be necessary to include conditional logic and / or a fallback mechanism similar to that being developed for the global styles work.


antonis (commented):
As discussed (internal ref: `pb3aDo-12I#comment-1912`) another approach will be followed for now https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2065

PR #2043, opened 2021-06-30
Add plugin support to wp org

planarvoid:
This PR adds calls to the WP Api. It uses similar approach as the React native store (and there is some extracted common functionality). The methods are to fetch, delete, install and configure plugins. This PR could be tested with [the WPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14966). The code is quite hacky but seems to work alright. Big part of this PR are the tests.

I'm pinging @mchowning as a reviewer because I touched the React native store. Everything should be OK though.


fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginCoroutineStore.kt
```diff
@@ -0,0 +1,227 @@
...
+        if (!payload.isError && payload.data != null) {
+            // Give a second to the server as otherwise the following configure call may fail
+            delay(1000)
+            val configureEvent = syncConfigureSitePlugin(site, payload.data.name, payload.data.slug, true)
+            dispatcher.emitChange(configureEvent)
+        }
+        return event
+    }
+
+    private suspend fun <T : Payload<BaseNetworkError?>> executeWPAPIRequest(
```
mchowning, line 147: There's a lot of duplication between this method and the `ReactNativeStore::executeWPAPIRequest` method. Just curious if there is a specific reason you decided against extracting out and sharing more of that common functionality.

planarvoid: I was thinking about that and you're right there is a lot of duplication. I was a bit afraid of breaking things but I think I can look into it.

renanferrari (approved):
@planarvoid Thanks for the changes! I've tested the solution here and it works pretty well, as far as I can see. 

To keep everyone in the loop: since @planarvoid is switching teams, I'm taking over the work related to this feature. I've checked with @mchowning and we agreed that it would be ok to merge this and tackle any further improvements to the code in a follow-up PR.

So I'm going to go ahead and do that, and once the follow-up PR is ready, I'll ping @mchowning again.

LGTM :shipit:

PR #2052, opened 2021-07-09
Order status optimistic update

### LINKED ISSUE #2050: Optimisticly update order status
This issue is about FluxC-side changes of https://github.com/woocommerce/woocommerce-android/issues/4248

wzieba:
## Description

For details what's an optimistic update please check https://github.com/woocommerce/woocommerce-android/issues/4248

TLDR; we update local state (in our case database) and propagate this change to the presentation layer. In case of error - we revert this change and propagate event again.

## Test cases

Please check Woo PR https://github.com/woocommerce/woocommerce-android/pull/4394 for test cases and additional context.

Closes: #2050


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -383,32 +383,29 @@ class OrderRestClient @Inject constructor(
      * [OrderErrorType.INVALID_ID] if an order by this id was not found on the server
      */
     fun updateOrderStatus(
-        localOrderId: Int,
-        remoteOrderId: Long,
+        orderToUpdate: WCOrderModel,
```
wzieba, line 386: The reason for this signature change is that I believe we should aim to design our APIs in a way that it's imposible or at least very difficult to model an invalid state.

`localOrderId` and `remoteOrderId` are not needed to be separated, it's quite easy to call this method with ids that both won't match the same order. So I think it's better to just pass `WCOrderModel` here.

Ideally, I'd love to not pass `SiteModel` as `WCOrderModel` can point precisely to the `SiteModel` (it's functional dependant) but I would have to refer to the database in `OrderRestClient` which would violate layers separation.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
```diff
@@ -314,4 +314,12 @@ object OrderSqlUtils {
...
                     .orderBy(WCOrderShipmentProviderModelTable.COUNTRY, SelectQuery.ORDER_ASCENDING)
                     .asModel
+
+    fun getOrder(localOrderId: Int): WCOrderModel =
+            WellSql.select(WCOrderModel::class.java)
+                    .where()
+                    .equals(WCOrderModelTable.ID, localOrderId)
+                    .endWhere()
+                    .asModel
+                    .first()
```
wzieba: There's already a method `OrderSqlUtils#getOrderForIdSet` but it accepts `OrderIdSet` as an argument. I've preferred to get the order by it's primary key, which is `WCOrderModelTable.ID`.

For a context of why `OrderIdSet` has been introduced, please see https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/800

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -504,7 +503,20 @@ class WCOrderStore @Inject constructor(
     }
 
     private fun updateOrderStatus(payload: UpdateOrderStatusPayload) {
-        with(payload) { wcOrderRestClient.updateOrderStatus(localOrderId, remoteOrderId, site, status) }
+        with(payload) {
+            val rowsAffected = updateOrderStatusLocally(payload.order.id, payload.status)
+
+            emitChange(OnOrderChanged(rowsAffected).apply { causeOfChange = WCOrderAction.UPDATED_ORDER_STATUS })
+
+            wcOrderRestClient.updateOrderStatus(payload.order, site, status)
```
wzieba, line 511: This is a core change for this PR. Update order status local database, emit a change and the perform update on remote.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
```diff
@@ -314,4 +314,12 @@ object OrderSqlUtils {
                     .endWhere()
                     .orderBy(WCOrderShipmentProviderModelTable.COUNTRY, SelectQuery.ORDER_ASCENDING)
                     .asModel
+
+    fun getOrder(localOrderId: Int): WCOrderModel =
```
kidinov: Not sure what convention is there, but looking into other methods it might be called `getOrderByLocalId`?

kidinov (changes requested):
Looks good! Just a minor question

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
wzieba: Good note, thanks! Renamed in 5198266. 

Just a friendly reminder: please don't merge this PR until woocommerce/woocommerce-android#4394 is approved.

Thanks for review!

kidinov (approved):

PR #2055, opened 2021-07-13
Map `Product`'s `total_sales` safely.

wzieba:
## Description

This PR adds a safety mechanism for cases when `total_sales` value was having a value that outreached `Integer` type in Kotlin. For details, please see: https://github.com/woocommerce/woocommerce-android/issues/3967

## Woo PR

https://github.com/woocommerce/woocommerce-android/pull/4419

### Not migrating database column

`TOTAL_SALES` column is in our SQLite database a type of Integer:
image

The max value we can put there is 2^63-1 ([source](https://www.sqlite.org/datatype3.html)) which is the same as Kotlin's Long max value ([source](https://kotlinlang.org/docs/basic-types.html)).

### ~String conversion~

It's not valid anymore, see [discussion](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2055#discussion_r668968367).

Deprecated:

I've decided to use `String` in response model and then parse it to Long (if it's possible) as our API can return us a numeric value with size bigger than maximum value of Kotlin's Long. Sample:

Having `total_sales` value set to `9999999999999999999999999999999999999999999999999`, we receive response of `9 223 372 036 854 776 000`. The max value for Long in Kotlin is `2^63 -1 = 9,223,372,036,854,775,807` and

```
(from API) 9 223 372 036 854 776 000 > (max Kotlin) 9 223 372 036 854 775 807
```

so with just converting to Kotlin's Long we would still be vulnerable for a `ParseError`.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponse.kt
```diff
@@ -28,7 +28,7 @@ class ProductApiResponse : Response {
     var regular_price: String? = null
     var sale_price: String? = null
     var on_sale = false
-    var total_sales = 0
+    var total_sales: String? = null
```
hichamboushaba: As discussed before, I don't think we should handle cases that are not normal, and having a `total_sales` above the Long limit is one of them.
I would prefer here to keep the field as a Long, as even with the workaround, we will be displaying wrong data.

But it's not a blocker 🙂

plugins/woocommerce/src/test/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponseTest.kt
```diff
@@ -0,0 +1,35 @@
...
+        assertThat(sut.asProductModel().totalSales).isEqualTo(Long.MAX_VALUE)
+    }
+
+    @Test
+    fun `should map total sales value to zero when received value is not a number`() {
+        sut.total_sales = "not a number"
+
+        assertThat(sut.asProductModel().totalSales).isZero
+    }
+}
```
hichamboushaba: Do you think this logic needs a test? I just want to hear your opinion 🙂 

Personally I think that tests are good, but they increase the cost of maintenance down the road, and IMO the logic of the ApiResponse is quite straightforward and doesn't require unit tests.

plugins/woocommerce/build.gradle
```diff
@@ -23,6 +23,7 @@ android {
 
     sourceSets {
         main.java.srcDirs += 'src/main/kotlin'
+        test.java.srcDirs += 'src/test/kotlin'
```
hichamboushaba: I think we should get the opinion of someone else before approving this change, as from what I see, the tests have been always part of the `example` module, and I want to make sure there isn't a reason we don't know about before adding tests to the other modules.

hichamboushaba (commented):
Looks good, I left some comments to discuss about them.

plugins/woocommerce/build.gradle
hichamboushaba: I think the tests have been always part of the `example` to use Robolectric, so we should be fine, but personally, I would prefer to add the above test (if it's necessary 🙂 ) to the example app just for the sake of consistency with the rest of tests.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponse.kt
wzieba: Sure but could you please define "not normal"? I mean I was able to set `total_sales` value that will outreach Kotlin's `Long` via standard, not modified by any plugin value in the web version of WooCommerce (see description).

hichamboushaba: You overrode its value using a custom field, but normally the regular field will count the total sales of the product, which shouldn't reach this value.
IMO If we think that we should support those cases as well, we should convert the value to `BigDecimal` instead of `Long`, because otherwise we are just showing wrong data 🙂

plugins/woocommerce/src/test/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponseTest.kt
wzieba: Sure! I think unit tests are completely natural way to verify certain behaviours and I think there's no really a maintenance cost. 

In one of the books about programming I read about [double-entry bookkeeping](https://en.wikipedia.org/wiki/Double-entry_bookkeeping) which is basically what unit tests do: we change the code and we assert that this change was an intentional one in unit tests.

Unit tests describe the intentions much better than anything else. Now, after reading just tests, I'm able to understand decision of why `total_sales` are `String` and not a `Long`. It's the best documentation which doesn't go out-of-date (comparing to comments in code).

About maintenance cost (in a meaning: maintenance of the whole codebase): I think there's no additional overhead + I think unit tests (all kind, even this small) lower maintenance cost. I'm super happy every time I have to fix unit tests after some of my change because I know that the behaviour changed and I can react to that.

> the logic of the ApiResponse is quite straightforward

IMO it's really difficult to ever determine what's straightforward and what's not. For some people having `Amount of sales` kept in `String` might be very not-straightforward so they can plainly change it to Long some time in the future.

WDYT about those arguments? What is your strategy to decide of writing or not writing unit tests?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponse.kt
wzieba: > You overrode its value using a custom field

Right, that's convincing argument for "not normal" behaviour :) 

Other way to reproduce is to create an order, add a product with quantity of "many, many 9s" and (then its quantity is transformed `9223372036,854775807` and then create a new order with any quantity. This way we outreach the limit again and we result with `9223372036854776000` value.

But I understand it's totally a not-realistic corner case we probably shouldn't be worry about.

I'll change this to simple `Long` 👍 .

plugins/woocommerce/build.gradle
wzieba: Good idea with keeping codebase in a consistante manner. I'll be removing the test though as you [convinced me](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2055#discussion_r668994824) to use `Long` instead of `String` 🙂.

plugins/woocommerce/src/test/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponseTest.kt
hichamboushaba: Thanks for the detailed response 🙂 

> think there's no additional overhead

Interesting, I always think, and I think that it's shared by many developers that there is a cost for unit tests, once you refactor something, you will have to spend time fixing the tests.

> It's the best documentation which doesn't go out-of-date (comparing to comments in code).

Good point 👍 

> IMO it's really difficult to ever determine what's straightforward and what's not

I agree, I just think that a simple conversion like this is straightforward, especially since I don't think it should be a String at all 😄.

Anyway, I wanted just to hear your opinion, but as I said, it's not a blocker.

wzieba: > once you refactor something, you will have to spend time fixing the tests.

Right and for me it's a time well spent - understanding what did the change caused and if it was intentional. Sometimes it happens to be a side effect and it saves a lot of time to discover it that early.

> I agree, I just think that a simple conversion like this is straightforward, especially since I don't think it should be a String at all 😄.

Okay, on this one you got me 😄 true - as it shouldn't be a String, I've removed the test as I can agree that plain passing Long value is not a testable feature :).

> Anyway, I wanted just to hear your opinion, but as I said, it's not a blocker.

Always a pleasure to discuss things with you, thanks!

wzieba:
Heads up: `Optional Connected Tests` failed due to reasons unrelated to this PR. See #fluxc-android channel for details.

hichamboushaba (approved):
Looks good, thanks @wzieba :shipit:

PR #2056, opened 2021-07-13
Add missing Room database migration path

renanferrari:
**Background context:** we are currently using Room to manage the database for two features: Blogging Reminders and Plan Offers. After the latter was added in #2031, no migration path was provided on the assumption that it was not needed. However, it seems that when a migration path is not defined, Room clears and recreates the entire database. This was causing an issue where Blogging Reminders data was being cleared out after an app update: https://github.com/wordpress-mobile/WordPress-Android/issues/15022

To fix that issue, this PR adds the missing migration path (fc3a1a3). I've reused the SQL statements provided by the schema that is exported to the `2.json` file on the assumption that's all that's needed to keep the Plan Offers feature working as expected. 

In addition to that, this PR also is also adding the aforementioned schema file (7516b31), since that's what's recommended by [Room's documentation](https://developer.android.com/training/data-storage/room/migrating-db-versions#export-schema):

>The exported JSON files represent your database's schema history. You should store these files in your version control system, as it allows Room to create older versions of the database for testing purposes.

Even though this is not needed to fix the issue at hand, we also have another good reason to do so, which is that from version `2.4.0-alpha01` and higher, Room is introducing support for [automated migrations](https://developer.android.com/training/data-storage/room/migrating-db-versions#automated) which are going to rely on these files (and some extra configuration) to be performed.

Two considerations we need to make in the future:

- Maybe we should remove `fallbackToDestructiveMigration`? This way, instead of silently deleting the database, the app would crash after a migration path was not found. Maybe something to consider, as it would allow us to catch this kind of problem earlier. On the other hand, this might not be necessary once we update Room to `2.4.0` and start using automated migrations.
- We should probably consider start adding [migration tests](https://developer.android.com/training/data-storage/room/migrating-db-versions#test).

Note: this PR is targeting a new `release/1.21.1` branch as we will probably need to cut a new release from this. Please use the corresponding WPAndroid PR to test this: https://github.com/wordpress-mobile/WordPress-Android/pull/15024

### DIFF
--- a/fluxc/schemas/org.wordpress.android.fluxc.persistence.WPAndroidDatabase/2.json
+++ b/fluxc/schemas/org.wordpress.android.fluxc.persistence.WPAndroidDatabase/2.json
@@ -1,0 +1,241 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 2,
+    "identityHash": "b096a48b9e2f42699f423e61fe4eb9ae",
+    "entities": [
+      {
+        "tableName": "BloggingReminders",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`localSiteId` INTEGER NOT NULL, `monday` INTEGER NOT NULL, `tuesday` INTEGER NOT NULL, `wednesday` INTEGER NOT NULL, `thursday` INTEGER NOT NULL, `friday` INTEGER NOT NULL, `saturday` INTEGER NOT NULL, `sunday` INTEGER NOT NULL, PRIMARY KEY(`localSiteId`))",
+        "fields": [
+          {
+            "fieldPath": "localSiteId",
+            "columnName": "localSiteId",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "monday",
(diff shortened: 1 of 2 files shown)


develric (approved):
Hi @renanferrari 👋 ! Thanks for looking into this 🙇 ; was able to reproduce and check that the added migration behaves as expected (reminders are preserved and plans works). Also checked by diffing the sql content of resulting db using `select SQL from sqlite_master` and it matches as expected 👍 

>Two considerations we need to make in the future: [...]

If we are going (as it seems) to further leverage room I agree it makes sense to explore the automated migrations that seems promising (when we upgrade to 2.4.0) and we should start to include migration tests (we are actually exploring those in another running project where we are going to include them 👍 ...and probably we can consider to remove the fallbackToDestructiveMigration as part of it 🤔  )

PS: I think I've read a discussion about this somewhere (maybe on some internal P2) but cannot find a reference right now (reporting just to share for future discussion as well, but no need to change anything here right now); once we continue to add tables or fields into the room db, to avoid to have all the migrations grow too much into a single file, we could consider to have each migration object into a dedicated file. Something we can further discuss going ahead 🙇 .

PR #2064, opened 2021-07-19
Refactor fetch plugin action

malinajirka:
Method `fetchWooCommerceServicesPluginInfo` used to fetch all plugins, store them into the database and filter out all plugins but "WooCommerceServices" plugin. This PR refactors the code so the method is a bit more reusable and doesn't cause side-effects.

WPAndroid is not affected
WCAndroid - PR [here](https://github.com/woocommerce/woocommerce-android/compare/update-fluxc-fetch-plugins?expand=1)

This PR
- renames `fetchWooCommerceServicesPluginInfo` to `fetchSitePlugins` - returns all fetched plugins instead of just the Service plugin (previously the method was causing a side-effect since it fetched all plugins and stored them into the database even though the name mentioned just the Service plugin).
- renames `getWooCommerceServicesPluginInfo` to `getSitePlugin` and introduces `WooPlugin` parameter to let the client filter - to make the method reusable and not tied just to the Service plugin.

To test:
1. Run the example app
2. Sign in and tap on "Woo"
3. Tap on Shipping labels
4. Select site
5. Tap on "Get Shipping plugin info"
6. Verify the log display the plugin info

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/shippinglabels/WooShippingLabelFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/shippinglabels/WooShippingLabelFragment.kt
@@ -664,7 +664,7 @@ class WooShippingLabelFragment : Fragment() {
                     result.error?.let {
                         prependToLog("${it.type}: ${it.message}")
                     }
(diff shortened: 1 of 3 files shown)


nbradbury (approved):
:shipit:

PR #2065, opened 2021-07-19
[Gutenberg] Pass the Gallery v2 Flag over from the editor

antonis:
### Related PRs
`WordPress-Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/15129
`gutenberg-mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3793
`gutenberg`: https://github.com/WordPress/gutenberg/pull/33544

### Description
This PR adds the Gallery v2 Flag to the EditorTheme to be passed down to the Editor

### To test
See https://github.com/WordPress/gutenberg/pull/33544


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/EditorThemeSqlUtils.kt
```diff
@@ -91,6 +91,7 @@ class EditorThemeSqlUtils {
             get
             @JvmName("setIsFSETheme")
             set
+        @Column var galleryWithImageBlocks: Boolean = false
```
chipsnyder, line 94: This works well and I don't have a problem with this approach, I would like your thoughts on the alternative approach I took for [iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/16832/files#diff-e67698cf9a52c271ce97ef3a21d59a92afe967789690b4cf23f4ec6657639961R76-R84).  

I didn't want to add a temporary key to the data model so I stored it in what would be an `EditorThemeElement` as a new type `experimentalFeatures` to make space for future experiments as well. My overall goal was I don't want to create multiple DB changes if we do a lot of experiments but the risk is my approach is added complexity if we don't do more experimental features. 

FSE is the space where I can see more experiments happening. WDYT?

antonis: This is indeed a good idea @chipsnyder towards future proofing our implementation.
On Android thought it seems that it is a bit [tricky to alter the type check](https://www.sqlite.org/lang_altertable.html) in the [`EditorThemeElement` table](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/43ec4755ba378fd6ed5c6df9b1d4da482ea4d872/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt#L1223) in order to add the new `experimentalFeature` type. An alternative would be to drop and recreate the table or create an entirely new table for the experiments. In any case I think this will add more complexity at this point. I would suggest to keep this as is for now and refactor when the need comes up with a future experiment. Wdyt?

chipsnyder: Ah yeah, I see the dilemma there. I agree that can be a better item to rework if need at the next instance of an experimental flag.

chipsnyder (commented):
I wasn't able to get the Gallery v2 to appear on this build also while testing the analytics the gallery key always showed up as false. 

I think the gap is that in EditorThemeSqlUtils.kt [makeEditorTheme](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/0193e64889a94b44c507c6aba5b83ce0fc96bc61/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/EditorThemeSqlUtils.kt#L64-L76) doesn't handle saving the boolean but perhaps there is something else I'm missing. 

I tested by generating a self-hosted site and confirmed I get the expect value from the API.

mkevins (approved):
Tested via: https://github.com/wordpress-mobile/WordPress-Android/pull/15134#pullrequestreview-744693701

Thanks Antonis! LGTM!

PR #2066, opened 2021-07-21
Add Product metadata and addon support to WCProductModel

ThomazFB:
Summary
==========
Fixes issue https://github.com/woocommerce/woocommerce-android/issues/4405 by extending the `WCProductModel` to support the Metadata group of data and filtering the Add-ons section with the desired model mapping.

How to Test
==========
**Prerequisites:** Have your site with the [add ons plugin](https://woocommerce.com/products/product-add-ons/) installed, configured, and with at least one product with add-ons. 

- Launch the example app.
- Navigate to the Product section inside the Woo area and select the `Fetch Product addons` button.
- See that add-ons info is being printed in the console.

Screenshots
==========
| Step 1  | Step 2 | Step 3 | Step 4 |
| ------------- | ------------- | ------------- | ------------- |
| Screenshot_20210722_184253 | Screenshot_20210722_184258 | Screenshot_20210722_184306 | Screenshot_20210722_184321 |


ThomazFB:
Hey @wzieba, I'm requesting your review from this PR since you will be joining the Add-ons product soon. Since this PR introduces the first data parsing for our future work, reviewing it may be a good entry point for technical context.

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
```diff
@@ -371,6 +377,26 @@ class WooProductsFragment : Fragment() {
...
+                    pendingFetchSingleProductRemoteId = editText.text.toString().toLongOrNull()
+                    pendingFetchSingleProductRemoteId?.let { id ->
+                        prependToLog("Submitting request to fetch product by remoteProductID $id")
+                        coroutineScope.launch {
+                            wcProductStore.fetchProductListSynced(site, listOf(id))
+                                    .takeUnless { it.isNullOrEmpty() }
+                                    ?.first()?.addons
+                                    ?.logAddons()
+                        }
+                    } ?: prependToLog("No valid remoteOrderId defined...doing nothing")
```
wzieba: ❓ If we have a log for `not valid remoteOrderId`, maybe we could have one for not finding `addons`?

wzieba (approved):
Thanks for adding tests! LGTM, one minor comment about logging - I think it could be a useful one to have log for getting no add-ons when requested.

About testing: at first I didn't get add-on as I've added a global one and global add-ons will be handled differently that added here support for single product add-on.

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
ThomazFB: Actually, we do have a log for this scenario, but it's inside the `logAddons()` method, do you think it's too hidden and should be handled at the method caller instead of inside the method itself?

Here's how implemented:
```
private fun Array<WCProductAddonModel>?.logAddons() {
         this?.forEachIndexed { index, addon ->
             prependToLog(addon.description?.let { "description: $it" }.orEmpty())
             prependToLog(addon.name?.let { "name: $it" }.orEmpty())
             prependToLog("========== Product Add-on #$index =========")
         } ?: prependToLog("No addons found for this product ID")
     }
```

ThomazFB:
Thanks for your review @wzieba!

> I think it could be a useful one to have log for getting no add-ons when requested.

Nice! I replied your comment, thanks for point this out!

> About testing: at first I didn't get add-on as I've added a global one and global add-ons will be handled differently that added here support for single product add-on.

I see! Awesome feedback! I'll try to differentiate better for the next PRs if the test is about Product, Order or Global add-ons, thank you for bringing this up.

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
wzieba: Oh missed that, thanks! But I think it might not work, please correct me if I'm wrong: 

if [`addons` are not found in payload, we return an emptyList](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2066/files#diff-62817f824846374ea4f2a7620ebef16d42efdb34f1990600f9c10b215c9c4f39R113). Next, in the snippet you've provided we log `No addons found for this product ID` for null `Array<WCProductAddonModel>` but it never happens.

I think `null`, if there are no add-ons in payload, is a valid state. I'd suggest to change `WCProductModel#addons` to return `null` if there are no add-ons, WDYT?

ThomazFB: Great catch @wzieba! You're 100% correct, looks like I got fooled by my own safety measure 😅. I agree with your suggestion, it's way better to use a nullable state to represent data failure instead of an empty array. I've submitted the change right here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2066/commits/65f612f39055b57ecae23a5ea8fd62141a9cff3e.

Let me know what you think and thanks for your time reviewing the PR!

hichamboushaba:
I have a remark about this @ThomazFB 🙂, we introduced a reusable class to represent all metadatas of WC objects https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/e1fc3eb87caa0a824d512dc5c902a6ed9e0d09a2/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCMetaData.kt, WDYT about reusing it instead of adding the new class `ProductMetadata`?

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
```diff
@@ -371,6 +377,26 @@ class WooProductsFragment : Fragment() {
...
+                        "Enter the remoteProductId of product to fetch the addons:"
+                ) { editText ->
+                    pendingFetchSingleProductRemoteId = editText.text.toString().toLongOrNull()
+                    pendingFetchSingleProductRemoteId?.let { id ->
+                        prependToLog("Submitting request to fetch product by remoteProductID $id")
+                        coroutineScope.launch {
+                            wcProductStore.fetchProductListSynced(site, listOf(id))
+                                    .takeUnless { it.isNullOrEmpty() }
+                                    ?.first()?.addons
+                                    ?.logAddons()
```
wzieba: I think here we also have to adjust the code: 
```suggestion
                                    ?.first()?.addons
                                    .logAddons()
```

Otherwise, when `addons` is `null`, `logAddons()` method won't be executed. I've tested it after removing `?` and now it seems to work 🙂:

image

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
wzieba: Thanks for change! I've added one more comment here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2066/files#r676488729

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
ThomazFB: This was a lack of attention of mine, sorry, it's fixed here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2066/commits/096b4455d600ace1465357ab5e896cb7505a9afd

ThomazFB:
@wzieba, once again, thanks for catching issues, amazing review work! The PR is ready for another round.

> WDYT about reusing it instead of adding the new class ProductMetadata?

Hey @hichamboushaba, thanks for letting me know about this, I wasn't aware that metadata is present in the same data format for other objects, great tip! I've deleted the internal `ProductMetadata` and adjusted the `WCProductModel` to take advantage of the existing `WCMetaData` 👍.

wzieba:
Thanks @ThomazFB ! It all looks great now! Also thank you @hichamboushaba for your suggestion 🙌 .

PR #2073, opened 2021-07-27
Add dont_change_homepage to activate theme API call

kashandanish:
## Part of: 
Issue: https://github.com/wordpress-mobile/WordPress-Android/issues/15014
Main PR: https://github.com/wordpress-mobile/WordPress-Android/pull/15055

## Description:
This PR adds a parameter `dont_change_homepage` to `POST /sites/$site/themes/mine` which can be set to `true` if we dont want to use theme's homepage layout when activating the theme.

## API Specs
https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/themes/mine/

## To Test
It can be tested with https://github.com/wordpress-mobile/WordPress-Android/pull/15055. There is also this `testActivateTheme()` test which tests theme activation but it is  not specific to `dont_change_homepage` param.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
@@ -85,6 +85,25 @@ public class ThemeStore extends Store {
         }
     }
 
+    /**
+     * A request and response payload class for theme activation call
+     */
+    public static class ActivateThemePayload extends Payload<ThemesError> {
+        public SiteModel site;
+        public ThemeModel theme;
+        public boolean dontChangeHomepage;
+
+        public ActivateThemePayload(SiteModel site, ThemeModel theme) {
+            this(site, theme, true);
+        }
+
+        public ActivateThemePayload(@NonNull SiteModel site, @NonNull ThemeModel theme, boolean dontChangeHomepage) {
+            this.site = site;
+            this.theme = theme;
+            this.dontChangeHomepage = dontChangeHomepage;
+        }
+    }
+
     public static class FetchStarterDesignsPayload extends Payload<BaseNetworkError> {
         @Nullable public Float previewWidth;
         @Nullable public Float previewHeight;
@@ -181,10 +200,12 @@ public class ThemeStore extends Store {
     public static class OnThemeActivated extends OnChanged<ThemesError> {
         public SiteModel site;
         public ThemeModel theme;
+        public boolean dontChangeHomepage;
 
-        public OnThemeActivated(SiteModel site, ThemeModel theme) {
+        public OnThemeActivated(SiteModel site, ThemeModel theme, boolean dontChangeHomepage) {
             this.site = site;
             this.theme = theme;
+            this.dontChangeHomepage = dontChangeHomepage;
         }
     }
 
@@ -267,10 +288,10 @@ public class ThemeStore extends Store {
                 handleCurrentThemeFetched((FetchedCurrentThemePayload) action.getPayload());
                 break;
             case ACTIVATE_THEME:
-                activateTheme((SiteThemePayload) action.getPayload());
+                activateTheme((ActivateThemePayload) action.getPayload());
                 break;
             case ACTIVATED_THEME:
-                handleThemeActivated((SiteThemePayload) action.getPayload());
+                handleThemeActivated((ActivateThemePayload) action.getPayload());
                 break;
             case INSTALL_THEME:
                 installTheme((SiteThemePayload) action.getPayload());
@@ -415,17 +436,17 @@ public class ThemeStore extends Store {
         emitChange(event);
     }
 
-    private void activateTheme(@NonNull SiteThemePayload payload) {
+    private void activateTheme(@NonNull ActivateThemePayload payload) {
         if (payload.site.isUsingWpComRestApi()) {
-            mThemeRestClient.activateTheme(payload.site, payload.theme);
+            mThemeRestClient.activateTheme(payload.site, payload.theme, payload.dontChangeHomepage);
         } else {
             payload.error = new ThemesError(ThemeErrorType.NOT_AVAILABLE);
             handleThemeActivated(payload);
         }
     }
 
-    private void handleThemeActivated(@NonNull SiteThemePayload payload) {
-        OnThemeActivated event = new OnThemeActivated(payload.site, payload.theme);
+    private void handleThemeActivated(@NonNull ActivateThemePayload payload) {
+        OnThemeActivated event = new OnThemeActivated(payload.site, payload.theme, payload.dontChangeHomepage);
         if (payload.isError()) {
             event.error = payload.error;
         } else {
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ThemeTestJetpack.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ThemeTestJetpack.java
@@ -23,6 +23,7 @@ import org.wordpress.android.fluxc.store.SiteStore.FetchSitesPayload;
 import org.wordpress.android.fluxc.store.SiteStore.OnSiteChanged;
 import org.wordpress.android.fluxc.store.SiteStore.OnSiteRemoved;
 import org.wordpress.android.fluxc.store.ThemeStore;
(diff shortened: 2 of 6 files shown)


kashandanish:
@antonis This is ready for review. Few things I would like to mention.

1.  Created a new class [ActivateThemePayload](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2073/files#diff-3be4f77b2c2e670769618b7d0e91b4c7f77115c4cbfa763a5041e63d2ec1395bR91) for theme. I did't like the name `SiteThemePayload` because it was based on its data members. Whereas everywhere else, `Payload` classed were named after their purpose like [here](https://github.com/kashandanish/WordPress-FluxC-Android/blob/1e396596cd691ccb42233eaa687fb6434b1cdb6d/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt#L128) and even in `ThemeStore` [here](https://github.com/kashandanish/WordPress-FluxC-Android/blob/1e396596cd691ccb42233eaa687fb6434b1cdb6d/fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java#L51)

2. I created a single class for request and response payload instead of creating two identical classes.  When I looked for conventions around this I saw that `SiteThemePayload` is being used in similar fashion.

3. I could not update [these](https://github.com/kashandanish/WordPress-FluxC-Android/blob/1e396596cd691ccb42233eaa687fb6434b1cdb6d/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ThemeTestWPCom.java#L78) [tests](https://github.com/kashandanish/WordPress-FluxC-Android/blob/1e396596cd691ccb42233eaa687fb6434b1cdb6d/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ThemeTestJetpack.java#L112) because response of the [API](https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/themes/mine/) with and without `dont_change_homepage` param is same. So there is no difference in persistent layer.

kashandanish:
One of the CI builds is failing due to:
```
> Task :applyConfiguration FAILED
Unable to initialize logging
thread 'main' panicked at 'Unable to decrypt and copy files: SecretsNotPresent', src/configure.rs:193:53
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

* What went wrong:
Execution failed for task ':applyConfiguration'.
> Process 'command '/home/circleci/project/vendor/configure/configure'' finished with non-zero exit value 101
```
This seems something not specific to my changes, but I can be wrong. If there is something I need to do to fix it, please let me know.

antonis:
> This seems something not specific to my changes, but I can be wrong. If there is something I need to do to fix it, please let me know.

@kashandanish Thank you for pointing this out. This seems to happen due to the fact that we are working on a forked repository and the access to the secret store is limited. We will be able to validate that all tests pass once this is merged on `wordpress-mobile:issue/15014-theme-activation-keep-homepage`

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
```diff
@@ -85,6 +85,25 @@ public SiteThemePayload(@NonNull SiteModel site, @NonNull ThemeModel theme) {
         }
     }
 
+    /**
+     * A request and response payload class for theme activation call
+     */
+    public static class ActivateThemePayload extends Payload<ThemesError> {
```
antonis, line 91: >  1. Created a new class [ActivateThemePayload](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2073/files#diff-3be4f77b2c2e670769618b7d0e91b4c7f77115c4cbfa763a5041e63d2ec1395bR91) for theme. I did't like the name `SiteThemePayload` because it was based on its data members. Whereas everywhere else, `Payload` classed were named after their purpose like [here](https://github.com/kashandanish/WordPress-FluxC-Android/blob/1e396596cd691ccb42233eaa687fb6434b1cdb6d/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt#L128) and even in `ThemeStore` [here](https://github.com/kashandanish/WordPress-FluxC-Android/blob/1e396596cd691ccb42233eaa687fb6434b1cdb6d/fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java#L51)

I like your rationale on this. On the other hand the `SiteThemePayload` is still used for other cases. I would suggest using `SiteThemePayload` as a base class and just add the new field. This is just a suggestion thought.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
```diff
@@ -85,6 +85,25 @@ public SiteThemePayload(@NonNull SiteModel site, @NonNull ThemeModel theme) {
...
+        public SiteModel site;
+        public ThemeModel theme;
+        public boolean dontChangeHomepage;
+
+        public ActivateThemePayload(SiteModel site, ThemeModel theme) {
+            this(site, theme, true);
```
antonis, line 97: I think your decision to set the default value to `true` is in the right direction even if the current default behaviour changes 👍

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
kashandanish: I thought of doing that but then I would need to add 3-4 new classes. Because `SiteThemePayload` is shared among many of theme related APIs. Just extending `ActivateThemePayload` and keeping everything else same would not have worked either. For example If something in `deleteTheme` changes which is using `SiteThemePayload` then we would need to rewrite `ActivateThemePayload`.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
kashandanish: I took it based on the the default selected option in theme action sheet.

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
antonis: I meant declaring the class like `class ActivateThemePayload extends SiteThemePayload` and keep the usages as is now.

antonis (commented):
Great work @kashandanish 👍 

I tested your changes and everything works as expected. The code also LGTM 🎉 
I just added a couple of suggestions but nothing blocking.

> 2. I created a single class for request and response payload instead of creating two identical classes.  When I looked for conventions around this I saw that `SiteThemePayload` is being used in similar fashion.

👍 

> 3. I could not update [these](https://github.com/kashandanish/WordPress-FluxC-Android/blob/1e396596cd691ccb42233eaa687fb6434b1cdb6d/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ThemeTestWPCom.java#L78) [tests](https://github.com/kashandanish/WordPress-FluxC-Android/blob/1e396596cd691ccb42233eaa687fb6434b1cdb6d/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ThemeTestJetpack.java#L112) because response of the [API](https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/themes/mine/) with and without `dont_change_homepage` param is same. So there is no difference in persistent layer.

I wonder if we can add extra tests with the parameters set to true/false just to validate that we are getting the same response. It might not make much sense now but it may help to detect breaking api changes in the future. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/store/ThemeStore.java
kashandanish: I understand that, this is what I meant by: _If something in `deleteTheme` changes which is using `SiteThemePayload` then we would need to rewrite `ActivateThemePayload`._
Or we do this for all of them like
```
class ActivateThemePayload extends SiteThemePayload
class DeleteThemePayload extends SiteThemePayload
```

antonis: Thank you for the clarification @kashandanish 🙏 
That make perfect sense 👍

antonis (approved):
Great work @kashandanish 👍 
Thank you for taking the time to make the suggested changes 🙇 
Everything LGTM 🎉

PR #2075, opened 2021-07-29
Fix json parser exception

### LINKED ISSUE #4440: ParseError: com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but w...
Sentry Issue: [WOOCOMMERCE-ANDROID-1XV](https://sentry.io/organizations/a8c/issues/2511997980/?referrer=github_integration)

```
IllegalStateException: Expected BEGIN_OBJECT but was BEGIN_ARRAY at line 1 column 10 path $.data
    at com.google.gson.stream.JsonReader.beginObject(JsonReader.java:385)
    at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:215)
    at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:131)
    at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:222)
    at com.google.gson.Gson.fromJson(Gson.java:927)
...

JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was BEGIN_ARRAY at line 1 column 10 path $.data
    at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:226)
    at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.read(ReflectiveTypeAdapterFactory.java:131)
    at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:222)
    at com.google.gson.Gson.fromJson(Gson.java:927)
    at com.google.gson.Gson.fromJson(Gson.java:892)
...
(5 additional frame(s) were not displayed)

ParseError: com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was BEGIN_ARRAY at line 1 column 10 path $.data
    at org.wordpress.android.fluxc.network.rest.GsonRequest.parseNetworkResponse(GsonRequest.java:98)
    at com.android.volley.NetworkDispatcher.processRequest(NetworkDispatcher.java:143)
    at com.android.volley.NetworkDispatcher.processRequest(NetworkDispatcher.java:111)
    at com.android.volley.NetworkDispatcher.run(NetworkDispatcher.java:90)

FluxC: com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was BEGIN_ARRAY at line 1 column 10 path $.data: API response parse error
```

Based on conversation in Slack p1626269544365400-slack-C6H8C3G23 , this was caused by a feature in WCPay that wasn't wrapped in feature flag.

malinajirka:
This PR fixes https://github.com/woocommerce/woocommerce-android/issues/4440

There is a bug in WCPay plugin (at least in versions 2.4.0-2.7.2) and when WCPay is installed but the account setup is not finished, the /payments/accounts endpoint returns an empty array instead of a meaningful object. This workaround turns this response into an empty/default NO_ACCOUNT response.

To test
1. Test this in WCAndroid - [PR](https://github.com/woocommerce/woocommerce-android/pull/4517)


kidinov:
Just for reference, I've created an issue on this matter 991-gh-Automattic/woocommerce-payments-server

Also, I've mentioned this before already, but don't you think that this is just hiding the problem? What if a popular plugin will modify the response in a similar way and we instead of seeing a rising number in sentry, won't have any metrics?

malinajirka:
> Just for reference, I've created an issue on this matter 991-gh-Automattic/woocommerce-payments-server

Perfect, thanks!

> Also, I've mentioned this before already, but don't you think that this is just hiding the problem? What if a popular plugin will modify the response in a similar way and we instead of seeing a rising number in sentry, won't have any metrics?

If a 3rd party plugin modifies the response there is nothing we can do about it anyway, right? Or do you think the logs in Sentry would help us in such scenario?

kidinov:
@malinajirka 

> If a 3rd party plugin modifies the response there is nothing we can do about it anyway, right? Or do you think the logs in Sentry would help us in such scenario?

Not sure, but I guess we could contact the owner of the plugin to fix it? Or maybe to fix it on our side. Anyway, I think it's way better to know that things are broken than not knowing that. To me, it's pretty much as try-catch-swallowing exceptions =)

kidinov:
@malinajirka Another case that let's say the backend will return an empty array, not because the setup of WCPay is not finished, but just because of another bug. And we won't still know about that with this workaround

malinajirka:
> Not sure, but I guess we could contact the owner of the plugin to fix it? Or maybe to fix it on our side. Anyway, I think it's way better to know that things are broken than not knowing that. To me, it's pretty much as try-catch-swallowing exceptions =)

I technically agree, but I can't think of any solution. This PR fixes an issue which breaks the behavior of in-person payments. Unfortunately, it breaks the onboarding flow which is probably the worst part it can break. Even if we fix it backend-side we can't guarantee that users will be using the updated version of the plugin. If we come up with a solution which would work and wouldn't hide potential issues, I'd be all up for it. But if we don't have an alternative I don't think there is much we can do. Wdyt?

malinajirka:
> Another case that let's say the backend will return an empty array, not because the setup of WCPay is not finished, but just because of another bug. And we won't still know about that with this workaround

I think that's the case for most of our code - for example if the backend returns "abcdefg" in WCPayAccountStatusEnum, we'll convert it to `UNKNOWN`, but I think we don't log any exceptions.

If we had a shared backend I'd definitely prefer to fix it backend side. But that's not the case in WordPress ecosystem. We could potential raise the minimal required version of WCPay to a version with the fix - that's the only solution I can think of.

malinajirka:
Closing - will be fixed in the plugin, more info p2-pdfdoF-1U#comment-54

PR #2076, opened 2021-07-29
PostFormatsError: Allow null value for message argument

AmandaRiu:
Fixes the NPE crash descripted in https://github.com/wordpress-mobile/WordPress-Android/issues/15071 by allowing the error message to be `null`.


planarvoid (approved):
👍

PR #2077, opened 2021-07-30
Change type of currentDeadline field

malinajirka:
When we implemented `/payments/accounts` endpoint we thought the endpoint uses ISO8601 format. As it turned out it uses a timestamp instead. This PR fixes this issue to prevent parsing errors.

To test:
Test this in [WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/4532)


nbradbury (approved):
:shipit:

PR #2078, opened 2021-07-30
Adding DAO for Comments in WPAndroid room database.

develric:
This PR is part of the phase 1 of the Unified Comments project aiming to unify the commenting experience in the WP app. 

This PR introduces a room database that should replace the CommentModel SQLite db that is based on Well SQL. 
- We are introducing a CommentsDao to interact with the comment entities stored in the Comments table.
- We have included a migration from version 2 to the current version 3 of the database.
- We tried to cover the dao with some unit testing and a migration test. To do this we included a new dependency `room-testing` but only as an androidTestImplementation, so should not affect the main implementation tree dependency graph.

To review this PR you should review the code and check the tests are green.

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -61,6 +61,11 @@ android {
             }
         }
     }
+
+    sourceSets {
+        // Adds exported schema location as test app assets.
+        getByName("androidTest").assets.srcDirs += "$projectDir/../fluxc/schemas"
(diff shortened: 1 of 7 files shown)


khaykov (approved):
Looks good to me, @develric ! And now I know why we those random json files :)

PR #2079, opened 2021-07-30
Porting network clients for comments to kotlin.

develric:
This PR is part of the unified comments project and aims to bring the rest and xmlrpc network clients to kotlin.

I tried to extract some small parts of logic from the clients (like restoring the local comment id into the network reponse or similar) and moved them to the comments store that is using the client since that seemed a better place (at least for me). The Comments Store is coming in a separated PR.

This PR is going to be merged in a feature branch and basically should be code reviewed + checking that the tests are green.


khaykov (approved):
Looks good to me, and looks like tests are passing! Learned about DTO vs DAO too :D

PR #2080, opened 2021-07-30
Porting comments store to kotlin with coroutines and backward compatible eventbus support

develric:
This PR is part of the unified comments project and aims to bring the comments store to kotlin, coroutines and room.

The well sql based CommentModel is still available while we move ahead with the project and bring into refactoring and unification all the parts that are related to comments in WPAndroid app. The changes for this phase 1 of the project will be available in WPAndroid behind a feature flag and when enabled all the parts will be able to use the data from/to the room cache using the new store via a CommentsStoreAdapter. This is why we kept in the new store event bus actions available. Regardless, turning the feature flag off in the client app, all the comments components will use the previous store.

This PR has a WPAndroid companion app and can be checked using that. We also included some unit testing for convenience and good measure :) .

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentsStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentsStore.kt
@@ -1,0 +1,830 @@
+package org.wordpress.android.fluxc.store
+
+import org.greenrobot.eventbus.Subscribe
+import org.greenrobot.eventbus.ThreadMode
+import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.Payload
+import org.wordpress.android.fluxc.action.CommentAction
+import org.wordpress.android.fluxc.action.CommentsAction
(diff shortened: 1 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentsStore.kt
```diff
@@ -0,0 +1,832 @@
...
+        site: SiteModel?,
+        orderByDateAscending: Boolean,
+        limit: Int,
+        vararg statuses: CommentStatus
+    ): CommentEntityList {
+        if (site == null) return listOf()
+
+        return commentsDao.getCommentsByLocalSiteId(
+                localSiteId = site.id,
+                statuses = if (statuses.asList().contains(ALL)) listOf() else statuses.map { it.toString() },
```
khaykov, line 100: This is not an issue at the moment, but we might want to differentiate `ALL` that means Approved and Pending and  all comments (where all means all of them).

khaykov (approved):
Looks good! Glad we still have compatibility with events. I left only one comment, but it can be addressed in the future since we don't really use that method with All fitler.

PR #2081, opened 2021-07-31
uc v1 feature branch

develric:
This PR mostly lands changes already reviewed in previous PRs. The only change I added that should be reviewed is this  https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/023719cae6362b731beb74a299616d07a06d5b82

In principle comments value could be null there, even though kotlin compiler is not complaining since the java side is not marking with nullable annotation. So I placed a guard returning empty list eventually.

This said we can use this PR code to final check/test the feature.

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -61,6 +61,11 @@ android {
             }
         }
     }
+
+    sourceSets {
+        // Adds exported schema location as test app assets.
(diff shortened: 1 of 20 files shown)


khaykov (approved):
Optional Connected Tests are also randomly failing in develop, and since we haven't touched anything related to them, I think it's ok to go ahead with this one 🚀

khaykov:
Feel free to merge this one whenever :)

PR #2082, opened 2021-08-02
Add support for restricted_soon status

malinajirka:
Issue https://github.com/woocommerce/woocommerce-android/issues/4413

**https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2077 needs to be merged first**

Adds support for "restricted_soon" status response from the backend which indicates that there are missing required business information but the merchant still has some time to complete the setup before the account gets moved to "restricted" state.

To test
Test in WCAndroid - PR tbd


nbradbury (approved):
@malinajirka I merged #2077 and verified this works with [the related WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/4538), so we're good to merge this. :shipit:

PR #2085, opened 2021-08-05
Update for Blogging Reminders time selection

ravishanker:
Updating for Blogging Reminders time selection feature by

- Adding hour and minute columns to `BloggingReminders` table
- Bumping up DB from 3 to 4 and manual migration logic to `WPAndroidDatabase`
- Updates to Mapper, Store and Migration tests


renanferrari (commented):
Thanks for the changes! The code around the updated model is looking good overall, but I do have some questions about the DB migration logic.

Correct me if I'm wrong, but these are the changes I understand we're making here:
1. Updating the DB to v4.
2. Adding a `Migration` that takes care of migrating the DB from v3 to v4.
3. Adding a test for that new migration logic.

Is that right? Either way, would you mind providing a more thorough description of the changes this PR is introducing? I just want to make sure I'm not missing anything and that we're able to easily identify these changes in the future.

Given my assumptions above, it's not clear to me why we had to change some of the existing migration tests and some of the existing schema files.

ravishanker:
> Thanks for the changes! The code around the updated model is looking good overall, but I do have some questions about the DB migration logic.
> 
> Correct me if I'm wrong, but these are the changes I understand we're making here:
> 
>     1. Updating the DB to v4.
> 
>     2. Adding a `Migration` that takes care of migrating the DB from v3 to v4.
> 
>     3. Adding a test for that new migration logic.
> 
> 
> Is that right? Either way, would you mind providing a more thorough description of the changes this PR is introducing? I just want to make sure I'm not missing anything and that we're able to easily identify these changes in the future.
> 
> Given my assumptions above, it's not clear to me why we had to change some of the existing migration tests and some of the existing schema files.

Thanks @renanferrari for a quick review.  Looks like it's messed up after I pulled in develop where db was already bumped up to 3.  Sorry, about that.  I've now fixed up the redundant changes and also the PR description.

example/src/androidTest/java/org/wordpress/android/fluxc/persistence/room/WPAndroidDatabaseMigrationTest.kt
```diff
@@ -82,4 +83,33 @@ class WPAndroidDatabaseMigrationTest {
...
+            // populate BloggingReminders with some data
+            execSQL(""" 
+                INSERT INTO BloggingReminders 
+                (localSiteId, monday, tuesday, wednesday, thursday, friday, saturday, sunday) 
+                VALUES (1000,1, 1, 0, 0, 1, 0, 1) 
+            """)
+            close()
+        }
+
+        // Re-open the database with version 3 and provide migration to check against.
```
renanferrari: Should we update this comment to say "version 4" instead? 

I know this is a copy of the other tests, but I think we could probably improve readability here a little bit if we extract these versions to variables like:

```kotlin
val oldVersion = 3
val newVersion = 4
``` 

Then we could update the comment to something more generic, which should make it easier to copy the test in the future anyway:

```
// Re-open the database with the new version and provide migration to check against.
```

It's a bit of a nitpick, but let me know what you think.

example/src/androidTest/java/org/wordpress/android/fluxc/persistence/room/WPAndroidDatabaseMigrationTest.kt
```diff
@@ -82,4 +83,33 @@ class WPAndroidDatabaseMigrationTest {
...
+        val db = helper.runMigrationsAndValidate(WP_DB_NAME, 4, true, MIGRATION_3_4)
+
+        // Validate that the data was migrated properly.
+        val cursor = db.query("SELECT * FROM BloggingReminders")
+
+        assertThat(cursor.count).isEqualTo(1)
+        cursor.moveToFirst()
+        assertThat(cursor.getInt(0)).isEqualTo(1000)
+        assertThat(cursor.getInt(2)).isEqualTo(1)
+        assertThat(cursor.getInt(4)).isEqualTo(0)
```
renanferrari, line 113: We should probably test that the new columns have been correctly inserted here as well, right?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WPAndroidDatabase.kt
```diff
@@ -113,5 +114,26 @@ abstract class WPAndroidDatabase : RoomDatabase() {
                 }
             }
         }
+
+        val MIGRATION_3_4 = object : Migration(3, 4) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                database.apply {
+                    execSQL(
+                            "CREATE TABLE IF NOT EXISTS `BloggingReminders` (" +
```
renanferrari: I believe we should use `ALTER TABLE` here instead. Otherwise, the table will remain the same if it already exists, right? This would probably cause a crash.

Edit: I tested this and it seems that it would indeed cause a crash. Left [a comment below](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2085#discussion_r686179292) with a suggestion. Let me know what you think.

renanferrari (commented):
Thanks for the changes! The diffs make more sense now 👍

Still found a few issues around the migration logic, so I left a few comments below.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/BloggingRemindersDao.kt
```diff
@@ -32,6 +32,8 @@ abstract class BloggingRemindersDao {
         var thursday: Boolean = false,
         var friday: Boolean = false,
         var saturday: Boolean = false,
-        var sunday: Boolean = false
+        var sunday: Boolean = false,
+        val hour: Int,
+        var minute: Int
```
renanferrari: Is there any reason why `hour` is a `val` while the other properties are `var`? Apparently, it doesn't make any difference, but I just want to be sure.

Also, what do you think of defining the default value here, like the other properties?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WPAndroidDatabase.kt
```diff
@@ -113,5 +114,26 @@ abstract class WPAndroidDatabase : RoomDatabase() {
...
+                                    "`tuesday` INTEGER NOT NULL, " +
+                                    "`wednesday` INTEGER NOT NULL, " +
+                                    "`thursday` INTEGER NOT NULL, " +
+                                    "`friday` INTEGER NOT NULL, " +
+                                    "`saturday` INTEGER NOT NULL, " +
+                                    "`sunday` INTEGER NOT NULL, " +
+                                    "`hour` INTEGER NOT NULL, " +
+                                    "`minute` INTEGER NOT NULL, " +
+                                    "PRIMARY KEY(`localSiteId`))"
+                    )
```
renanferrari:
```suggestion
                    execSQL("ALTER TABLE BloggingReminders ADD COLUMN hour INTEGER DEFAULT 10 NOT NULL")
                    execSQL("ALTER TABLE BloggingReminders ADD COLUMN minute INTEGER DEFAULT 0 NOT NULL")
```

example/src/androidTest/java/org/wordpress/android/fluxc/persistence/room/WPAndroidDatabaseMigrationTest.kt
ravishanker: It makes sense, it was confusing indeed before.  Updated.  Thanks

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/BloggingRemindersDao.kt
ravishanker: I always start with val, probably looked up while typing and used var for the second one, didn't notice the difference.  Thanks for noticing it.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WPAndroidDatabase.kt
ravishanker: I was wondering the same, it should be ALTER TABLE.  Fixed it.  Thanks for spotting it.

renanferrari (approved):
Thanks for the changes! Everything looks good to go now. 

LGTM :shipit: 

Let me know if you need any help creating a new release tag!

PR #2088, opened 2021-08-05
Fetch Global Add-on Groups

wzieba:
Summary
============
This PR is part of issue https://github.com/woocommerce/woocommerce-android/issues/4410 - networking layer only.

I'll add tests and persistence in next PRs.

How to Test
===========
**Prerequisites:** Have your site with the [add ons plugin](https://woocommerce.com/products/product-add-ons/) installed, configured, and with at least one global add-on group defined. 

1. Launch the example app.
1. Navigate to the Product section inside the Woo area and select the `Fetch Product addons` button.
1. See that global add-ons groups info is being printed in the console.

Screenshot
===========
Screenshot 2021-08-05 at 20 29 00

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
@@ -2,30 +2,30 @@ package org.wordpress.android.fluxc.model.addons
 
 import com.google.gson.annotations.SerializedName
 
-class WCProductAddonModel {
+data class WCProductAddonModel(
     @SerializedName("title_format")
-    val titleFormat: AddOnTitleFormat? = null
+    val titleFormat: AddOnTitleFormat? = null,
     @SerializedName("description_enable")
-    val descriptionEnabled: String? = null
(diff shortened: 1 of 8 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
```diff
@@ -2,30 +2,30 @@ package org.wordpress.android.fluxc.model.addons
 
 import com.google.gson.annotations.SerializedName
 
-class WCProductAddonModel {
+data class WCProductAddonModel(
```
wzieba, line 5: I've changed this and `ProductAddonOption` to `data class` es for a trivial reason of ready `toString()` method but eventually I think that those classes could be `data` either way. WDYT?

ThomazFB: I think it makes perfect sense, it's a good addition!

ThomazFB (approved):
Hey @wzieba! Great work here, code is nicely done and the feature works as expected here. I have no changes to request. LGTM :shipit:

PR #2089, opened 2021-08-05
Expose and set `CookieManager` instance as the system default `CookieHandler`

renanferrari:
This PR exposes a `CookieManager` instance to the Dagger graph and sets it as the system-wide default `CookieHandler,` allowing us to share that same instance with `OkHttpClient` via its `CookieJar`.

This will be used on `WPAndroid` as a way to manually add and remove cookies to requests for debugging and testing. Check this internal ref for an example of what we're trying to achieve: pbArwn-2yE-p2

@wzieba I'm requesting your review here because you recently edited this module, but mostly because I think it would be great to have this double-checked by someone from Woo. Hope you don't mind! 😄


wzieba (approved):
Looks good!

> This will be used on WPAndroid as a way to manually add and remove cookies to requests for debugging and testing. Check this internal ref for an example of what we're trying to achieve: pbArwn-2yE-p2

I like this idea, might be helpful also for Woo in the future. Thanks for adding this!

PR #2092, opened 2021-08-06
Allow `ReleaseStack_WCBase` tests to specify test site

### LINKED ISSUE #2091: Allow `ReleaseStack_WCBase` tests to select a site.
### Current state

When we create a new site for testing WooCommerce, we create a new .com account. E.g. in sample `tests.properties` there's:
- `TEST_WPCOM_USERNAME_WOO_JETPACK`
- `TEST_WPCOM_USERNAME_WOO_JETPACK_EXTENSIONS`
- `TEST_WPCOM_USERNAME_WOO_JP_WCPAY`

So we end up with user->site association. Our tests just picks the first site we fetch from the api ([see](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/81be0ed2ca0f017be9485b1c3f4e8408af68b24e/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCBase.kt#L21)).

### Why is it problematic

We would like rather to have them all under one .com user and point to the specific site because it reduces overhead of creating a new .com account every time.

If we won't point test to the specific site, we can end up with test choosing wrong side without e.g. installed a proper plugin which will lead to test failure.

### Solution

We have to find some way to point test to a specific site of one .com user like e.g. [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/260f4f9fb6ed38b67e539446a65a103d1e915d80/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ScanTestJetpack.kt#L214) but more generic one for easier usage of future tests for WooCommerce.

cc: @malinajirka

wzieba:
Closes #2091 

This draft is ready to review. I mark it as draft as I didn't add correct values to `test.properties` as I'm not sure if this is direction we would like to resolve this issue.

### Description

This PR allows to select test site for every site depending on `ReleaseStack_WCBase`, but preserving old, default behaviour of picking first available site that was before.

### How to test?

Please update site id declared in `ReleaseStack_WCPayTest#22` with correct site id (please DM me for it if you don't have it by hand) and run:
- `ReleaseStack_WCPayTest`
- Any other `ReleaseStack_WC*` test (if it fails, make sure it doesn't on `develop`)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCBase.kt
```diff
@@ -4,23 +4,40 @@ import com.wellsql.generated.SiteModelTable
...
-    private val authenticatePayload by lazy {
-        AuthenticatePayload(BuildConfig.TEST_WPCOM_USERNAME_WOO_JETPACK,
-                BuildConfig.TEST_WPCOM_PASSWORD_WOO_JETPACK)
-    }
+    private val defaultAuthenticatePayload = AuthenticatePayload(
+            BuildConfig.TEST_WPCOM_USERNAME_WOO_JETPACK,
+            BuildConfig.TEST_WPCOM_PASSWORD_WOO_JETPACK
+    )
+
+    protected open val testSite: TestSite = FirstAvailable
```
wzieba, line 18: Default behaviour is perceived - we select first available site.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCBase.kt
```diff
@@ -4,23 +4,40 @@ import com.wellsql.generated.SiteModelTable
...
 
-    override fun getSiteFromDb(): SiteModel {
-        val wcSites = siteSqlUtils.getSitesWith(SiteModelTable.HAS_WOO_COMMERCE, true).asModel
-        if (wcSites.isEmpty()) {
-            throw AssertionError("This test account doesn't seem to have any WooCommerce sites!")
-        } else {
-            return wcSites[0]
+    override fun getSiteFromDb(): SiteModel = when (val site = testSite) {
```
wzieba, line 20: I set `val` in `when` as without this smart-casting doesn't work. TBH I don't know the reason for that, what am I missing? `testSite` is a `val` so it shouldn't be an issue 🤔

malinajirka: I'm not sure why is that either - I mean it's because the value is "open" -> but why an "open" val doesn't work with smart-casting is a mystery.

malinajirka: I might know the reason -> `open` val can have a custom getter -> the custom getter can return a different value every time you call it. Wdyt?

malinajirka (approved):
Looks good and works as expected. Thanks for working on this!!

I'm approving the PR, but we need to update the tests.properties (our secret repo) before we merge this PR.

example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCBase.kt
wzieba: > I mean it's because the value is "open"

You're right! After making it non-open, smart cast starts to work.

> (...) open val can have a custom getter

But here I'm not sure - not-open vals can also have custom getters (backing fields).

Or you mean inheriting class? If so, that's correct - inheriting class could override `testSite` and make it a class with `backing field`. This way or another you're right that custom getter seems to be an issue here.

What's strange, I've tried to have a reproduction for that on Kotlin playground, and smart cast seems to work there even with `open val` 🤔  (or am I missing something there?). See here: https://pl.kotl.in/4ET5_kvB7

malinajirka: > Or you mean inheriting class? If so, that's correct - inheriting class could override testSite and make it a class with backing field.

Yes, that's what I meant.

> What's strange, I've tried to have a reproduction for that on Kotlin playground, and smart cast seems to work there even with open val 🤔 (or am I missing something there?)

Strange indeed - I'm not sure why 😕

wzieba:
On CI 3 tests failed, all of them are from `ReleaseStack_WPApiPluginTest`:

- `testUninstallAndInstallPlugin` x2
- `testActivateAndDeactivatePlugins`

They do fail also on `develop` ([link](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-FluxC-Android/5274/workflows/4ffeb654-4059-45e7-9d3b-af0f8a92cae0/jobs/13697/tests)) but also pass locally. If it was about configuration change, they would fail locally too.

So I think we're safe to merge.

PR #2093, opened 2021-08-09
Support fetching and getting products filtered by category

hafizrahman:
### What this PR is about:

The PR adds support on the FluxC side to 
- fetch a list of products, filtered by category
- get a list of products from the database filtered by category

This PR is related to https://github.com/woocommerce/woocommerce-android/issues/4578 

### What the PR does:
- Add `CATEGORY` as one of the `ProductFilterOption` (later usable in Woo Mobile).
- Update the SQL utility to get product with filter, to include filtering with category.
- Update unit test related to getting product with filter, to include filtering with category.
- Add an example case for testing in the Example app (more below in the testing area).

Finally, this does not modify the REST client part as it's already flexible enough to accept the additional category parameter.

### Testing instruction:
 **Unit test:** 
- Run `testGetProductsByFilterOptions` in `WCProductStoreTest.kt` and make sure it passes.

**Example app:**
- In the Example app, go to Woo -> Products. 
- Tap "Select Site" and pick a site containing products with categories set on them.
- On the next screen, select "Fetch Products With Filters".
- Tap "Filter By A Random Category". This will show "Selected category:  [name] id:[id number]" on the log area if a category is selected properly.
- Tap "Apply Filter". If it finds some product, it will show "From category ID [id number]" and "Fetched X products".
- Go to the wp-admin of the test site selected on the second step, and open Product > Categories. Find the category name with the same name as on the Example app, and make sure the "Count" value shown there is the same as what's reported by the Example app.


ThomazFB (approved):
Nice work, @hafizrahman! I have no changes to request. LGTM :shipit:

PR #2096, opened 2021-08-12
Add support for test_mode parameter on payments/accounts endpoint

malinajirka:
Parent issue https://github.com/woocommerce/woocommerce-android/issues/4579

This PR adds support for "test_mode" flag on "payments/accounts" endpoint. The flag is "true" when the test mode on the site is enabled, "false" when it's disabled and "null" when it's unknown.

To test:
Test in WCAndroid PR (tbd)


nbradbury (approved):
Looks good! :shipit:

PR #2097, opened 2021-08-12
Add entities for storing Global Add-on Groups locally

wzieba:
### Description

This PR adds entities for storing Global Add-on Groups locally.

Here's how ERD diagram looks like:

Screenshot 2021-08-12 at 13 43 26

### How to test

It's not testable code change - rather skeleton/introduction to later PRs with the whole integration of database for Add-ons.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/AddonEntity.kt
```diff
@@ -0,0 +1,82 @@
...
+    @Embedded val addon: AddonEntity,
+    @Relation(
+            parentColumn = "addonLocalId",
+            entityColumn = "addonLocalId"
+    )
+    val options: List<AddonOptionEntity>
+)
+
+@Entity(
+        foreignKeys = [ForeignKey(
```
wzieba, line 20: Foreign keys here and below will keep our database clean - after removing `GlobalAddonGroupEntity`, all linked `AddonEntity` objects and its `AddonOptionEntity`s will be also removed.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/AddonEntity.kt
```diff
@@ -0,0 +1,82 @@
...
+    val position: Int?,
+    val restrictions: Boolean?,
+    val restrictionsType: RestrictionsType?,
+    val adjustPrice: String?,
+    val priceType: PriceType?,
+    val price: String?,
+    val min: Int?,
+    val max: Int?
+) {
+    enum class Type {
```
wzieba, line 46: Here I copied values from remote model. It'll require some not-so beautiful, boilerplate mapper but this way we're safe from API changes. I think it's worthy to keep separate models for API and local databases. WDYT @ThomazFB ?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/GlobalAddonGroupEntity.kt
```diff
@@ -0,0 +1,25 @@
...
+    )
+    val addons: List<AddonWithOptions>
+)
+
+@Entity
+data class GlobalAddonGroupEntity(
+    @PrimaryKey(autoGenerate = true) val globalGroupLocalId: Long = 0,
+    val remoteId: Long,
+    val name: String,
+    val restrictedCategoriesIds: List<Long>,
```
wzieba, line 23: Ideally, we shouldn't keep `List<Long>` in database but have a many-to-many relation with `WCProductCategoryModel`. But this model is kept in a separate, `WellSql` database so joining those two will be introduced later.

ThomazFB:
Hey @wzieba! Nice work with this PR! I tested your changes but I'm still about to review the code. I'll submit feedback this Friday yet, I hope this doesn't block your tasks, sorry for that 😓

wzieba:
Sure @ThomazFB , thanks for heads-up! No problem, it doesn't block me at all.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/AddonEntity.kt
```diff
@@ -0,0 +1,82 @@
...
+    val display: Display?,
+    val name: String?,
+    val titleFormat: TitleFormat?,
+    val descriptionEnabled: Boolean,
+    val description: String?,
+    val required: Boolean?,
+    val position: Int?,
+    val restrictions: Boolean?,
+    val restrictionsType: RestrictionsType?,
+    val adjustPrice: String?,
```
wzieba: @ThomazFB I've just realised that according to API, this value is a `Boolean`, not a `String`. I'll fix that in a minute!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/AddonEntity.kt
ThomazFB: Nice, I agree! When you mention the mapper do you mean the requirements to parse the enum types and the Booleans represented as Int? If that's the case, I guess this will probably make the parsing to the `ProductAddon` model at the Woo side easier since right now we're converting a lot of data from the remote model directly to the presentation model inside the `toAppModel()`. 

I guess this is one of the points you mentioned at Slack about bringing to FluxC some of the `ProductAddon` model behavior, right?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/AddonEntity.kt
ThomazFB: Nice catch! I decided to do a second check on the Woo side and found out that I've left the `adjustPrice` field as String too, I'll also fix that in the upcoming PR 😅

ThomazFB (approved):

ThomazFB:
@wzieba, nice work! I did some tests with playground code just parsing everything to see if the AddonEntity data class is fully reflecting the `WCProductAddonModel`, and with the exception of the IDs which we don't possess at the remote model, everything is looking good. LGTM :shipit:

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/AddonEntity.kt
ThomazFB: OK, I'm reviewing [this PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2098) now, and it kind of answers my questions already about the mapping 😆

PR #2099, opened 2021-08-17
Fetch addons in sample app

wzieba:
### Description 

This PR adds a test feature to `example` project for testing fetching of Global Add-ons Groups. In the beginning I thought we should be able to fetch and observe data from sample app, but browsing this data in small screen of sample app is quite difficult. I removed it and suggest to use Database Inspector.

It also moves `CacheGlobalAddonsGroups` logic directly into `AddonsDao` so we can perform insert operation in transaction. Without this we could encounter timing-based issues while inserting entities into db (e.g. foreign key validation).

### How to test

1. Run the `example` app
1. Go to `Woo` -> `Products` 
1. Select site
1. Press `Fetch and cache global addons groups`
1. In Android Studio: go to App Inspection.
1. Make sure there are at least 3 tables in `wc-android-database`: 
    - `AddonEntity`
    - `AddonOptionEntity`
    - `GlobalAddonGroupEntity`
1. Browse this data, **make sure you see correct one** (according to what you set in Woo site)
1. Click multiple times (basically spam) `Fetch and cache global addons groups` button
1. Make sure that app didn't crash and there's still the same number of every entity. Tip: `Live updates` is handy in this case.

### Screens from App Inspector

image


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCAddonsStore.kt
```diff
@@ -1,22 +1,35 @@
...
+    suspend fun fetchAllGlobalAddonsGroups(site: SiteModel): WooResult<Unit> {
         return coroutineEngine.withDefaultContext(AppLog.T.API, this, "fetchGlobalAddonsGroups") {
-            restClient.fetchGlobalAddOnGroups(site).asWooResult()
+            val response = restClient.fetchGlobalAddOnGroups(site)
+            when {
+                response.isError -> WooResult(response.error)
+                response.result != null -> {
+                    val globalAddonGroups = response.result
+                    dao.cacheGroups(globalAddonGroups, site.siteId)
+                    WooResult(Unit)
```
ThomazFB, line 29: I'm curious about this `WooResult(Unit)`. I understand that we're not returning to the example app the data fetched because we're not logging anything there, but wouldn't we want to place the result inside the `WooResult` because of the Woo app that will take advantage to receive the data right away after the fetching? 🤔

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
```diff
@@ -402,9 +402,14 @@ class WooProductsFragment : Fragment() {
...
-                        error?.let { prependToLog("${it.type}: ${it.message}") }
-                            prependToLog("Global addons: ${this.model}")
+                    addonsStore.fetchAllGlobalAddonsGroups(site).let { result ->
+                        prependToLog(
+                                if (result.error != null) {
+                                    "${result.error.type}: ${result.error.message}"
+                                } else {
+                                    "Global addons fetch successful"
+                                }
+                        )
```
ThomazFB: Do you think it would be more direct if we operate the error directly? Something like the previous code was, maybe like this:

```suggestion
                    addonsStore.fetchAllGlobalAddonsGroups(site).error
                            ?.let { prependToLog("${it.type}: ${it.message}") }
                            ?: prependToLog("Global addons fetch successful")
```

I have no strong opinions about this, just a nitpick.

ThomazFB (approved):
Great work here @wzieba, really nice to see the Woo plugin receiving the very first entities using Room! Code looks good, tests are green and the feature works flawlessly.

I left some suggestions and questions, but no big deal, I'm approving this PR beforehand!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCAddonsStore.kt
wzieba: I set `Unit` return type intentionally to have more CQRS structure here. Eventually, as we use local database as cache, so we won't be taking results directly from remote but rather from cache. So instead:
```
remote -> presentation
remote -> database
```
we'll have roughly 
```
remote -> database -> presentation
```

flow. This will save us some additional mappers and we'll have one source of (validated) true. If that's okay - I'd keep `Unit` as return type here. I'll go ahead and merge this PR (because of timeline) but I'm totally open to discuss it further (here or anywhere else)!

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
wzieba: Looks good, neater - thanks!

PR #2100, opened 2021-08-18
Update Product Addon remote and local models

wzieba:
Follow-up to https://github.com/woocommerce/woocommerce-android/issues/4405

### Description

Working longer with Add-on API and domain made me understand it more and I think we can improve our models for:
- storing less fields in database, which reduces its complexity
- improve nullability of remote object fields to not worry about cases that, according to API, don't happen

### To test

This is not a testable PR but please take a second look at api description to double-check me if I didn't make any mistake. I'll try to provide also a rich description in this PR of decision process behind changed.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
```diff
@@ -4,41 +4,38 @@ import com.google.gson.annotations.SerializedName
 
 data class WCProductAddonModel(
     @SerializedName("title_format")
-    val titleFormat: AddOnTitleFormat? = null,
-    @SerializedName("description_enable")
-    val descriptionEnabled: String? = null,
```
wzieba: `descriptionEnabled` seems to be redundant: `description` itself can be empty string. I don't feel any possible usage for `descriptionEnabled`

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
```diff
@@ -4,41 +4,38 @@ import com.google.gson.annotations.SerializedName
...
+    val titleFormat: AddOnTitleFormat,
     @SerializedName("restrictions_type")
     val restrictionsType: AddOnRestrictionsType? = null,
     @SerializedName("adjust_price")
-    val adjustPrice: String? = null,
+    val adjustPrice: Int,
```
wzieba, line 13: It's a `boolean` represented by an integer (0 or 1)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
```diff
@@ -4,41 +4,38 @@ import com.google.gson.annotations.SerializedName
...
     @SerializedName("price_type")
     val priceType: AddOnPriceType? = null,
 
-    val type: AddOnType? = null,
+    val type: AddOnType,
```
wzieba, line 17: Add-on always has to have a type - otherwise we don't know how to treat it.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
```diff
@@ -4,41 +4,38 @@ import com.google.gson.annotations.SerializedName
...
     val display: AddOnDisplay? = null,
-    val name: String? = null,
-    val description: String? = null,
-    val required: String? = null,
-    val position: String? = null,
-    val restrictions: String? = null,
+    val name: String,
+    val description: String,
+    val required: Int,
```
wzieba, line 21: It's also a `boolean` represented by integer

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
```diff
@@ -4,41 +4,38 @@ import com.google.gson.annotations.SerializedName
...
+    val position: Int,
     val price: String? = null,
-    val min: String? = null,
-    val max: String? = null,
+    val min: Long,
+    val max: Long,
```
wzieba, line 25: Those values have different meaning based on `type` of the Add-on but they are always numeric type.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
```diff
@@ -4,41 +4,38 @@ import com.google.gson.annotations.SerializedName
...
     val options: List<ProductAddonOption>? = null
 ) {
     enum class AddOnType {
-        @SerializedName("description_enable") MultipleChoice,
+        @SerializedName("multiple_choice") MultipleChoice,
         @SerializedName("checkbox") Checkbox,
         @SerializedName("custom_text") CustomText,
-        @SerializedName("custom_text_area") CustomTextArea,
+        @SerializedName("custom_textarea") CustomTextArea,
```
wzieba, line 32: Here were just typos

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
```diff
@@ -65,7 +62,7 @@ data class WCProductAddonModel(
 
     data class ProductAddonOption(
         @SerializedName("price_type")
-        val priceType: AddOnPriceType? = null,
+        val priceType: AddOnPriceType,
```
wzieba, line 67: `AddonOption` also always has to have type of the price adjustment.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
ThomazFB: I did some tests at `wp-admin` here and meanwhile I agree with your statement, I think this may cause confusion to the merchant. I went to the `wp-admin` and tried to uncheck the `Add description` box for one specific Product add-on, and the text description continues to be returned inside the metadata and be displayed in our app in Product Details. If I check back again the same box, the description comes back with the text it was before the unchecking. So our only way to follow this behavior will be probably following what `description enabled dictates.

What do you think about this?

ThomazFB (approved):
Changes looks good, @wzieba. Thanks for your attention to details and fixing the models! Besides the `descriptionEnabled` suggestion, I'm fine with everything else, so I'm approving before hand this PR and leave to you if think it should be merged or not.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/WCProductAddonModel.kt
wzieba: > and the text description continues to be returned inside the metadata and be displayed in our app in Product Details

True, great eye @ThomazFB !! That's surprising behaviour but it also means that we have to have this field. I'm moving back `descriptionEnabled` to the models, thanks!

PR #2101, opened 2021-08-18
Introduce domain models for Add-ons feature

wzieba:
Part of https://github.com/woocommerce/woocommerce-android/issues/4645

### Description

This PR introduces domain models for the "Add-ons world". This models are based on what I could understand from how the features works on the live site and what features are available in each of the add-on type.

I'm pretty content with those `sealed class` models as they help to:
- offer only valid states to presentation layer
- add much more context without looking into API or live site, things like:
    - `display` has only sense in case of `MultipleChoice` add-on
    - `restrictions` has only sense in case of `CustomText`
    - not every add-on has `options`
    - not every add-on has `price`
    - `min` and `max` fields differ in meaning in different add-ons types. They can be `characterLength` as well as `quantityRange`

and other. I believe such modelling will make our lives much easier when working on presentation as well as in the future, during creating an order with add-ons in the app.

### Flow

This PR introduces only models but I think it can be valuable to present idea where those models will be in use:

This way we'll have a single source of getting data from (cache/database) of valid data.

### How to test

This PR doesn't contain tests as mappers but it is valuable to test correctness of the model.

To do this, please:
- refer to API specification as well as live site
- create each type of the add-on and validate if all fields/features are covered by domain model

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/domain/Addon.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/domain/Addon.kt
@@ -1,0 +1,127 @@
+package org.wordpress.android.fluxc.domain
+
+import org.wordpress.android.fluxc.domain.Addon.HasAdjustablePrice.Price
+
+sealed class Addon {
+    abstract val name: String
+    abstract val titleFormat: TitleFormat
+    abstract val description: String
+    abstract val required: Boolean
+    abstract val position: Int
+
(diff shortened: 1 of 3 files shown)


ThomazFB (approved):
This is a great step forward, thanks for doing this, @wzieba! About field correctness, I think this is something we can go discovering what needs to be changed when we start mapping all this to the presentation layer, doing it now would be just predictions from my part. 

Awesome job! LGTM :shipit:

PR #2102, opened 2021-08-18
Store product-specific add-ons in local database

### LINKED ISSUE #4591: [Product Add-ons] [FluxC] Refactor single Product addons storage to match Global addons table structure
Currently, we're storing addons as a JSON string inside FluxC `WCProductModel` class and we serialize it to an object when the property gets called:
https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5c420b05cfff5b44e66d1b538ded7a11c5642188/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt#L113

Our objective here is that once #4410 is complete, we want to stop doing the JSON string strategy and store all Product addon data at the same table of Global addons, normalizing how every addon info is fetched at database level.

image

wzieba:
Fixes: https://github.com/woocommerce/woocommerce-android/issues/4591

### Description

This PR introduces logic for storing Product-specific add-ons in local database (in addition to already stored Global Add-ons).

### Disclaimer

Sorry I didn't fit to write unit tests for this behaviour. We can do this later or feel free @ThomazFB to update this PR. This behaviour can be tested though manually in next PR.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/AddonsDao.kt
```diff
@@ -30,28 +36,51 @@ internal abstract class AddonsDao {
...
+    abstract suspend fun deleteGlobalAddonsForSite(siteRemoteId: Long)
+
+    @Query("DELETE FROM AddonEntity WHERE productRemoteId = :productRemoteId AND siteRemoteId = :siteRemoteId")
+    abstract suspend fun deleteAddonsForSpecifiedProduct(productRemoteId: Long, siteRemoteId: Long)
 
     @Transaction
     open suspend fun cacheGroups(
         globalAddonGroups: List<AddOnGroupDto>,
-        remoteSiteId: Long
+        siteRemoteId: Long
```
wzieba, line 48: I've renamed `remoteSiteId` here and in other places as I think it's more readable. There's no thing like "remoteSite" and I sometimes was confused and thought about this as "id of remote site". It's rather "remote id" of "site" and I think `siteRemoteId` reflects it better. Sorry for this inconvenience in review!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/mappers/AddonEntityMapper.kt
```diff
@@ -7,10 +7,23 @@ import org.wordpress.android.fluxc.model.addons.WCProductAddonModel.AddOnRestric
 import org.wordpress.android.fluxc.model.addons.WCProductAddonModel.AddOnTitleFormat
 import org.wordpress.android.fluxc.model.addons.WCProductAddonModel.AddOnType
 import org.wordpress.android.fluxc.persistence.entity.AddonEntity
+import java.lang.Exception
+
+data class ProductBasedIdentification(
+    val productRemoteId: Long,
+    val siteRemoteId: Long
+)
```
wzieba, line 14: As we don't store `Product` in Room database, we have to find some identification other than primary key. This tuple is minimal set of information needed for identify a Product.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/mappers/AddonEntityMapper.kt
```diff
@@ -8,9 +8,23 @@ import org.wordpress.android.fluxc.model.addons.WCProductAddonModel.AddOnTitleFo
...
+
+internal fun WCProductAddonModel.toAddonEntity(
+    globalGroupLocalId: Long? = null,
+    productBasedIdentification: ProductBasedIdentification? = null
+): AddonEntity {
+    if (globalGroupLocalId == null && productBasedIdentification == null) {
+        throw IllegalStateException("Addon has to be identified with a Group or a Product")
```
wzieba, line 21: I can't see a way for this to happen in user-based scenario. WDYT about keeping this for now, especially for test phase as this will directly tell us what went wrong?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/AddonsDao.kt
```diff
@@ -9,17 +9,23 @@ import org.wordpress.android.fluxc.model.addons.WCProductAddonModel
...
+import org.wordpress.android.fluxc.persistence.entity.AddonWithOptions
 import org.wordpress.android.fluxc.persistence.entity.GlobalAddonGroupEntity
 import org.wordpress.android.fluxc.persistence.entity.GlobalAddonGroupWithAddons
+import org.wordpress.android.fluxc.persistence.mappers.ProductBasedIdentification
 import org.wordpress.android.fluxc.persistence.mappers.toAddonEntity
 import org.wordpress.android.fluxc.persistence.mappers.toAddonGroupEntity
 import org.wordpress.android.fluxc.persistence.mappers.toAddonOptionEntity
 
 @Dao
-internal abstract class AddonsDao {
```
wzieba, line 19: I had to remove `internal` modifier as we start to use it in `WCProductStore` which has tests in different module and its constructor can't be `internal`.

ThomazFB (approved):
Changes looks good so I don't need to adjust anything here, but it will require adding the `WCDatabaseModule` to the Woo Module mapping in order to successfully map our Addons DAO, so I'm placing this PR to be followed on Woo: https://github.com/woocommerce/woocommerce-android/pull/4674

Nevertheless, LGTM :shipit:

PR #2103, opened 2021-08-18
Combine Global and Product-specific Add-ons

### LINKED ISSUE #4647: [Product Add-ons] [FluxC] Offer combined Global and single Product Add-ons for a product
Having two types of add-ons: global and product-specific, we have to offer to presentation layer all Add-ons needed to be shown. After #4591 it'll be even easier as both type of add-ons will be kept in the same data source.

wzieba:
Fixes: https://github.com/woocommerce/woocommerce-android/issues/4647

### Description

This PR offers a logic for receiving combined Global and Product-specific Add-ons.

### Disclaimer

Like in #2102 I unfortunately didn't make it to write tests for added behaviour. I think it can be very valuable to do this but I think it's better this way to unblock further progress. 

### Demo

On the demo you can see that after pressing "start observing db", the view is updated in case of any change after fetching a product or global add-ons.

❗️I didn't test scenarios with global-addons specific to a product category. 

https://user-images.githubusercontent.com/5845095/129912967-d959c97f-1803-4a2a-8b12-f02b807f8f53.mov

### To test

1. Please go to example app (you might need to reinstall it due to changes in database schema in previous PR)
2. Go to `Woo` -> `Products`
3. Select site
4. Click on `Test Add-ons`
5. Enter remote product it you want to observe add ons on
6. Click on `Start observing DB`
7. **Assert there's list of currently stored add-ons there**
8. Add/remove/update specific-product add-ons on site. Click on `Fetch product` and assert that list has been updated.
9. Add/remove/update global add-ons on site. Click on `Fetch global addons` and assert that list has been updated.
10. Please also test categories restrictions for global add-ons (declared on site).

## Note about add-on option

I've noticed that to every Add-on, API adds  an empty option (even if its add-on without possibility to specify options). This problem will be resolved by domain mapping (#2101) but for now every Add-on will have at least one option.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
@@ -10,7 +10,6 @@ import dagger.android.support.AndroidSupportInjection
 import kotlinx.android.synthetic.main.fragment_woo_products.*
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.launch
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
 import org.wordpress.android.fluxc.Dispatcher
@@ -38,7 +37,6 @@ import org.wordpress.android.fluxc.generated.WCProductActionBuilder
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.WCProductCategoryModel
 import org.wordpress.android.fluxc.model.WCProductImageModel
-import org.wordpress.android.fluxc.model.addons.WCProductAddonModel
 import org.wordpress.android.fluxc.store.MediaStore
 import org.wordpress.android.fluxc.store.WCAddonsStore
 import org.wordpress.android.fluxc.store.WCProductStore
@@ -379,33 +377,9 @@ class WooProductsFragment : Fragment() {
             }
         }
 
-        fetch_product_addons.setOnClickListener {
+        test_add_ons.setOnClickListener {
             selectedSite?.let { site ->
-                showSingleLineDialog(
-                        activity,
-                        "Enter the remoteProductId of product to fetch the addons:"
-                ) { editText ->
-                    pendingFetchSingleProductRemoteId = editText.text.toString().toLongOrNull()
-                    pendingFetchSingleProductRemoteId?.let { id ->
-                        prependToLog("Submitting request to fetch product by remoteProductID $id")
-                        coroutineScope.launch {
-                            wcProductStore.fetchProductListSynced(site, listOf(id))
-                                    .takeUnless { it.isNullOrEmpty() }
-                                    ?.first()?.addons
-                                    .logAddons()
-                        }
-                    } ?: prependToLog("No valid remoteOrderId defined...doing nothing")
-                }
-            }
-        }
-
-        fetch_global_addons_groups.setOnClickListener {
-            selectedSite?.let { site ->
-                coroutineScope.launch {
-                    addonsStore.fetchAllGlobalAddonsGroups(site).error
-                            ?.let { prependToLog("${it.type}: ${it.message}") }
-                            ?: prependToLog("Global addons fetch successful")
-                }
+                WooAddonsTestFragment.show(childFragmentManager, site.siteId)
             }
         }
 
@@ -697,13 +671,5 @@ class WooProductsFragment : Fragment() {
                 else -> { }
             }
         }
-    }
-
-    private fun Array<WCProductAddonModel>?.logAddons() {
-        this?.forEachIndexed { index, addon ->
-            prependToLog(addon.description?.let { "description: $it" }.orEmpty())
-            prependToLog(addon.name?.let { "name: $it" }.orEmpty())
-            prependToLog("========== Product Add-on #$index =========")
-        } ?: prependToLog("No addons found for this product ID")
     }
 }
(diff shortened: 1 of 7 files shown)


example/src/main/res/layout/fragment_woo_addons.xml
```diff
@@ -0,0 +1,54 @@
...
+        android:id="@+id/addons_fetch_global"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="Fetch global addons" />
+
+    <TextView
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:padding="8dp"
+        android:text="Only brief informations of addons are presented here. For detailed list, please use Database Inspector." />
```
wzieba, line 46: I'm not sure if this is needed, I added it to give a hint for anyone in the future working on this feature but we can remove it as well.

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooProductsFragment.kt
```diff
@@ -698,12 +672,4 @@ class WooProductsFragment : Fragment() {
...
         }
     }
-
-    private fun Array<WCProductAddonModel>?.logAddons() {
-        this?.forEachIndexed { index, addon ->
-            prependToLog(addon.description?.let { "description: $it" }.orEmpty())
-            prependToLog(addon.name?.let { "name: $it" }.orEmpty())
-            prependToLog("========== Product Add-on #$index =========")
-        } ?: prependToLog("No addons found for this product ID")
-    }
```
wzieba, line 708: I've moved presenting Product add-ons to a single screen with Global-addons - I hope this way it can be also readable. Having those two in the FluxC log screen was a little messy.

shiki:
10 points for making a video! 👏

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCAddonsStore.kt
```diff
@@ -32,4 +38,32 @@ class WCAddonsStore @Inject internal constructor(
...
+
+    private fun GlobalAddonGroupEntity.appliesToEveryProduct(): Boolean {
+        return this.restrictedCategoriesIds.isEmpty()
+    }
+
+    private fun GlobalAddonGroupEntity.appliesToProduct(product: WCProductModel): Boolean {
+        val productCategoriesIds = product.getCategoryList().map { it.id }
+
+        return restrictedCategoriesIds.intersect(productCategoriesIds).isNotEmpty()
+    }
```
ThomazFB, line 68: Hey @wzieba, I have some concerns about locking our only way to get Global add-ons using Flow and overall data streams to acquire our needed info for this scenario here. 

From my perspective, Global Add-ons data is something that needs to be actively requested inside our app to get a data refresh, we don't have an always-on connection with the site, and our Add-ons views inside Order Details and Product Details doesn't have any refresh action, they're currently static when we join all necessary data since we expect that the update comes from the Product Details and Order Details.

Also, we should consider the nature of this data itself, Global Add-ons doesn't change frequently, and it's something we're joining with Order data, which currently aren't supported through Flow streams, we would end up having a data acquisition where half of it is actively requested, and the other half acquired through reactive stream events. So I think that considering our current scenario, this streaming event will trigger just once, so it feels like we would want a basic I/O return to get the data.

I can adapt our current `OrderedAddonsViewModel` to act on a hybrid data merging using active requests and reactive events, but that may overcomplicate even further what's currently a workaround using string matching to render the UI data. I could also just flatten the result of the Flow data to transform it into an I/O response instead of a stream event, but that would just kill the reason for using Flows 😅. WDYT about all this?

ThomazFB: I decided to explore more options on how we could handle this scenario without affecting too much our current structure (considering how Room is nowadays adapted to work with Flow) and trying to find a solution for the points I described above, and I think that maybe using the `firstOrNull()` method from the Flow standard functions could be a good solution instead of using `collect`, so instead of observing something that will trigger a stream event just once, we catch the data once and finish the Flow connection. With that said, the `AddonsRepository` would have a method doing something like this:

```
    private suspend fun List<Attribute>.joinWithAddonsFrom(productID: Long) =
        productStore.getProductByRemoteId(selectedSite.get(), productID)
            ?.let { addonsStore.observeAddonsForProduct(selectedSite.get().siteId, it) }
            ?.firstOrNull()
            ?.map { it.toAppModel() }
            ?.let { addons -> Pair(addons, this) }
```

This way we don't affect structure from both parts, and still manage to properly use Flow combined with Room. WDYT?

ThomazFB (approved):
Hey @wzieba, after exploring a lot on [WCAndroid side](https://github.com/woocommerce/woocommerce-android/pull/4680) the changes introduced here, I think everything is working great and we can discuss if we want to change anything related to how to map everything later. Thank you so much for these changes here! LGTM :shipit:

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCAddonsStore.kt
wzieba: hi @ThomazFB , thank you for those comments! I'm glad you found and suggested a compromise with reactive and non-reactive (or rather: non `Flow`-based?) words. I think it's good to keep our database (and other data sources) flow-based, the same as WP does. For the `firstOrNull` vs `collect` and other approaches, I'll comment on the Woo PR https://github.com/woocommerce/woocommerce-android/pull/4680/.

PR #2105, opened 2021-08-23
This commit adds order_key column into WCOrderModel table and maps order_key into the order model.

AnirudhBhat:
parent issue: https://github.com/woocommerce/woocommerce-android/issues/4030 (Add order_key to PaymentIntent's metadata)
### Description
This PR adds the order_key column into the WCOrderModel table and maps order_key into the order model.

### Testing instructions
Since this PR adds a new column into the WCOrder model table. Here are some of the things we need to test to see if adding a new column didn't break anything.

1. Install the app on top of an existing app
2. See that the app doesn't crash on startup
3. You can use Flipper for making sure there is a column called ORDER_KEY created in WCOrderModel
4. Run the unit test and ensure it passes

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCOrdersTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCOrdersTest.kt
@@ -196,6 +196,24 @@ class MockedStack_WCOrdersTest : MockedStack_Base() {
     }
 
     @Test
+    fun testFetchSingleOrderOrderKeySuccess() {
+        val remoteOrderId = 88L
+        val orderKey = "wc_order_5a77766b88986"
+        interceptor.respondWith("wc-fetch-order-response-success.json")
+        orderRestClient.fetchSingleOrder(siteModel, remoteOrderId)
+
+        countDownLatch = CountDownLatch(1)
+        assertTrue(countDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS.toLong(), TimeUnit.MILLISECONDS))
+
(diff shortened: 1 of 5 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCOrdersTest.kt
```diff
@@ -195,6 +195,24 @@ class MockedStack_WCOrdersTest : MockedStack_Base() {
...
 
+    @Test
+    fun testFetchSingleOrderOrderKeySuccess() {
+        val remoteOrderId = 88L
+        val orderKey = "wc_order_5a77766b88986"
+        interceptor.respondWith("wc-fetch-order-response-success.json")
+        orderRestClient.fetchSingleOrder(siteModel, remoteOrderId)
+
+        countDownLatch = CountDownLatch(1)
+        assertTrue(countDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS.toLong(), TimeUnit.MILLISECONDS))
```
kidinov: nit: TimeUnit is redundant here. Maybe to remove the import? 

```
import java.util.concurrent.TimeUnit.MILLISECONDS
```

kidinov (approved):
Great job! Let's merge it after we check that the integration works with the main app!

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCOrdersTest.kt
AnirudhBhat: Removed redundant TimeUnit since we already had import added.

PR #2107, opened 2021-09-02
Add mappers for domain layer

wzieba:
### Description

This PR adds mappers for remote -> domain <-> database layers communication. Those mappers are not a trivial ones due to nature of Product Add-ons - there are several different types and each of them differs.

I've also renamed some models like `WCProductAddonModel` to `RemoteAddonDto` for improving redability - it's easy to get lost with 3 models for the Add-ons here and there.

### Diagram 

Just a note that database and related entities should be placed in a separate module. I plan to work on that later, accordingly to priorities. 

### Tests

The best way to test mappers will be to write integration tests of the whole flow. I plan to do this (in a separate PR) before merging this PR but before doing that, I'd love to have initial review from you @ThomazFB if you agree with changes made in this PR. **I'll keep this as a draft but it's ready for review.** Thank you in advance!

### Woo PR

For the same reason as mentioned above, I didn't yet prepare Woo PR. I plan to do this though, before merging this PR of course.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/domain/Addon.kt
```diff
@@ -1,16 +1,23 @@
 package org.wordpress.android.fluxc.domain
 
 import org.wordpress.android.fluxc.domain.Addon.HasAdjustablePrice.Price
+import org.wordpress.android.fluxc.domain.Addon.HasOptions.Option
 
 sealed class Addon {
     abstract val name: String
     abstract val titleFormat: TitleFormat
-    abstract val description: String
+    abstract val description: String?
```
wzieba, line 9: Following up to one of our conversations: domain models (and database too) don't have to have `descriptionEnabled` field as we can handle it during mapping from remote model.

See https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2107/files#diff-1f56ad35bc6f9037762b90cbf7e5009dbf8ae81cfd0c3118b0ae770173807388R96-R102

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/RemoteAddonDto.kt
```diff
@@ -2,30 +2,30 @@ package org.wordpress.android.fluxc.model.addons
 
 import com.google.gson.annotations.SerializedName
 
-data class WCProductAddonModel(
+data class RemoteAddonDto(
```
wzieba, line 5: I've renamed this model and its enums to make it easier to ~easily~ understand what's the purpose of it.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/mappers/FromDatabaseAddonsMapper.kt
```diff
@@ -0,0 +1,163 @@
...
+import org.wordpress.android.fluxc.persistence.entity.AddonEntity.LocalType.CustomPrice
+import org.wordpress.android.fluxc.persistence.entity.AddonEntity.LocalType.CustomText
+import org.wordpress.android.fluxc.persistence.entity.AddonEntity.LocalType.CustomTextArea
+import org.wordpress.android.fluxc.persistence.entity.AddonEntity.LocalType.FileUpload
+import org.wordpress.android.fluxc.persistence.entity.AddonEntity.LocalType.Heading
+import org.wordpress.android.fluxc.persistence.entity.AddonEntity.LocalType.InputMultiplier
+import org.wordpress.android.fluxc.persistence.entity.AddonEntity.LocalType.MultipleChoice
+import org.wordpress.android.fluxc.persistence.entity.AddonWithOptions
+
+object FromDatabaseAddonsMapper {
```
wzieba, line 30: I've divided `*DatabaseAddonsMapper` to separate objects as those two mappers are not-so-small.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/addons/mappers/RemoteAddonMapper.kt
```diff
@@ -0,0 +1,166 @@
...
+                RemoteRestrictionsType.AnyText -> Addon.CustomText.Restrictions.AnyText
+                RemoteRestrictionsType.OnlyLetters -> Addon.CustomText.Restrictions.OnlyLetters
+                RemoteRestrictionsType.OnlyNumbers -> Addon.CustomText.Restrictions.OnlyNumbers
+                RemoteRestrictionsType.OnlyLettersNumbers -> Addon.CustomText.Restrictions.OnlyLettersNumbers
+                RemoteRestrictionsType.Email -> Addon.CustomText.Restrictions.Email
+            }
+        } else {
+            throw Exception()
+        }
+    }
```
ThomazFB: Hey Wojtek! I want to discuss the use of Exceptions as a map failure signal and get some thoughts from you. I see multiples `throw Exception()` through the mapping code here, and none of them are handled inside the FluxC Library. 

I usually try to think that if the language supports nullability handling, it's best to deal with null values than risk breaking stuff to the user, especially considering that this is code from a library, the Woo app will request something and may receive an Exception in return for an Illegal state that was caused by the site, not the app.

But I also understand that you want to avoid null values inside the domain. Still, between risking receiving an Unhandled exception and having a nullable field, I think kotlin would deal way better with the nullability handling option. We could maybe use an Enum value that represents an `Unknown` option for the scenario where we can't tell what's happening, but I don't think it's a good idea to serve Exceptions to our app because of a plugin data inconsistency. WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/mappers/ToDatabaseAddonsMapper.kt
```diff
@@ -0,0 +1,136 @@
...
+            val price = this.price
+            if (price is Adjusted) {
+                price.value
+            } else {
+                null
+            }
+        } else {
+            null
+        }
+    }
```
ThomazFB: np: it's possible to greatly improve this function readability by taking advantage of kotlin null returning operations
```suggestion
    private fun Addon.mapPrice(): String? =
            (this as? HasAdjustablePrice)
                    ?.let { it.price as? Adjusted }
                    ?.value
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/mappers/ToDatabaseAddonsMapper.kt
```diff
@@ -0,0 +1,136 @@
...
+                    PriceType.QuantityBased -> AddonEntity.LocalPriceType.QuantityBased
+                    PriceType.PercentageBased -> AddonEntity.LocalPriceType.PercentageBased
+                }
+            } else {
+                null
+            }
+        } else {
+            null
+        }
+    }
```
ThomazFB: I would say that the same idea of [this comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2107/files#r701478290) could improve this part here too. WDYT?

ThomazFB:
Great work here, @wzieba, thanks for tackling this! It will be truly beneficial for us at handling Add-ons types and makes it easier to understand what to expect from the plugin API. I also agree with your thoughts about tackling the three Woo issues you mentioned before after this PR, it will make it much easier for us to fix the problem, and avoid rework.

For now, I reviewed only the code itself, didn't tested anything yet as the PR is still in progress. I've left some comments around the code with some toughts on error handling and code refactoring. Let me know what you think!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/mappers/ToDatabaseAddonsMapper.kt
wzieba: love it, much more readable!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/addons/mappers/RemoteAddonMapper.kt
wzieba: Hi Thomaz, completely agree that exposing exceptions (especially those related to mapping!) is not a good idea, great catch. What I wanted to achieve is to "drop" invalid add-ons from our flow: e.g. if there's an Add-on of type `MultipleChoice` we know that it has to have `display` type.

I've added exceptions handling in 76ae2bf3120415a10a5a6298bdbe4d022ead31ef - does this solution work for you?

ThomazFB (approved):
Tested in integration with PR https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2108, working perfectly. LGTM :shipit:

PR #2109, opened 2021-09-06
Fixes crash appearing when the site software version is null

antonis:
`WordPress-Android` PR: https://github.com/wordpress-mobile/WordPress-Android/pull/15286

### Description
This PR fixes a [crash](https://github.com/wordpress-mobile/WordPress-Android/issues/15278) appearing when the site software version is `null`

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/EditorThemeStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/EditorThemeStore.kt
@@ -20,6 +20,7 @@ import org.wordpress.android.fluxc.store.ReactNativeFetchResponse.Error
 import org.wordpress.android.fluxc.store.ReactNativeFetchResponse.Success
 import org.wordpress.android.fluxc.tools.CoroutineEngine
 import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.VersionUtils
 import org.wordpress.android.util.helpers.Version
 import javax.inject.Inject
 import javax.inject.Singleton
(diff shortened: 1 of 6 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/EditorThemeStore.kt
```diff
@@ -192,13 +192,18 @@ class EditorThemeStore
...
-    private fun SiteModel.hasRequiredWordPressVersion(requiredVersion: String) = try {
-        val version = if (softwareVersion.contains("-")) {
-            // strip semantic versioning information (alpha, beta etc)
-            softwareVersion.substringBefore("-")
-        } else softwareVersion
-        Version(version) >= Version(requiredVersion)
-    } catch (e: IllegalArgumentException) {
-        false // if version parsing fails return false
+    private fun hasRequiredWordPressVersion(softwareVersion: String?, requiredVersion: String): Boolean {
+        if (softwareVersion == null) {
```
antonis: Note that this function duplicates logic introduced about the same time in [VersionUtils](https://github.com/wordpress-mobile/WordPress-Android/blob/feeacbc8ac15a903608ee0fc7671e03d2e10a014/WordPress/src/main/java/org/wordpress/android/util/VersionUtils.kt#L17). Reusing this implementation is not possible in `WordPress-FluxC-Android` or would have required more wide code changes.

antonis: Moved VersionUtils to `WordPress-Utils-Android` with https://github.com/wordpress-mobile/WordPress-Android/pull/15287

develric (approved):
LGTM, checked using the companion WPAndroid PR, looking for the standard behaviour and forcing software version to null using debugger. Everything worked well 👍

PR #2111, opened 2021-09-09
Woo Shipping Labels: pass is_letter value when purchasing shipping labels

hichamboushaba:
This change is only to align with the requirements of the API, even though the call succeeds without it.


hafizrahman (approved):
This LGTM. Thanks for working on it!

PR #2112, opened 2021-09-09
Prevents null styles/features from being converted to string

antonis:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/3936

`WordPress-Android` PR: https://github.com/wordpress-mobile/WordPress-Android/pull/15312

### Description
Prevents null styles and features from being converted to string


geriux (approved):
LGTM! I tested this by using the build from https://github.com/wordpress-mobile/WordPress-Android/pull/15312 and also by checking locally that this data is not being sent to the editor when it's not available. 🚀

PR #2116, opened 2021-09-15
Database and model setup for SSR in WC Android

hafizrahman:
## PR Description

I'm working on allowing merchants to share System Status Report (SSR) directly via app. The main issue thread is on https://github.com/woocommerce/woocommerce-android/issues/4781 .

This PR is the first part of that. It adds:

1. A new entity for it in the WooCommerce Room database
2. A DAO and its unit test
2. A model

## To test

Run `SSRDaoTest` and ensure it is successful.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSSRModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSSRModel.kt
@@ -1,8 +1,7 @@
 package org.wordpress.android.fluxc.model
 
 data class WCSSRModel(
-    val id: Int,
-    val localSiteId: Int,
+    val remoteSiteId: Int,
     val environment: String? = null,
     val database: String? = null,
     val activePlugins: String? = null,
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSSRModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSSRModel.kt
@@ -1,6 +1,7 @@
 package org.wordpress.android.fluxc.model
 
 data class WCSSRModel(
+    val id: Int,
     val localSiteId: Int,
     val environment: String? = null,
     val database: String? = null,
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSSRModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSSRModel.kt
@@ -1,7 +1,7 @@
 package org.wordpress.android.fluxc.model
 
 data class WCSSRModel(
-    val remoteSiteId: Int,
+    val remoteSiteId: Long,
     val environment: String? = null,
     val database: String? = null,
     val activePlugins: String? = null,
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSSRModel.kt
```diff
@@ -0,0 +1,13 @@
+package org.wordpress.android.fluxc.model
+
+data class WCSSRModel(
```
hafizrahman, line 3: The API call[ returns a JSON object with seven keys](https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#list-all-system-status-items), with each column in the table representing a key. Here I just wanted to be super defensive and don't want to assume that any of those key/item will exist at all (e.g: maybe it will be altered/hidden by a security plugin), so I decided to let everything possibly be null.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSSRModel.kt
```diff
@@ -0,0 +1,13 @@
...
+    val id: Int,
+    val localSiteId: Int,
```
wzieba: ❓ WDYT about making this `siteRemoteId`? I think that "remote id of the site" is globally a primary key for a site and might be easier to manage (query/observe) data later in the app.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/SSREntity.kt
```diff
@@ -0,0 +1,17 @@
+package org.wordpress.android.fluxc.persistence.entity
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity
+data class SSREntity(
+    @PrimaryKey(autoGenerate = true) val id: Int = 0,
+    val localSiteId: Int,
```
wzieba: ❓ I wonder - do you think we need more than one `SSREntity` for a given site?

Because if we don't need it, I'd suggest to make `siteRemoteId` (changed from `localSiteId`, see comment above) a `@PrimaryKey` and remove the "casual" id.

My thought is that making `siteRemoteId` as a PrimaryKey will save you an effort with deleting and updating the values in db. Every time with a new `SSREntity` for a given site you'll replace the old value (as you set `OnConflictStrategy.REPLACE`)

But it won't work of course if we need many SSR reports for the same site.

wzieba (approved):
Look good! I left minor questions and concernes.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCSSRModel.kt
hafizrahman: This makes sense, thanks! Removing `id` was also something I've been thinking about as I worked on the REST API part.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/SSREntity.kt
hafizrahman: I agree, I think a site will just have one `SSREntity`.

> My thought is that making siteRemoteId as a PrimaryKey will save you an effort with deleting and updating the values in db. Every time with a new SSREntity for a given site you'll replace the old value (as you set OnConflictStrategy.REPLACE)

Can you clarify what you meant with this? Initially I thought to update the value from the same site, I could just use the original `@Insert` function. Do you mean with `remoteSiteId` as primary key, `OnConflictStrategy.REPLACE` is not needed?

hafizrahman: Oh I think I see what you mean. With the current setup with an autogenerated id, a second insert on the same site will create a new row instead of replacing.

PR #2118, opened 2021-09-15
FetchSingleOrder - replace event bus with suspendable

malinajirka:
This PR refactors `fetchSingleOrder` into a suspendable function - removes usage of the event bus. 

2. WCAndroid app uses stateful repositories which need to be registered to the event bus - this leads to memory leaks when the repository doesn't unregister from the event bus. Also it causes side-effects when multiple instances of a repository are created - eg. duplicate tracking events.

To Test:
Test "fetchSingleOrder" action in the example app or test in WCAndroid (PR https://github.com/woocommerce/woocommerce-android/pull/4791)

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
@@ -141,7 +141,7 @@ class WooOrdersFragment : Fragment(), WCAddOrderShipmentTrackingDialog.Listener 
                     coroutineScope.launch {
                         enteredRemoteId?.let { id ->
                             prependToLog("Submitting request to fetch order by remoteOrderID: $enteredRemoteId")
-                            wcOrderStore.fetchSingleOrder(site, id).takeUnless {   it.isError }?.let {
+                            wcOrderStore.fetchSingleOrder(site, id).takeUnless { it.isError }?.let {
                                 prependToLog("Single order fetched successfully!")
                             } ?: prependToLog("WARNING: Fetched order not found in the local database!")
                         } ?: prependToLog("No valid remoteOrderId defined...doing nothing")
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
@@ -9,6 +9,9 @@ import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import dagger.android.support.AndroidSupportInjection
 import kotlinx.android.synthetic.main.fragment_woo_orders.*
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
 import org.wordpress.android.fluxc.Dispatcher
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -269,37 +274,44 @@ class OrderRestClient @Inject constructor(
...
-                            orderError,
-                            WCOrderModel().apply { this.remoteOrderId = remoteOrderId },
-                            site
-                    )
-                    dispatcher.dispatch(WCOrderActionBuilder.newFetchedSingleOrderAction(payload))
-                },
-                { request: WPComGsonRequest<*> -> add(request) })
-        add(request)
+                    RemoteOrderPayload(newModel, site)
+                } ?: RemoteOrderPayload(
```
malinajirka, line 300: 👉 This is a change from the previous behavior -> it was not returning anything when the response from the server was successful but didn't contain any data. The new behavior returns an error response in this scenario.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -678,20 +685,6 @@ class WCOrderStore @Inject constructor(
...
-
-        if (payload.isError) {
-            onOrderChanged = OnOrderChanged(0).also { it.error = payload.error }
-        } else {
-            val rowsAffected = OrderSqlUtils.insertOrUpdateOrder(payload.order)
-            onOrderChanged = OnOrderChanged(rowsAffected)
-        }
-
-        onOrderChanged.causeOfChange = WCOrderAction.FETCH_SINGLE_ORDER
-        emitChange(onOrderChanged)
```
malinajirka, line 692: 👉  WCAndroid is not using this event anywhere else but OrderDetailRepository -> it is imho safe to remove the event completely.

wzieba (approved):
This looks good to me, nice improvement with 

> not returning anything when the response from the server was successful but didn't contain any data

not mentioning the overall improvement after we'll migrate from event bus.

Thank you, tomorrow I'll be checking the Woo PR. Leaving not merged so @hichamboushaba can also take a look.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
```diff
@@ -135,13 +137,22 @@ class WooOrdersFragment : Fragment(), WCAddOrderShipmentTrackingDialog.Listener
...
+                            wcOrderStore.fetchSingleOrder(site, id)
+                            wcOrderStore.getOrderByIdentifier(
+                                    OrderIdentifier(
+                                            WCOrderModel().apply {
+                                                remoteOrderId = enteredRemoteId
+                                                localSiteId = site.id
+                                            })
+                            )?.let {
+                                prependToLog("Single order fetched successfully!")
+                            } ?: prependToLog("WARNING: Fetched order not found in the local database!")
```
hichamboushaba: Just a question, why not use the `fetchSingleOrder`'s result to know if the call succeeded or not instead of depending on `getOrderByIdentifier`, since the order may be cached before, and we'll get a success even if the call failed.

hichamboushaba (approved):
LGTM, good job @malinajirka.

I didn't test the app, I'll test it when reviewing the WCAndroid's PR.

I left a small suggestion, but it doesn't have to hold this PR.

And a small thing that we need to watch out from when reviewing the WCAndroid PR, we need to ensure that the WCAndroid app doesn't depend on the event `OnOrderChanged` anywhere outside the Repository, since it won't be dispatched now(one of the reasons why I hate EventBus, everyone can listen to everything)

malinajirka:
Thank you both for the reviews ;)!

> And a small thing that we need to watch out from when reviewing the WCAndroid PR, we need to ensure that the WCAndroid app doesn't depend on the event OnOrderChanged anywhere outside the Repository, since it won't be dispatched now(one of the reasons why I hate EventBus, everyone can listen to everything)

Yes, I checked that and the cause - FETCH_SINGLE_ORDER or "case without any cause" is not used anywhere so it should be safe to remove.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
malinajirka: I mechanically transformed the previous code into the suspendable approach - I didn't want to change the behavior unless necessary to limit the chance we'll miss something when converting all the methods. Having said that, this is just an example app and it's a simple change - updated in d1d6be254c405811bbfc7f9a86c20fc8f03e006f

malinajirka:
I'll be afk next week - please feel free to fix issues found during the review but also feel free to leave it up to me when I get back. Thanks 🙇

PR #2119, opened 2021-09-16
Store functionalities for SSR

hafizrahman:
## PR Description

I'm working on allowing merchants to share System Status Report (SSR) directly via app. The main issue thread is on https://github.com/woocommerce/woocommerce-android/issues/4781 .

This PR is the second part of that. It adds:

1. A fetch SSR function in `WooCommerceStore`
2. REST API call in `WooSystemRestClient`
3. Unit tests for above
4. A new section in Example app to test it. 

## To test
**Example app:**
1. Assuming you're logged in, go to Woo -> Help & Support. 
2. Select a WooCommerce site, then tap "Fetch System Status Report". 
3. Observe the log and ensure it outputs fetched data.

**Unit tests**
Run `WooCommerceStoreTest` and ensure everything is successful.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
@@ -16,7 +16,6 @@ import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
-import org.robolectric.RuntimeEnvironment
 import org.robolectric.annotation.Config
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.SingleStoreWellSqlConfigForTests
@@ -48,11 +47,11 @@ class WooCommerceStoreTest {
         const val TEST_SITE_REMOTE_ID = 1337L
     }
 
-    private val appContext = RuntimeEnvironment.application.applicationContext
+    private val appContext = ApplicationProvider.getApplicationContext<Application>()
     private val restClient = mock<WooSystemRestClient>()
 
     private val roomDB = Room.inMemoryDatabaseBuilder(
-            ApplicationProvider.getApplicationContext<Application>(),
+            appContext,
             WCAndroidDatabase::class.java
     )
             .allowMainThreadQueries()
@@ -184,11 +183,13 @@ class WooCommerceStoreTest {
     }
 
     @Test
-    fun `when fetching ssr succeeds, then success returned`() = test {
-        val result = fetchSSR(isError = false)
+    fun `when fetching ssr succeeds, then success returned`() {
+        runBlocking {
+            val result = fetchSSR(isError = false)
 
-        Assertions.assertThat(result.isError).isFalse
-        Assertions.assertThat(result.model).isNotNull
+            Assertions.assertThat(result.isError).isFalse
+            Assertions.assertThat(result.model).isNotNull
+        }
     }
 
     @Test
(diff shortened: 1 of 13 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
```diff
@@ -25,7 +25,7 @@ import org.wordpress.android.fluxc.persistence.entity.SSREntity
 @TypeConverters(value = [LongListConverter::class])
 abstract class WCAndroidDatabase : RoomDatabase() {
     internal abstract fun addonsDao(): AddonsDao
-    internal abstract fun ssrDao(): SSRDao
+    abstract fun ssrDao(): SSRDao
```
hafizrahman, line 28: ❓ Had to remove `internal` here so that `roomDB.ssrDao()` in `WooCommerceStoreTest` can work (otherwise the function couldn't be accessed).

I wonder if there's a better way to do this (or whether this is something not to worry about)?

wzieba: Unfortunately, we can't have `SSRDao` not `internal` because `WooCommerceStoreTest` lies in `example` module, not in the `fluxc` module.

I don't know the reasons for this and I think it's rather unwanted behaviour. But it's of course out of scope of this PR - moving the test to correct module _might_ be more complicated.

So we have to leave it without `internal` for now.

example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
```diff
@@ -33,18 +44,34 @@ import kotlin.test.assertEquals
...
 class WooCommerceStoreTest {
+    private companion object {
+        const val TEST_SITE_REMOTE_ID = 1337L
+    }
+
     private val appContext = RuntimeEnvironment.application.applicationContext
     private val restClient = mock<WooSystemRestClient>()
+
+    private val roomDB = Room.inMemoryDatabaseBuilder(
+            ApplicationProvider.getApplicationContext<Application>(),
```
wzieba: 🔧 Can we extract `ApplicationProvider.getApplicationContext<Application>()` and use it in other places, where currently `RuntimeEnvironment.application.applicationContext` is used? The second one is deprecated and it would be nice to use the same app context in all places in the test.

example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
```diff
@@ -120,6 +176,32 @@ class WooCommerceStoreTest {
...
+    fun `when fetching ssr succeeds, then success returned`() = test {
+        val result = fetchSSR(isError = false)
+
+        Assertions.assertThat(result.isError).isFalse
+        Assertions.assertThat(result.model).isNotNull
+    }
+
+    @Test
+    fun `when fetching ssr succeeds, then data is saved to database`() {
+        runBlocking {
```
wzieba, line 197: ⛏️  Nitpick but is there a reason why `runBlocking` is used here and `test` is used above?

wzieba (approved):
Looks and works good! I left some minor comments related to tests only.

example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
hafizrahman: Good catch, thanks!

example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
hafizrahman: Ah, I forgot to replace the first one. Thanks!

PR #2130, opened 2021-09-28
Fetch site domains

ravishanker:
Fetch list of site domains for display on new domains dashboard

Please see this [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/15377) for result

- Add a new action FETCH_DOMAINS
- Fetches domains for a given site from `sites/:siteId/domains` endpoint

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
@@ -704,7 +704,7 @@ class SiteRestClient @Inject constructor(
 
     suspend fun fetchSiteDomains(site: SiteModel): Response<DomainsResponse> {
         val url = WPCOMREST.sites.site(site.siteId).domains.urlV1_1
-         return  wpComGsonRequestBuilder.syncGetRequest(this, url, mapOf(), DomainsResponse::class.java)
+        return wpComGsonRequestBuilder.syncGetRequest(this, url, mapOf(), DomainsResponse::class.java)
     }
 
     fun designatePrimaryDomain(site: SiteModel, domain: String) {
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
@@ -23,6 +23,7 @@ import org.wordpress.android.fluxc.network.UserAgent
 import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Error
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Success
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -1238,6 +1263,8 @@ open class SiteStore
             FETCHED_JETPACK_CAPABILITIES -> handleFetchedJetpackCapabilities(
                     action.payload as FetchedJetpackCapabilitiesPayload
             )
+            FETCH_DOMAINS -> fetchSiteDomains(action.payload as SiteModel)
+            FETCHED_DOMAINS -> handleFetchedSiteDomains(action.payload as FetchedDomainsPayload)
```
renanferrari: Since we're introducing a new call, we should probably replace `EventBus` usage with Coroutines, as per [the new architecture guidelines](https://github.com/wordpress-mobile/WordPress-FluxC-Android/wiki/Architecture), unless we really need it for some reason (I don't think we do in this case, but let me know if I'm missing something). 

After that's done, I'd also recommend:
- Wrapping the new method in the store with [`coroutineEngine.withDefaultContext`]( https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/9f6351af24f45379f96013ee4b44fe002220c385/fluxc/src/main/java/org/wordpress/android/fluxc/tools/CoroutineEngine.kt#L20-L28) to add some extra logging.
- Adding some tests. Doesn't need to be super comprehensive, but anything is better than none in this case 😉

renanferrari (changes requested):
Thanks for the changes! I've tested this with the WPAndroid PR and it seems to work as expected.

The code also looks good, but I left a comment below with a few suggestions.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
ravishanker: Thanks @renanferrari for a great suggestion as always.  Refactored to use coroutines, also added tests.  I was indeed feeling weird using EventBus 😏  Thank you.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -1238,6 +1262,9 @@ open class SiteStore
             FETCHED_JETPACK_CAPABILITIES -> handleFetchedJetpackCapabilities(
                     action.payload as FetchedJetpackCapabilitiesPayload
             )
+            FETCH_DOMAINS -> coroutineEngine.launch(T.MAIN, this, "Fetch site domains") {
+                emitChange(fetchSiteDomains(action.payload as SiteModel))
+            }
```
renanferrari: I'd actually just remove this too. We want to call the `fetchSiteDomains` function directly, so we shouldn't need to dispatch an action to use it. The reason other functions in this class do it this way is that they are trying to maintain backward compatibility.

Then you can move the `coroutineEngine` call inside the `fetchSiteDomains` function and use `withDefaultContext` instead of `launch`, similar to how we do it on the [`SiteOptionsStore`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/d32de86947dea4d0bc1b6ea247eb10fa73cc2777/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteOptionsStore.kt), for example.

Makes sense?

renanferrari (commented):
Thanks for the changes! Coroutines makes everything look so much better, doesn't it? 😄

I still left a couple of suggestions below, but other than that, this is pretty much good to go.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
ravishanker: @renanferrari Thanks again.  It certainly does make sense to go all the way.  Updated 👍

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -702,24 +702,9 @@ class SiteRestClient @Inject constructor(
...
-                val siteErrorType = when (response.error.apiError) {
-                    "unauthorized" -> UNAUTHORIZED
-                    "unknown_blog" -> UNKNOWN_SITE
-                    else -> SiteErrorType.GENERIC_ERROR
-                }
-                val domainsError = SiteError(siteErrorType, response.error.message)
-                FetchedDomainsPayload(site, domainsError)
-            }
-        }
+        return wpComGsonRequestBuilder.syncGetRequest(this, url, mapOf(), DomainsResponse::class.java)
```
renanferrari: Nice cleanup! 👍

renanferrari (approved):
@ravishanker Thank you for all the changes! It's looking great now.

PR #2134, opened 2021-10-01
Woo: identify Jetpack CP sites, and check WooCommerce's availability

hichamboushaba:
As part of https://github.com/woocommerce/woocommerce-android/issues/4856, we need to be able to identify sites that use Jetpack Connection Package instead of the full Jetpack plugin.

WCAndroid's PR: https://github.com/woocommerce/woocommerce-android/pull/4933

This PR adds the following changes:
- It adds a new column to `SiteModel` to identify sites using Jetpack CP.
- Since we can't rely on the option `woocommerce_is_active` for sites using Jetpack CP, we need to check if WooCommerce is available manually.
In order to avoid performing this check from the apps other than WCAndroid, I added a helper function  `fetchWooCommerceSites` to WooCommerceStore, to fetch sites and perform the check. And for cases where we fetch sites using the old way (for example during login), WooCommerceStore will listen to those changes, and perform the Woo availability check.

#### Testing
1. You need a new site that doesn't have Jetpack, or if using an old one, please uninstall Jetpack, then disconnect the store from the [dashboard](https://dashboard.wordpress.com/wp-admin/index.php?page=my-blogs) to make sure the flag `jetpack` has been reset.
2. Install WCPay and WooCommerce in the store.
3. Go to Payments, then click on `Setup`, and perform the WordPress.com authorization (you don't need the rest of steps).
4. Open the example app.
5. Click on Woo.
6. Click on `Log Sites`
7. Confirm that you can see the new added site printed in the log.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/WooCommerceFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/WooCommerceFragment.kt
@@ -63,9 +63,17 @@ class WooCommerceFragment : Fragment() {
         super.onViewCreated(view, savedInstanceState)
 
         log_sites.setOnClickListener {
-            for (site in wooCommerceStore.getWooCommerceSites()) {
-                prependToLog(site.name + ": " + if (site.isWpComStore) "WP.com store" else "Self-hosted store")
-                AppLog.i(T.API, LogUtils.toString(site))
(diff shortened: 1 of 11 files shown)


hafizrahman (approved):
This looks good to me. I also tried with WooCommerce and Jetpack Backup combination (instead of WCPay) and it works correctly as well.

PR #2135, opened 2021-10-01
Issue/4806 convert update order status to suspend

malinajirka:
"Not ready for merge" - Should be merged together with a PR in WCAndroid

This PR refactors updateOrderStatus into a suspendable function - removes usage of the event bus.

To Test:
1. Test updateOrderStatus action in the example app or test in WCAndroid ([PR](https://github.com/woocommerce/woocommerce-android/pull/4955))

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
@@ -401,7 +401,7 @@ class OrderRestClient @Inject constructor(
         val url = WOOCOMMERCE.orders.id(orderToUpdate.remoteOrderId).pathV3
         val params = mapOf("status" to status)
 
-        val response = jetpackTunnelGsonRequestBuilder.syncGetRequest(
+        val response = jetpackTunnelGsonRequestBuilder.syncPutRequest(
             this,
             site,
             url,
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
@@ -388,41 +388,50 @@ class OrderRestClient @Inject constructor(
      * Makes a PUT call to `/wc/v3/orders/<id>` via the Jetpack tunnel (see [JetpackTunnelGsonRequest]),
      * updating the status for the given [order] to [status].
      *
-     * Dispatches a [WCOrderAction.UPDATED_ORDER_STATUS] with the updated [WCOrderModel].
      *
      * Possible non-generic errors:
      * [OrderErrorType.INVALID_PARAM] if the [status] is not a valid order status on the server
      * [OrderErrorType.INVALID_ID] if an order by this id was not found on the server
      */
-    fun updateOrderStatus(
+    suspend fun updateOrderStatus(
         orderToUpdate: WCOrderModel,
         site: SiteModel,
         status: String
-    ) {
+    ): RemoteOrderPayload {
         val url = WOOCOMMERCE.orders.id(orderToUpdate.remoteOrderId).pathV3
         val params = mapOf("status" to status)
 
-        val request = JetpackTunnelGsonRequest.buildPutRequest(url, site.siteId, params, OrderApiResponse::class.java,
-                listener = { response: OrderApiResponse? ->
-                    response?.let {
-                        val newModel = orderResponseToOrderModel(it).apply {
-                            id = orderToUpdate.id
-                            localSiteId = orderToUpdate.localSiteId
-                        }
-                        val payload = RemoteOrderPayload(newModel, site)
-                        dispatcher.dispatch(WCOrderActionBuilder.newUpdatedOrderStatusAction(payload))
+        val response = jetpackTunnelGsonRequestBuilder.syncGetRequest(
+            this,
+            site,
+            url,
+            params,
+            OrderApiResponse::class.java
+        )
+
+        return when (response) {
+            is JetpackSuccess -> {
+                response.data?.let {
+                    val newModel = orderResponseToOrderModel(it).apply {
+                        id = orderToUpdate.id
+                        localSiteId = orderToUpdate.localSiteId
                     }
-                },
-                errorListener = { networkError ->
-                    val orderError = networkErrorToOrderError(networkError)
-                    val payload = RemoteOrderPayload(
-                            orderError,
-                            orderToUpdate,
-                            site
-                    )
-                    dispatcher.dispatch(WCOrderActionBuilder.newUpdatedOrderStatusAction(payload))
-                })
-        add(request)
+                    RemoteOrderPayload(newModel, site)
+                } ?: RemoteOrderPayload(
+                    OrderError(type = GENERIC_ERROR, message = "Success response with empty data"),
+                    orderToUpdate,
+                    site
+                )
+            }
+            is JetpackError -> {
+                val orderError = networkErrorToOrderError(response.error)
+                RemoteOrderPayload(
+                    orderError,
+                    orderToUpdate,
+                    site
+                )
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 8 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -504,13 +514,28 @@ class WCOrderStore @Inject constructor(
...
+                val rowsAffected = OrderSqlUtils.insertOrUpdateOrder(remotePayload.order)
+                OnOrderChanged(rowsAffected)
+            }
 
-            wcOrderRestClient.updateOrderStatus(payload.order, site, status)
+            remoteUpdateResult.causeOfChange = WCOrderAction.UPDATE_ORDER_STATUS
+
+            emit(RemoteUpdateResult(remoteUpdateResult))
+            // Needs to remain here until all event bus observables are removed from the client code
+            emitChange(remoteUpdateResult)
```
wzieba, line 538: As I'm currently working on optimistic update for customer notes I already browse this PR for some time 😄 .

Overall, great improvement, thank you for that and discussing it on Slack before.

My feedback on this solution is: 
1. WDYT about not including `OnOrderChanged` event in `UpdateOrderStatusResult`? I'm afraid that we couple here some event-bus-like class. 
2. Second thing is: do we need `rowsAffected` in `UpdateOrderStatusResult`?

malinajirka: > WDYT about not including OnOrderChanged event in UpdateOrderStatusResult? I'm afraid that we couple here some event-bus-like class.

I think that's actually an intention - as we are refactoring the event-bus architecture we sometime needs to still emit the original events (such as in this case, since the client app relies on them). I think it's nice that the event/model returned from the suspendable function is identical to the event emitted into the events bus. It's even the recommended practice mentioned in [the architecture overview](https://github.com/wordpress-mobile/WordPress-FluxC-Android/wiki/Architecture) - `Store returns the OnChanged event back to the app.`.
I mean I'm not technically against changing this, but I think it's a bigger decision that'd need to be made among all FluxC devs. Wdyt?

> Second thing is: do we need rowsAffected in UpdateOrderStatusResult?

It is currently part of the `OnOrderChanged` and is used eg [here](https://github.com/woocommerce/woocommerce-android/blob/1bf576d8d75708dc860e04c932b31da39301c623/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/mystore/MyStorePresenter.kt#L278). I think changing that is out-side of the scope of this PR. Wdyt?

wzieba:
@malinajirka can I ask you to merge/rebase this PR on top of `develop`? With recently included change (#2139 ) I'll be able to test this change with example app. Currently I can't (unfortunately, the first site I have doesn't include WooCommerce 😅 ).

malinajirka:
> can I ask you to merge/rebase this PR on top of develop? With recently included change (#2139 ) I'll be able to test this change with example app. Currently I can't (unfortunately, the first site I have doesn't include WooCommerce 😅 ).

@wzieba Done ;). Thanks for looking into this. Btw I worked-around this issue a while back by changing title of my main site to "AAAAA..." :P.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
wzieba: > I think it's nice that the event/model returned from the suspendable function is identical to the event emitted into the events bus.

Right, that's a good point. On a second thought it sounds good and will cause less changes in Woo (in this and other cases). Thankfully `OnOrderChanged` doesn't have dependencies on EventBus. 

For future: I wonder if we should delete `OnOrderChanged#causeOfChange` in a long term? If yes, maybe it's good idea to already deprecate it? (not necessarily in scope of this PR).

> It is currently part of the OnOrderChanged and is used eg here.

True, thanks for the link. Definitely it's out-side of the scope of this PR.

malinajirka: > For future: I wonder if we should delete OnOrderChanged#causeOfChange in a long term? If yes, maybe it's good idea to already deprecate it? (not necessarily in scope of this PR).

Yes, I think that's a good idea - we could deprecate even `WCOrderAction` and `@ActionEnum`. But as you mentioned, I think it's currently in the "plan" stage as it's still used without an alternative in some places.

wzieba:
@malinajirka I think I've found an issue - I've looked briefly at code but I can't see the reason 🤔 . 

### Before
On current `develop` if I try to update status with some dummy nonsense value, I receive error:

https://user-images.githubusercontent.com/5845095/136221931-eb6dc782-f8e5-4ae3-8a6f-4653c878ca9e.mov

### After
But with this PR I don't receive error

https://user-images.githubusercontent.com/5845095/136222054-f3dabc1f-3f22-4346-84db-cdcbc3405150.mov

but confirmation for both optimistic and remote updates.

In both cases order status is not updated on remote.

---

This is strange as I think the error is propagated in `remoteUpdateResult`, can you see the cause?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -388,41 +388,50 @@ class OrderRestClient @Inject constructor(
...
-        val request = JetpackTunnelGsonRequest.buildPutRequest(url, site.siteId, params, OrderApiResponse::class.java,
-                listener = { response: OrderApiResponse? ->
-                    response?.let {
-                        val newModel = orderResponseToOrderModel(it).apply {
-                            id = orderToUpdate.id
-                            localSiteId = orderToUpdate.localSiteId
-                        }
-                        val payload = RemoteOrderPayload(newModel, site)
-                        dispatcher.dispatch(WCOrderActionBuilder.newUpdatedOrderStatusAction(payload))
+        val response = jetpackTunnelGsonRequestBuilder.syncGetRequest(
```
wzieba: Oh I see @malinajirka - here should be `syncPutRequest` I think!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -166,6 +169,13 @@ class WCOrderStore @Inject constructor(
         constructor(error: OrderError, order: WCOrderModel, site: SiteModel) : this(order, site) { this.error = error }
     }
 
+    sealed class UpdateOrderStatusResult {
```
wzieba: Also one more question: WDYT about changing this to `UpdateOrderResult`?

It doesn't have to be done in this PR, more like a question - as I'm working on updating notes and in future we'll be updating even more fields. All of those results would look the same. Do you think it's wise to reuse it or do you see arguments against?

malinajirka: Good idea, fixed in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2135/commits/6044af2929a430b52d1650fed5ccf3126b1ff057

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
malinajirka: 😞 I should have noticed this, sorry. Fixed in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2135/commits/af7673a33a8b562aa4cdd17476d9a30d9f1cabe8

P.S. I even remember changing the status to some random string and thinking "wow interesting, I can change the status to whatever I want". It didn't even cross my mind that it's not working but returning 200 due to the wrong method.

malinajirka:
Thanks for the review - great catch! I'm glad we didn't merge this into production 😨. It's ready for another round 🙇

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
wzieba: I had a similar experience (especially as client app supports plaintext statuses IIRC) but then I've tested `develop` and noticed errors so it made me think that it's something wrong. I agree - backend should return something else than "200 - OK" if update didn't happen.

wzieba (approved):
Looks good, thank you @malinajirka . Yet another step in a great direction!

PR #2140, opened 2021-10-07
Throw exception on null `type` or `title format` in Add-on DTO.

### LINKED ISSUE #4941: NullPointerException: Attempt to invoke virtual method 'int java.lang.Enum.ordinal()' on a null object reference
Sentry Issue: [WOOCOMMERCE-ANDROID-2CT](https://sentry.io/organizations/a8c/issues/2671870886/?referrer=github_integration)

```
NullPointerException: Attempt to invoke virtual method 'int java.lang.Enum.ordinal()' on a null object reference
    at org.wordpress.android.fluxc.network.rest.wpcom.wc.addons.mappers.RemoteAddonMapper.toDomain(RemoteAddonMapper.kt:23)
    at org.wordpress.android.fluxc.store.WCProductStore.mapProductAddonsToDomain(WCProductStore.kt:1078)
    at org.wordpress.android.fluxc.store.WCProductStore.access$getDEFAULT_CATEGORY_SORTING$cp(WCProductStore.kt:48)
    at org.wordpress.android.fluxc.store.WCProductStore.access$mapProductAddonsToDomain(WCProductStore.kt:48)
    at org.wordpress.android.fluxc.store.WCProductStore$handleFetchProductsCompleted$2.invokeSuspend(WCProductStore.kt:1135)
...
```

wzieba:
Fixes https://github.com/woocommerce/woocommerce-android/issues/4941

### Description

This PR is about dropping invalid Add-on DTOs and not consider it for passing outside network layer. Invalid add-on DTO is one that does not meet requirements described in API specification: `type` and `title format` are required parameters.

### How to test

I couldn't replicate behaviour so I had to SSP into malformed site. The best way would be to firstly:
1. Log-in to site with app based on `develop` of FluxC and see that app crashes
2. Then build app based on this PR and run app again - it should not crash anymore

### Logging events

Those exceptions are consumed in the callers of the mapper and logged via `AppLog`. I've decided to not send Sentry issues when this happens ([what I suggested in discussion](https://github.com/woocommerce/woocommerce-android/issues/4941#issuecomment-936562181)) as I've realised that those malformed add-ons are completely empty and won't give us any information on how to fix things. Even site [seems to not have add-ons plug-in enabled](https://github.com/woocommerce/woocommerce-android/issues/4941#issuecomment-937605856) (maybe had in the past?).

I've added some more context to thrown exception so it might be easier in the future to spot malformed DTO right away.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/addons/mappers/RemoteAddonMapper.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/addons/mappers/RemoteAddonMapper.kt
@@ -29,7 +29,7 @@ object RemoteAddonMapper {
                     position = dto.position,
                     options = dto.mapOptions(),
                     display = dto.display?.toDomainModel() ?: throw MappingRemoteException(
-                                    "Can't map ${dto.name}. MultipleChoice add-on type has to have `display` defined."
+                                    "MultipleChoice add-on type has to have `display` defined."
                             )
             )
             Checkbox -> Addon.Checkbox(
@@ -137,7 +137,7 @@ object RemoteAddonMapper {
                 RemoteRestrictionsType.Email -> Addon.CustomText.Restrictions.Email
             }
         } else {
-            throw MappingRemoteException("Can't map $name. CustomText Add-on has to have restrictions defined.")
+            throw MappingRemoteException("CustomText Add-on has to have restrictions defined.")
         }
     }
 
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/addons/mappers/RemoteAddonMapper.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/addons/mappers/RemoteAddonMapper.kt
@@ -92,6 +92,7 @@ object RemoteAddonMapper {
                     required = dto.required.asBoolean(),
                     position = dto.position
             )
+            else -> throw MappingRemoteException("Add-on has to have type")
         }
     }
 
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/addons/mappers/RemoteAddonMapper.kt
```diff
@@ -109,11 +110,12 @@ object RemoteAddonMapper {
         }.orEmpty()
     }
 
-    private fun RemoteTitleFormat.toDomainModel(): Addon.TitleFormat {
+    private fun RemoteTitleFormat?.toDomainModel(): Addon.TitleFormat {
         return when (this) {
             Label -> Addon.TitleFormat.Label
             RemoteTitleFormat.Heading -> Addon.TitleFormat.Heading
             Hide -> Addon.TitleFormat.Hide
+            null -> throw MappingRemoteException("Add-on has to have title")
```
ThomazFB, line 118: Just to be sure, this part is to fix https://github.com/woocommerce/woocommerce-android/issues/4843 too, is that correct?

ThomazFB (approved):
Nice work here @wzieba! Thank you so much for tackling this! And awesome job logging the parse failures and testing everything! LGTM! :shipit: 

By the way, should we create a PR on Woo to include these changes here directly into release 7.7? This way we can generate a new beta cut and have fast feedback of the changes. WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/addons/mappers/RemoteAddonMapper.kt
wzieba: Right, thanks for awareness! I'll close the ticket then. I resolved this together because the app was still crashing after making type nullable.

wzieba:
Hi @ThomazFB ! That's good idea, I forgot about this and unfortunately I'm not close to the computer now. If you have time, it'd be awesome if you could create PR. If not I can handle it tomorrow around morning!

ThomazFB:
Leave it to me @wzieba! You already solved the issue, so go enjoy your weekend 😄 Thanks!

ThomazFB:
Okay, I just verified that the current state of FluxC conflicts with the release/7.7 code base state after the Order Status improvements and would require us to do some git operations on a week worth of team commits just to make everything compatible. So I think it's best to wait for 7.8 and get feedback over the fix since it's not affecting many users right now.

PR #2141, opened 2021-10-08
Expose "temporary" parameter for cart creation

renanferrari:
This PR exposes the "temporary" parameter in [`TransactionsRestClient::createShoppingCart`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/0fe3490b81e3b099f2329229dfc9a79c7422d48c/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/transactions/TransactionsRestClient.kt#L57-L62) and updates its related store method accordingly.

This is necessary in order to purchase domains through a persistent shopping cart.

There is no companion PR in WPAndroid for this change, but pulling this into a local FluxC build and building the app should be enough to make sure it isn't breaking anything.


ravishanker (approved):
LGTM 👍

PR #2149, opened 2021-10-11
Use order `local id` for order updates

wzieba:
Part of fixing https://github.com/woocommerce/woocommerce-android/issues/4982

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
@@ -8,6 +8,7 @@ import com.wellsql.generated.WCOrderStatusModelTable
 import com.wellsql.generated.WCOrderSummaryModelTable
 import com.yarolegovich.wellsql.SelectQuery
 import com.yarolegovich.wellsql.WellSql
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.WCOrderModel
@@ -81,6 +82,16 @@ object OrderSqlUtils {
                     .put(order, UpdateAllExceptId(WCOrderModel::class.java)).execute()
         }
     }
+
+    fun getOrderOrNull(orderId: RemoteId, siteId: LocalId): WCOrderModel? = WellSql.select(WCOrderModel::class.java)
+            .where()
+            .beginGroup()
+            .equals(WCOrderModelTable.REMOTE_ORDER_ID, orderId.value)
+            .equals(WCOrderModelTable.LOCAL_SITE_ID, siteId.value)
+            .endGroup()
+            .endWhere()
+            .asModel
+            .firstOrNull()
 
     fun getOrderForIdSet(orderIdSet: OrderIdSet): WCOrderModel? {
         val (id, remoteOrderId, localSiteId) = orderIdSet
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
@@ -83,15 +83,16 @@ object OrderSqlUtils {
         }
     }
 
-    fun getOrderForSiteByRemoteId(orderId: RemoteId, siteId: LocalId): WCOrderModel? = WellSql.select(WCOrderModel::class.java)
-            .where()
-            .beginGroup()
-            .equals(WCOrderModelTable.REMOTE_ORDER_ID, orderId.value)
-            .equals(WCOrderModelTable.LOCAL_SITE_ID, siteId.value)
-            .endGroup()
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/OrderSqlUtils.kt
```diff
@@ -82,6 +83,16 @@ object OrderSqlUtils {
         }
     }
 
+    fun getOrderOrNull(orderId: RemoteId, siteId: LocalId): WCOrderModel? = WellSql.select(WCOrderModel::class.java)
```
nbradbury: WDYT about renaming this to `getOrderByRemoteId`, to be consistent with `getOrdersForSiteByRemoteIds` and `getOrderByLocalId`?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -21,29 +22,45 @@ class OrderUpdateStore @Inject internal constructor(
     private val orderSqlDao: OrderSqlDao
 ) {
     suspend fun updateOrderNotes(
-        initialOrder: WCOrderModel,
+        orderId: RemoteId,
```
nbradbury: I mentioned this on Slack but thought I should bring it up here, too. The function is misnamed because it's not updating the order notes - that's a separate thing - but instead updating a single **customer** order note. So this would be more accurate if it was `updateCustomerOrderNote` (singular).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -21,29 +22,45 @@ class OrderUpdateStore @Inject internal constructor(
...
+                    OnOrderChanged(orderSqlDao.insertOrUpdateOrder(updateRemoteOrderPayload.order))
+                }
+                emit(RemoteUpdateResult(remoteUpdateResult))
             }
-            emit(RemoteUpdateResult(remoteUpdateResult))
         }
     }
+
+    private companion object {
+        const val NO_ROW_AFFECTED = -1
```
nbradbury: I'm confused by why this is `-1`. Shouldn't it return `0` if no rows are affected?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/wrappers/OrderSqlDao.kt
```diff
@@ -14,4 +16,8 @@ internal class OrderSqlDao @Inject constructor() {
                 .apply(updateOrder)
         return OrderSqlUtils.insertOrUpdateOrder(updatedOrder)
     }
+
+    fun getOrder(orderId: RemoteId, siteId: LocalId): WCOrderModel? {
```
nbradbury: Similar to the above comment, how about `getOrderByRemoteId`?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
wzieba: True, good point! I'm changing this in next commit, thank you.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
wzieba: Thanks! True - it should be ~one~ zero. I thought that it's "row affected" but its indeed `rowS affected`. WellSql uses `SQLiteDatabase` under the hood ant this is update method with documentation: 
> @ return the number of rows affected

https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/database/sqlite/SQLiteDatabase.java#1750

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/wrappers/OrderSqlDao.kt
wzieba: I agree for changing in `OrderSqlUtils.kt` to keep convention in this object but for `OrderSqlDao` I'd opt to leave simple `getOrder` - I think it's shorter and we already know (by looking at method signature) what is needed to obtain the order. WDYT?

nbradbury: Similar to my WCAndroid comment, I just think we should be explicit about it being a remote ID.

wzieba:
@nbradbury thank you for the review - I've applied your suggestions.

Because of new arguments under discussion (p91TBi-6dF-p2#comment-6691) I've also changed the idea of using `remote id` to `local id` - I think it might work well.

I didn't yet sync this change with Woo PR - could you please roughly check the PR and confirm that it's okay to you to continue with using `local id`?

nbradbury:
> I didn't yet sync this change with Woo PR - could you please roughly check the PR and confirm that it's okay to you to continue with using `local id`?

Yes, I think it's fine.

nbradbury (approved):
Looks good! :shipit:

PR #2150, opened 2021-10-13
Issue/woo 4806 suspendable add order shipment tracking

malinajirka:
"Not ready for merge" - Should be merged together with a PR in WCAndroid

This PR refactors addOrderShipmentTracking into a suspendable function - removes usage of the event bus.

To Test:
1. Test addOrderShipmentTracking action in the example app or test in WCAndroid ([PR](https://github.com/woocommerce/woocommerce-android/pull/5005))


hichamboushaba (approved):
Nice work, thanks @malinajirka

PR #2151, opened 2021-10-14
Issue/woo 4806 suspendable delete shipment tracking

malinajirka:
This PR refactors deleteOrderShipmentTracking into a suspendable function - removes usage of the event bus.

- This PR also moves "AddOrderShipmentTracking" to a bg thread, using a coroutine engine

To Test:
1. Test deleteOrderShipmentTracking action in the example app  (you first need to fetch orders and fetch trackings for an order) or test in WCAndroid ([PR](https://github.com/woocommerce/woocommerce-android/pull/5008))


0nko (approved):
The test worked as expected. Nice work! 👍

0nko:
We can merge this once the conflict's fixed 👇 .

PR #2152, opened 2021-10-14
Issue/woo 4806 suspendable add note

malinajirka:
This PR refactors addOrderNote into a suspendable function - removes usage of the event bus.

Other changes
1. The app now returns an error response when the server returns success with an empty body - [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/34004e9927a505f3ad0a90126f217a34cdce9068/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt#L504) - instead of silently consuming the event.
2. The rest client now returns `newNote` instead of the `original/local` note - [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/34004e9927a505f3ad0a90126f217a34cdce9068/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt#L503) - I suspect this was a bug in the previous implementation, the note which the app received from the server was completely ignored in the previous implementation.

To Test:
1. Test addOrderNote action in the example app ( or test in WCAndroid ([PR](https://github.com/woocommerce/woocommerce-android/pull/5012)))
2. Check wpadmin if the note is visible in wpadmin


ThomazFB (approved):
Hey @malinajirka, nice work here, thanks for tackling this huge work of removing EventBus usage out of the Orders requests. I had some questions about keeping using payloads, but your previous PR discussion with Wojtek already answered everything, so LGTM :shipit:

PR #2154, opened 2021-10-15
Add domains type

ravishanker:
Adding claim domain and get domain types to to show in quick start next steps for domains feature


renanferrari (commented):
Thanks for the changes! Assuming we want the domain actions to be a part of Quick Start, this is looking good. The problem is that I'm not sure if that's what we really want, especially considering that according to the current design specs, the domain actions should appear even when Quick Start has not started yet or has finished.

I've raised a few questions about this here: p1634316764310800/1634074917.274500-slack-C0280085R7W

We should put this on hold until we get further clarification.

ravishanker:
Closing it as it is no longer required for the new design

PR #2159, opened 2021-10-20
Enable updating both order addresses on the same request

ThomazFB:
Summary
==========
Fixes issue https://github.com/woocommerce/woocommerce-android/issues/4978 by creating a way to update both Shipping and Billing Addresses of a given Order on a single request. Because of the `use as` toggle inside the Address editing form, we need a way to update both addresses meanwhile avoiding creating duplicated optimistic requests for something that should be a single updating operation.

How to Test
==========
1. Go to the latest store Order on your test site wp-admin dashboard and leave the Billing and Shipping Addresses with different information from each other
image

2. Go to the example app and select to update the Billing and Shipping address and hit the `send update replicating both addresses`

3. Go back to your latest store Order on your test site wp-admin dashboard and verify if the Shipping and Billing Addresses are now the same

Screen Shot 2021-10-20 at 11 42 31

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
@@ -34,6 +34,7 @@ class AddressEditDialogFragment : DaggerFragment() {
     private lateinit var binding: FragmentAddressEditDialogBinding
 
     var selectedOrder = MutableStateFlow<WCOrderModel?>(null)
+    var originalBillingEmail = ""
 
     override fun onCreateView(
         inflater: LayoutInflater,
@@ -51,6 +52,7 @@ class AddressEditDialogFragment : DaggerFragment() {
 
         CoroutineScope(Dispatchers.Main).launch {
             currentAddressType.combine(selectedOrder) { addressType, order ->
+                originalBillingEmail = order?.billingEmail.orEmpty()
                 when (addressType) {
                     SHIPPING -> {
                         binding.addressTypeSwitch.text = "Edit shipping address for order ${order?.remoteOrderId}"
@@ -134,6 +136,8 @@ class AddressEditDialogFragment : DaggerFragment() {
             country = binding.country.text.toString(),
             phone = binding.phone.text.toString(),
             email = binding.email.text.toString()
+                    .takeIf { it.isNotEmpty() }
+                    ?: originalBillingEmail
     )
 
     private fun generateShippingAddressModel() = Shipping(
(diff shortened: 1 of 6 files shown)


nbradbury:
@ThomazFB I'm getting this crash when trying to update the order addresses. Was a database migration perhaps missed?

```
ProductSqlUtils.insertOrUpdateProducts(ProductSqlUtils.kt:65)
2021-10-20 11:41:02.171 24202-25648/? E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-3
    Process: org.wordpress.android.fluxc.example, PID: 24202
    android.database.sqlite.SQLiteException: no such column: SHIPPING_PHONE
```

ThomazFB:
@nbradbury I had the same issue after @wzieba PR, it's a migration issue after the adaptation work to support the Shipping and Billing domain models!

I'll check if there's a migration missing or maybe it's just a reinstall case.

ThomazFB:
@nbradbury It's fixed, I forgot to update the DB version, it should work now!

nbradbury:
@ThomazFB When I try to update both addresses I see "Invalid parameter: billing" in the log.

invalid

Flipper shows the response body to be:

```
{
  "error": "rest_invalid_param",
  "message": "Invalid parameter(s): billing"
}
```

ThomazFB:
@nbradbury after some investigations I understood what's going on since I couldn't reproduce this error from the beginning: this only happens when we try to clear the email field by replicating the Shipping address (doesn't contain any email info) into the Billing one in a Store using any version 5.8.1 and below. It relates to an API limitation discovered by the iOS folks here: https://github.com/woocommerce/woocommerce-ios/pull/5242

[The issue was merged into Woo Core already for version 5.9.0](https://github.com/woocommerce/woocommerce/pull/30850), but we will need a good way to handle this specific error on Woo side, so this will require a good bunch of work. With that said, I would like to suggest that we move forward with this limitation now and tackle it in a dedicated PR. I can create a separate issue to tackle all the improvements needed and evaluate how iOS handled this problem for Woo store versions below 5.8.1. WDYT?

And as a temporary workaround, I can handle on the Woo side [in this PR](https://github.com/woocommerce/woocommerce-android/pull/5017) a way to never erase the Billing email info and avoid this error for now. I don't think FluxC should be responsible to control this API constraint since it's just sending forward what was handed to it.

ThomazFB:
@nbradbury I've also updated the example app to handle empty email address inputs here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2159/commits/c1c741642f5b815502104388b68e3c69dda8cacb. So now you probably shouldn't see this error happening again.

nbradbury (approved):
Nice work! :shipit:

PR #2161, opened 2021-10-21
Fix phone info order shipping address

ThomazFB:
Summary
==========
This is a quick fix to allow the Shipping Address models to support Phone information since the update was missing some parts to effectively update the info at the site side.

How to Test
==========
1. Go to the Order Address update section
2. Update the Shipping Address phone
3. Verify if the latest order in the Woo Site got updated with the inserted information


nbradbury (approved):
Looks good! :shipit:

PR #2162, opened 2021-10-22
Feature/15481 fetch products for sales cost

ravishanker:
Fetch list of products from products endpoint for finding domains on sale for display on domains search suggestions screen

Fixes: #15481

- Adds a new endpoint /products
- Adds a products rest client with related product data class and adapter
- Add a new action FETCH_PRODUCTS
- Add products store with **fetchProducts** method


antonis (approved):
Great work @ravishanker 👍 
The code looks good and works as expected. The test also cover the added functionality pretty well 🎉

PR #2163, opened 2021-10-25
Jetpack CP: add support for uploading media using the WPCom's `/wp/v2` API

hichamboushaba:
Jetpack CP sites can't access the regular WPCom APIs that rely on syncing from Jetpack, we need to rely instead on the mirrored `/wp/v2` APIs (more details: p91TBi-67H-p2), this PR adds support for this by adding a new `RestClient` that use this API.

The PR is long because of the tests and the json sample file.

#### Testing
1. You need a site that uses a Jetpack CP plugin (WCPay or Backup) and doesn't have full Jetpack.
2. Open the example app, and login.
3. Click on Media.
4. Select the Jetpack CP site.
5. Test media upload and confirm that it works as expected.

You can also use the WCAndroid's PR for testing: https://github.com/woocommerce/woocommerce-android/pull/5071

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/wpv2/WPV2MediaRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/wpv2/WPV2MediaRestClient.kt
@@ -11,6 +11,7 @@ import okhttp3.Callback
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import org.wordpress.android.fluxc.Dispatcher
+import org.wordpress.android.fluxc.generated.UploadActionBuilder
 import org.wordpress.android.fluxc.generated.endpoint.WPAPI
 import org.wordpress.android.fluxc.model.MediaModel
 import org.wordpress.android.fluxc.model.MediaModel.MediaUploadState.FAILED
@@ -36,7 +37,7 @@ import kotlin.coroutines.resume
 class WPV2MediaRestClient @Inject constructor(
     dispatcher: Dispatcher,
     private val coroutineEngine: CoroutineEngine,
-    private val okHttpClient: OkHttpClient,
+    @Named("regular") private val okHttpClient: OkHttpClient,
     appContext: Context?,
     @Named("regular") requestQueue: RequestQueue,
     accessToken: AccessToken,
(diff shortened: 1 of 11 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/wpv2/WPV2MediaRestClient.kt
```diff
@@ -0,0 +1,198 @@
...
+        currentUploads[media.id]?.let { scope ->
+            scope.cancel()
+            val payload = ProgressPayload(media, 0f, false, true)
+            mDispatcher.dispatch(MediaActionBuilder.newCanceledMediaUploadAction(payload))
+        }
+    }
+
+    private fun syncUploadMedia(site: SiteModel, media: MediaModel): Flow<ProgressPayload> {
+        return callbackFlow {
+            val url = WPAPI.media.getWPComUrl(site.siteId)
```
hafizrahman, line 89: ❓ Non-blocking: Should we maybe put a code comment on why we're able to do it this way for media upload on Jetpack CP sites?

I understand the reasoning because I've read the conversations about it, but for somebody coming into this part of the code in the future, it might be unclear. WDYT?

hafizrahman (approved):
Approving because code looks good to me, and sorry it took me a while to figure out the flow logic part here. Just left a non-blocking question above.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/wpv2/WPV2MediaRestClient.kt
hichamboushaba: If we add a comment like this it would be misleading, since this new client works for all jetpack sites, as the new endpoint is a WPCom one, and that's one I called the function `getWPComUrl`.
There have been mentions by the Jetpack team that this endpoint is newer and might even offer better reliability than the one we use currently, so I'm planning to run some tests to confirm, and we may switch to this endpoint for all the uploads.

hafizrahman: > There have been mentions by the Jetpack team that this endpoint is newer and might even offer better reliability than the one we use currently

Ah, that's cool and good to know 👍🏼

PR #2164, opened 2021-10-26
Jetpack CP: Add Jetpack-tunneled plugins-related functionalities.

hafizrahman:
## PR Description

This PR is adding functionalities to:
-  Install a plugin,
- Fetch a plugin's information
  
to .org sites that are connected using Jetpack Connection Package.

## To Test

1. Prepare a site connected with Jetpack CP. An easy way is to start a .org site, don't install Jetpack, and instead install Jetpack Backup plugin. Open the plugin's setting and connect it to your wpcom account.
2. Open Example app, then sign in to your wpcom account.
3. Go to "Plugins", then select the site created on step 1.
4. Use the "Install and activate a plugin" button and try installing a plugin. You will need to enter a plugin's slug, which is the `http://wordpress.org/support/plugin/<SLUG>` part on the Plugin Directory. Observe the log and ensure it says the plugin is installed and activated.
5. Use the "Fetch a plugin from site" button. Try fetching for `jetpack-backup/jetpack-backup`, and observe the log. It will show the plugin's name and description.

Extra: you can also use Flipper to observe the network calls on both buttons's action and check the request/response there.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
@@ -84,12 +84,12 @@ class PluginsFragment : Fragment() {
         fetch_plugin.setOnClickListener {
             selectedSite?.let { site ->
 
-                showSingleLineDialog(activity, "Enter a plugin slug.") { pluginSlugText ->
-                    if (pluginSlugText.text.isEmpty()) {
-                        prependToLog("Slug is null so doing nothing")
+                showSingleLineDialog(activity, "Enter the plugin name.") { pluginNameText ->
+                    if (pluginNameText.text.isEmpty()) {
+                        prependToLog("Name is null so doing nothing")
                         return@showSingleLineDialog
                     }
-                    pluginSlugText.text.toString().apply {
+                    pluginNameText.text.toString().apply {
                         prependToLog("Fetching plugin: $this")
 
                         val payload = FetchJetpackSitePluginPayload(site, this)
--- a/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
@@ -1,16 +1,21 @@
 package org.wordpress.android.fluxc.example
 
+import android.content.Context
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.fragment.app.Fragment
+import dagger.android.support.AndroidSupportInjection
+import kotlinx.android.synthetic.main.fragment_notifications.*
 import kotlinx.android.synthetic.main.fragment_plugins.*
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
 import org.wordpress.android.fluxc.Dispatcher
-import org.wordpress.android.fluxc.example.ui.StoreSelectingFragment
+import org.wordpress.android.fluxc.example.ui.common.showSiteSelectorDialog
 import org.wordpress.android.fluxc.example.utils.showSingleLineDialog
 import org.wordpress.android.fluxc.generated.PluginActionBuilder
+import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.store.PluginStore
 import org.wordpress.android.fluxc.store.PluginStore.FetchSitePluginPayload
 import org.wordpress.android.fluxc.store.PluginStore.InstallSitePluginPayload
@@ -19,9 +24,12 @@ import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginFetched
 import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginInstalled
 import javax.inject.Inject
 
-class PluginsFragment : StoreSelectingFragment() {
+class PluginsFragment : Fragment() {
     @Inject internal lateinit var dispatcher: Dispatcher
     @Inject internal lateinit var pluginStore: PluginStore
+
+    private var selectedSite: SiteModel? = null
+    private var selectedPos: Int = -1
 
     override fun onStart() {
         super.onStart()
@@ -33,11 +41,26 @@ class PluginsFragment : StoreSelectingFragment() {
         dispatcher.unregister(this)
     }
 
+    override fun onAttach(context: Context) {
+        AndroidSupportInjection.inject(this)
+        super.onAttach(context)
+    }
+
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? =
             inflater.inflate(R.layout.fragment_plugins, container, false)
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+
+        plugins_select_site.setOnClickListener {
+            showSiteSelectorDialog(selectedPos, object : SiteSelectorDialog.Listener {
+                override fun onSiteSelected(site: SiteModel, pos: Int) {
+                    selectedSite = site
+                    selectedPos = pos
+                    plugins_selected_site.text = site.name ?: site.displayName
+                }
+            })
+        }
 
         install_activate_plugin.setOnClickListener {
             selectedSite?.let { site ->
@@ -99,6 +122,7 @@ class PluginsFragment : StoreSelectingFragment() {
         }
     }
 
+    @Suppress("unused")
     @Subscribe(threadMode = ThreadMode.MAIN)
     fun onSitePluginConfigured(event: OnSitePluginConfigured) {
         if (!event.isError) {
--- a/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
@@ -17,10 +17,10 @@ import org.wordpress.android.fluxc.example.utils.showSingleLineDialog
 import org.wordpress.android.fluxc.generated.PluginActionBuilder
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.store.PluginStore
-import org.wordpress.android.fluxc.store.PluginStore.FetchSitePluginPayload
+import org.wordpress.android.fluxc.store.PluginStore.FetchJetpackSitePluginPayload
 import org.wordpress.android.fluxc.store.PluginStore.InstallSitePluginPayload
+import org.wordpress.android.fluxc.store.PluginStore.OnJetpackSitePluginFetched
 import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginConfigured
-import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginFetched
 import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginInstalled
 import javax.inject.Inject
 
@@ -92,8 +92,8 @@ class PluginsFragment : Fragment() {
                     pluginSlugText.text.toString().apply {
                         prependToLog("Fetching plugin: $this")
 
-                        val payload = FetchSitePluginPayload(site, this)
-                        dispatcher.dispatch(PluginActionBuilder.newFetchSitePluginAction(payload))
+                        val payload = FetchJetpackSitePluginPayload(site, this)
+                        dispatcher.dispatch(PluginActionBuilder.newFetchJetpackSitePluginAction(payload))
                     }
                 }
             } ?: prependToLog("Please select a site first.")
@@ -114,7 +114,7 @@ class PluginsFragment : Fragment() {
 
     @Suppress("unused")
     @Subscribe(threadMode = ThreadMode.MAIN)
-    fun onSitePluginFetched(event: OnSitePluginFetched) {
+    fun onSitePluginFetched(event: OnJetpackSitePluginFetched) {
         if (!event.isError) {
             prependToLog("${event.plugin.displayName}: ${event.plugin.description}")
         } else {
--- a/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
@@ -43,7 +43,7 @@ class PluginsFragment : StoreSelectingFragment() {
             selectedSite?.let { site ->
                 showSingleLineDialog(
                         activity,
-                        "Enter a plugin slug.\n(Note: If testing Jetpack CP sites, don't enter `jetpack`)"
+                        "Enter a plugin slug.\n(Hint: If testing Jetpack CP sites, don't use `jetpack`)"
                 ) { pluginSlugText ->
                     if (pluginSlugText.text.isEmpty()) {
                         prependToLog("Slug is null so doing nothing")
@@ -99,11 +99,11 @@ class PluginsFragment : StoreSelectingFragment() {
         }
     }
 
-    @Suppress("unused")
     @Subscribe(threadMode = ThreadMode.MAIN)
     fun onSitePluginConfigured(event: OnSitePluginConfigured) {
         if (!event.isError) {
-            prependToLog("${event.pluginName} is configured.")
+            // If there is no error, we can assume that the configuration (activating in our case) is successful.
+            prependToLog("${event.pluginName} is activated.")
         }
     }
 }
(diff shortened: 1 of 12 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/plugin/PluginJetpackTunnelRestClient.kt
```diff
@@ -0,0 +1,182 @@
...
+
+        val request = JetpackTunnelGsonRequest.buildGetRequest(
+                url,
+                site.siteId,
+                emptyMap(),
+                PluginResponseModel::class.java,
+                { response: PluginResponseModel? ->
+                    response?.let {
+                        val payload = FetchedSitePluginPayload(sitePluginModelFromResponse(site, it))
+                        dispatcher.dispatch(PluginActionBuilder.newFetchedSitePluginAction(payload))
```
wzieba: hi @hafizrahman and @hichamboushaba ! 👋 I hope you don't mind jumping into your PRs with one comment but, [as Event Bus is a deprecated](https://github.com/wordpress-mobile/WordPress-FluxC-Android/wiki/Architecture), would you mind to change those new RestClients (`PluginJetpackTunnelRestClient` and `WPV2MediaRestClient` in #2163 ) to not use `Dispatcher` to communicate?

You could use coroutines like e.g. in [CommentsRestClient](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/766e0ae544de5d8b5e5d769d27f6bbb54bdf4f1d/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/stats/insights/CommentsRestClient.kt) or [AddonsRestClient](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/329454c58c643c5e1d50ee5c6e669422614a0698/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/addons/AddOnsRestClient.kt).

WDYT?

hafizrahman: 👋🏼  @wzieba Thank you for your suggestion! Personally the coroutines way is simpler and easier to understand for me, and I would have loved to use it here. In fact if you check [the first commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2164/commits/24d6cffb681a53602925e76cddb4d9f0ba585abc), you can see that my first iteration in the rest client already used coroutines.

However, the rest client is used to add functionalities to the existing `PluginStore`, that already exclusively uses EventBus. For the install plugin case, for example, the `PluginStore` already has a function that deals with installing plugin on:
- a wpcom site,
- a .org site with Jetpack plugin

And now this PR adds another case, to install to a .org site with Jetpack CP plugins.

I discussed this quite a bit with @hichamboushaba and we thought this would be the way to go for now. In the near future, the `PluginStore` and its related rest clients will have to be rewritten to not use Event Bus anymore.

wzieba: Ok, gotcha, thank you @hafizrahman for more context and thank you and @hichamboushaba for discussing this! 🙌 . 

> we thought this would be the way to go for now. In the near future, the PluginStore and its related rest clients will have to be rewritten to not use Event Bus anymore.

sounds good!

example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
```diff
@@ -0,0 +1,109 @@
...
+import org.wordpress.android.fluxc.generated.PluginActionBuilder
+import org.wordpress.android.fluxc.store.PluginStore
+import org.wordpress.android.fluxc.store.PluginStore.FetchSitePluginPayload
+import org.wordpress.android.fluxc.store.PluginStore.InstallSitePluginPayload
+import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginConfigured
+import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginFetched
+import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginInstalled
+import javax.inject.Inject
+
+class PluginsFragment : StoreSelectingFragment() {
```
hichamboushaba: Inhereting from `StoreSelectingFragment` would make this fragment compatible with Woo fragments only, for better compatibility, you can probably add the button yourself to the layout, and use `showSiteSelectorDialog` to handle the logic of the selected site by yourself, WDYT?

fluxc-processor/src/main/resources/wp-api-endpoints.txt
```diff
@@ -13,4 +13,7 @@
 /settings/
 
 /users/
-/users/me/
\ No newline at end of file
+/users/me/
+
+/plugins/
+/plugins/<slug>#String
```
hichamboushaba: I would rename `slug` on this context to `name` to align with the rest of `PluginStore` logic, and the `SitePluginModel`.
This would also make the parameter in `fetchPlugin` clearer, without needing the explanation on this being different than the `slug`.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/plugin/PluginJetpackTunnelRestClient.kt
```diff
@@ -0,0 +1,182 @@
...
+                    dispatcher.dispatch(PluginActionBuilder.newConfiguredSitePluginAction(payload))
+                }
+        )
+        add(request)
+    }
+
+    private fun sitePluginModelFromResponse(siteModel: SiteModel, response: PluginResponseModel): SitePluginModel {
+        val sitePluginModel = SitePluginModel().apply {
+            localSiteId = siteModel.id
+            name = response.name
```
hichamboushaba: The name in the database matches the `plugin` field, while the `displayName` is what matches the `name` field, check the existing mapping we have now: https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/abaefb3ae628cec6080ccf4be8ce523ce69ce2d2/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/plugin/PluginWPAPIRestClient.kt#L119-L132

I would also extract this logic and put it in a common place, for example in the `PluginResponseModel` file as an extension, to reuse the same logic in both places.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginStore.java
```diff
@@ -77,6 +78,17 @@ public FetchPluginDirectoryPayload(PluginDirectoryType type, @Nullable SiteModel
         }
     }
 
+    @SuppressWarnings("WeakerAccess")
+    public static class FetchSitePluginPayload extends Payload<BaseNetworkError> {
+        public SiteModel site;
+        public String slug;
+
+        public FetchSitePluginPayload(SiteModel site, String slug) {
```
hichamboushaba: Same as above, I would rename the `slug` parameter, to make it clear this is not a slug, to align with `ConfigureSitePluginPayload`, you can call it `pluginName`.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginStore.java
```diff
@@ -847,11 +914,22 @@ private void fetchWPOrgPlugin(String pluginSlug) {
         mPluginWPOrgClient.fetchWPOrgPlugin(pluginSlug);
     }
 
+    /* Fetch a single plugin from a site, to get its information and whether it exists or not.
+       Currently this is only supported on sites connected using Jetpack Connection Package.
+     */
+    private void fetchSitePlugin(FetchSitePluginPayload payload) {
+        if (payload.site.isJetpackCPConnected()) {
```
hichamboushaba: This can work for all jetpack connected sites, so IMO we should do two things here:
1. Update the condition to allow it to work for Jetpack sites too.
2. Add an else that falls back to using `fetchWPOrgPlugin` for the other cases, because the name `FETCH_SITE_PLUGIN` is generic.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginStore.java
```diff
@@ -847,11 +914,22 @@ private void fetchWPOrgPlugin(String pluginSlug) {
...
+    }
+
     private void installSitePlugin(InstallSitePluginPayload payload) {
         if (payload.site.isUsingWpComRestApi() && payload.site.isJetpackConnected()) {
             mPluginRestClient.installSitePlugin(payload.site, payload.slug);
         } else if (!payload.site.isUsingWpComRestApi()) {
             mPluginCoroutineStore.installSitePlugin(payload.site, payload.slug);
+        } else if (payload.site.isJetpackCPConnected()) {
+            mPluginJetpackTunnelRestClient.installPlugin(payload.site, payload.slug);
         } else {
```
hichamboushaba: The chaining of those conditions will be broken once https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2163 is merged, and the cause is that there is a bug right now that makes jetpack CP sites return `true` when we check `isUsingWpComRestApi`, it's fixed here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2163/files#diff-ac504807ac3755802bc839947ec85b4b045a906264eaf35ab97885016704df8aR1010-R1012

So to fix this, you need to move the check `isJetpackCPConnected` above the condition `if (!payload.site.isUsingWpComRestApi())`

hichamboushaba (changes requested):
Really great work @hafizrahman, the functionality works well.

I left some comments for discussion

example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
hafizrahman: Yup! I also realized from your recent PR how you did it for media upload test. Will change it.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginStore.java
hafizrahman: That makes sense, thanks!

Aside: How do you generate a permalink to a / some line of codes in the diff of a PR like that? I've been trying to figure that out for a while.

hichamboushaba: > How do you generate a permalink to a / some line of codes in the diff of a PR like that? I've been trying to figure that out for a while.

When you select the lines (by clicking on the line numbers and swiping or using the shift key), then the URL in the address bar gets updated, you just need to copy it, and it will contain the parameter of the highlighted lines.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginStore.java
hafizrahman: > Add an else that falls back to using fetchWPOrgPlugin for the other cases, because the name FETCH_SITE_PLUGIN is generic.

My purpose for this method is more to allow checking whether a plugin exists on a site, plus its current state (version, is it active or not, etc). The fallback will fetch a plugin's information from the repository. The informations from repository ([example here for Akismet](https://api.wordpress.org/plugins/info/1.0/akismet.json)) are different and seems to be for different purposes, it's more for the plugin's status in the repo (like ratings, changelogs).

What would you suggest in this case? I feel like probably `fetchSitePlugin()` might not be a clear enough name for its purpose.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginStore.java
hafizrahman: Woah nice, thanks for that!

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginStore.java
hichamboushaba: OK, I see, `fetchWPOrgPlugin` is different.

So for our case, maybe we can rename the function and payload to something that let's us know this is for Jetpack sites: `fetchPluginForJetpackSite`, and just remove the condition on `isJetpackCPConnected` to let it work for other jetpack connected sites, WDYT?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/plugin/PluginJetpackTunnelRestClient.kt
```diff
@@ -0,0 +1,160 @@
...
+    companion object {
+        private const val PLUGIN_ALREADY_EXISTS = "Destination folder already exists."
+    }
+
+    /**
+     * Fetch a plugin's information from a site.
+     *
+     * @param [pluginName] Note that this is not the same as the WordPress.org plugin directory slug that can be used
+     * to install a plugin. Instead, it needs to be the value of the `plugin` key inside a plugin object as returned
+     * by the `GET wp/v2/plugins` endpoint. For example, for Jetpack, the correct value should be `jetpack/jetpack`.
```
hichamboushaba: I think this comment can be simplified now, we don't need to mention that it's different from the `slug`, since they don't have the same name anymore 🙂

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/plugin/PluginJetpackTunnelRestClient.kt
```diff
@@ -0,0 +1,160 @@
...
+        add(request)
+    }
+
+    /**
+     * Configure a plugin's status in a site. This supports making it 'active', or 'inactive'. The API also supports
+     * the 'network-active' status, but it is not supported yet here.
+     *
+     * @param [pluginName] Note that this is not the same as the WordPress.org plugin directory slug that can be used
+     * to install a plugin. Instead, it needs to be the value of the `plugin` key inside a plugin object as returned
+     * by the `GET wp/v2/plugins` endpoint. For example, for Jetpack, the correct value should be `jetpack/jetpack`.
```
hichamboushaba: Same as above.

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginStore.java
hafizrahman: Great suggestion, thank you. I've made the changes.

hafizrahman:
This is ready for another round of review @hichamboushaba Thank you!

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginStore.java
```diff
@@ -921,11 +922,18 @@ private void fetchWPOrgPlugin(String pluginSlug) {
     }
 
     /* Fetch a single plugin from a site, to get its information and whether it exists or not.
-       Currently this is only supported on sites connected using Jetpack Connection Package.
+       Currently this is only supported on sites connected using Jetpack plugin or Jetpack Connection Package.
      */
-    private void fetchSitePlugin(FetchSitePluginPayload payload) {
-        if (payload.site.isJetpackCPConnected()) {
+    private void fetchPluginForJetpackSite(FetchJetpackSitePluginPayload payload) {
+        if (payload.site.isJetpackConnected()) {
```
hichamboushaba: This fetch should work also for Jetpack CP connected sites, no?
When I said "just remove the condition on isJetpackCPConnected to let it work for other jetpack connected sites", I meant let the client check on this, now that you added the error `NOT_JETPACK_SITE`, I think the condition should be: "payload.site.isJetpackConnected || payload.site.isJetpackCPConnected", WDYT?

hichamboushaba (commented):
The changes looks good, I just left one last comment 🙂

example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
```diff
@@ -0,0 +1,133 @@
...
+                        dispatcher.dispatch(PluginActionBuilder.newInstallSitePluginAction(payload))
+                    }
+                }
+            } ?: prependToLog("Please select a site first.")
+        }
+
+        fetch_plugin.setOnClickListener {
+            selectedSite?.let { site ->
+
+                showSingleLineDialog(activity, "Enter a plugin slug.") { pluginSlugText ->
```
hichamboushaba: A small nitpick too, should we say here "Enter the plugin name" instead?

hafizrahman: Thanks! I missed this part.

hichamboushaba (approved):
Great job @hafizrahman 💯

PR #2165, opened 2021-10-26
[1/4] Use Room database to store Orders data

wzieba:
### Description

This PR is about changing source of data for cached Orders from WellSql to Room database.

It also adds improvements to `OrderError` signature making it easier to work with (data class).

This PR has target into `woo/5070_orders_to_room` (feature branch) and **it's not meant to pass CI build**.

Please feel free to share feedback, nitpicking is welcome as its first PR that moves some entities from WellSql to Room - so probably it might be a template/pattern for future migrations.

PR statistics not including generated schema file: 359 additions and 387 deletions.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -1,84 +1,75 @@
...
-
-    @Column var feeLines = ""
-
-    @Column var metaData = ""
 
+@Entity
+data class WCOrderModel(
+    @PrimaryKey(autoGenerate = true) val id: Int = 0,
+    val localSiteId: Int = 0,
+    val remoteOrderId: Long = 0L, // The unique identifier for this order on the server
```
hichamboushaba, line 15: @wzieba just a question, did you consider using a [composite primary key](https://developer.android.com/training/data-storage/room/defining-data#composite-key) here using `localSiteId` and `remoteOrderId` without any local id? I don't know if this would work for our usage, and what are the impacts, I just wanted to consider it before going forward 🙂

wzieba: Good idea @hichamboushaba thanks! I've been thinking about this but I have concern about `Order Creation` project. For drafts of orders, I think we'll want to use the same table but we won't have `remoteOrderId` for them so I think we have to stick to auto generated local id. WDYT?

hichamboushaba: If it's like products creation, then we will have a maximum of one order draft that's saved locally and not present in the server, in this case, we can use a constant as a temporary value for `remoteOrderId` that let us know that the order is not present in the server yet 🤔.
But if we are planning to support multiple draft orders locally, then it's a blocker.

wzieba: You're right but I can imagine that we would like to support multiple orders drafts locally. That's a good idea to ask this question to the team.

I've tried to check how it works for products creation but I can't create a local-only draft. When I disable network, I just have to discard product. Or am I missing some steps, does local-only draft work for you there?

hichamboushaba: > I've tried to check how it works for products creation but I can't create a local-only draft. When I disable network, I just have to discard product. Or am I missing some steps, does local-only draft work for you there?

On products, we can't even save the current draft locally and leave the screen to edit it later, we need to push it to the server as draft before leaving the screen, so there is no need of saving it to the DB locally only.

> You're right but I can imagine that we would like to support multiple orders drafts locally

Unless it's for better offline support, I don't see a real need for it from the user experience perspective, but I may be wrong 🙂

hichamboushaba: But now that I think about it, even if we don't need multiple drafts now, there may be a need for them in the future, so this is to consider too 🙂

wzieba: > On products, we can't even save the current draft locally and leave the screen to edit it later, we need to push it to the server as draft before leaving the screen, so there is no need of saving it to the DB locally only.

That's a pity! I can imagine that users would like to create some drafts even if they're not connected to the internet in this very moment - there are many scenarios for loosing internet connectivity.

> Unless it's for better offline support, I don't see a real need for it from the user experience perspective, but I may be wrong 🙂

Yes, I thought about better offline support 🙂.

> But now that I think about it, even if we don't need multiple drafts now, there may be a need for them in the future

True, it'd be easier later to add them. But I still think we should support it in the first place - I'll raise a discussion about it on P2!

wzieba: To close a loop: after internal discussion with other team members too, we all agreed that we'll go with composite primary key of (`remoteOrderId` x `localSiteId`) tuple. I'm starting to refactor this PR.

Thanks @hichamboushaba once again!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -1,84 +1,75 @@
...
-    @Column var shippingLines = ""
-
-    @Column var feeLines = ""
-
-    @Column var metaData = ""
 
+@Entity
+data class WCOrderModel(
+    @PrimaryKey(autoGenerate = true) val id: Int = 0,
+    val localSiteId: Int = 0,
```
wzieba, line 14: [According to our discussions](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2165/files#r737336749) this will be changed into composite primary key and local id will be removed, just not in this PR as the change is quite big so I wanted to split it.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
```diff
@@ -5,34 +5,38 @@ import androidx.room.Database
...
     abstract fun ssrDao(): SSRDao
+    abstract fun ordersDao(): OrdersDao
 
     companion object {
         fun buildDb(applicationContext: Context) = Room.databaseBuilder(
                 applicationContext,
                 WCAndroidDatabase::class.java,
                 "wc-android-database"
-        ).fallbackToDestructiveMigration()
+        ).allowMainThreadQueries()
```
wzieba, line 39: This is needed _for now_. After migration, it'll be my priority to refactor every main thread db query to use I/O thread.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.model.WCOrderModel
+import org.wordpress.android.fluxc.model.order.OrderIdSet
+
+typealias AutoGeneratedId = Long
+
+@Dao
+abstract class OrdersDao {
```
wzieba, line 18: Most of the methods here will be marked as `@Deprecated` in next PRs. The preferable approach will be to use reactive approach of `suspend fun` / `Flow` to insert/query data.

For now its rather replacement for `OrderSqlUtils`.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -317,11 +321,14 @@ class WCOrderStore @Inject constructor(
...
+        val statusFilter: String? = null,
+        val canLoadMore: Boolean = false,
+        val causeOfChange: WCOrderAction? = null,
+        private val orderError: OrderError? = null
     ) : OnChanged<OrderError>() {
-        var causeOfChange: WCOrderAction? = null
+        init {
+            super.error = orderError
+        }
     }
```
wzieba, line 332: Personally I like this part pretty much 🙂! From
```
OnOrderChanged(0).also { it.error = result.error }
```

to

```
OnOrderChanged(orderError = result.error)
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -1,84 +1,75 @@
...
-    @Column var lineItems = ""
-
-    @Column var shippingLines = ""
-
-    @Column var feeLines = ""
-
-    @Column var metaData = ""
 
+@Entity
+data class WCOrderModel(
```
wzieba, line 12: I'm wondering if we should keep `WC*` prefix. Woo now will have a separate database so probably we could go with `OrderEntity` - WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -739,65 +737,58 @@ class OrderRestClient @Inject constructor(
     }
 
     private fun orderResponseToOrderModel(response: OrderDto): WCOrderModel {
```
hichamboushaba, line 739: I'm wondering if we shouldn't pass the localSiteId as parameter her too, since from what I saw, we always add a `copy` statement to set it.

hichamboushaba (approved):
Great work @wzieba, I just took a look at the code, I didn't test the app yet, I'll try to do it tomorrow, but I'm approving anyway since this is to be merged in a feature branch.

Feel free to merge, or to wait for someone else from the team to take a look.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
hichamboushaba: IMO we should keep it, even though it's part of the Woo database, the `WC` prefix makes it easier to see which classes belong to Woo from the rest, without having to dig more, but it's just an opinion :)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
wzieba: Good point! I'll cover it in next (3rd) PR, thanks!

wzieba:
Thanks @hichamboushaba ! Merging this one to feature branch

> I didn't test the app yet, I'll try to do it tomorrow

If I may, I'd recommend to test app later, when all feature-branch PRs will be merged and Woo PR will be in its final form (moved from draft state).

PR #2168, opened 2021-10-29
Emit `empty billing email` order error

wzieba:
### Description

This PR will make `OrderUpdateStore` map `rest_invalid_param` to `EMPTY_BILLING_EMAIL` if update request contains empty billing email. 

For motivation and more context please see corresponding Woo PR: https://github.com/woocommerce/woocommerce-android/pull/5127

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
@@ -171,6 +171,10 @@ class OrderUpdateStore @Inject internal constructor(
             updateRemoteOrderPayload,
             initialOrder,
             mapError = { originalOrderError: OrderError? ->
+                /**
+                 * It's *likely* as INVALID_PARAM can be caused by probably other cases too and
+                 * empty billing address email in future releases of WooCommerce will be not relevant.
+                 */
                 val isLikelyEmptyBillingEmailError =
                         updateRemoteOrderPayload.error.type == OrderErrorType.INVALID_PARAM &&
                                 billingAddress.email.isBlank()
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
@@ -35,7 +35,6 @@ class OrderUpdateStore @Inject internal constructor(
 ) {
     suspend fun updateCustomerOrderNote(
         orderLocalId: LocalId,
-            // todo wzieba: Remove site argument. As we have local order id we can get site from database. No need to bother client with that.
         site: SiteModel,
         newCustomerNote: String
     ): Flow<UpdateOrderResult> {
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -158,13 +163,39 @@ class OrderUpdateStore @Inject internal constructor(
         this.billingPhone = newAddress.phone
     }
 
+    private suspend fun FlowCollector<UpdateOrderResult>.emitRemoteUpdateOfUpdateContainingBillingAddress(
```
wzieba: Sorry for this name 😬 do you have other ideas?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -158,13 +163,39 @@ class OrderUpdateStore @Inject internal constructor(
...
+        updateRemoteOrderPayload: RemoteOrderPayload,
+        initialOrder: WCOrderModel,
+        billingAddress: Billing
+    ) = emitRemoteUpdateResultOrRevertOnError(
+            updateRemoteOrderPayload,
+            initialOrder,
+            mapError = { originalOrderError: OrderError? ->
+                val isLikelyEmptyBillingEmailError =
```
wzieba, line 178: It's _likely_ because we can't be 100% sure. `INVALID_PARAM` can be caused by probably other cases too and empty billing address email in future releases of WooCommerce will be not relevant.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
ThomazFB: maybe just `emitRemoteUpdateContainingBillingAddress`? 🤔 Not sure if we need to enforce that we're emitting a `RemoteUpdate` since this is inferred by the generics type defined at the `FlowCollector` declaration.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
ThomazFB: do you think we should also say this as a comment inside the `val` declaration too? I think someone touching this may ask himself why it's `likely` and may feel it's unsafe to modify it.

wzieba: Sure! Comment added in e733b7aca867a5b2b2f4ca4734bb00f513e74b76

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
wzieba: I think we can enforce the `RemoteUpdate` name as `UpdateOrderResult` is also used in local update. I've renamed to `emitRemoteUpdateContainingBillingAddress` - it's better thanks!

ThomazFB (approved):
Thanks for implementing all suggestions, Wojtek! LGTM :shipit:

PR #2172, opened 2021-11-02
Fix NPE and set consumerProguardFiles to propagate rules to client apps

malinajirka:
Review from one reviewer should be enough.

This PR is an attempt at fixing https://github.com/woocommerce/woocommerce-android/issues/4868, which we are not able reproduce. I've left [some comments](https://github.com/woocommerce/woocommerce-android/issues/4868#issuecomment-957273359) on the issue. 

This PR:
1. Sets proguard rules for client/consumer apps
2. Updates proguard rules to include [rules recommended for gson library](https://github.com/google/gson/blob/master/examples/android-proguard-example/proguard.cfg)
3. Explicitly specifies type for generic type passed to `toJson` method

To test:
Test in WCAndroid

### DIFF
--- a/fluxc/build.gradle
+++ b/fluxc/build.gradle
@@ -22,6 +22,7 @@ android {
                         "room.expandProjection":"true"]
             }
         }
+        consumerProguardFiles 'proguard-rules.pro'
     }
     buildTypes {
         release {
--- a/fluxc/proguard-rules.pro
+++ b/fluxc/proguard-rules.pro
@@ -36,6 +36,7 @@
 
 # Application classes that will be serialized/deserialized over Gson
 -keep class org.wordpress.android.fluxc.model.** { <fields>; }
+-keep class org.wordpress.android.fluxc.network.** { <fields>; }
 
 # Prevent proguard from stripping interface information from TypeAdapter, TypeAdapterFactory,
 # JsonSerializer, JsonDeserializer instances (so they can be used in @JsonAdapter)
--- a/fluxc/proguard-rules.pro
+++ b/fluxc/proguard-rules.pro
@@ -21,3 +21,36 @@
 -keep class com.sun.jna.* { *; }
 -keepclassmembers class * extends com.sun.jna.* { public *; }
 ## Encrypted Logging End
+
+##---------------Begin: proguard configuration for Gson  ----------
+# Gson uses generic type information stored in a class file when working with fields. Proguard
+# removes such information by default, so configure it to keep all of it.
+-keepattributes Signature
+
+# For using GSON @Expose annotation
+-keepattributes *Annotation*
+
+# Gson specific classes
+-dontwarn sun.misc.**
+#-keep class com.google.gson.stream.** { *; }
+
+# Application classes that will be serialized/deserialized over Gson
+-keep class org.wordpress.android.fluxc.model.** { <fields>; }
+
+# Prevent proguard from stripping interface information from TypeAdapter, TypeAdapterFactory,
+# JsonSerializer, JsonDeserializer instances (so they can be used in @JsonAdapter)
+-keep class * extends com.google.gson.TypeAdapter
+-keep class * implements com.google.gson.TypeAdapterFactory
+-keep class * implements com.google.gson.JsonSerializer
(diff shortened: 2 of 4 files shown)


fluxc/proguard-rules.pro
```diff
@@ -20,4 +20,37 @@
...
+-keep class * implements com.google.gson.TypeAdapterFactory
+-keep class * implements com.google.gson.JsonSerializer
+-keep class * implements com.google.gson.JsonDeserializer
+
+# Prevent R8 from leaving Data object members always null
+-keepclassmembers,allowobfuscation class * {
+  @com.google.gson.annotations.SerializedName <fields>;
+}
+
+# Retain generic signatures of TypeToken and its subclasses with R8 version 3.0 and higher.
```
malinajirka, line 53: This rule might very theoretically fix https://github.com/woocommerce/woocommerce-android/issues/4868. I can't explain how could proguard be related to the crashes (since they happen just sometime), but the crashes are related to `gson`, `reflection` and `Generics` - same as `TypeTokens`, so there is a chance/hope :P.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequest.kt
```diff
@@ -244,7 +244,7 @@ object JetpackTunnelGsonRequest {
             put("path", "$path&_method=get")
             put("json", "true")
             if (params.isNotEmpty()) {
-                put("query", gson.toJson(params))
+                put("query", gson.toJson(params, object : TypeToken<Map<String, String>>() {}.type))
```
malinajirka, line 247: This is another blind attempt to fix https://github.com/woocommerce/woocommerce-android/issues/4868. The JavaDoc for `toJson` method says `typeOfSrc – The specific genericized type of src. You can obtain this type by using the TypeToken class. For example, to get the type for Collection<Foo>, you should use:
  Type typeOfSrc = new TypeToken<Collection<Foo>>(){}.getType(); `. [The docs also mention](https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.5/com/google/gson/Gson.html#toJson-java.lang.Object-java.lang.reflect.Type-) that for generic types, `toJson(Object, Type)` should be used.

Props to @hichamboushaba for pointing me to this direction 🙇.

fluxc/build.gradle
```diff
@@ -22,6 +22,7 @@ android {
                         "room.expandProjection":"true"]
             }
         }
+        consumerProguardFiles 'proguard-rules.pro'
```
malinajirka, line 25: Props to @wzieba for pointing me to this direction - I completely forgot that `consumerProguardFiles` existed.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequest.kt
hichamboushaba: Thanks for using the right syntax for keeping the generic type 🙏

hichamboushaba (approved):
LGTM, thanks for working on this @malinajirka.

I won't merge, to give the others a chance to review it too.

wzieba (approved):
Looks good to me!

green-nick:
Hi @malinajirka !
Sorry for ping (I'm not sure if this is proper place to ask such questions).
I've found this PR while researching similiar issue with gson (https://github.com/google/gson/issues/1778 and https://github.com/woocommerce/woocommerce-android/issues/4868) and can you tell me, if this fix (providing TypeToken explicitly) realy helped you?
Thanks in advance!

wzieba:
hi @green-nick ! I know @malinajirka is AFK now so let me answer.

I've looked at our crash analytics and after merging this PR we no longer experience this issue. But I can't tell if it's because of providing `TypeToken` or by providing proguard rules.

green-nick:
@wzieba great, thanks!

PR #2174, opened 2021-11-03
[3/4] Identify Order by (`localSiteId` x `remoteOrderId`) tuple

wzieba:
### Description

In sync with the internal decision ( p91TBi-6dF-p2 ) we decided to use tuple of (`localSiteId` x `remoteOrderId`) as identification for locally stored order.

This PR adds a support for that by adding `unique` constraint on that tuple. Leaving autogenerated `id` is dictated only by legacy support and is marked as deprecation error. 

In next PRs (out of scope of migration) I plan to remove all references to autogenerated `id` and remove this column from local db.


wzieba:
~Left TODO:~
- [ ] ~Add 3->4 Room migration~

Migration will be added in a separate PR.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -1,18 +1,36 @@
...
+                unique = true
+        )]
+)
 data class WCOrderModel(
+    @Deprecated(
+            message = "Please do not use local id for order identification. Use tuple of localSiteId x remoteOrderId." +
+                    "See p91TBi-6dF-p2#comment-7013",
+            level = ERROR
+    )
     @PrimaryKey(autoGenerate = true) val id: Int = 0,
```
wzieba, line 29: This `id` is used in many places. I didn't want to remove it yet as it would cause great many changes and the WooCommerce PR is already quite big. I think it's okay to leave it, mark it as deprecation error (so we're sure we don't use it anymore) and add uniqueness index for desired pair: `localSiteId x remoteOrderId`.

plugins/woocommerce/schemas/org.wordpress.android.fluxc.persistence.WCAndroidDatabase/4.json
```diff
@@ -507,7 +507,17 @@
...
+          {
+            "name": "index_OrderEntity_localSiteId_remoteOrderId",
+            "unique": true,
+            "columnNames": [
+              "localSiteId",
+              "remoteOrderId"
+            ],
+            "createSql": "CREATE UNIQUE INDEX IF NOT EXISTS `index_OrderEntity_localSiteId_remoteOrderId` ON `${TABLE_NAME}` (`localSiteId`, `remoteOrderId`)"
+          }
+        ],
```
wzieba, line 520: The key part of this PR.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -1,18 +1,36 @@
...
-    val localSiteId: Int = 0,
-    val remoteOrderId: Long = 0L, // The unique identifier for this order on the server
+    @ColumnInfo(name = "localSiteId")
+    val localSiteId: LocalId,
+    @ColumnInfo(name = "remoteOrderId")
+    val remoteOrderId: RemoteId, // The unique identifier for this order on the server
```
wzieba, line 33: `LocalId` and `RemoteId` are mapped via `LocalIdConverter` and `RemoteIdConverter`.

At some moment I was sceptical to the idea of using those types as we add a dependency to FluxC directly into database model but, at the end of the day, it's just some compile-time safety net of using a correct type of id.

Eventually, if we'll ever need to change this type, nothing bad will happen - under the hood those two are just simple SQLs `INTEGER` types.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -1,18 +1,36 @@
...
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
 import org.wordpress.android.fluxc.model.order.OrderAddress
 import org.wordpress.android.fluxc.model.order.OrderIdentifier
-
-@Entity
+import kotlin.DeprecationLevel.ERROR
+
+@Entity(
+        tableName = "OrderEntity",
```
wzieba, line 17: Despite [suggestion](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2165#discussion_r740254950) to keep `WCOrderModel` name, I decided to propose name change to `OrderEntity`.

My argument is that `*Model` is ambiguous in terms of its usage - is it remote, local or domain model? I think that `Entity` gives instant association to databases so it makes easier to navigate in the codebase.

I didn't change the class name yet as I didn't want to shadow more important changes. Table name change is here to skip the need of having database migration later.

wzieba:
@hichamboushaba @ThomazFB - as this is quite closely tied to Order Creation, can I ask you for review? Thank you in advance!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
hichamboushaba: I agree that the `Entity` suffix is good, but still I feel like the `WC` prefix was good for distincting Woo stuff from regular WordPress stuff, but it's not a blocker 🙂

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
hichamboushaba: Having the index manually means we can't use the syntax perks of Room:
```kotlin
  primaryKeys = ["localSiteId", "remoteOrderId"]
```
Or we'll need at least a migration later on to add it.
But I agree that removing the `id` will require bigger PRs, I'm just mentioning it 🙂

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
```diff
@@ -210,7 +214,7 @@ class WooOrdersFragment : StoreSelectingFragment(), WCAddOrderShipmentTrackingDi
                             note = editText.text.toString()
                         }
                         coroutineScope.launch {
-                            val payload = PostOrderNotePayload(order.id, order.remoteOrderId, site, newNote)
+                            val payload = PostOrderNotePayload(123, order.remoteOrderId.value, site, newNote)
```
hichamboushaba: Why removing the reference for `order.id` here and hardcoding the value?
I understand you deprecated it, but the hardcoded value doesn't make sense to me, so either we keep using `order.id`, or we remove the local Id argument from the `PostOrderNotePayload` completely.

hichamboushaba (approved):
Nice work @wzieba, I left one question about a hardcoded order id, but the rest looks good, so I'm pre-approving.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
wzieba: True. Just to emphasise reason: unfortunately `Room` doesn't allow to have multiple primary keys at the same time 😕  that's why I added manual index.

```
You cannot have multiple primary keys defined in an Entity. If you want to declare a composite primary key, you should use @Entity#primaryKeys and not use @PrimaryKey.
```

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
wzieba: Right, thanks for spotting that! I overlooked this, fixed in 3041decbd8fa4c15e8db81e5488d5ae1a43c7e85

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
wzieba: I see, but I have to admit that I'm not a fan of `WC` prefix - especially as it would go out of sync with currently stored entities in Room like `AddonEntity`, `AddonOptionEntity` or `GlobalAddonGroupEntity` and that `Order` exists only in domain of WooCommerce. 

As it's not a blocker, I'll stick to `OrderEntity` if that's okay!

wzieba:
Thanks @hichamboushaba ! Merging this as I resolved hardcoded issue.

PR #2175, opened 2021-11-04
Remove legacy isCardPresentEligible flag

malinajirka:
The `isCardPresentEligible` feature flag is not used anymore in the apps. The IPP feature is released to all users now. The server always returns `isCardPresentEligible=false` for mobile clients and uses the flag internal for a different purpose.


kidinov (approved):
👍

PR #2180, opened 2021-11-08
Add a column to expose active Jetpack connection plugins in the SiteModel table

hichamboushaba:
For the issue https://github.com/woocommerce/woocommerce-android/issues/5186, we need to pass the list of the currently active jetpack connection plugins of the selected site, this PR just adds a new column to the table, and fills it from the network response.

This doesn't bring any breaking changes, so it can be merged after the review directly.

#### Testing
I didn't add any new parts to the example app for this, as it seems not worth it, please use the database inspector from Android studio to confirm that the column has been added, and has the correct data (after forcing fetching sites from the example app)

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
@@ -133,6 +133,7 @@ public class SiteModel extends Payload<BaseNetworkError> implements Identifiable
 
     @Column private String mActiveModules;
     @Column private boolean mIsPublicizePermanentlyDisabled;
+    @Column private String mActiveJetpackConnectionPlugins;
 
     // Zendesk meta
     @Column private String mZendeskPlan;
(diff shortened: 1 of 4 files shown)


wzieba (approved):
Looks good, `ACTIVE_JETPACK_CONNECTION_PLUGINS` is filled with some data for some sites.

image

One observation (if not spotted earlier): one of the sites hadn't any data despite having `WooCommerce Payments` plugin installed and enabled. I've went to site settings, updated all plugins (`WooCommerce`, `Jetpack`, `WooCommerce Payments` etc.), re-fetched sites and then data appeared in app database. Of course it's API-only thing but just dropping this as it might be useful at some point.

Merging!

hichamboushaba:
Thanks for the review @wzieba, if the site had only `WooCommerce Payments`, then it's a known issue for the current version, as it doesn't sync this column to the cache site, and it stays empty.
But if it had Jetpack before you did your test, then it's just a delay in the sync probably.

PR #2181, opened 2021-11-08
Issue/2173 fetch order shipment providers

### LINKED ISSUE #2173: Woo: Migrate FETCH_ORDER_SHIPMENT_PROVIDERS action to coroutines
The FETCH_ORDER_SHIPMENT_PROVIDERS request in `WCOrderStore` needs to be refactored from using an event bus to coroutines. In the WooCommerce app, this fetch action is called from the `OrderShipmentProvidersRepository`.

0nko:
Resolves #2173.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -764,37 +764,44 @@ class OrderRestClient @Inject constructor(
...
                     }
-                },
-                WPComErrorListener { networkError ->
-                    val providersError = networkErrorToOrderError(networkError)
-                    val payload = FetchOrderShipmentProvidersResponsePayload(providersError, site, order)
-                    dispatcher.dispatch(WCOrderActionBuilder.newFetchedOrderShipmentProvidersAction(payload))
-                },
-                { request: WPComGsonRequest<*> -> add(request) })
-        add(request)
+                } ?: FetchOrderShipmentProvidersResponsePayload(site, order)
```
malinajirka: ❓ I'm wondering if we want to return a successful empty response when the json doesn't contain `data` field. AFAIK it's not expected, should we perhaps return INVALID_RESPONSE or GENERIC_ERROR?

malinajirka (commented):
It looks good overall 👏

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
0nko: Yeah, that makes sense. I'll fix it.

malinajirka (approved):

PR #2184, opened 2021-11-09
Woo/2183 create quick order

### LINKED ISSUE #2183: Woo: Create quick order (FluxC)
In order to create a "[quick order](https://github.com/woocommerce/woocommerce-android/issues/5113)," we need a way to pass just the site ID and the payment amount. Based on [the iOS code](https://github.com/woocommerce/woocommerce-ios/blob/1e4e1e6ed42f5e2da7db679573dc276b5c8b673f/Yosemite/Yosemite/Stores/OrderStore.swift#L250-L264), the only field we need to pass is `fee_lines`.

nbradbury:
Closes #2183 by adding the ability to create a "quick order," which is an empty order with an amount attached. To test, visit Woo > Orders in the example app and tap "Create quick order" at the bottom.

You can test this functionality with [this WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/5211).


ThomazFB (approved):
Working as expected, Order created with the `Pending Payment` status, code looks good too. :shipit:

PR #2185, opened 2021-11-09
Add support for `WCTracker`

wzieba:
Relates to https://github.com/woocommerce/woocommerce-android/issues/5196

### Description
This PR adds an endpoint to send `WCTracker` telemetry. For test instructions please see WooCommerce PR.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/tracker/TrackerStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/tracker/TrackerStore.kt
@@ -1,0 +1,13 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.tracker
+
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class TrackerStore @Inject internal constructor(private val restClient: TrackerRestClient) {
+    suspend fun sendTelemetry(appVersion: String, site: SiteModel): WooPayload<Unit> {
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/tracker/TrackerStore.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class TrackerStore @Inject internal constructor(private val restClient: TrackerRestClient) {
+    suspend fun sendTelemetry(appVersion: String, site: SiteModel): WooPayload<Unit> {
+        return restClient.sendTelemetry(appVersion, site)
```
wzieba, line 11: Worth to mention that iOS implementation doesn't send this request when WooCommerce instance is below supported version (`5.9.0`). They do this in `RequirementsChecker.swift` - we unfortunately don't have similar mechanism out of the box. 

We have something like `DiscoveryWPAPIRestClient` but it doesn't save results in any place I could take this from and adding support to it would exceed scope of this PR.

ThomazFB: Considering your comment, it's safe to assume that when we send the telemetry for a site below 5.9.0, woo core will just ignore the request just like happened in the TC3 scenario you described [here](https://github.com/woocommerce/woocommerce-android/pull/5209)?

ThomazFB (approved):
Code looks good and considering that I've already tested this through https://github.com/woocommerce/woocommerce-android/pull/5209, I'm approving this PR. Thanks for the great work on the `WCTracker`, @wzieba!

PR #2186, opened 2021-11-10
Crash tests when api.properties is not setup

malinajirka:
This PR crashes tests when `api.properties` file is not correctly setup. Original, the app would fail with `The required "client_id" parameter is missing` later and figuring out what is wrong took unnecessary time.

No need to test anything, but if you really wanted, you can
1. Remove content of `api.properties`
2. Run eg `ReleaseStack_WCPayTest`
3. Notice that they fail with `OAUTH_APP_ID or OAUTH_APP_SECRET is empty, check your api.properties` error


kidinov (approved):
Good improvement 👍

PR #2188, opened 2021-11-10
Convert FetchProductReviews into suspendable method

malinajirka:
Merge Instructions:
1. Review this PR
2. [Review PR in WCAndroid](https://github.com/woocommerce/woocommerce-android/pull/5207)
3. Merge this PR
4. Update FluxC hash in WCAndroid
5. Merge WCAndroid PR

This PR refactors fetchProductReviews into a suspendable function - removes usage of the event bus.

Other changes
1. The app now returns an error response when the server returns success with an empty body - [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/fb2de0803654d74bf1c0b5395ae23bd0e944694f/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt#L1009) - instead of silently consuming the event => the main benefit of this approach is that the client app always receives a result and doesn't need to manually implement timeout behavior.
2. Makes JetpackTunnelGsonRequestBuilder and CoroutineEngine non-nullable => I didn't find any reason why they were nullable

To Test:
1. Test fetch reviews for specific product and fetch reviews for site actions in the example app ( or test in WCAndroid ([PR](https://github.com/woocommerce/woocommerce-android/pull/5207)))

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
@@ -1,6 +1,7 @@
 package org.wordpress.android.fluxc.mocked
 
 import com.google.gson.JsonArray
+import kotlinx.coroutines.runBlocking
 import org.greenrobot.eventbus.Subscribe
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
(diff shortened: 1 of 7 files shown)


anitaa1990 (approved):
Looks and works great :shipit: Thank you for working on this @malinajirka!

PR #2189, opened 2021-11-10
Refactor StatsRepository into stateless repository. Mark I: fetchHasOrders

ctarda:

This PR refactors fetchHasOrders into a suspendable function - removes usage of the event bus.

## Changes
* Refactored fetchHasOrders in OrdersRestClient to a suspendible function
* Updated fetchHasOrders in WCOrderStore
* Added a couple of tests
* Removed `FETCHED_HAS_ORDERS` in WCOrderAction, plus changes in cascade motivated by that removal

## How to test
1. Test fetchHasOrders action in the example app ( or test in WCAndroid ([PR](https://github.com/woocommerce/woocommerce-android/pull/5215)))


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -490,8 +487,17 @@ class WCOrderStore @Inject constructor(
         with(payload) { wcOrderRestClient.fetchOrderCount(site, statusFilter) }
     }
 
-    private fun fetchHasOrders(payload: FetchHasOrdersPayload) {
-        with(payload) { wcOrderRestClient.fetchHasOrders(site, statusFilter) }
+    suspend fun fetchHasOrders(site: SiteModel, status: String?): OnOrderChanged {
```
kidinov, line 490: Not your change, but it took me a while to understand why `fetchHasOrders` returns an event `OnOrderChanged` =)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -771,21 +777,21 @@ class WCOrderStore @Inject constructor(
...
-            with(payload) {
-                // set 'rowsAffected' to non-zero if there are orders, otherwise set to zero
-                val rowsAffected = if (payload.hasOrders) 1 else 0
-                OnOrderChanged(rowsAffected, statusFilter)
-            }
-        }.also { it.causeOfChange = WCOrderAction.FETCH_HAS_ORDERS }
-        emitChange(onOrderChanged)
-    }
+//    /**
+//     * This is a response to a request to determine whether any orders matching a filter exist
```
kidinov: Probably should be removed now?

kidinov (commented):
Your changes look very good! Good job 👍

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCOrdersTest.kt
```diff
@@ -366,18 +365,14 @@ class MockedStack_WCOrdersTest : MockedStack_Base() {
...
-        assertTrue(countDownLatch.await(TestUtils.DEFAULT_TIMEOUT_MS.toLong(), MILLISECONDS))
-
-        assertEquals(WCOrderAction.FETCHED_HAS_ORDERS, lastAction!!.type)
-        val payload = lastAction!!.payload as FetchHasOrdersResponsePayload
-        assertNull(payload.error)
-        assertTrue(payload.hasOrders)
-        assertNull(payload.statusFilter)
+        val payload = orderRestClient.fetchHasOrders(siteModel, filterByStatus = null)
+        with(payload) {
+            assertNull(payload.error)
```
malinajirka: 💡 When we use `with(payload)` operator, we don't need to explicitly reference `payload` in the block anymore. So this code could be updated to:
```
with(payload) {
            assertNull(error)
            assertTrue(hasOrders)
            assertNull(statusFilter)
        }
```
Another option is to remove the `with(payload)` block. Wdyt?

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
```diff
@@ -136,8 +134,11 @@ class WooOrdersFragment : StoreSelectingFragment(), WCAddOrderShipmentTrackingDi
 
         fetch_has_orders.setOnClickListener {
             selectedSite?.let {
-                val payload = FetchHasOrdersPayload(it)
-                dispatcher.dispatch(WCOrderActionBuilder.newFetchHasOrdersAction(payload))
+                coroutineScope.launch {
+                    wcOrderStore.fetchHasOrders(it, null).takeUnless { it.isError }?.let {
+                        prependToLog("Has orders ${it.rowsAffected != 0}")
+                    }
```
malinajirka: 💡 Originally, the log printed `Has orders false` in case of an error. The current solution doesn't print anything, we might want to consider adding `} ?: prependToLog("Fetching hasOrders failed.")`. Wdyt?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -358,36 +358,48 @@ class OrderRestClient @Inject constructor(
...
+                Array<OrderDto>::class.java
+        )
+
+        return when (response) {
+            is JetpackSuccess -> {
+                response.data?.let {
+                    FetchHasOrdersResponsePayload(
+                            site,
+                            filterByStatus,
+                            true
```
malinajirka: ⚠️ It seems that originally the orders in the response were converted/mapped into OrderModel and counted -> `hasOrders` was `true` only when there was at least one order in the response. It seems that currently, we always return `true` in case of a successful response from the server. Is that an intentional change?

If the idea was that we now have the 
```OrderError(type = GENERIC_ERROR, message = "Success response with empty data")```
case, so we don't need to count them anymore, I think it might not cover the following:

```
HTTP 200 { "data" : [] }
```
Is a valid response but `hasOrder` should be `false`.

```
HTTP 200 { }
```
Is an invalid response since `data` key is not present and we are correctly returning `GENERIC_ERROR`.


I'm thinking we might want to update the code to
```
FetchHasOrdersResponsePayload(
                            site,
                            filterByStatus,
                            it.count() > 0
                    )
```
I'm not sure why we were converting/mapping them to OrderModel before, it seems unnecessary. What do you think about all this?

P.S. Before changing this, I'd consider adding two more tests in `MockedStack_WCOrdersTest.kt`. One for `wc-has-orders-response-success-no-orders.json` and another one for `wc-has-orders-response-success-missing-data-key.json`, just to be sure it works as expected.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
malinajirka: Yeah, this is a bit misleading, I agree. However, I would suggest keeping the return type as is in this PR - that's what we did in all the order related PRs.

malinajirka (commented):
Really great job @ctarda! You got oriented in the Android codebase very quickly 👏 I've left a few comments, let me know what you think :).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
ctarda: This one is embarrassing. I guess I hardcoded it to true to be able to make progress and then I forgot about it. Good catch, thanks!

ctarda:
Thanks a lot for the feedback @kidinov @malinajirka 

I pushed a few more commits that, unless I've missed something, I think implement all your feedback. I have also updated the PR description and submitted the PR formally

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
malinajirka: It happens even to the best :)

malinajirka (approved):
Thanks @ctarda! It all looks good to me. I'm not merging the PR since we always try to keep FluxC's develop compatible with WP/WCAndroid's develop. We can merge this PR when a PR in WCAndroid gets approved.

PR #2190, opened 2021-11-12
Dashboard Cards: Fetch Dashboard Cards

ParaskP7:

This PR:
- Adds the `/sites/$site/dashboard/cards/` API endpoint.
- Updates the `CardsRestClient` to handle this API endpoint.
- Updates the `CardsStore` to convert the API response into a model.

-----

Note:

- Storing and getting the `CardsModel` data to and from the db via the `CardsStore` will be done in an upcoming PR.
- At this point, the `CardsModel` might also change depending on the above to-be implemented persistence solution.

-----

To test:

- There is actually nothing to test in this PR, just review the base `CardsRestClient` and `CardsStore` classes and make sure that the accompanying tests are working as expected.
- When [WordPress-Android#15497](https://github.com/wordpress-mobile/WordPress-Android/issues/15497) will get deployed and then the `ReleaseStack_CardsTest` test suite gets implemented, then triggering this connected test suite will be another way of verifying that this solution works as expected.


zwarm (approved):
@ParaskP7 - All tests are working as expected. 
Is there a companion PR for WP to use this branch? I won't merge until I hear back.

Thanks 🎸

ParaskP7:
👋 @zwarm !

Thank you for this review! 🙏 

> @ParaskP7 - All tests are working as expected.
Is there a companion PR for WP to use this branch? I won't merge until I hear back.

Actually no, there is no companion PR for WP to use this branch. Were you expecting one? Should I create one?

The way I was planning on this work is to first create and merge the skeleton classes, then create and merge the network functionality, then do the same for the persistence functionality and only afterward to create the companion PR for WP along with maybe one more PR on the FluxC side to implement any adjustments that might become necessary for the FluxC solution to become totally complete and work as expected with the WP solution.

zwarm:
> Actually no, there is no companion PR for WP to use this branch. Were you expecting one? Should I create one?

Doubling checking, that's why I asked. I can merge now. Thanks. :)

PR #2192, opened 2021-11-17
Migrate fetchSingleProductReview to suspendable function

malinajirka:
Merge Instructions:
1. Review this PR
2. Review [PR in WCAndroid](https://github.com/woocommerce/woocommerce-android/pull/5259)
3. Remove "Not ready for merge" label and merge this PR
4. Update FluxC hash in WCAndroid
5. Merge WCAndroid PR

This PR refactors fetchSingleProductReview into a suspendable function - removes usage of the event bus.

Other changes
1. The app now returns an error response when the server returns success with an empty body instead of silently consuming the event => the main benefit of this approach is that the client app always receives a result and doesn't need to manually implement timeout behavior.

To Test:
1. Test fetch product review by id in the example app ( or test in [WCAndroid](https://github.com/woocommerce/woocommerce-android/pull/5259))


0nko (approved):
LGTM 👍

PR #2194, opened 2021-11-23
Refactor StatsRepository into stateless repository. Mark II: fetchVisitorStats

ctarda:

This PR refactors fetchVisitorStats into a suspendable function - removes usage of the event bus.

## Changes
* Refactored fetchNewVisitorStats in OrderStatsRestClient to a suspendible function
* Updated fetchNewVisitorStats in WCStatsStore
* Updated tests

## How to test
1. Test fetchNewVisitorStats action in the example app ( or test in WCAndroid ([PR](https://github.com/woocommerce/woocommerce-android/pull/5323)))


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -315,42 +319,47 @@ class OrderStatsRestClient @Inject constructor(
...
-                            }
-                            val payload = FetchNewVisitorStatsResponsePayload(site, granularity, model)
-                            mDispatcher.dispatch(WCStatsActionBuilder.newFetchedNewVisitorStatsAction(payload))
-                        },
-                        { networkError ->
-                            val orderError = networkErrorToOrderError(networkError)
-                            val payload = FetchNewVisitorStatsResponsePayload(orderError, site, granularity)
-                            mDispatcher.dispatch(WCStatsActionBuilder.newFetchedNewVisitorStatsAction(payload))
-                        })
+        println("URL $url")
```
kidinov: Probably worth removing?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -315,42 +319,47 @@ class OrderStatsRestClient @Inject constructor(
...
+        val response = wpComGsonRequestBuilder.syncGetRequest(
+                this,
+                url,
+                params,
+                VisitorStatsApiResponse::class.java
+        )
 
-        request.enableCaching(BaseRequest.DEFAULT_CACHE_LIFETIME)
-        if (force) request.setShouldForceUpdate()
```
kidinov, line 351: This is gone in the new code and param `force` is unused. From what I see we actually pass sometimes true sometimes false here, so apparently we should replicate this functionality in the new code

plugins/woocommerce/src/main/java/org/wordpress/android/fluxc/action/WCStatsAction.java
```diff
@@ -51,6 +51,7 @@
     FETCHED_VISITOR_STATS,
 
     @Action(payloadType = FetchNewVisitorStatsResponsePayload.class)
+    @Deprecated // Use suspendable WCStatsStore.fetchNewVisitorStats(..) directly.
```
kidinov: If anyone will fire this action it won't work now anyway, so maybe to remove this now and have an error during compilation than during runtime?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
```diff
@@ -315,42 +319,47 @@ class OrderStatsRestClient @Inject constructor(
...
+                        endDate?.let { this.endDate = it }
+                        startDate?.let {
+                            this.startDate = startDate
+                            this.isCustomField = true
+                        }
+                    }
 
-        add(request)
+                    FetchNewVisitorStatsResponsePayload(site, granularity, model)
+                } ?: FetchNewVisitorStatsResponsePayload(site, granularity)
```
kidinov: I think this elvis operator is not needed here

```
?: FetchNewVisitorStatsResponsePayload(site, granularity)
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCStatsStore.kt
```diff
@@ -494,16 +493,35 @@ class WCStatsStore @Inject constructor(
...
-                apiUnit,
-                payload.granularity,
-                getFormattedDateByOrderStatsApiUnit(payload.site, apiUnit, endDate),
-                quantity,
-                payload.forced,
-                startDate,
-                endDate
-        )
+        return coroutineEngine.withDefaultContext(T.API, this, "fetchNewVisitorStats") {
+            val result = wcOrderStatsClient.fetchNewVisitorStats(
```
kidinov, line 497: Maybe would be a good idea to use named parms here, so you don't pass `startDate` as `endDate` or something like that

```
            val result = wcOrderStatsClient.fetchNewVisitorStats(
                    site = payload.site,
                    unit = apiUnit,
                    granularity = payload.granularity,
                    date = getFormattedDateByOrderStatsApiUnit(payload.site, apiUnit, endDate),
                    quantity = quantity,
                    force = payload.forced,
                    startDate = startDate,
                    endDate = endDate
            )
```

kidinov (commented):
Good job! Thank you for this improvement! I left a couple of comments - the only one with "force" update is probably important

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt
ctarda: lol. Elvis. Now I can not unsee that.

malinajirka: ~I believe we should return an error not a successful response when the backend returns http 200 with an empty `data`. That's what we do for the other endpoints at least - example~ https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/a32bc8e71f7991a9a9c33c57573277fd59cff9d0/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt#L308. ~Wdyt?~

malinajirka: Ohh, we are using [WPComGsonRequestBuilder](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/9876e32b73819d2a091881ca218be9d9b53a487c/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComGsonRequestBuilder.kt#L112) here which has non-nullable `data` field, as oppose to [JetpackTunnelGsonRequestBuilder](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/48069374b1f23056af27bacfbd373267d2515ff6/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpacktunnel/JetpackTunnelGsonRequestBuilder.kt#L179) which has the same field marked as nullable. TIL

ctarda:
Thanks for the review @kidinov  @malinajirka . I think I implemented everything you suggested, so this should be ready for another round of review.

ctarda:
🤔  There is one failing test: `testCancelImageUpload()`

However, the test passes locally, on my machine. In fact all the tests in `MockedStack_MediaTest` pass on mu machine. 

What is the best way forward?

kidinov (approved):

kidinov:
@ctarda apparently this flaky =(. I restarted and it's ok now

PR #2196, opened 2021-11-25
Refactor StatsRepository into stateless repository. Mark III: fetchRevenueStats

ctarda:

This PR refactors fetchVisitorStats into a suspendable function - removes usage of the event bus.

## Changes
* Refactored fetchNewVisitorStats in OrderStatsRestClient to a suspendible function
* Updated WooRevenueStatsFragment
* Updated fetchRevenueStats in WCStatsStore
* Updated tests

## How to test
1. Test fetchNewVisitorStats action in the example app ( or test in WCAndroid ([PR](https://github.com/woocommerce/woocommerce-android/pull/5351)))


plugins/woocommerce/src/main/java/org/wordpress/android/fluxc/action/WCStatsAction.java
```diff
@@ -21,9 +20,6 @@
     @Action(payloadType = FetchOrderStatsPayload.class)
     FETCH_ORDER_STATS,
 
-    @Action(payloadType = FetchRevenueStatsPayload.class)
-    FETCH_REVENUE_STATS,
```
malinajirka, line 25: 🔍 Nitpick: I believe we can remove even `FETCHED_REVENUE_STATS` since it's not used anymore, wdyt?

malinajirka (approved):
Really amazing job @ctarda! 👏 I've left one minor comment, feel free to merge the PR no matter if you implement the suggestion or not.

PR #2197, opened 2021-11-26
Dashboard Cards: Store Dashboard Cards

ParaskP7:

This PR:
- Updates the post card response (based on the newly agreed structure).
- Updates the post card model.
- Adds the cards database access object (dao).
- Adds cards dao to database and Increments version.
- Stores dashboard cards into the database.
- Retrieves dashboard cards from the database.

-----

To test:

- There is actually nothing to test in this PR, just review again the updated `CardsRestClient` and `CardsStore` classes and make sure that the accompanying tests are working as expected.
- When [WordPress-Android#15497](https://github.com/wordpress-mobile/WordPress-Android/issues/15497) will get deployed and then the `ReleaseStack_CardsTest` test suite gets implemented, then triggering this connected test suite will be another way of verifying that this solution works as expected.

-----

To manual test:

- If you really want to test the solution:
  - I can provide you with a patch that connects the dashboard cards store with its source. Let me know!
  - Or, you can wait for the accompanying `WordPress-Android` PR to be created, which will implement the connection between the dashboard cards store and its source.
- However, mind the fact that in both the above cases you would need to:
  - Either apply the [My Site Dashboard: Cards Endpoint](https://code.a8c.com/D69742) differential on your sandbox, proxy your device on it. Let me know if you need any help with that!
  - Or, you can wait for the [My Site Dashboard: Cards Endpoint](https://code.a8c.com/D69742) differential to be deployed. Then, you would be able to test the solution as usual.

-----

Merge instructions:
- [x] ~~Wait for the new API endpoint to be deployed (see [D69742](https://code.a8c.com/D69742)).~~
- [x] Wait for the accompanying [WordPress-Android#15637](https://github.com/wordpress-mobile/WordPress-Android/pull/15637) PR to be reviewed (and possibly approved).
- [x] Remove the `[DO NOT MERGE]` from the title.
- [x] Remove the `[PR] Not Ready For Merge` label.
- [x] Merge the PR.


fluxc/src/main/java/org/wordpress/android/fluxc/model/dashboard/CardModel.kt
```diff
@@ -0,0 +1,27 @@
...
+    ) {
+        POSTS(PostsCardModel::class.java)
+    }
+
+    data class PostsCardModel(
+        val hasPublished: Boolean,
+        val draft: List<PostCardModel>,
+        val scheduled: List<PostCardModel>
+    ) : CardModel(Type.POSTS) {
+        data class PostCardModel(
```
zwarm, line 19: ❓ two questions on naming ...
(1) If `id` actually represents `postId` should it be called `postId`? I know it's not the local id.
(2) A similar question on the `date` field. It is being transformed from UTC to local format in `CardsRestClient` using `CardUtils`. Should the name be reflective of that fact - `displayDate` so that it's clear that no further transformation is needed?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/dashboard/CardsRestClient.kt
```diff
@@ -59,26 +59,26 @@ class CardsRestClient @Inject constructor(
...
     ) {
-        fun toPost() = CardsModel.PostsModel.PostModel(
+        fun toPost() = PostCardModel(
                 id = id,
                 title = title,
                 content = content,
-                date = date,
-                featuredImage = featuredImage
+                featuredImage = featuredImage,
+                date = CardsUtils.fromDate(date)
```
zwarm, line 81: It feels like we are making a UI decision on how we'd like to display this date in the API layer, rather than in the UI layer. wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/store/dashboard/CardsStore.kt
```diff
@@ -57,7 +61,7 @@ class CardsStore @Inject constructor(
 
     /* ACTIONS */
 
-    data class OnCardsFetched<T>(
+    data class CardsResult<T>(
```
zwarm, line 64: Like `CardsResult` so much better than just` OnCardsFetched`. it's really clear to see the purpose now. ❤️

zwarm (approved):
Looking good @ParaskP7 . Walking us through your plan-of-attack (last week), was a big help while doing this review. 🙏   I left a couple of questions, but nothing that stops me from giving this the thumbs up, but I do want you to have the opportunity to think about the ❓ . 

I would love to test this all the way through, but I always get stumped on applying the differential in my sandbox and then proxying my device. And I'm not sure if we really have time for your suggestion.

>  Let me know if you need any help with that!

Let us see what tomorrow brings. Other than that, 🙇  nice job!

fluxc/src/main/java/org/wordpress/android/fluxc/store/dashboard/CardsStore.kt
ParaskP7: Indeed, and many thanks to you and @ashiagr for the suggestion and for talking me into it! ❤️

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/dashboard/CardsRestClient.kt
ParaskP7: 👋 @zwarm !

Thank for the comment and hmmm... Why do you think that? The only thing this does is converting the `String` object into a `Date` object, right from the start of it, right before creating the model, in order for the model to always contain and be worked on based on a `Date`, instead of a `String`. Also, the API layer knows the date format of the response and as such converts it based on that, thus reducing that work on the UI layer, who will now not care about doing that or knowing about that.

On the UI layer we would anyway need to parse the `Date` and convert it to the actual display format that we want, in this case something like: `Tomorrow at 3:04 PM` or `Oct 28, 2021 at 1:04 PM`

In both cases, the above UI result, IMHO, would be much easier achieved with a `Date` object, rather that a raw `String` object, don't you think? 🤔

fluxc/src/main/java/org/wordpress/android/fluxc/model/dashboard/CardModel.kt
ParaskP7: 👋 @zwarm !

(1) Good question, I was thinking of that as well. I went with a plain `id` because we are already working with a `PostCardModel`, thus going with a `postId` for a `Post` data class is kind of over-information IMHO. Now, in case this data class were to contain two different ids, then it would be necessary to name them both with something that would differentiate them. Wdyt?
(2a) What do you mean to `local format`? The `String` UTC is being transformed to a `Date` UTC instance, am I missing something?
(2b) About naming it a `displayDate`, I am not sure if that is related to the comment you posted [further below](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2197#discussion_r758627253). However, I think I am in favor of keeping it simple and at the same time generic, letting the UI decide how it is going to be using this `date`, either just for displaying purposes, or otherwise. Also, I think keeping it `date` would preserve the API information, since on the API response for the post we can chose between two dates, the `date` and `modified` dates. Wdyt?

ParaskP7:
👋 @zwarm !

> Looking good @ParaskP7 . Walking us through your plan-of-attack (last week), was a big help while doing this review. 

Thanks so much for going through the review! 🥇 

> 🙏 I left a couple of questions, but nothing that stops me from giving this the thumbs up, but I do want you to have the opportunity to think about the ❓ .

Thank you for sharing your thoughts, I have answered them to the best of my abilities. Let me know and we can definitely fine-tune all that as we arrive into an agreement. 🙇 

> I would love to test this all the way through, but I always get stumped on applying the differential in my sandbox and then proxying my device. And I'm not sure if we really have time for your suggestion.

Please do let me know and we can sit together to setup the sandbox and proxy for you. It really is not that big of a deal, I was also very afraid of that, but now that I am using it every day I know that it is just a matter of few minutes. And then, when this gets configured, it is a matter of few seconds to switch on the sandbox and connect the proxy. 💯  PS: I also think that this is a superpower, thus, the quicker you get it working for you, the better!

fluxc/src/main/java/org/wordpress/android/fluxc/model/dashboard/CardModel.kt
zwarm: > Good question, I was thinking of that as well. I went with a plain id because we are already working with a PostCardModel, thus going with a postId for a Post data class is kind of over-information IMHO. Now, in case this data class were to contain two different ids, then it would be necessary to name them both with something that would differentiate them. Wdyt?


I looked at it and thought "gee, is that `PostId` or is that an identity `id`". Obviously I have context and made the assumption that it's postId because it's in the `PostModel`, but because the serialized name is also `id`, I needed dig deeper into the PHP code to really be sure that it represented `postId`.  Using a generic `id` that can refer to many different things, doesn't convey much detail to the next developer, even in the context of this model. I'm not asking for a change, but I did want to bring it up.

PR #2199, opened 2021-11-29
Rename in person payments classes

malinajirka:
This PR renames classes related to In Person Payments. There are two reasons for this change
1. Some of the classes were not following naming-conventions
2. More importantly, In Person Payments are currently supported only through WCPay extension. However, we plan to add support for Stripe Extension. Using "WCPay" prefix was a bit misleading since the classes will be used for both WCPay and Stripe Extension.

To Test:
Smoke test WCAndroid app - refactoring tools were used, so if the PR builds it should be safe.

P.S. If you have proposals for better names, please share them :).
P.S.2. Sorry for the number of commits, I kept changing my mind. The final file comparison should be easy to follow, so hopefully you won't need to review commit-by-commit.


AnirudhBhat:
Thanks for working on this @malinajirka I've tested the changes and it works fine. The naming changes also sound good to me.

AnirudhBhat (approved):

PR #2202, opened 2021-12-01
Woo: create order

hichamboushaba:
This PR adds the ability to create orders from the app, we support currently just: status, a list of products, and customer addresses (shipping and billing).

I created a separate class for the request [CreateOrderRequest](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/218d9523bd33cf04b5e0f151354cfe2e32aa8ab2/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/CreateOrderRequest.kt#L3) because:
1. WCOrderModel has a ton of stuff that we won't need for the creation, and we need to keep passing null arguments for it.
2. WCOrderModel doesn't play well with the OrderAddress type.
3. The new class would make it more flexible in organizing the classes of different sections later on (for example grouping the payment options in the same class).

The new class meant that I had to move the subclasses of WCOrderModel to their own files (for a better organization), sorry @wzieba for this, I know it will create conflicts with your Room changes.

### Testing
1. Open the example app.
2. Navigate to Woo then Orders.
3. Select a site.
4. Click on create order.
5. Enter the order's information.
6. Confirm that the created order's id is printed.
7. Check the order in wp-admin, and confirm that it's matching what you entered.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
@@ -4,127 +4,163 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.lifecycle.lifecycleScope
 import dagger.android.support.DaggerFragment
 import kotlinx.android.synthetic.main.fragment_address_edit_dialog.*
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.collect
-import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.launch
 import org.wordpress.android.fluxc.example.databinding.FragmentAddressEditDialogBinding
 import org.wordpress.android.fluxc.example.prependToLog
-import org.wordpress.android.fluxc.example.ui.orders.AddressEditDialogFragment.EditTypeState.BILLING
-import org.wordpress.android.fluxc.example.ui.orders.AddressEditDialogFragment.EditTypeState.SHIPPING
+import org.wordpress.android.fluxc.example.ui.orders.AddressEditDialogFragment.AddressType.BILLING
+import org.wordpress.android.fluxc.example.ui.orders.AddressEditDialogFragment.AddressType.SHIPPING
+import org.wordpress.android.fluxc.example.ui.orders.AddressEditDialogFragment.Mode.Add
+import org.wordpress.android.fluxc.example.ui.orders.AddressEditDialogFragment.Mode.Edit
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
 import org.wordpress.android.fluxc.model.WCOrderModel
+import org.wordpress.android.fluxc.model.order.OrderAddress
 import org.wordpress.android.fluxc.model.order.OrderAddress.Billing
 import org.wordpress.android.fluxc.model.order.OrderAddress.Shipping
 import org.wordpress.android.fluxc.store.OrderUpdateStore
 import javax.inject.Inject
 
 class AddressEditDialogFragment : DaggerFragment() {
-    @Inject lateinit var orderUpdateStore: OrderUpdateStore
+    companion object {
+        @JvmStatic
+        fun newInstanceForEditing(order: WCOrderModel): AddressEditDialogFragment =
+                AddressEditDialogFragment().apply {
+                    this.selectedOrder = order
+                    this.mode = Edit
+                }
 
-    enum class EditTypeState {
-        SHIPPING, BILLING
+        @JvmStatic
+        fun newInstanceForCreation(addressType: AddressType, listener: (OrderAddress) -> Unit): AddressEditDialogFragment =
+                AddressEditDialogFragment().apply {
+                    this.selectedOrder = WCOrderModel()
+                    this.mode = Add
+                    this.addressListener = listener
+                    this.currentAddressType.value = addressType
+                }
     }
 
-    private var currentAddressType = MutableStateFlow(SHIPPING)
-    private lateinit var binding: FragmentAddressEditDialogBinding
+    @Inject lateinit var orderUpdateStore: OrderUpdateStore
 
-    var selectedOrder = MutableStateFlow<WCOrderModel?>(null)
+    enum class AddressType {
+        SHIPPING, BILLING
+    }
+    enum class Mode {
+        Edit, Add
+    }
+
+    private lateinit var mode: Mode
+    private var currentAddressType = MutableStateFlow(SHIPPING)
+
+    private lateinit var selectedOrder: WCOrderModel
     var originalBillingEmail = ""
+
+    private var addressListener: ((OrderAddress) -> Unit)? = null
 
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
         savedInstanceState: Bundle?
-    ): View = FragmentAddressEditDialogBinding.inflate(inflater).apply {
-        binding = this
-        addressTypeSwitch.setOnCheckedChangeListener { _, isChecked ->
-            currentAddressType.value = if (isChecked) BILLING else SHIPPING
-        }
-    }.root
+    ): View = FragmentAddressEditDialogBinding.inflate(inflater).root
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        val binding = FragmentAddressEditDialogBinding.bind(view)
 
-        CoroutineScope(Dispatchers.Main).launch {
-            currentAddressType.combine(selectedOrder) { addressType, order ->
-                originalBillingEmail = order?.billingEmail.orEmpty()
+        if (mode == Add) {
+            binding.addressTypeSwitch.visibility = View.GONE
+            binding.sendBothAddressesUpdate.visibility = View.GONE
+            binding.sendUpdate.text = "Save"
+            binding.addressTypeTitle.visibility = View.VISIBLE
+            binding.addressTypeTitle.text = if (currentAddressType.value == BILLING) {
+                "Enter a billing address"
+            } else {
+                "Enter a shipping address"
+            }
+        }
+
+        binding.addressTypeSwitch.setOnCheckedChangeListener { _, isChecked ->
+            currentAddressType.value = if (isChecked) BILLING else SHIPPING
+        }
+
+        viewLifecycleOwner.lifecycleScope.launch {
+            currentAddressType.collect { addressType ->
+                originalBillingEmail = selectedOrder.billingEmail
                 when (addressType) {
                     SHIPPING -> {
-                        binding.addressTypeSwitch.text = "Edit shipping address for order ${order?.remoteOrderId}"
-                        binding.firstName.setText(order?.shippingFirstName)
-                        binding.lastName.setText(order?.shippingLastName)
-                        binding.company.setText(order?.shippingCompany)
-                        binding.address1.setText(order?.shippingAddress1)
-                        binding.address2.setText(order?.shippingAddress2)
-                        binding.city.setText(order?.shippingCity)
-                        binding.state.setText(order?.shippingState)
-                        binding.postcode.setText(order?.shippingPostcode)
-                        binding.phone.setText(order?.shippingPhone)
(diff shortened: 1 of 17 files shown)


example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
```diff
@@ -4,127 +4,163 @@ import android.os.Bundle
...
+            } else {
+                "Enter a shipping address"
+            }
+        }
+
+        binding.addressTypeSwitch.setOnCheckedChangeListener { _, isChecked ->
+            currentAddressType.value = if (isChecked) BILLING else SHIPPING
+        }
+
+        viewLifecycleOwner.lifecycleScope.launch {
```
hichamboushaba, line 92: I switched to the lifecycleScope provided by androidx, since we weren't properly cancelling the created CoroutineScope previously.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
```diff
@@ -4,127 +4,163 @@ import android.os.Bundle
...
+            val newAddress = when (currentAddressType.value) {
+                SHIPPING -> generateShippingAddressModel(binding)
+                BILLING -> generateBillingAddressModel(binding)
+            }
 
-                    orderUpdateStore.updateOrderAddress(
-                            orderLocalId = LocalId(order.id),
-                            newAddress = newAddress
-                    ).collect {
-                        CoroutineScope(Dispatchers.Main).launch {
```
hichamboushaba, line 102: The flow collection happens on the scope's default dispatcher (check the "Collector section" of this [article](https://elizarov.medium.com/execution-context-of-kotlin-flows-b8c151c9309b)), so no need to do this here.

And as a reminder to all of us, if we had to switch context here, we should've done it using [`withContext`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-context.html), and not by instantiating a new scope, since this will break the cancellation mechanism, and will be leaking stuff 🙂.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
```diff
@@ -4,127 +4,165 @@ import android.os.Bundle
...
+import org.wordpress.android.fluxc.model.order.OrderAddress
 import org.wordpress.android.fluxc.model.order.OrderAddress.Billing
 import org.wordpress.android.fluxc.model.order.OrderAddress.Shipping
 import org.wordpress.android.fluxc.store.OrderUpdateStore
 import javax.inject.Inject
 
 class AddressEditDialogFragment : DaggerFragment() {
+    companion object {
+        // These instantiations won't work with screen rotation or process death scenarios.
+        // They are just for the sake of simplification of the example
```
hichamboushaba, line 30: I added this comment to make it clearer to anyone who reads the code later that we are not properly passing the arguments to the Fragment just for the sake of the example.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDto.kt
```diff
@@ -70,3 +72,68 @@ class OrderDto : Response {
     val total_tax: String? = null
     val meta_data: JsonElement? = null
 }
+
+fun OrderDto.toDomainModel(): WCOrderModel {
```
hichamboushaba, line 76: I moved the mapping function to be an extension, instead of a private function inside the RestClient, this is because the mapping is generally a responsibility of the Store (a Repository), and not the RestClient.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDto.kt
```diff
@@ -70,3 +72,69 @@ class OrderDto : Response {
     val total_tax: String? = null
     val meta_data: JsonElement? = null
 }
+
+fun OrderDto.toDomainModel(localSiteId: Int): WCOrderModel {
```
hichamboushaba, line 76: I moved the mapping function to be an extension, instead of a private function inside the RestClient, this is because the mapping is generally a responsibility of the Store (a Repository), and not the RestClient.

I kept the other private function `orderResponseToOrderModel` to limit the impact on your work @wzieba, since this extension is being used only for the order creation only, which mean fewer conflicts with your other work, but let me know if it's OK to convert the other calls, I can do it.

hichamboushaba:
@ThomazFB @wzieba I'm marking this as ready for review to get your feedback as early as possible, please don't hesitate on giving any remarks regarding the choices I made 🙂.
I'll continue testing the functionality, and then I'll start working on the WCAndroid's PR to adapt to the breaking changes introduced by this PR.
Thanks in advance 🙏

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
```diff
@@ -392,6 +403,47 @@ class WooOrdersFragment : StoreSelectingFragment(), WCAddOrderShipmentTrackingDi
...
             }
         }
+
+        create_order.setOnClickListener {
+            selectedSite?.let { site ->
+                lifecycleScope.launch {
+                    val products = showSingleLineDialog(
+                            activity = requireActivity(),
+                            message = "Please type a comma separated list of product IDs",
+                            isNumeric = false
```
wzieba, line 413: I can see why it's `numeric = false` - otherwise, we can't enter `,` even if it's visible on keyboard, right?

I found that we could resolve this [by adding `digits`](https://nkaushik.com/android/android-edittext-comma-number/) but we would have to do this [in code.](https://stackoverflow.com/questions/7300490/set-edittext-digits-programmatically).

I know it's super minor but on the other hand, it's only one line of code:

```
    if (isNumeric) {
        editText.inputType = InputType.TYPE_CLASS_NUMBER
        editText.keyListener = DigitsKeyListener.getInstance("0123456789,")
    }
```

in `AlertDialogUtils#40` seems to work fine, WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDto.kt
wzieba: Thank you for thinking about it! It's good idea with keeping `orderResponseToOrderModel` (maybe deprecate it?) until migration to Room will be merged - less conflicts, the better.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
wzieba: Thank you for the additional info on `withContext`! I missed that completely, TIL.

wzieba (approved):
Looks and works great, thanks!

One thing to remember about testing: billing address has to have mail address if testing on not the newest WooCommerce instance (I forgot exact version needed).

> The new class meant that I had to move the subclasses of WCOrderModel to their own files (for a better organization), sorry @wzieba for this, I know it will create conflicts with your Room changes.

It works for me - good idea with moving subclasses to own files.

Added some minor comments - nothing blocking.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
hichamboushaba: I'm afraid this will break the logic in other areas in the app, since the this AlertDialog is used in multiple places.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDto.kt
hichamboushaba: Good idea, I'll deprecate it 🙂

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/CreateOrderRequest.kt
```diff
@@ -0,0 +1,10 @@
+package org.wordpress.android.fluxc.model.order
+
+import org.wordpress.android.fluxc.model.WCOrderStatusModel
+
+data class CreateOrderRequest(
+    val status: WCOrderStatusModel,
```
hichamboushaba, line 6: I don't know if using WCOrderStatusModel here is a good idea, or just stick to a String like what we do in `updateOrderStatus`?

The `Order` class in the presentation layer that will hold the draft for our case will hold an `OrderStatus` object, so if we keep the WCOrderStatusModel, we'll need to make a call to `getOrderStatusForSiteAndKey` before making the call to pass it to the request, but I'm not sure it's worth it, WDYT @wzieba @ThomazFB ?

wzieba: I like passing here `String` and not `WCOrderStatusModel`. It'd be even better to me personally if we could use `typealias` or `inline class` (eventually another name for the variable) to explicitly determine that we're asking for "status key".

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
```diff
@@ -4,127 +4,167 @@ import android.os.Bundle
...
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
 import org.wordpress.android.fluxc.model.WCOrderModel
+import org.wordpress.android.fluxc.model.order.OrderAddress
 import org.wordpress.android.fluxc.model.order.OrderAddress.Billing
 import org.wordpress.android.fluxc.model.order.OrderAddress.Shipping
 import org.wordpress.android.fluxc.store.OrderUpdateStore
 import javax.inject.Inject
 
 class AddressEditDialogFragment : DaggerFragment() {
+    companion object {
```
wzieba, line 28: Is it okay to move `companion object` back to the bottom? It's in sync with official Kotlin convention: https://kotlinlang.org/docs/coding-conventions.html#class-layout

hichamboushaba: AFAIK not everyone follows the official convention for this, for Android the rule is: if it's important then put it to the top, otherwise bottom (https://github.com/android/kotlin-guides/issues/11), for WCAndroid and even if it's not documented, it's been always the top (we had the same in my previous employer, but it was specifically documented).
I've moved it to the top to stay consistent with our codebase, and if we follow the Android guideline, it should stay at the top, since it's a factory object with the `newInstance` functions, but it's s a very small thing, if you want to move it back to the bottom, feel free to do it 🙂.

wzieba: I'm okay with both top and bottom I just don't like ambiguity in such static matters as code style. In other words I'd prefer to not decide it every time if we should put it top or bottom, I thought bottom is the official placement. Top is also good for me if it's preferred (aligned in https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/17d02449dbeaf030b016d4e3c64a831a755fd3ec) 

---

I'm wondering about `WooCommerce` codebase. We have Detekt and the default class ordering rule is forcing `companion object` to the bottom (https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/17d02449dbeaf030b016d4e3c64a831a755fd3ec). On the other hand we have a lot of classes with `companion object` on top, although I don't see that `ClassOrdering` is disabled or baselined 🤔 .

hichamboushaba: > On the other hand we have a lot of classes with companion object on top, although I don't see that ClassOrdering is disabled or baseline

[ClassOrdering](https://detekt.github.io/detekt/style.html#classordering) is not enabled by default, so it's disabled too for our case.

wzieba: Oh I see `Active by default: No` - mystery solved, thanks 🙂

Issue #2143, opened 2021-10-08
Move module unit tests from `example` to respective modules


wzieba:
#### Challenge I try to resolve:

When working in WooCommerce module, there are cases when I'd like to visibility modifier of a class to `internal`. I can't do this though because currently unit tests for WooCommerce related classes are placed in `example` module.

Even if I write a new test for WooCommerce module I can't re-use current tests setup which is placed in `example` module.

#### Solution

`example` module is one for example app and I think we shouldn't place unit tests there. I'd like propose to:
- [ ] create a separate module containing common tests setup (`tests-common`)
- [ ] move unit tests to respective modules (WooCommerce tests to WooCommerce module, WordPress tests to WordPress module etc.)

PR #2203, opened 2021-12-01
[4/4] Add Woo database migration to support `OrderEntity`

wzieba:
Relates to: https://github.com/woocommerce/woocommerce-android/issues/5070

### Description

This PR adds a migration about adding `OrderEntity` to database schema + instrumented test for it.

### DIFF
--- a/plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
+++ b/plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
@@ -1,0 +1,33 @@
+package org.wordpress.android.fluxc.persistence
+
+import androidx.room.testing.MigrationTestHelper
+import androidx.sqlite.db.framework.FrameworkSQLiteOpenHelperFactory
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_3_4
+
+@RunWith(AndroidJUnit4::class)
+class MigrationTests {
+    @Rule
+    @JvmField
+    val helper: MigrationTestHelper = MigrationTestHelper(
+            InstrumentationRegistry.getInstrumentation(),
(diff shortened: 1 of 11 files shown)


plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
```diff
@@ -0,0 +1,33 @@
...
+    @Rule
+    @JvmField
+    val helper: MigrationTestHelper = MigrationTestHelper(
+            InstrumentationRegistry.getInstrumentation(),
+            WCAndroidDatabase::class.java.canonicalName,
+            FrameworkSQLiteOpenHelperFactory()
+    )
+
+    @Test
+    fun testMigrate3To4() {
```
wzieba, line 23: This test _does not run on CI_.

The reason for that is that I intentionally didn't add it into `example` app project. I believe we shouldn't add all possible tests to the `example` module as it blocks as from using correct visibility modifiers and makes executing tests longer - when I work on Woo tests I don't want to re-run FluxC tests every time too. Relates to #2143 .

I'll prepare a Platform9 request to support executing instrumented tests from `:plugins:woocommerce` module. For now, please execute this test locally - if you want to check failure, please edit basically anything in `MIGRATION_3_4` sql statements - the test should fail.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
```diff
@@ -45,6 +46,8 @@ abstract class WCAndroidDatabase : RoomDatabase() {
                 WCAndroidDatabase::class.java,
                 "wc-android-database"
         ).allowMainThreadQueries()
+                .fallbackToDestructiveMigrationOnDowngrade()
+                .addMigrations(MIGRATION_3_4)
```
wzieba, line 50: I really wanted to use [`AutoMigration`](https://developer.android.com/reference/kotlin/androidx/room/AutoMigration) for this, but unfortunately it was forcing me to bump other dependencies (like Kotlin) and `compileSdkVersion`. I didn't want to block migration by this as it'll probably take some time.

hichamboushaba (approved):
Works well, nice work @wzieba

PR #2205, opened 2021-12-03
Order migration - post merge sync

wzieba:
### Description

This PR is set of changes needed to make `woo/5070_orders_to_room` (main Order -> Room migration branch) compile with current `develop`.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -323,6 +323,11 @@ class WCOrderStore @Inject constructor(
         }
     }
 
+    sealed class HasOrdersResult {
+        data class Success(val hasOrders: Boolean) : HasOrdersResult()
+        data class Failure(val error: FluxCError) : HasOrdersResult()
+    }
```
wzieba: This is something that should address @kidinov [concern](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2189#discussion_r747242064) in previous PR #2189 .

For the context: One of the migration changes is that we're removing `rowsAffected` field which was a bit overused in previous `HasOrders` implementation (returning `1` for having orders and `0` for not having any).

I believe that not making `HasOrdersResult` an `OnChanged`* does not violate [FluxC architecture](https://github.com/wordpress-mobile/WordPress-FluxC-Android/wiki/Architecture) as after PR #2189 we don't use event bus approach but rather `getter` from FluxC architecture chart as a way to communicate Woo with FluxC.

@malinajirka @kidinov as reviewers of #2189 WDYT about this? Is it okay?

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
```diff
@@ -37,13 +38,12 @@ class AddressEditDialogFragment : DaggerFragment() {
...
         fun newInstanceForCreation(addressType: AddressType, listener: (OrderAddress) -> Unit):
-                AddressEditDialogFragment =
-                AddressEditDialogFragment().apply {
-                    this.selectedOrder = WCOrderModel()
-                    this.mode = Add
-                    this.addressListener = listener
-                    this.currentAddressType.value = addressType
-                }
+                AddressEditDialogFragment = AddressEditDialogFragment().apply {
+            this.selectedOrder = WCOrderModel(localSiteId = LocalId(-1), remoteOrderId = RemoteId(-1))
```
wzieba, line 42: I'm not a fan of this (as well as in other places where we were using an empty `WCOrderModel` [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2205/files#diff-1faa106d0316334dc71e2013c675ab9009954e267691d14d16bf4d84b8c61fe9R516) and [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2205/files#diff-1faa106d0316334dc71e2013c675ab9009954e267691d14d16bf4d84b8c61fe9R526).

I believe that ideally we wouldn't have to ever create empty `WCOrderModel` which is not really a valid state. But I didn't want to change those things in this PR as they're out of scope. Commenting for visibility and opinions.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
malinajirka: I agree, I don't see any benefit in returning OnChanged* events. During the refactorings from events bus to suspendable functions, we just decided to make as little changes as possible so if the method was returning OnChanged* we kept it.

wzieba:
👋 @malinajirka @kidinov requesting your review [in matter](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2205#discussion_r761768915) of #2189 PR. Thanks! 

Edit: ✅ in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2205#discussion_r761777960

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
wzieba: Sure, I understand the decision in #2189. Thanks for confirmation!

malinajirka: Thanks for improving it! :)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -325,7 +325,7 @@ class WCOrderStore @Inject constructor(
 
     sealed class HasOrdersResult {
         data class Success(val hasOrders: Boolean) : HasOrdersResult()
-        data class Failure(val error: FluxCError) : HasOrdersResult()
+        data class Failure(val error: OrderError) : HasOrdersResult()
```
wzieba, line 327: I'm not fan of doing this as `OrderError` is `OnChangedError` and we don't have any change here but it seems correct that `OrderError` is an error related to Order calls. 

We can't use `FluxCError` as in Woo app we use error's message.

hichamboushaba: np, Why not use Kotlin's Result? we can have an exception that encapsulates the `OrderError` and use it in the failure.
Kotlin's Result has some nice APIs that will help the code make easier to read.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
hichamboushaba: I feel like this is an issue because WCOrderModel is currently a database entity, since you renamed the database entity to `OrderEntity`, do you think it'll be easier to have a proper model class here, and we can keep the name `WCOrderModel` for it, and we can make all fields optional for it, since it can be a draft?

I'm just thinking aloud, if it would require lots of changes, or if it doesn't make sense, ignore my comment 😄

hichamboushaba (approved):
Looks good @wzieba, thanks.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
wzieba: Good idea, I'd like to but when trying to do that I'm prompted that 
> `'kotlin.Result' cannot be used as a return type`

This is until Kotlin 1.5. I could add compiler arg that would ignore this but I feel like it's too much overhead for this particular case. But generally speaking I agree that using `kotlin.Result` in this and other cases will make code easier to read.

--- 

Regarding this matter - WDYT about moving `WooException` from WooCommerce database into FluxC? Is that okay to you?

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
wzieba: > I feel like this is an issue because WCOrderModel is currently a database entity

True!

> , do you think it'll be easier to have a proper model class here, and we can keep the name WCOrderModel for it, and we can make all fields optional for it

Yes, or better to me, model this in a way that `Order` is something different than `draft Order`, something like

```
sealed class WCOrderModel(fields) {
    data class Draft(with nullable fields or generally without fields like `remoteOrderId`)
    data class Regular(without nullable fields)
}
```

WDYT?

So generally I agree but I'd keep it outside of this PR. I added context of this into task https://github.com/woocommerce/woocommerce-android/issues/5386

wzieba:
Thank you @hichamboushaba !

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
hichamboushaba: Yes, I forgot FluxC is still not using kotlin 1.5, I think we should handle this migration.

> Regarding this matter - WDYT about moving WooException from WooCommerce database into FluxC? Is that okay to you?

Yes, totally fine.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/AddressEditDialogFragment.kt
hichamboushaba: > Yes, or better to me, model this in a way that Order is something different than draft Order

I think this needs a bigger reflection to think about the impact of a change like this, for sure it would require way more changes than just making WCOrderModel a domain model. 
So I can't really say right now, I'll be more than happy to re-discuss this after my AFK 🙂

PR #2207, opened 2021-12-06
Dashboard Cards: Convert Date to GMT Timezone

ParaskP7:
Parent: [WordPress-Android#15200](https://github.com/wordpress-mobile/WordPress-Android/issues/15200)

The API endpoint date for the post cards are in GMT format and as such this PR makes sure that this is being respected.

As such:
- First, when converting the `date` post response string using the `fromDate(...)` function, the date format timezone is set to `GMT`.
- Then, when converting the card entity from and to a JSON using GSON in order to save and retrieve the card model from the database, a newly created, private to cards `GsonDateAdapter` is used which also sets the date format timezone to `GMT`.

Both additions, makes sure that the date ends up being the expected one when it is being utilized later on and converted to the user's timezone.

-----

To test:
- First, install the `WordPress-Android` app without with branch and verify that any scheduled post time is incorrectly displayed (use latest `develop`).
- Then, install the `WordPress-Android` app with with branch and verify that any scheduled post time is now correctly displayed (use latest `develop`, but point to `issue/dashboard-cards-date-gmt-timezone` branch through `local-build.gradle`).

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/dashboard/CardsUtils.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/dashboard/CardsUtils.kt
@@ -2,18 +2,31 @@ package org.wordpress.android.fluxc.network.rest.wpcom.dashboard
 
 import com.google.gson.Gson
 import com.google.gson.GsonBuilder
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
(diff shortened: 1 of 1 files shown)


ashiagr (approved):
Tested using accompanying WordPress-Android PR, everything works as expected! 🙌

ParaskP7:
👋 @ashiagr !

> Tested using accompanying WordPress-Android PR, everything works as expected! 🙌

Great, thanks you so much! 🥇

PR #2212, opened 2021-12-09
Dashboard Cards: Add Room Database Version 5 Schema

ParaskP7:

This PR completes the previously already merged and incomplete #2197 PR, which was actually incomplete because the `5.json` room database version 5 schema was missing from it.

PS: Without that, most probably the migration and as such upgrading the app from one version to another might fail. See previous such [PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2085/files) for reference purposes.
PSS: After some research this update is actually not mandatory and will not fail the migration and as such upgrading the app. It is just a good practice to have this file committed in the version control system and have it in the codebase. For more info check the [documentation](https://developer.android.com/reference/android/arch/persistence/room/Database#exportSchema()).

### DIFF
--- a/fluxc/schemas/org.wordpress.android.fluxc.persistence.WPAndroidDatabase/5.json
+++ b/fluxc/schemas/org.wordpress.android.fluxc.persistence.WPAndroidDatabase/5.json
@@ -1,0 +1,420 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 5,
+    "identityHash": "30ad07883666cf9a4121a173b49edebd",
+    "entities": [
+      {
+        "tableName": "BloggingReminders",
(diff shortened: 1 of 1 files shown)


ashiagr (approved):
As discussed, the file doesn't affect migration and [is not mandatory](https://developer.android.com/reference/android/arch/persistence/room/Database#exportschema). 

I confirmed there were no migration issues by installing `WordPress-Android` `18.7` (without `DashboardCards` table) and then `develop`(without this fix). 

Yet as suggested [here](https://developer.android.com/reference/android/arch/persistence/room/Database#exportschema) and as an established practice in `FluxC` codebase, it's ok to include it in version history.

All 🟢 from my side. 👍 Merging it now.

Feel free to create a corresponding `WordPress-Android` PR to update the `FluxC` version after merge.

PR #2213, opened 2021-12-10
Threat jetpack cp sites as self-hosted.

khaykov:
This PR prevents sites that are using Jetpack Connection Package from being treated as wpcom sites.


khaykov:
This is not a viable solution after all, so closing.

PR #2215, opened 2021-12-13
Add two more parameters to WCRevenueStatsModel.kt 

rfermontero:
As part of the [analytics hub issue ](https://github.com/woocommerce/woocommerce-android/issues/5237) we need to parse avg value of order and the number of items sold in WCRevenueStatsModel total field response.

This pull request adds those 2 new params to the model.


anitaa1990 (approved):
:shipit:

PR #2219, opened 2021-12-16
Remove SSREntity table from WCAndroidDatabase, as well as its related usages.

hafizrahman:
The System Status Report feature in WooCommerce Android ended up not needing to use any persistence, since we always want the SSR info to be fresh for troubleshooting purposes.

The database content has never been used in WooCommerce Android in the first place, and this PR removes the table and its usages.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
```diff
@@ -47,7 +43,7 @@ abstract class WCAndroidDatabase : RoomDatabase() {
                 "wc-android-database"
         ).allowMainThreadQueries()
                 .fallbackToDestructiveMigrationOnDowngrade()
-                .addMigrations(MIGRATION_3_4)
+                .addMigrations(MIGRATION_4_5)
```
wzieba, line 48: We have to add all migrations, not change the current one. Currently, if someone has app with version 3 and then installs with version 5, the app will crash.

wzieba (changes requested):
Thanks @hafizrahman for keeping DB clean! I've added one comment but crucial one. Other than that, all looks and works good.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
hafizrahman: > We have to add all migrations

Aha! I misunderstood this before, thank you!

wzieba (approved):
Thanks @hafizrahman !

PR #2222, opened 2021-12-21
[Woo] Order Creation: handle customer note

hichamboushaba:
This PR just updates the `OrderCreationRequest` model to include a customer note.

WCAndroid's PR: https://github.com/woocommerce/woocommerce-android/pull/5494

#### Testing
Test order creation in the sample, and confirm the customer note is working as expected.


ThomazFB (approved):
Looking good! :shipit: 

I'll just leave the PR approved so you can merge it when you find it's best to update just in case you want to leave the Woo counterpart with the develop checkout instead of the branch number version.

PR #2223, opened 2021-12-21
Issue/wc simple payments taxable

nbradbury:
This simple PR renames FluxC's "Quick Order" to "Simple Payment" and adds the `tax_status` field when posting a simple payment.  This is for use with [this WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/5498).

The reason `tax_status` is added is so when the order is created, core will automatically assign taxes to the order. Otherwise we wouldn't know how to calculate taxes.


ThomazFB (approved):
Changes are looking good! I'll just approve so you can merge this when you think it's the best moment.

PR #2225, opened 2021-12-23
Issue/5387 Steps to remove order identifier from the Woo app

kidinov:
Part of  https://github.com/woocommerce/woocommerce-android/issues/5387

The PR adds `getOrderByRemoteIdAndSite` method that used in the Woo app as a replacement for `getOrderByIdentifier` which will be removed soon

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
@@ -409,6 +409,14 @@ class WCOrderStore @Inject constructor(
     }
 
     /**
+     * Given an [RemoteId] and [SiteModel],
+     * returns the corresponding order from the database as a [WCOrderModel].
+     */
+    fun getOrderByRemoteIdAndSite(orderRemoteId: RemoteId, site: SiteModel): WCOrderModel? {
+        return ordersDao.getOrder(orderRemoteId, site.localId())
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -408,6 +408,14 @@ class WCOrderStore @Inject constructor(
         return ordersDao.getOrderForIdSet((orderIdentifier.toIdSet()))
     }
 
+    /**
+     * Given an [RemoteId] and [SiteModel],
+     * returns the corresponding order from the database as a [WCOrderModel].
+     */
+    fun getOrderByRemoteIdAndSite(orderRemoteId: RemoteId, site: SiteModel): WCOrderModel? {
```
wzieba: As `WCOrderStore` is a class exposed to client and in client we agreed to not use `RemoteId` for `Order` identification, WDYT about changing signature to accept primitive?

I would see benefit of `RemoteId` here if this method would be used in `FluxC` but I think it's not the case here.

```suggestion
    fun getOrderByRemoteIdAndSite(orderRemoteId: Long, site: SiteModel): WCOrderModel? {
```

kidinov: Good point! Done - https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2225/commits/da3c3c237e74e77d25b07fc1e6cb640997ac7b14

wzieba (approved):
Thanks!

PR #2234, opened 2021-12-27
Issue/5381 Removes main thread Order room DAO usage

kidinov:
Part of: https://github.com/woocommerce/woocommerce-android/issues/5381

The PR removes main thread Order room DAO usage

Please notice that the following methods have not been made suspended:  `insertOrUpdateOrder`, `getOrdersForSiteByRemoteIds`,  `deleteOrdersForSite`, `getOrderCountForSite`. Those methods seem to be not used from the main thread. They ideally should be suspended too but because they called from the event bus methods so conversation gets out of the goal of the PR


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
```diff
@@ -15,39 +15,42 @@ typealias AutoGeneratedId = Long
 @Dao
 abstract class OrdersDao {
     @Query("SELECT * FROM OrderEntity")
-    abstract fun getAllOrders(): List<WCOrderModel>
+    abstract suspend fun getAllOrders(): List<WCOrderModel>
 
     @Insert(onConflict = REPLACE)
     abstract fun insertOrUpdateOrder(order: WCOrderModel): AutoGeneratedId
```
wzieba, line 21: > They ideally should be suspended too but because they called from the event bus methods so conversation gets out of the goal of the PR

Gotcha. WDYT about deprecating them and introducing `suspend` counterparts to make sure that someone next using those methods will use thread-safe option?

wzieba (approved):
I've tested it mostly be checking it along with https://github.com/woocommerce/woocommerce-android/pull/5562/ .

I've left one suggestion, otherwise LGTM, thank you!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
kidinov: Like the idea, just not sure how to implement this technically. In this case, we have to have 2 identical methods in `OrdersDao.kt` for each case (one "normal" another suspend). Also, several unused methods in WCOrderStore have to be added. 

I would probably tackle this in a "horizontal" manner. Same way as we convert eventbus-based communication to coroutines. One action per PR. Changes both in Woo and Fluxc, so everything can be tested and we don't leave Fluxc in weird state in between

Wdyt?

wzieba: > In this case, we have to have 2 identical methods in OrdersDao.kt for each case (one "normal" another suspend). Also, several unused methods in WCOrderStore have to be added.

I think it's fine (especially as we'd mark those non-suspendable methods as deprecated) but I don't insist. I agree with your horizontal approach and I think those two (deprecating and horizontal) doesn't exclude each other. I'm just worried that before we'll have time to fully take care of all non-suspandable methods, they'll be used in new places giving us more work to do.

But that's totally optional, please feel free to merge this PR without adding deprecations if you think it'll add more overhead than value.

PR #2235, opened 2022-01-03
Fix for ReleaseStack_WCOrderTest

kidinov:
`sSite` variable was accessed before it was initialized. This fixes the main issue, but one test still failing due to unknown reason (may be that the test is wrong or the code behavior has changed)

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_Base.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_Base.java
@@ -1,8 +1,8 @@
 package org.wordpress.android.fluxc.release;
 
-import static androidx.test.InstrumentationRegistry.getInstrumentation;
-
 import android.content.Context;
+
+import static androidx.test.InstrumentationRegistry.getInstrumentation;
 
 import com.android.volley.RequestQueue;
 import com.yarolegovich.wellsql.WellSql;
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_Base.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_Base.java
@@ -1,5 +1,7 @@
 package org.wordpress.android.fluxc.release;
 
+import static androidx.test.InstrumentationRegistry.getInstrumentation;
+
 import android.content.Context;
 
 import com.android.volley.RequestQueue;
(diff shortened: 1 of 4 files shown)


wzieba (commented):
Looks good but `ReleaseStack_WCOrderTest` is still failing - I understand it's not because your changes but maybe we could fix it in scope of this PR?

kidinov:
@wzieba We definitely should, it's just I have a little of the context of the changes which were implemented and after spending a couple of hours didn't find any reasonable explanation if the problem in the test itself (it's weird, to be honest) or the code was broken in one of the recent refactoring. That's why I asked you if you can take a look too 😃

wzieba:
Oh sorry I missed your comment 🤦 .

I'll skip debugging details but yeah, it took a while to realise the root cause. So apparently I changed `causeOfChange` to incorrect one here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/0540cf8#diff-78ca32c7d01076b3625428759cd7cb0a614d6ad7c7e8a75081046526a7af5bc4R678 (it's link to line 678 in `WCOrderStore` but Github doesn't link it right at least for me).

What does it really change in implementation in Woo - now after fetching completed, we'll also fetch "unfilled order count" (`MainPresenter#203`).

I've allowed myself to already push fix to your branch - I've tested Woo app briefly and it looks all good. Could you please confirm it's all good?

wzieba:
Some other tests are failing, I'm checking that

wzieba:
Similar case in another place: https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/675a97d5e2ce13a0d25ce57da5935174ee4843ae#diff-78ca32c7d01076b3625428759cd7cb0a614d6ad7c7e8a75081046526a7af5bc4L826

I moved setting `causeOfChange` outside `also`.

wzieba:
Rest of the failing tests seem to be flaky or just passes for me locally. @kidinov could you please check if `testFetchOrdersById` passes for you?

kidinov:
@wzieba I think the test fails because we don't reset room DB after/before each individual test. I added this here - 0ae9e1c2e2b287f22ad738c0055d8f4cd726076f. Logically to do it before, but because of the current approach to inject dependencies it doesn't look too easy to do

kidinov:
@wzieba it looks like that `testFetchOrdersById` is fixed now, but apparently there are something broken with `plugin*` tests 🤦

wzieba (approved):
Gotcha - let's merge this PR as it fixes all fails caused by Order->Room migration and handle `plugins`* tests separately. 

Thanks!

PR #2238, opened 2022-01-06
Issue/5382 Remove WcOrderModel id. Part 2

kidinov:
Part 2  of woocommerce/woocommerce-android#5382


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
```diff
@@ -1835,6 +1835,12 @@ open class WellSqlConfig : DefaultWellConfig {
                 168 -> migrateAddOn(ADDON_WOOCOMMERCE, version) {
                     db.execSQL("ALTER TABLE WCProductModel ADD PURCHASABLE INTEGER")
                 }
+                169 -> migrateAddOn(ADDON_WOOCOMMERCE, version) {
+                    db.execSQL("DROP TABLE IF EXISTS WCOrderSummaryModel")
+                    db.execSQL("DROP TABLE IF EXISTS WCOrderNoteModel")
+                    db.execSQL("DROP TABLE IF EXISTS WCOrderShipmentTrackingModel")
+                    db.execSQL("DROP TABLE IF EXISTS WCOrderSummaryModel")
```
wzieba: I'm not sure if I understand this change - why do we drop those tables?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -3,37 +3,28 @@ package org.wordpress.android.fluxc.model
...
-        tableName = "OrderEntity",
-        indices = [Index(
-                value = ["localSiteId", "remoteOrderId"],
-                unique = true
-        )]
+    tableName = "OrderEntity",
+    indices = [Index(
+        value = ["localSiteId", "orderId"],
+        unique = true
+    )],
```
wzieba: I believe that if `localSiteId` and `orderId` tuple is now primary key, we can remove unique constraint 
```suggestion
    
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
```diff
@@ -7,32 +7,29 @@ import androidx.room.Query
 import androidx.room.Transaction
 import kotlinx.coroutines.flow.Flow
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
-import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
 import org.wordpress.android.fluxc.model.WCOrderModel
 
-typealias AutoGeneratedId = Long
```
wzieba, line 13: Awesome you removed this 🙌

wzieba (commented):
Looks good but I'm not sure about dropping some entities from WellSql table.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
kidinov: Do you know how to verify that the index is created even in the case when the primary key is complex?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
kidinov: This is done because of the changes similar to that one - https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2238/files#diff-fe59a7e4fb779dc4325f1244b69faaa138c20f9a67d60db93ff7d3cbabf4e3d0R27

I could probably make a migration that renames the field, it's a bit more complex. Also we anyway drop Orders table so probably it doesn't make sense to keep this anyway too

wzieba: But here you drop tables of different entities, I still don't get it :/. E.g. `WCOrderNoteModel` is still used in WooCommerce and it's defined in `WellSql` - we didn't move it anywhere. Why should we remove it?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
wzieba: I think it's just a feature of primary key but if you want to check that - I think the fastest way would be to add a short test that adds 2 entities with the same `localSiteId` and `orderId` and see if database has 1 or 2 entities. It should have 1.

kidinov: Sorry, I meant `index`, not the PK. Looking into `8.json` I see there is a

```
        "indices": [
          {
            "name": "index_OrderEntity_localSiteId_orderId",
            "unique": true,
            "columnNames": [
              "localSiteId",
              "orderId"
            ],
            "createSql": "CREATE UNIQUE INDEX IF NOT EXISTS `index_OrderEntity_localSiteId_orderId` ON `${TABLE_NAME}` (`localSiteId`, `orderId`)"
          }
        ]
```

Which is what we need. I'm just not sure if sqlite always creates an index by PK or not 🤔 . In any case, I think explicitly creating an index won't hurt, wdyt?

wzieba: >  I'm just not sure if sqlite always creates an index by PK or not 🤔

True, I just checked it and it doesn't. Previous `index` was added in order to achieve uniqness, currently you'd prefer to leave it due to performance optimalization, is that correct? If so, I'm all good with that but would suggest to remove `unique = true`

My argument is that it's unexpected (at least to me) to have uniqness constraint *and* primary key on the same field. If I was browsing this code, I'd be surprised and started to look for any reason of such constraint.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
kidinov: 1. 🤦‍♂️ It is supposed to be not `DROP TABLE` but `TRUNCATE` or `DELETE` in sqlite
2. 🤦‍♂️ `db.execSQL("DROP TABLE IF EXISTS WCOrderSummaryModel")` I run it twice
3. I remove the data from that not only because we drop `Orders` anyway, but also because `WCOrderNoteModel` and `WCOrderShipmentTrackingModel` used `local_order_id` as FK to orders, which is removed, so the value stored there is not relevant and can not be used

Changes - b0abe5d0a7068025e99bce63f22119e56ae3f7c7

wzieba: > 🤦‍♂️ It is supposed to be not DROP TABLE but TRUNCATE or DELETE in sqlite

Gotcha!

> I remove the date from that not only because we drop `Orders` anyway, but also because `WCOrderNoteModel` and `WCOrderShipmentTrackingModel` used `local_order_id` as FK to orders, which is removed

I'm not sure if `local_order_id` was used as a real FK to orders (I think there's no such constraint) but I agree they were dependant. But, e.g. in `WCOrderNoteModel`, you persisted column name to `LOCAL_ORDER_ID` which should make everything work like before.

What's the reason for deleting this data? What would happen if we wouldn't delete it?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
kidinov: Yep, making a full scan when you want to get an entry by id is not ideal.

> My argument is that it's unexpected (at least to me) to have uniqness constraint and primary key on the same field.

I see! I was confused here, It's just constraints are one thing, indices another one, but it looks like that in Room (or maybe it's SQLITE design) uniqueness are made via indexes which is weird. What to do if you want a unique field but don't want to build an index? 🤷‍♂️

Changed is here - 50e46e67413c8c65f76963103ae67231dac91a85

wzieba: Thanks! Looks good to me, yeah I agree it's a confusing with index and unique constraint being tied together.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
kidinov: > I'm not sure if local_order_id was used as a real FK to orders (I think there's no such constraint) but I agree they were dependant. But, e.g. in WCOrderNoteModel, you persisted column name to LOCAL_ORDER_ID which should make everything work like before.
> What's the reason for deleting this data? What would happen if we wouldn't delete it?

Example for better understanding. 
Lets say In the current version:
* `Order( local_id = 13, remote_id = 10)`
* `WCOrderNoteModel(local_order_id=13)` (a not linked to the order with remote_id=10)
* `WCOrderNoteModel(local_order_id=10)` (Note for another order)

After the change:
* `Order(order_id = 10)` (local_id is gone, we have only `order_id` which is used to be `remote_id`)
Without deleting the data will try to fetch `WCOrderNoteModel` for `order_id=10` and the model for another order will be returned. Because before we used `local_id` as FK here, and now it's gone.

In short - we don't have `local_order_id` anymore that's why we can not link the notes to the orders.

wzieba: > Without deleting the data will try to fetch `WCOrderNoteModel` for `order_id=10` and the model for another order will be returned. Because before we used local_id as FK here, and now it's gone.

To my knowledge we don't use foreign keys in WellSql - data junction/linking happens in `OrderSqlUtils`. So e.g. `OrderSqlUtils#getOrderNotesForOrder` have line 
```kotlin                 
   .where()
   .equals(WCOrderNoteModelTable.LOCAL_ORDER_ID, orderId)
```
which seems to be not affected by any database schema changes. 

Sorry if I'm still missing something!

wzieba: oh nevermind @kidinov sorry I just realised the issue 🤦 Thanks for patience 😅 

We remove data because we've been linking them by local id which is removed. We need to re-fetch it and link by remote id.

But we can't do this really because remote id is ambiguous. We need to identify it by pair of `remoteOrderId x siteId`. Should we do this in this or next PR? Because in current form it'll cause problems for users with many sites.

kidinov: > But we can't do this really because remote id is ambiguous. We need to identify it by pair of remoteOrderId x siteId. Should we do this in this or next PR? Because in current form it'll cause problems for users with many sites.

Great point. I actually got tricked by my own renaming -> `orderId` is not really an id of the order 😃. Yes, it has to be done. I can do it in the next PR. The current one doesn't target the trunk so we can merge this anyway

wzieba: Gotcha, thanks!

wzieba (approved):
Looks good, thank you for working on this @kidinov ! Merging this knowing we have an issue that has to be addressed before merging to trunk (https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2238#discussion_r803629641)

PR #2239, opened 2022-01-06
Dashboard Cards: Handle Authorization Required

ParaskP7:
Related: [WordPress-Android#15736](https://github.com/wordpress-mobile/WordPress-Android/issues/15736)

This PR:
- Adds the `clear()` function to `CardsDao`, which deletes all rows from the `DashboardCards` table.
- Implements the `authorization required` case on `CardsStore`, which makes sure that when a user gets downgraded from an 'administrator', 'editor', 'author' or 'contributor' to a 'follower' or 'subscriber', on an authorization required error:
  1) The database is being cleaned of any cards model from previous requests, and
  2) Instead of having the user seeing an error message, no error message is shown. Instead, the dashboard cards feed just disappears from the 'My Site' tab.

-----

To test:
- Start with a site with `draft` and `scheduled` posts.
- Make sure you are an `administrator` of that site.
- Add another `editor` user to your side.
- Make sure that this `editor` user is able to see all the `draft` and `scheduled` posts.
- Downgrade this user to a `follower` (or `subscriber`).
  - Make sure that this `follower` user is no longer able to see any of the `draft` or `scheduled` posts, or the dashboard feed in general.
  - Make sure that this `follower` user is not seeing any dashboard feed related error card, snackbar or stale message.

-----

Note:
1) Before trying all the above, maybe try it once without applying this solution and see how it works atm.
2) You will need to have two instances of the app running side-by-side on your device, or two devices, each with an instance of the app. Then, when the user is downgraded on one instance, or a device, you should pull-to-refresh on this other instance/device to see the result. Otherwise, if you close the app and swipe it off from running in the background, the next time you open the app that site will not be visible altogether. Thus, this PR only solves this edge case where the user gets downgraded while their app is currently open or running on the background.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/store/dashboard/CardsStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/dashboard/CardsStoreTest.kt
@@ -1,5 +1,6 @@
 package org.wordpress.android.fluxc.store.dashboard
 
+import com.nhaarman.mockitokotlin2.verify
 import com.nhaarman.mockitokotlin2.whenever
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.single
@@ -117,7 +118,17 @@ class CardsStoreTest {
     }
 
     @Test
-    fun `given cards response, when fetch cards gets triggered, then null no error cards model is returned`() = test {
+    fun `given cards response, when fetch cards gets triggered, then cards model is inserted into db`() = test {
+        val payload = CardsPayload(CARDS_RESPONSE)
+        whenever(restClient.fetchCards(siteModel)).thenReturn(payload)
+
+        cardsStore.fetchCards(siteModel)
+
+        verify(dao).insertWithDate(siteModel.id, CARDS_MODEL)
+    }
+
+    @Test
+    fun `given cards response, when fetch cards gets triggered, then empty cards model is returned`() = test {
         val payload = CardsPayload(CARDS_RESPONSE)
         whenever(restClient.fetchCards(siteModel)).thenReturn(payload)
 
--- a/example/src/test/java/org/wordpress/android/fluxc/store/dashboard/CardsStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/dashboard/CardsStoreTest.kt
@@ -165,6 +165,29 @@ class CardsStoreTest {
     }
 
     @Test
+    fun `given authorization required, when fetch cards gets triggered, then db is cleared of cards model`() = test {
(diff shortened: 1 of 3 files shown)


zwarm (approved):
@ParaskP7 - Love the refactoring in cardStore to separate out the error from success. 👏 

I'm not 100% sure if I need to test this using the sandbox (to pick up the new API logic), the instructions didn't specify so I proceeded off of the sandbox.

The **testing results** are from the "**user**" perspective, not the site owner. I toggled the permissions using the admin account.

`Administrator/Editor` - Able to see drafts/scheduled post cards and posts were available on the posts lists (drafts, scheduled). ✅ 
`Contributor` - Able to see drafts/scheduled post cards; however posts were NOT available on the posts lists (drafts, scheduled) ❓ 
`Subscriber` - Post cards were removed and no error message was shown. Posts were not available on the posts lists (drafts/scheduled) ✅ 

I don't think I should see any post cards for the `contributor` role. Is this due to the API endpoint not being deployed yet?  I am going to approve this PR because I believe the contributor experience falls outside of it; however I am not going to merge until you give it the double ok.

Thanks for all the work!

ParaskP7:
👋 @zwarm !

Thanks so much for the review!

> @ParaskP7 - Love the refactoring in cardStore to separate out the error from success. 👏

🙇 

> I'm not 100% sure if I need to test this using the sandbox (to pick up the new API logic), the instructions didn't specify so I proceeded off of the sandbox.

You are right, there is no need to test using the sandbox. Thus, I didn't specify that in the PR.

> The testing results are from the "user" perspective, not the site owner. I toggled the permissions using the admin account.

👍 

> Administrator/Editor - Able to see drafts/scheduled post cards and posts were available on the posts lists (drafts, scheduled). ✅

👍 

> Contributor - Able to see drafts/scheduled post cards; however posts were NOT available on the posts lists (drafts, scheduled) ❓

👍 This is expectable as you are not using the sandbox.

> Subscriber - Post cards were removed and no error message was shown. Posts were not available on the posts lists (drafts/scheduled) ✅

👍 🥇 

> I don't think I should see any post cards for the contributor role. Is this due to the API endpoint not being deployed yet? 

Yes, exactly, don't worry about that as the API endpoint is not yet deployed. For you to check that you would need to use a sandbox, but this is not necessary for this PR, which is all about `followers` and `subscribers`, or else users that do not have any edit rights.

> I am going to approve this PR because I believe the contributor experience falls outside of it; however I am not going to merge until you give it the double ok.

Yes, thank you, I'll merge it myself as this PR has nothing to do with the `contributor` and `author` experience. 🙇 

> Thanks for all the work!

Thank YOU! 🏅

PR #2240, opened 2022-01-07
Fix precision of total refunds calculation

hichamboushaba:
Please check the WAndroid's PR for more details: https://github.com/woocommerce/woocommerce-android/pull/5612

Even though the number of lines is big, it's mostly just the database schema, the code changes are small.

And please don't merge until the above PR is approved.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -48,7 +49,7 @@ data class WCOrderModel(
     val customerNote: String = "", // Note left by the customer during order submission
     val discountTotal: String = "",
     val discountCodes: String = "",
-    val refundTotal: Double = 0.0, // The total refund value for this order (usually a negative number)
+    val refundTotal: BigDecimal = BigDecimal.ZERO, // The total refund value for this order (usually a negative number)
```
hichamboushaba, line 52: I was undecided between saving the `refundTotal` as String or as BigDecimal, but opted for a BigDecimal representation using a converter that converts it to its String representation in the database, let me know what you think about this?

IMO it's better, and we can even start migrating the other fields to be BigDecimals, to avoid the mapping on the domain level.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/converters/BigDecimalConverter.kt
```diff
@@ -0,0 +1,12 @@
...
+import androidx.room.TypeConverter
+import java.math.BigDecimal
+
+class BigDecimalConverter {
+    @TypeConverter
+    fun bigDecimalToString(input: BigDecimal): String = input.toPlainString()
+
+    @TypeConverter
+    fun stringToBigDecimal(input: String): BigDecimal = input.toBigDecimalOrNull() ?: BigDecimal.ZERO
+}
```
hichamboushaba, line 12: I'm using the String representation for the conversion instead of converting it to a Double, to avoid any precision loss, check the [docs](https://developer.android.com/reference/java/math/BigDecimal#BigDecimal(double)) of BigDecimal for more context.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -855,69 +855,6 @@ class OrderRestClient @Inject constructor(
...
-                shippingState = response.shipping?.state ?: "",
-                shippingPostcode = response.shipping?.postcode ?: "",
-                shippingCountry = response.shipping?.country ?: "",
-                lineItems = response.line_items.toString(),
-                shippingLines = response.shipping_lines.toString(),
-                feeLines = response.fee_lines.toString(),
-                metaData = response.meta_data.toString()
-        )
-    }
-
```
hichamboushaba, line 920: This is not particularly related to this PR, but since I needed to change the mapping, I decided to remove this deprecated function, and make the change only in one place.

plugins/woocommerce/schemas/org.wordpress.android.fluxc.persistence.WCAndroidDatabase/6.json
```diff
@@ -0,0 +1,530 @@
+{
```
shiki, line 1: How do you usually review this? Do you compare it to `5.json`? I'm just curious whether it's possible to check what has changed.

hichamboushaba: This is actually automatically [generated](https://developer.android.com/training/data-storage/room/migrating-db-versions#export-schema) by Room according to what we defined in the entities, and the database version, so just checking the kotlin code is enough.
And for now, I don't think there is any use for them, but they are useful in case we wanted to use  Room's auto migration in the future.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
wzieba: Yeah, to me it's okay to delegate `String <-> BigDecimal` mapping to Room. 

I have question about:
> avoid the mapping on the domain level.

What do you mean by that? Link to some part in code will be all good, I just want to fully understand your decision.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
wzieba: Thank you for that 💯 . Closes: https://github.com/woocommerce/woocommerce-android/issues/5404

wzieba (approved):
Looks good 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
hichamboushaba: > What do you mean by that? Link to some part in code will be all good, I just want to fully understand your decision.

I just mean I don't want to have a String representation of the price in the database entity, and have to convert to `BigDecimal` in the WCAndroid's `toAppModel`

PR #2244, opened 2022-01-17
 MSD - Stats Card - Handle Success Response

ashiagr:
Parent https://github.com/wordpress-mobile/WordPress-Android/issues/15749

This is a draft PR that includes the following changes for adding the `Stats Card` to the `cards` result:
- JSON deserialization for the success response 
- Mapping success response to model

----------------------------------------

There's nothing to test in this PR as the `dashboard/cards` endpoint doesn't yet include the `Stats Card` yet. So that existing tests work as expected,  I've updated the following test classes: 
- `CardsRestClientTest` is updated to add `StatsResponse`and `cards.json` is updated to include the `Stats Card`.
- `CardsStoreTest` is updated to add `STATS_ENTITY` to `CARDS_ENTITY`.
----------------------------------------

Note: We might want to keep this PR in draft mode until the `dashboard/cards` endpoint is updated for the `Stats Card` as the final structure might vary. With the current prod version of the endpoint, the stats card response being `null` is not added to the cards list.


zwarm:
> We might want to keep this PR in draft mode

@ashiagr , yes, let's do that! 
(1) I added `comments` to the endpoint because it's needed for iOS; not sure if we want to do anything with it. 
(2) Let's sync on what the success looks like ... I was basing it off of the following, but that can be changed.
```

Success: return array(
        'ok'           => true,
	'error'      => '',
	'stats'      => stats data,
      );
Failure: return array(
        'ok'     => false,
	 'error' => 'error message',
      );
```
NOTE:: We decided against the above approach, this is here for history purposes

zwarm:
@ashiagr - The `cards-data` endpoint has been deployed - you can swap out the mock data responses.

zwarm (approved):
@ashiagr Very well done 👏 

❤️  for the patience in changing all those tests.

I like the improvement that added "label" to types so the arguments build out works smoothly.

I tested using the WPAndroid#[15893 ]([MSD - Handle Single Dashboard Card Error in UI](https://github.com/wordpress-mobile/WordPress-Android/pull/15893#)) and was able to view the stats card as expected. I also visited a few web sites and PTR's, and the counts updated. 

Please feel free to merge when you feel comfortable.

PR #2246, opened 2022-01-19
[Order Update] ability to update more fields

hichamboushaba:
⚠️ Please don't merge until the WCAndroid's [PR](https://github.com/woocommerce/woocommerce-android/pull/5671) is reviewed

This PR adds support for updating orders using a single entry point to the OrderUpdateStore, this is actually needed to support Order Creation and Simple Payments.

We can probably refactor the Store to remove the repeated functionality between the new function and the other functions in an upcoming PR.

#### Testing
Just confirm that there is no regression in order creation in the example app.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
@@ -17,4 +17,14 @@ class FeeLine {
 
     @SerializedName("total_tax")
     val totalTax: String? = null
+
+    @SerializedName("tax_status")
+    val taxStatus: FeeLineTaxStatus? = null
+}
+
+enum class FeeLineTaxStatus {
+    @SerializedName("taxable")
+    Taxable,
+    @SerializedName("none")
+    None
 }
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/UpdateOrderRequest.kt
```diff
@@ -0,0 +1,13 @@
+package org.wordpress.android.fluxc.model.order
+
+import org.wordpress.android.fluxc.model.WCOrderStatusModel
+
+data class UpdateOrderRequest(
```
hichamboushaba, line 5: This will be used for both creation and update.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
```diff
@@ -17,4 +17,14 @@ class FeeLine {
 
     @SerializedName("total_tax")
     val totalTax: String? = null
+
+    @SerializedName("tax_status")
+    val taxStatus: FeeLineTaxStatus? = null
```
hichamboushaba, line 22: Since this is used in simple payments, I added it

nbradbury:
@ThomazFB Since you're reviewing the related WCAndroid PR, maybe you'd want me to take this one? I would've grabbed it yesterday but I was having build problems (unrelated tot this PR).

ThomazFB:
@nbradbury I already started the review but mostly on the Woo part, I can pass this PR to you so you can unblock yourself on Simple Payments. Thanks for stepping in!

nbradbury (approved):
@hichamboushaba This all looks, works, and tests just fine. I'll approve it, and you can merge it when the WCAndroid part is approved.

PR #2248, opened 2022-01-24
Woo/add product purchasable field

ThomazFB:
This PR simply adds the `purchasable` field at the `WCProductModel` in order to support better product configuration and filters inside the Woo app, testing this just requires us to verify if the field is correctly added to the database and reflects the API response as expected.


hichamboushaba (approved):
Looks good, thanks @ThomazFB

PR #2251, opened 2022-01-27
MSD - Stats Card - Handle Error Inside Cards Response

ashiagr:
Parent https://github.com/wordpress-mobile/WordPress-Android/issues/15749

This draft PR:
- Updates the `today's stats` and `posts` cards responses to include error.
- Updates the `today's stats` and `posts` card model to include error.
- Maps error in responses to an error in models.
- Makes no changes to storing and retrieving logic for dashboard cards into the database. This means the error inside the card model will get saved into the database whenever it is present and returned to the client as part of the card model. We can discuss any changes to this behavior.

------------------------------------------------------------------------------------------

To test:

- There is actually nothing to test in this PR, just review the updated `CardsRestClient` and `CardsStore` classes and make sure that the accompanying tests are working as expected.

------------------------------------------------------------------------------------------

Notes: 

- As the differential is still to be deployed, the `fetchCards` endpoint is not replaced with `v1_1` version yet to avoid rest route not found errors.
- To manual test: Apply D73653-code on your sandbox, proxy your device on it. Replace the `fetchCards` endpoint with `v1_1` version.
- Let's keep the PR in draft mode till the corresponding `WordPress-Android` PR is up and we've tested cards with errors feature.


zwarm (commented):
@ashiagr All looks good. The initial data values have been set to “0” and “empty” as we discussed. The errors are specific to the card type and embedded as such, In addition the errors will flow through to the client. Even if we only show a generic card today, we are ready to handle the specific errors in the future.

Tests work as expected. +1

zwarm:
@ashiagr - I retested this PR using WPAndroid#[15893 ]([MSD - Handle Single Dashboard Card Error in UI](https://github.com/wordpress-mobile/WordPress-Android/pull/15893#)).

I was able to see the stats card properly AND no stats card when using a site in which I don't have permission to view stats. I intercepted the response to verify this.

I'm going to leave the merging for you because I'm not 100% sure which PRs need to go first. 
Thanks for all the great work! 🙇

PR #2253, opened 2022-01-27
Issue/2252 wcordermodel taxlines

nbradbury:
Closes #2252 - This PR adds support for fetching and parsing `tax_lines` in the order model. For a description of why this is necessary, please refer to [this WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/5731).

PR #2255, opened 2022-01-31
Issue/2252 wc taxlines migration

nbradbury:
This PR adds the Room migration missing from #2253.

### DIFF
--- a/plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
+++ b/plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
@@ -41,7 +41,7 @@ class MigrationTests {
     @Test
     fun testMigrate5to6() {
         helper.apply {
-            createDatabase(TEST_DB, 6).close()
+            createDatabase(TEST_DB, 5).close()
             runMigrationsAndValidate(TEST_DB, 6, true, MIGRATION_5_6)
         }
     }
@@ -49,7 +49,7 @@ class MigrationTests {
     @Test
     fun testMigrate6to7() {
         helper.apply {
-            createDatabase(TEST_DB, 7).close()
+            createDatabase(TEST_DB, 6).close()
             runMigrationsAndValidate(TEST_DB, 7, true, MIGRATION_6_7)
         }
     }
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
```diff
@@ -139,3 +139,9 @@ internal val MIGRATION_5_6 = object : Migration(5, 6) {
         }
     }
 }
+
+internal val MIGRATION_6_7 = object : Migration(6, 7) {
```
kidinov, line 143: We should probably add a test? Check `MigrationTests`

kidinov (commented):
Adding a test would be nice (`MigrationTests`), otherwise looks good

nbradbury:
@kidinov Thanks for the pointer about `MigrationTests`. I added a 6-7 migration test in 897c419.

plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
```diff
@@ -45,6 +46,14 @@ class MigrationTests {
         }
     }
 
+    @Test
+    fun testMigrate6to7() {
+        helper.apply {
+            createDatabase(TEST_DB, 7).close()
```
kidinov: I think it should be 6, because we are testing migration from 6 to 7, not from 7 to 7. The same mistake was done with version 6, so probably worth fixing it there too

nbradbury: Fixed in 473683b, but now the migration fails.

nbradbury: ~~It looks like the tests are passing CI but the migration test from 6 to 7 is failing locally. Does it work for you?~~

nbradbury: @kidinov Scratch that. The migration error message was unhelpful, but it turned out I used the wrong column name in the migration and neglected to specify `NOT NULL`. It's working now. 🤦

kidinov: It's nice that you wrote a test 😃

kidinov (approved):

PR #2206, opened 2021-12-03
Migrate `WCOrderModel` from `WellSql` to `Room`

wzieba:
### Description

This is a compose PR of all code related of migrating `WCOrderModel` from `WellSql` to `Room` database.

It contains code from:

- https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2165
- https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2166
- https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2174
- https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2203
- https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2205

### Reviewing
As included PRs are reviewed, please feel free to just browse this code to see if you have any concerns or questions.

This PR is not as huge as stats my show as it contains a lot of auto-generated JSON database schema (`3.json` and `4.json`).

### Testing

Please see: https://github.com/woocommerce/woocommerce-android/pull/5092/

PR #2264, opened 2022-02-03
Fix WellSql's DB version number

hichamboushaba:
The last two migrations of WellSql had a wrong version number, where it's lower than its expected value by 1, this is causing an issue for future migrations, where the migration code might be executed twice.
Today while I tried adding a migration, I ran into this exception:
```     
Caused by: android.database.sqlite.SQLiteException: duplicate column name: PURCHASABLE (code 1 SQLITE_ERROR): , while compiling: ALTER TABLE WCProductModel ADD PURCHASABLE INTEGER
```

So I'm writing this small PR to avoid shipping the broken version into the next code freeze.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
@@ -30,7 +30,7 @@ open class WellSqlConfig : DefaultWellConfig {
     annotation class AddOn
 
     override fun getDbVersion(): Int {
-        return 168
+        return 169
     }
 
     override fun getDbName(): String {


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
```diff
@@ -30,7 +30,7 @@ open class WellSqlConfig : DefaultWellConfig {
     annotation class AddOn
 
     override fun getDbVersion(): Int {
-        return 168
+        return 169
```
hichamboushaba, line 33: This number wasn't incremented on https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2206, which resulted in using the wrong number in a later PR too.
A good thing is that the `167` migration doesn't hurt if it's executed multiple times, so if we can manage to get this merged before the next code freeze, the crash caused the `168` migration I mentionedd above won't cause any crashes to our users.

wzieba: Thanks for spotting this 🙇 .

wzieba (approved):

PR #2266, opened 2022-02-03
Consolidate SQLite storage for Woo plugins and WP plugins

hichamboushaba:
⚠️ Please don't merge before the WCAndroid [PR](https://github.com/woocommerce/woocommerce-android/pull/5775) is approved.

Currently, WooCommerceStore offers a way to fetch plugins using the `/system_status` endpoint, to allow users with the role "shop manager" to fetch the plugins list, since they can't use the WPApi one.
But I've found that we are using a different database table to store the plugins, where we don't need to, since the used table in the WP part is enough (they are the same plugins).
And this was problematic because we started supporting plugin installation from the app, and the installation uses the default `PluginStore`, which saves its results to the table `SitePluginModel`, and this makes the tables out of sync after each installation, and would require more requests than needed to update them.

If you don't agree with my thinking, please let me know, this is not necessary for the task I'm working on, and it's just an improvement that I noticed.

Changes done on this PR:
1. Remove the WCPlugins table, and the correspondent code.
2. Use `SitePluginModel` for the fetching functions on WooCommerceStore.
3. Update the filtering logic to use the plugin's "full name" (which is actually its path), instead of the display name, since the display name is subject to change, and relying on it is fragile (an [example](https://github.com/Automattic/woocommerce-services/pull/2179))
4. Add `WOO_CORE` to the enum `WooPlugin`, since it's needed for this [task](https://github.com/woocommerce/woocommerce-android/issues/5774).

#### Testing
Test the fetching of the WCShip plugin in the example app (Woo -> Shipping Labels -> Get shipping plugin info).

But for more deep dive testing, please the following parts using WCAndroid:
1. Display of the "Create shipping labels" button on eligible orders.
2. Payments eligibility.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WCSystemPluginResponse.kt
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Maps a System endpoint plugin item to [SitePluginModel]
+ *
+ * This function follows the same logic as Core to retrieve the name and slug from the plugin's path,
+ * Check class-wp-rest-plugins-controller.php and plugin.php in WPCore for more information.
+ */
+fun SystemPluginModel.toDomainModel(siteId: Int): SitePluginModel {
+    return SitePluginModel().apply {
+        localSiteId = siteId
+        name = this@toDomainModel.plugin.substringBeforeLast(".")
```
hichamboushaba, line 35: WooCommerce [returns](https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-system-status-v2-controller.php#L965) the full file path of the plugin, so we are following the same [logic](https://github.com/WordPress/WordPress/blob/b14d7a814923369d0290a0f5a771a41f85bedd90/wp-includes/rest-api/endpoints/class-wp-rest-plugins-controller.php#L583) as WP Core for removing the extension.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WCSystemPluginResponse.kt
```diff
@@ -0,0 +1,43 @@
...
+        displayName = this@toDomainModel.name
+        authorName = this@toDomainModel.authorName
+        authorUrl = this@toDomainModel.authorUrl
+        slug = this@toDomainModel.plugin.substringBeforeLast("/")
```
hichamboushaba, line 39: The `slug` is just the parent dir of the plugin file according to [this](https://github.com/WordPress/WordPress/blob/b14d7a814923369d0290a0f5a771a41f85bedd90/wp-admin/plugins.php#L311)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -49,10 +47,11 @@ open class WooCommerceStore @Inject constructor(
...
 ) : Store(dispatcher) {
-    enum class WooPlugin(val displayName: String) {
-        WOO_SERVICES("WooCommerce Shipping &amp; Tax"),
-        WOO_PAYMENTS("WooCommerce Payments"),
-        WOO_STRIPE_GATEWAY("WooCommerce Stripe Gateway"),
+    enum class WooPlugin(val pluginName: String) {
+        WOO_CORE("woocommerce/woocommerce"),
+        WOO_SERVICES("woocommerce-services/woocommerce-services"),
+        WOO_PAYMENTS("woocommerce-payments/woocommerce-payments"),
+        WOO_STRIPE_GATEWAY("woocommerce-gateway-stripe/woocommerce-gateway-stripe"),
```
hichamboushaba, line 57: I was torn between using the `slug` or the full plugin name here for filtering, but since all the endpoints use the name for fetching and not the `slug`, I decided to use it too.

0nko (approved):
Looks good 👍. Thanks for taking care of this, it's always a good idea to remove duplicate functionality.

PR #2267, opened 2022-02-03
Woo/update simple payment

nbradbury:
This PR addresses the final FluxC part of simple payments by adding the ability to update a simple payment. A test is included, along with the addition of a "Update simple payment" button in the orders page of the example app.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
@@ -19,7 +19,7 @@ class FeeLine {
     val totalTax: String? = null
 
     @SerializedName("tax_status")
-    val taxStatus: FeeLineTaxStatus? = null
+    val taxStatus: String? = null
 }
 
 enum class FeeLineTaxStatus {
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
@@ -22,9 +22,9 @@ class FeeLine {
     val taxStatus: FeeLineTaxStatus? = null
 }
 
-enum class FeeLineTaxStatus {
+enum class FeeLineTaxStatus(val value: String) {
     @SerializedName("taxable")
-    Taxable,
+    Taxable("taxable"),
     @SerializedName("none")
-    None
+    None("none")
 }
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
@@ -19,7 +19,7 @@ class FeeLine {
     val totalTax: String? = null
 
     @SerializedName("tax_status")
-    val taxStatus: String? = null
+    val taxStatus: FeeLineTaxStatus? = null
 }
 
 enum class FeeLineTaxStatus {
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
@@ -7,19 +7,19 @@ import com.google.gson.annotations.SerializedName
  */
 class FeeLine {
     @SerializedName("id")
-    val id: Long? = null
+    var id: Long? = null
 
     @SerializedName("name")
-    val name: String? = null
+    var name: String? = null
 
     @SerializedName("total")
-    val total: String? = null
+    var total: String? = null
 
     @SerializedName("total_tax")
-    val totalTax: String? = null
+    var totalTax: String? = null
 
     @SerializedName("tax_status")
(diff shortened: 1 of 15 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/FeeLine.kt
```diff
@@ -7,24 +7,24 @@ import com.google.gson.annotations.SerializedName
  */
 class FeeLine {
     @SerializedName("id")
-    val id: Long? = null
+    var id: Long? = null
```
hichamboushaba, line 10: np, why not keep all the properties immutable, but pass them as constructor parameters instead? I think it would be more concise for creating instances in the code in other areas.
We can also make the class as `data class` to have equality checks, which might come in handy in the future.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -483,19 +485,12 @@ class OrderRestClient @Inject constructor(
     )
 
     /**
-     * Creates a "simple payment," which is an empty order assigned the passed amount
+     * Creates a "simple payment," which is an empty order assigned the passed amount. The backend will
+     * return a new order with the tax already calculated.
      */
     suspend fun postSimplePayment(site: SiteModel, amount: String, isTaxable: Boolean): RemoteOrderPayload {
```
hichamboushaba, line 490: Just a question, do we need this separate function? I think calling `createOrder` from the store should be enough when passing the appropriate `UpdateOrderRequest` having a single `FeeLine` 🤔

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -1009,5 +1004,37 @@ class OrderRestClient @Inject constructor(
...
+         */
+        fun generateSimplePaymentFeeLineList(amount: String, isTaxable: Boolean, feeId: Long? = null): List<FeeLine> {
+            FeeLine().also { feeLine ->
+                feeId?.let {
+                    feeLine.id = it
+                }
+                feeLine.name = "Simple Payment"
+                feeLine.total = amount
+                feeLine.taxStatus = if (isTaxable) FeeLineTaxStatus.Taxable else FeeLineTaxStatus.None
+                return ArrayList<FeeLine>().also { it.add(feeLine) }
```
hichamboushaba: np, this can be simplified using `listOf(feeLine)`

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -1009,5 +1004,37 @@ class OrderRestClient @Inject constructor(
                 "tracking_number",
                 "tracking_provider"
         ).joinToString(separator = ",")
+
+        /**
+         * Generates the feeLines for a simple payment order containing a single fee line item with
+         * the passed information. Pass null for the feeId if this is a new fee line item, otherwise
+         * pass the id of an existing fee line item to replace it.
+         */
+        fun generateSimplePaymentFeeLineList(amount: String, isTaxable: Boolean, feeId: Long? = null): List<FeeLine> {
```
hichamboushaba: Since this function is used only in the store, I think we can move it there 🤔

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -260,4 +322,8 @@ class OrderUpdateStore @Inject internal constructor(
                 OnOrderChanged(orderError = OrderError(message = message))
         ))
     }
+
+    companion object {
+        private val gson by lazy { Gson() }
+    }
```
hichamboushaba: I don't think this is used anywhere 🙂

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -116,6 +117,67 @@ class OrderUpdateStore @Inject internal constructor(
...
     }
 
+    suspend fun updateSimplePayment(
+        site: SiteModel,
+        orderId: Long,
+        amount: String,
+        customerNote: String,
+        billingEmail: String,
+        isTaxable: Boolean
+    ): Flow<UpdateOrderResult> {
```
hichamboushaba, line 128: 💯

hichamboushaba (approved):
This looks good @nbradbury, I'm pre-approving, I left some comments, but nothing blocking.

And regarding the unit test, it's good, I have nothing to add.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
nbradbury: You're right, this isn't really necessary. My plan is to remove this as part of my "clean up" when simple payments is done.

PR #2268, opened 2022-02-04
Add timezone option parameter to site creation request

### LINKED ISSUE #14400: Site timezone defaults to UTC instead of user timezone
From user feedback session:

This comes up very often in testing so it's obvious it confuses people.

You schedule a post for later and it publishes immediately but you have no idea why. It's because new site timezones default to UTC instead of the user's location. This is not an intuitive experience.

ovitrif:
# Add timezone option parameter to site creation request
Fixes [#14400 Site timezone defaults to UTC instead of user timezone](https://github.com/wordpress-mobile/WordPress-Android/issues/14400)

## Description

Adds the possibility to specify the time zone programatically for sites created via the WP.com REST API.
The changes enable projects that depend on `WordPress-Flux-Android` to create WP.com sites in the time zone of the device.

Before this change, the default time zone of a new site was always UTC, which caused issues with posts that are scheduled to publish in the future.

**Related:**
- Integrated in [WordPress-Android PR # 15904](https://github.com/wordpress-mobile/WordPress-Android/pull/15904)

## To Test

I've updated existing unit tests to assert on the value of the `options.timezone_string` property inside the payload for site creation requests to the WP.com REST API.

**See changes in the following tests:**
- `SiteRestClientTest`
	- `creates new site without params with dry run`
	- `creates new site with all params`

**Integration**
- Integration can be tested with the [WordPress-Android PR # 15904](https://github.com/wordpress-mobile/WordPress-Android/pull/15904).

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
@@ -156,6 +156,21 @@ open class SiteStore
         constructor(
             siteName: String,
             language: String,
+            visibility: SiteVisibility,
+            dryRun: Boolean
+        ) : this(siteName, language, null, visibility, null, null, dryRun)
+
+        constructor(
+            siteName: String,
+            language: String,
+            visibility: SiteVisibility,
+            segmentId: Long?,
+            dryRun: Boolean
+        ) : this(siteName, language, null, visibility, segmentId, null, dryRun)
+
+        constructor(
+            siteName: String,
+            language: String,
             timeZoneId: String,
             visibility: SiteVisibility,
             dryRun: Boolean
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
@@ -156,21 +156,6 @@ open class SiteStore
         constructor(
             siteName: String,
             language: String,
-            visibility: SiteVisibility,
-            dryRun: Boolean
-        ) : this(siteName, language, null, visibility, null, null, dryRun)
-
-        constructor(
-            siteName: String,
-            language: String,
-            visibility: SiteVisibility,
-            segmentId: Long?,
-            dryRun: Boolean
-        ) : this(siteName, language, null, visibility, segmentId, null, dryRun)
-
-        constructor(
-            siteName: String,
-            language: String,
             timeZoneId: String,
             visibility: SiteVisibility,
             dryRun: Boolean
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
@@ -1,6 +1,7 @@
 package org.wordpress.android.fluxc.store
 
 import android.text.TextUtils
+import androidx.annotation.VisibleForTesting
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode.ASYNC
 import org.wordpress.android.fluxc.Dispatcher
@@ -1450,6 +1451,7 @@ open class SiteStore
         return rowsAffected
     }
 
+    @VisibleForTesting
     suspend fun createNewSite(payload: NewSitePayload): OnNewSiteCreated {
         val result = siteRestClient.newSite(
                 payload.siteName,
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
@@ -146,6 +146,7 @@ open class SiteStore
     data class NewSitePayload(
         @JvmField val siteName: String,
         @JvmField val language: String,
+        @JvmField val timeZoneId: String?,
         @JvmField val visibility: SiteVisibility,
         @JvmField val segmentId: Long? = null,
         @JvmField val siteDesign: String? = null,
@@ -154,6 +155,7 @@ open class SiteStore
         constructor(siteName: String, language: String, visibility: SiteVisibility, dryRun: Boolean) : this(
                 siteName,
                 language,
+                null,
                 visibility,
                 null,
                 null,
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -146,27 +147,34 @@ open class SiteStore
     data class NewSitePayload(
         @JvmField val siteName: String,
         @JvmField val language: String,
+        @JvmField val timeZoneId: String?,
```
ovitrif, line 150: I considered it intuitive (for me and other contributors) for the `timeZoneId` to be specified near `language`.

Another solution could be for FluxC to set the default `timeZoneId` automatically:
1. As default value when it's not specified from outside
2. As constant value, since we don't need the timezone to be specified from outside to fix [#14400](https://github.com/wordpress-mobile/WordPress-Android/issues/14400).
   Such level of extendability could be a sign of premature optimisation.

Both approaches can reduce the complexity of the fix.

I'm asking the reviewer of these changes for a second opinion on this.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -146,27 +147,34 @@ open class SiteStore
...
-                dryRun
-        )
+        constructor(
+            siteName: String,
+            language: String,
+            visibility: SiteVisibility,
+            dryRun: Boolean
+        ) : this(siteName, language, null, visibility, null, null, dryRun)
 
         constructor(
```
ovitrif, line 166: In Android Studio, with the WordPress-Android project open and composite builds enabled to point to a local clone of WordPress-FluxC-Android I see both this constructor and the one starting at line 156 as not being used.

I'm going to check also with [woocommerce-android](https://github.com/woocommerce/woocommerce-android) to see what I get and if they still end up as not used it would seem safe to remove them.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -146,27 +147,19 @@ open class SiteStore
...
         @JvmField val dryRun: Boolean
     ) : Payload<BaseNetworkError>() {
-        constructor(siteName: String, language: String, visibility: SiteVisibility, dryRun: Boolean) : this(
-                siteName,
-                language,
-                visibility,
-                null,
-                null,
-                dryRun
-        )
```
mkevins, line 164: I'm not quite sure why this is here, since the overloaded forms of this will be available anyway for the 2 (and now 3) optional parameters. My guess is that it is for convenience: allowing calling without names, and without needing to fill the "gaps" with `null` in that form. Having not found such usage makes me wonder if our search is exhaustive throughout all consumers, though. Perhaps it is safest to leave this in for now, and we could address cleanup in a separate PR to give time for wider feedback without blocking the feature in this PR. Wdyt?

mkevins (commented):
The code looks good (with just one suggestion about keeping the constructor). Also, I tested this via the steps in the related PR: https://github.com/wordpress-mobile/WordPress-Android/pull/15904 on a Pixel 3a (physical device) and the time zone is being set correctly.

I did notice that sometimes the scheduled posts still did not get published, though the time zone was set correctly, so I think this is a different issue, and not related with this PR.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
ovitrif: @mkevins Thank you for looking into this! 🙇 

I agree that it's best to keep this separate and review the search process in order to make sure we don't miss any consumers!

ovitrif:
> I did notice that sometimes the scheduled posts still did not get published, though the time zone was set correctly, so I think this is a different issue, and not related with this PR.

@mkevins Thank you for reviewing this PR! 🙇‍♂️

One of the reasons why the scheduled posts might not get published could be what was also explained on the companion [WP-Android PR #15904](https://github.com/wordpress-mobile/WordPress-Android/pull/15904):
> Due to a known API issue you have to visit the site to trigger the publishing of scheduled posts. You can tap the site url under the Site Title on the My Site screen for this.
> The issue was originally reported for the WordPress-iOS repo https://github.com/wordpress-mobile/WordPress-iOS/issues/10456

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
ovitrif: I reverted the removal of the additional constructors.

One thing I'd like to mention is that currently instead of updating the previously used constructor I have added another one to be sure these changes don't break integration from any consumers of `WordPress-FluxC-Android`.

Another approach could have been to update the constructor that was used before (the one starting at line 159).

mkevins:
> One of the reasons why the scheduled posts might not get published could be what was also explained on the companion [WP-Android PR #15904](https://github.com/wordpress-mobile/WordPress-Android/pull/15904):

I should have mentioned, I read that, and for the first post I tried, I visited the site after scheduling the post, but before it's scheduled time. That post never published (presumably because the site should be visited _prior to scheduling_?). Then, on my next attempt (having already visited the site before even scheduling the post) I scheduled a post, two minutes ahead, and that post almost immediately was published. The minute hand on my clock turned very soon after tapping publish, so there was likely only ~ a one minute difference between when I tapped "Schedule" and when it was scheduled. Perhaps the server's clock is slightly ahead of mine :man_shrugging: . So, I tried a few more times, with longer wait times: ~ 5 mins, for example. And those posts did not publish. Given that there are already API issues we are aware of with the scheduling mechanism, and that the focus of this PR is clearly met (i.e. updating the correct timezone on the site during site creation), this is not a blocker for this PR, imo, but worth noting and investigating as a separate issue.

ovitrif:
> Then, on my next attempt (having already visited the site before even scheduling the post) I scheduled a post, two minutes ahead, and that post almost immediately was published.

I had the same experience on my end one time, it was like one minute was not counted from the two that I set in advance to the current time.
Thank you for sharing this! 🙇

mkevins (approved):
Approved via https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2268#issuecomment-1033053919 (forgot to request myself as a re-reviewer :sweat_smile: ).

PR #2270, opened 2022-02-07
woo/simple-payment-billing-fix

nbradbury:
This small PR corrects a bug that sends an empty billing model when creating a simple payment without an email address. This causes the `POST` to fail with "Invalid parameter(s): billing." The fix is to skip sending a billing model if the email address is empty.


ThomazFB (approved):
LGTM :shipit:

PR #2271, opened 2022-02-07
Migrate newUpdateProductReviewStatusAction to a suspend function 

atomtec:
Closes: #[5647](https://github.com/woocommerce/woocommerce-android/issues/5647) . This is a dependency in FluxC for the main fix 

**Description**

The main change for  5647 is to decouple the   tight coupling between ReviewDetailFragment and ReviewListFragment. As one of the item for the same newUpdateProductReviewStatusAction was also required to be migrated to suspend function. This will allow removal of EventBus from the class implementing the logic of moderation.

**Changes** 
Migrate the function to suspend
Update the tests

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
@@ -30,7 +30,6 @@ import org.wordpress.android.fluxc.store.WCProductStore.RemoteAddProductTagsResp
 import org.wordpress.android.fluxc.store.WCProductStore.RemoteDeleteProductPayload
 import org.wordpress.android.fluxc.store.WCProductStore.RemoteProductCategoriesPayload
 import org.wordpress.android.fluxc.store.WCProductStore.RemoteProductListPayload
-import org.wordpress.android.fluxc.store.WCProductStore.RemoteProductReviewPayload
 import org.wordpress.android.fluxc.store.WCProductStore.RemoteProductShippingClassListPayload
 import org.wordpress.android.fluxc.store.WCProductStore.RemoteProductShippingClassPayload
 import org.wordpress.android.fluxc.store.WCProductStore.RemoteProductSkuAvailabilityPayload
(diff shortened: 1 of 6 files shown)


hichamboushaba (changes requested):
Nice work @atomtec 💯

But the PR is not ready for merge yet, there are still usages of the old API in the tests (check those [lines](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/e9da3c36c3b82591ab395b724b935bb2a9fa3a13/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCProductTest.kt#L479-L495))

And it would be good to remove the unused `Actions` now that we migrated everything to coroutines.

And one remark regarding the commit messages, please try to put some good expressive messages, it would be better to avoid messages like "first one" 🙏

atomtec:
> Nice work @atomtec 100
> 
> But the PR is not ready for merge yet, there are still usages of the old API in the tests (check those [lines](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/e9da3c36c3b82591ab395b724b935bb2a9fa3a13/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_WCProductTest.kt#L479-L495))
Yup
> 
> And it would be good to remove the unused `Actions` now that we migrated everything to coroutines.
Yes missed that 
> 
> And one remark regarding the commit messages, please try to put some good expressive messages, it would be better to avoid messages like "first one" pray
Ahh !! thanks for pointing out I was initially planning this as temp branch so was not careful there . Will keep in mind.

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
```diff
@@ -512,6 +512,7 @@ class MockedStack_WCProductsTest : MockedStack_Base() {
                 assertEquals(3, reviewerAvatarUrlBySize.size)
             }
         }
+        Unit
```
hichamboushaba, line 515: np, another way to fix this would be to explicitly specify the return type on the function's signature.

hichamboushaba (approved):
Nice work @atomtec, I won't merge this until the WCAndroid's PR is ready.

hichamboushaba:
Thanks again @atomtec, but since we can't merge this now without the WCAndroid changes, I'm closing this PR for now.

PR #2272, opened 2022-02-08
Update build.gradle

ravishanker:
This PR applies a work around to avoid the following error that crops up when doing a Composite build with local FluxC repo due to [sqlite-jdbc library used in Room](https://issuetracker.google.com/issues/174695268?pli=1)
```
 error: cannot find symbol
    private final XMLRPC mMethod;
```

It is fixed in Room ~ v2.4.0.  Until FluxC is upgraded to it,  this would prevent the issue for anyone going to use a M1 Mac (as more people are moving to M1 now) and hopefully saves a lot of time and frustration.

Thanks to @AjeshRPai and @JorgeMucientes who first investigated this issue (p1634023130210300-slack-C02QANACA, p1634815713005700-slack-C02QANACA).  Surprisingly I didn't face this issue till now, even though I have been on a M1 since the beginning!

To test:

Test 1
- Using an M1 Mac, checkout WordPress-FluxC-Android repo
- Build and Run example app on trunk
- Notice the above error

Test 2
- Using an M1 Mac
- Checkout this branch
- Build and Run example app 
- Notice it will build and run without the error

Test 3
- For completeness, Please test this fix on an Intel Mac to ensure there are no regression issues and works fine

NOTE:  I couldn't perform Test 3 as I don't have an Intel Mac.  Greatly appreciate it if you could please try it out.  Thanks

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,6 +46,8 @@ allprojects {
         configFile file("${project.rootDir}/config/checkstyle.xml")
     }
 
+    // See https://issuetracker.google.com/issues/174695268?pli=1,
+    // remove this when Room version is upgraded to 2.4.0 or later
     configurations.all {
         resolutionStrategy {
             force 'org.xerial:sqlite-jdbc:3.34.0'
--- a/build.gradle
+++ b/build.gradle
@@ -45,14 +45,6 @@ allprojects {
         toolVersion = '8.3'
         configFile file("${project.rootDir}/config/checkstyle.xml")
     }
-
-    // See https://issuetracker.google.com/issues/174695268?pli=1,
-    // remove this when Room version is upgraded to 2.4.0 or later
-    configurations.all {
-        resolutionStrategy {
-            force 'org.xerial:sqlite-jdbc:3.34.0'
-        }
-    }
 }
 
 subprojects {
--- a/build.gradle
+++ b/build.gradle
@@ -45,6 +45,12 @@ allprojects {
         toolVersion = '8.3'
         configFile file("${project.rootDir}/config/checkstyle.xml")
     }
+
+    configurations.all {
+        resolutionStrategy {
+            force 'org.xerial:sqlite-jdbc:3.34.0'
+        }
+    }
 }
 
 subprojects {
--- a/fluxc/build.gradle
+++ b/fluxc/build.gradle
@@ -87,6 +87,11 @@ dependencies {
     implementation 'org.apache.commons:commons-text:1.1'
     api "androidx.paging:paging-runtime:2.1.2"
     implementation "androidx.room:room-runtime:$roomVersion"
+
+    // See https://issuetracker.google.com/issues/174695268?pli=1,
+    // remove this when Room version is upgraded to 2.4.0 or later
+    kapt("org.xerial:sqlite-jdbc:3.36.0")
+
     kapt "androidx.room:room-compiler:$roomVersion"
     implementation "androidx.room:room-ktx:$roomVersion"
 
--- a/plugins/woocommerce/build.gradle
+++ b/plugins/woocommerce/build.gradle
@@ -75,6 +75,11 @@ dependencies {
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlinCoroutinesVersion"
 
     implementation "androidx.room:room-runtime:$roomVersion"
+
+    // See https://issuetracker.google.com/issues/174695268?pli=1,
+    // remove this when Room version is upgraded to 2.4.0 or later
+    kapt("org.xerial:sqlite-jdbc:3.36.0")
+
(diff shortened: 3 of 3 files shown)


build.gradle
```diff
@@ -45,6 +45,14 @@ allprojects {
         toolVersion = '8.3'
         configFile file("${project.rootDir}/config/checkstyle.xml")
     }
+
+    // See https://issuetracker.google.com/issues/174695268?pli=1,
+    // remove this when Room version is upgraded to 2.4.0 or later
```
AliSoftware: 👍 for the comment; for visibility I'd suggest to add a similar comment next to the Room dependency declaration (I didn't check where that was), referencing that line/workaround, so that it'd be more visible when we change said line about Room version later (otherwise I feel I'd be pretty easy to miss this comment here while updating Room version from on a completely different place in gradle?) 🤔

ParaskP7: 👍 from me as well. I would also recommend creating an GitHub issue for upgrading Room to 2.4.0 where everything will be explained there as well. Then, maybe adding a comment next to the Room dependency declaration that point to that newly created issue, just like Olivier recommended. Wdyt?

ParaskP7 (dismissed):
👋 @ravishanker !

> Test 3
> 
> For completeness, Please test this fix on an Intel Mac to ensure there are no regression issues and works fine
> 
> NOTE: I couldn't perform Test 3 as I don't have an Intel Mac. Greatly appreciate it if you could please try it out. Thanks

As I don't have M1, but do have an Intel Mac, I only tested this `Test 3` scenario. Everything seems to be working as expected. I build/installed both the `FluxC` example app and the `WordPress` app, through composite build, and was able to play around a bit with both. Thus, it is a partial approval from my side.

build.gradle
JorgeMucientes: Thanks for tackling this @ravishanker. While I don't have a strong opinion on applying this changes I think we might run into potential side effects or risks of forcing this specific version of `sqlite-jdbc` for everyone, and specially releasing the app with this forced version. I mean there's probably no issue at all, but I'd prefer to fix this locally for the devs experiencing the issue. 

One way to do that is how I suggested in the P2 troubleshooting this problem: pbMoDN-2Zf-p2 

Basically, using Gradle initialization scripts. The tradeoff is that everyone will have to deal with this manually, and be aware of the existing solution.
The pros are we don't have to change production code, assuming unknown side effects, and we won't have to take care of the migration/removal of this code once we update to room `2.4.0`.

Let me know what you think about it 😃

ParaskP7: I actually agree with Jorge on that. 😊 👍 

PS: I thought that the situation changed and now with Jetpack Compose we might want to be more aggressive about that, but it is better to be safe and do it right after all.

ParaskP7 (commented):
Pointing To: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2272#discussion_r801504044

build.gradle
ravishanker: Thanks all for reviewing this.  I hope the Room upgrade  happens sooner than later now given @ParaskP7 has joined Platform9 😀

I've now moved the comments and dependency before the room dependency.  

Also [created an issue](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/2279) to upgrade Room version to 2.4.0 or later

I'm going to leave this PR here for sometime.  Please feel free to close it when  unnecessary.

ovitrif: Great initiative! Though I'm also leaning towards going for the room `2.4.0` update if that's feasible :)

I can say the pbMoDN-2Zf-p2 internal post was easy to discover by myself 🙇

ParaskP7: 👋 @ravishanker !

> Thanks all for reviewing this. I hope the Room upgrade happens sooner than later now given @ParaskP7 has joined Platform9 😀

😅 🤞 

> Also https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/2279 to upgrade Room version to 2.4.0 or later

Thanks so much for creating this issue Ravi. 🙇

Since I am a bit busy with Platform9 onboarding and the `Gradle+AGP` upgrade, which takes priority over everything else, I recommend someone from `WP/WCAndroid` to pick this Room upgrade instead of waiting for it to happen by me, especially since this seems to becoming more and more of a blocker with more people onboarding to M1.

> I'm going to leave this PR here for sometime. Please feel free to close it when unnecessary.

I recommend you closing this PR @ravishanker , pointing to the new Room upgrade issue you opened. We can reopen this PR later on if necessary.

PS: Thank you so much for working on this Ravi, and everyone else for your inputs! 🥇

ravishanker:
Closing it and created an issue [created an issue](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/2279) to upgrade Room version to ~ 2.4.0 or latest

PR #2274, opened 2022-02-08
Add default value for `taxLines` migration

### LINKED ISSUE #2273: SQLiteException: Cannot add a NOT NULL column with default value NULL (code 1): , while compiling: ALTER TABLE Ord...
Sentry Issue: [WOOCOMMERCE-ANDROID-390](https://sentry.io/organizations/a8c/issues/2995780157/?referrer=github_integration)

```
SQLiteException: Cannot add a NOT NULL column with default value NULL (code 1): , while compiling: ALTER TABLE OrderEntity ADD taxLines TEXT NOT NULL
    at android.database.sqlite.SQLiteConnection.nativePrepareStatement(SQLiteConnection.java)
    at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:890)
    at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:501)
    at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:588)
    at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:58)
...
```

wzieba:
Fixes: #2273 

### Description

We’re altering database with a new non-null column in `MIGRATION_6_7` but we don’t provide a default value for that. We do have a default value but it's declared in `WCOrderModel` Kotlin class but during SQL migration, if there are existing values, SQL wouldn’t know how to behave (at the moment of migration).

### Testing

1. Checkout this branch
2. Remove any changes in `MIGRATION_6_7` (remove ` DEFAULT ''` part)
3. Run `testMigrate6to7()` test (on any device)
4. **Assert that test is crashing**
5. Revet change made in 2
6. Run `testMigrate6to7()` again
7. **Assert that test is not crashing**

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
@@ -142,6 +142,6 @@ internal val MIGRATION_5_6 = object : Migration(5, 6) {
 
 internal val MIGRATION_6_7 = object : Migration(6, 7) {
     override fun migrate(database: SupportSQLiteDatabase) {
-        database.execSQL("ALTER TABLE OrderEntity ADD taxLines TEXT NOT NULL")
+        database.execSQL("ALTER TABLE OrderEntity ADD taxLines TEXT NOT NULL DEFAULT ''")
     }
 }
--- a/plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
+++ b/plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
@@ -49,8 +49,23 @@ class MigrationTests {
     @Test
     fun testMigrate6to7() {
         helper.apply {
-            createDatabase(TEST_DB, 6).close()
-            runMigrationsAndValidate(TEST_DB, 7, true, MIGRATION_6_7)
+            val existingDb = createDatabase(TEST_DB, 6).apply {
+                execSQL(
+                        // language=RoomSql
+                        """
+                            INSERT INTO OrderEntity VALUES(1, 2, 3, '123', 'processing', '$', 'key', 'date of creation', 'date of modification', '123', '456', '789', 'card', 'by card', 'date paid', TRUE, 'sample customer note', '213', 'CODE', 123, 'billing first name', 'billing last name', 'billing company', 'billing address1', 'billing address2', 'billing city', 'billing state', 'billing postcode', 'billing country', 'billing email', 'billing phone', 'shipping first name', 'shipping last name', 'shipping company', 'shipping address1', 'shipping address2', 'shipping city', 'shipping state', 'shipping postcode', 'shipping country', 'shipping phone', 'line items', 'shipping lines', 'fee lines', 'meta data')
+                        """.trimIndent()
+                )
+            }.close()
+
+            val migratedDb = runMigrationsAndValidate(TEST_DB, 7, true, MIGRATION_6_7)
+
+            migratedDb.query(
+                    // language=RoomSql
+                    """
+                        SELECT * FROM OrderEntity
+                    """.trimIndent()
+            )
         }
     }
 


hafizrahman:
Tested this on a Samsung S21 device.

Can confirm that `testMigrate6to7` fails when ` DEFAULT ''` is removed, and succeeds with it added.

wzieba:
That's awesome. Could you please take a look at code too @hafizrahman ? So we'll have ready fix for `1.35.1` FluxC release.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
```diff
@@ -142,6 +142,6 @@ internal val MIGRATION_5_6 = object : Migration(5, 6) {
 
 internal val MIGRATION_6_7 = object : Migration(6, 7) {
     override fun migrate(database: SupportSQLiteDatabase) {
-        database.execSQL("ALTER TABLE OrderEntity ADD taxLines TEXT NOT NULL")
+        database.execSQL("ALTER TABLE OrderEntity ADD taxLines TEXT NOT NULL DEFAULT ''")
```
hichamboushaba, line 145: I wonder if this the best solution for our case, as explained [here](https://developer.android.com/training/data-storage/room/migrating-db-versions#handle-default-values-migrations), this might create a discrepancy between migration path, and new install paths.
The suggested solution is to add the `ColumnInfo` annotations, then add a migration that drops and recreates the table, and I think this would make passing to auto-migration in the future easier, otherwise, we may have the same issue when we enable it, since Room won't know how to handle default value.

hichamboushaba: I think for the release, we can go ahead and merge this, but we need to define a strategy for migrations in the future, taking into consideration auto-migrations that will come with 2.4.0.
It's pretty easy to forget about the `DEFAULT` annotation here, so this might come and bite us again if we don't define good rules about the migration process.

wzieba: Yeah, it's exactly our case described in the link you provided, thanks! I'll iterate on that.

wzieba: @hichamboushaba if I understood correctly, the discrepancy we'll experience would be that
|                      | Fresh installation schema | Post migration schema                |
|----------------------|---------------------------|--------------------------------------|
| OrderEntity#taxLines | `taxLines TEXT NOT NULL`  | `taxLines TEXT NOT NULL  DEFAULT ''` |

and I agree that we should've remove any discrepancies just for sake of having not confusing setup but also as you say it might influence auto-migration.

Although according to documentation we should experience schema validation. Do you see why we don't experience this? It should happen I believe when running `testMigrate6to7` as we start with database in version 6 (so `Fresh installation schema` e.g.) and then we migrate to 7 which creates discrepancy 🤔 .

---

Second matter would be: as version schema 7 wasn't released publicly, does it make sense to you to forcefully re-generate schema 7 with `ColumnInfo` annotation over `taxLines` to save overhead of drop & recreate strategy or not especially? What's your opinion on this?

hichamboushaba: > Although according to documentation we should experience schema validation. Do you see why we don't experience this?

Because we don't have the `ColumnInfo` right now, Room doesn't know we added a default value in the migration but not on the fresh creation, so it ignore the discrepancy, once we add the annotation, we should be warned by Room that we need to update the version.

> as version schema 7 wasn't released publicly, does it make sense to you to forcefully re-generate schema 7 with ColumnInfo annotation over taxLines to save overhead of drop & recreate strategy or not especially? What's your opinion on this?

I don't really have a strong opinion on this, my gut says we should annotate all properties that have default values with `ColumnInfo`, not just `taxLines`, but maybe this can wait.
WDYT about going with the merge of your PR as is, since it fixes the issue, and it's an urgent fix? then we can add a second PR that updates the whole table to annotate any column that needs a default value with `ColumnInfo`, and add a drop/recreate migration. Later, we can document this to avoid having this issue in the future.

wzieba: > WDYT about going with the merge of your PR as is, since it fixes the issue, and it's an urgent fix? then we can add a second PR that updates the whole table to annotate any column that needs a default value with ColumnInfo, and add a drop/recreate migration. Later, we can document this to avoid having this issue in the future.

Sounds good - it'd be the best to annotate every having-default-value column with `ColumnInfo` in one PR.

So merging this one and preparing ticket for `ColumnInfo` work. Thank you for your awareness on that!

PR #2275, opened 2022-02-08
[Woo] ability to delete order

hichamboushaba:
This PR just adds support for deleting an order, this is needed for the task https://github.com/woocommerce/woocommerce-android/issues/5823 to cleanup the draft order created to calculate taxes.

#### Test
1. Open the example app.
2. Click on Woo, then Orders.
3. Click on "Delete an order"
4. Type an order id, and choose whether to move the order to trash or to delete it permanently.
5. Confirm that the order is deleted.


ThomazFB (approved):
Looks good and works as expected! Thanks, Hicham! Leaving this approved and moving to the Woo counterpart PR now.

PR #2280, opened 2022-02-09
Xmlrpc login errors.

develric:
IMPORTANT: this is ready to be reviewed; but since it has implications for Woo as well, keeping it as a draft so it is not accidentally merged before the validation from WPAndroid and WooAndroid point of view.

This is part of [12401 issue](https://github.com/wordpress-mobile/WordPress-Android/issues/12401).

I tried to keep the current behaviors not affected, only adding a couple of enums mapping the various errors conditions for xml-rpc, that are used/parsed only in the login flow for self-hosted for now. This should reduce the possibility of regressions but appreciate testing of the various log-in/sign-up scenarios 🙇 .

See the [companion WPMobile PR](https://github.com/wordpress-mobile/WordPress-Android/pull/15919) for more details.


fluxc/src/main/java/org/wordpress/android/fluxc/utils/SiteErrorUtils.java
```diff
@@ -14,6 +15,23 @@ public static SiteError genericToSiteError(BaseNetworkError error) {
                     break;
             }
         }
-        return new SiteError(errorType, error.message);
+
+        SiteError siteError = new SiteError(errorType, error.message);
```
khaykov, line 19: A very minor NP - maybe we can avoid a double initialization of `siteError` here? If it makes sense :)

khaykov (approved):
Looks good to me! No issues when testing with wpandroid client.

khaykov:
Looks like there are no changes since my review, so good to go :)

PR #2250, opened 2022-01-27
Renaming of WCOrderModel to OrderEntity

kidinov:
The PR renames `WCOrderModel` to `OrderEntity`. Can be tested with https://github.com/woocommerce/woocommerce-android/pull/5717

PR #2281, opened 2022-02-09
[Do not merge] POC for moving Order's Line Items to a seperate database entity

wzieba:
This PR is only for purpose of discussions. I've timeboxed some time for this POC so before reviewing it mind that:

1. There are a few changes that make no sense - just for sake of making it compile
2. Naming is not perfect - probably we could come with better names
3. I made some exsiting classes name changes for sake of clarity

### Description

This PR is a POC of spotting how much effort would be to move `WCOrderModel#*Items` into a separate database entity. For starter, I choose `LineItems`. 

### Screenshot of database content

Running this PR and launching `Fetch Orders` from example menu should result with having content in `LineItemEntity` table in database.

Screenshot 2022-02-09 at 12 17 41

### Database diagram

`OrderEntity` (aka `WCOrderModel`) is connected with `LineItemEntity` via foreign key:

Screenshot 2022-02-09 at 12 19 20


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -9,13 +10,18 @@ import com.google.gson.reflect.TypeToken
...
 import org.wordpress.android.fluxc.model.order.ShippingLine
 import org.wordpress.android.fluxc.model.order.TaxLine
+import org.wordpress.android.fluxc.model.order.entities.LineItemEntity
 import java.math.BigDecimal
 import kotlin.DeprecationLevel.ERROR
 
+data class RichOrderEntity(
+    val orderInfo: WCOrderModel,
+    val lineItems: List<LineItemEntity>
+)
```
wzieba, line 23: Ideally, I'd use `@Embeded` annotation and let Room to handle joining `orderInfo` with its `lineItems` but, unfortunately, this strategy doesn't work if entities are connected via composite primary key.

It's not a blocker though, we can do this manually.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
```diff
@@ -101,19 +107,12 @@ data class WCOrderModel(
      */
     fun getShippingAddress() = OrderAddress.Shipping(this)
 
-    /**
-     * Deserializes the JSON contained in [lineItems] into a list of [LineItem] objects.
-     */
-    fun getLineItemList(): List<LineItem> {
-        val responseType = object : TypeToken<List<LineItem>>() {}.type
-        return gson.fromJson(lineItems, responseType) as? List<LineItem> ?: emptyList()
-    }
```
wzieba, line 110: That's the key part. Line items are now not deserialized every time we need them.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/entities/LineItemEntity.kt
```diff
@@ -0,0 +1,30 @@
...
+    val name: String? = null,
+    val parentName: String? = null,
+    var productId: Long? = null,
+    val variationId: Long? = null,
+    var quantity: Float? = null,
+    val subtotal: String? = null,
+    val total: String? = null, // Price x quantity
+    val totalTax: String? = null,
+    val sku: String? = null,
+    val price: String? = null, // The per-item price
```
wzieba, line 29: In this entity, comparing to `LineItemDto`, `List<WCMetaData>` is missing.

I run out of time but I believe, due to unpredictable nature of `WCMetaData` and its `Any` fields, we should keep raw JSON string for it. I can't see a way to handle it otherwise.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/entities/LineItemSerializer.kt
```diff
@@ -0,0 +1,34 @@
+package org.wordpress.android.fluxc.model.order.entities
+
+import com.google.gson.Gson
+import com.google.gson.JsonElement
+import com.google.gson.reflect.TypeToken
+import org.wordpress.android.fluxc.model.WCOrderModel
+import org.wordpress.android.fluxc.model.order.LineItemDto
+import javax.inject.Inject
+
+class LineItemSerializer @Inject constructor(private val gson: Gson) {
```
wzieba, line 10: (A little out of topic) This is part I personally like pretty much. I believe that in `FluxC` we construct too many `Gson` instances. I believe there should be only one, with the same configuration.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -103,14 +103,11 @@ class OrderRestClient @Inject constructor(
                 "_fields" to ORDER_FIELDS)
         val request = JetpackTunnelGsonRequest.buildGetRequest(url, site.siteId, params, responseType,
                 { response: List<OrderDto>? ->
-                    val orderModels = response?.map { orderDto ->
-                        orderDto.toDomainModel(site.localId())
-                    }.orEmpty()
```
wzieba, line 108: We should remove mapping to database type from `OrderRestClient` anyway.

This part is my fault because in the past I wrongly interpreted that `WCOrderModel` is treated as `domain` object. In fact, it's not - that's just database entity and we'll adjust it's name in #2250. We do not have `domain` object anywhere.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
```diff
@@ -60,4 +73,20 @@ abstract class OrdersDao {
...
+    open fun observeRichOrderEntity(remoteOrderId: RemoteId, localSiteId: LocalId): Flow<RichOrderEntity?> {
+        return combine(
+                observeOrder(remoteOrderId, localSiteId),
+                observeLineItems(remoteOrderId, localSiteId)
+        ) { order, lineItems ->
+            if (order != null)
+                RichOrderEntity(order, lineItems)
+            else
+                null
+        }
```
wzieba, line 90: This is other key part: joining `OrderEntity` with its `List<LineItem>`

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
hichamboushaba: I don't think `@Embeded` would work with a list of items, it's just for single items, and it's very basic, it would add all of the object's properties as columns on the parent table.
Room does [support](https://developer.android.com/training/data-storage/room/relationships#one-to-many) other relationships though, I [used](https://github.com/hichamboushaba/WCStoreApp/blob/5e2c5686b1952042b661c02223ea219bcf0c0b99/app/src/main/java/com/hicham/wcstoreapp/data/db/entities/CartEntity.kt#L24-L32) them in the WCStoreApp for the cart entity with the items, and they work fine.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
hichamboushaba: > We do not have domain object anywhere.

I don't know if this is the case, my understanding is that `WCOrderModel` was intended as a domain model, but it was leaked to the storage layer. Now if we are going to rename the database model to `OrderEntity` (which I like btw), I think we should introduce a new domain model to replace the usage of `WCOrderModel`, this would allow not exposing the database entity through the `Stores`.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
hichamboushaba: If we use Room's relationships, we'll be able to directly reference the `Rich` class through the Dao, and it will be done on a single SQLite transaction, which is better performance wise (for example on WCStoreApp https://github.com/hichamboushaba/WCStoreApp/blob/5e2c5686b1952042b661c02223ea219bcf0c0b99/app/src/main/java/com/hicham/wcstoreapp/data/db/daos/CartDao.kt#L12-L14)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
hichamboushaba: > this strategy doesn't work if entities are connected via composite primary key.

After checking more, even `@Relation` doesn't work with composite keys, but I would still prefer to support the relationship by Room instead of doing it manually (IMHO).
I think we can either keep the local id (while marking it as internal maybe?), since now we know it's not used anywhere on the business code, and won't cause any confusion as before.
The other choice would be to keep the manual code, until we update to Room 2.4.0, where it will have a proper support for JOIN queries directly in the dao (https://issuetracker.google.com/issues/64247765), which would be equivalent to the `@Relation` annotation.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
hichamboushaba: Giving this more thought, I think you are right, even without adding another class, exposing `OrderEntity` from the Stores won't be harmful, and we already have a domain model in WCAndroid (even though it's more of a presentation model due to the use of @Parcelize 😄), but less layering would mean less complicated code.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
```diff
@@ -60,4 +73,20 @@ abstract class OrdersDao {
 
     @Query("SELECT COUNT(*) FROM OrderEntity WHERE localSiteId = :localSiteId")
     abstract fun getOrderCountForSite(localSiteId: LocalId): Int
+
+    @Query("SELECT * FROM LineItemEntity WHERE parentOrderId = :parentOrderId AND parentOrderLocalSiteId = :parentOrderLocalSiteId")
+    abstract fun observeLineItems(parentOrderId: RemoteId, parentOrderLocalSiteId: LocalId): Flow<List<LineItemEntity>>
+
+    @Transaction
+    open fun observeRichOrderEntity(remoteOrderId: RemoteId, localSiteId: LocalId): Flow<RichOrderEntity?> {
```
kidinov, line 81: What if we won't have `RichOrderEntity` at all, but the client, if needed will just fetch `LineItems` for order separately? I feel this way API exposed by fluxc will be easier to follow and a bit simpler

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt
wzieba: > I don't think @Embeded would work with a list of items, it's just for single items, and it's very basic, it would add all of the object's properties as columns on the parent table.

Yeah I was thinking about `@Embeded` *and* `@Relation` - the same as in e.g. `AddonWithOptions`

Thanks for research on that! Awesome it'll be available in Room 2.4.0. I would rather go with this solution than keeping local id for order. Especially that Room 2.4.0 will bring other, awaited improvements (M1 support, AutoMigration)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
wzieba: Nice thoughts, I feel the same. Part of me *really* wants to introduce platform-agnostic, pure domain model but other part sees that it might be not extremely valuable in this very moment. Maybe exposing entity from `*Store` and mapping it to "presentation/wannabe-domain" model in WCAndroid is good for now. As you say, it's for sure less complicated.

But I have hopes that some day we'll introduce domain layer in app either by introducing new model or removing platform-specific code from WCAndroid's `Order`.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
wzieba: Yes, this is something we have to establish. I see above @hichamboushaba preferring to leave this joins to Room.

But maybe we don't ever want to have `LineItems` joined with `Order`. So, just to make sure I understand, you're suggesting removing `lineItems` field from WCAndroid's `Order` and request `lineItems` only when needed right?

kidinov: > But maybe we don't ever want to have LineItems joined with Order. So, just to make sure I understand, you're suggesting removing lineItems field from WCAndroid's Order and request lineItems only when needed right?

Yep, exactly this

hichamboushaba: Just my 2 cents, personally, I prefer leaving the join to the DB since it's more performant, as it will be executed on a single SQL transaction.
Also, I think for most cases where we need an `Order`, we also need its line items. The exception is the usage in lists, and for those, we have the special model `WCOrderSummaryModel`, which contains only minimal data.

kidinov: @hichamboushaba Shouldn't it be slower because we run it inside a transaction? Don't know the specific, but we might be blocking at least 2 tables in this case, or maybe even the whole DB. And we might avoid transactions at all in case of fetching those 2 tables separately, at least we don't retrieve orders inside a transaction now and apparently, we didn't have any issues with consistency of the data just due to the low concurrency nature of how we use that

hichamboushaba: > @hichamboushaba Shouldn't it be slower because we run it inside a transaction?

I think this depends on what you mean, if you mean it's slower than just fetching Orders without the LineItems, then sure it will be slower, but if you mean slower than fetching orders, then joining the line items in code behind, I think it will be faster, since it will have a single round-trip to the database.
Just to be clear, I'm not saying this implementation will be faster, for this one I'm not sure, as I didn't check the code that Room generates behind. What I mean is the implementation that Room optimizes when either using the `@Relation` annotation, or when supporting the SQL `JOIN` statements in 2.4.0.

> at least we don't retrieve orders inside a transaction now and apparently, we didn't have any issues with consistency of the data just due to the low concurrency nature of how we use that

But now we have everything in a single table 😕, I don't see how this applies now, can you explain what you mean by this part? 🙂

kidinov: >   then joining the line items in code behind, I think it will be faster, since it will have a single round-trip to the database.

Won't argue here, because didn't check, but my gut feeling is that it will be slower because transactions are usually expensive. Maybe it won't  be slow on the query itself, but because of the locking it might slow other query still

> But now we have everything in a single table 😕, I don't see how this applies now, can you explain what you mean by this part? 🙂

From the doc:

When used on a Query method that has a SELECT statement, the generated code for the Query will be run in a transaction. There are 2 main cases where you may want to do that:

* If the result of the query is fairly big, it is better to run it inside a transaction to receive a consistent result. Otherwise, if the query result does not fit into a single [CursorWindow](https://developer.android.com/reference/android/database/CursorWindow.html), the query result may be corrupted due to changes in the database in between cursor window swaps.
* If the result of the query is a POJO with [Relation](https://developer.android.com/reference/androidx/room/Relation) fields, these fields are queried separately. To receive consistent results between these queries, you also want to run them in a single transaction.

Apparently, we want to run this query inside a transaction because we are afraid of potential data inconsistency (case 2)? I'm saying that if we didn't fetch `Orders with lineItems` inside a transaction (because of case 1), then I assume we don't need transactions if we would fetch them in 2 separate queries too (CursorWindow will be even smaller than we have right now)

hichamboushaba: I think we are discussing two different subjects, I'm not saying the usage of `@Transaction` will make it faster, I'm saying having a single SQL statement will make it faster, and this can be achieved using the `JOIN` statement in 2.4.0, and unless I'm mistaken, it's already done by the `@Relation` in the current version (I didn't check the generated code behind to confirm this).
So I prefer keeping the joining on the Dao as @wzieba did, to make simpler to migrate to the `JOIN` statement after updating to 2.4.0.

Anyway, I think the performance impact will be minimal and not noticeable on any case, so we should just lean towards the clearest and easiest to read solution, and personally, I prefer hiding the joining part on the `DAO`, but that's subjective.

> Apparently, we want to run this query inside a transaction because we are afraid of potential data inconsistency (case 2)? I'm saying that if we didn't fetch Orders with lineItems inside a transaction (because of case 1), then I assume we don't need transactions if we would fetch them in 2 separate queries too (CursorWindow will be even smaller than we have right now)

I see, thanks for the clarification.

wzieba:
Thanks @hichamboushaba @kidinov for suggestions and discussion. For now though, I'm closing this POC PR as there seem to be much easier and straightforward solution to our problem. More details: p91TBi-7qQ-p2#comment-8528

PR #2283, opened 2022-02-09
Make statementDescriptor on PaymentAccount nullable

malinajirka:
Stripe Extension version 6.2.0+ returns null instead of an empty string when the statement descriptor is not set.

[WPAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/5836)


AnirudhBhat:
@malinajirka Thanks for taking care of this! The changes look good to me :shipit:

AnirudhBhat (approved):

PR #2276, opened 2022-02-08
Woo/simple payments cleanup

nbradbury:
This is the first of two FluxC PRs to clean up handling of simple payments. In this PR, creating simple payments is added to `OrderUpdateStore` and simple payment creation is marked as deprecated in `OrderRestClient`. Once WCAndroid switches to the `OrderUpdateStore` function, we can drop the one in `OrderRestClient` and complete the cleanup.

PR #2284, opened 2022-02-09
Woo/simple payments cleanup p2

nbradbury:
This is a companion to #2276, which added creating a simple payment to the order update store. In this PR, the deprecated `postSimplePayment` code has been removed (as of [this PR](https://github.com/woocommerce/woocommerce-android/pull/5837), it's no longer used by WCAndroid so it's safe to be removed).


hichamboushaba (approved):
Looks good :shipit:

PR #2287, opened 2022-02-10
Bump Kotlin to `1.6.10` and Room to `2.4.2`

### LINKED ISSUE #2279: Upgrade Room version to 2.4.0 or latest
At the moment FluxC uses Room v2.3.0.  

This needs to be upgraded to v2.4.0 or later to prevent the following error that crops up when doing a Composite build with local FluxC repo due to [sqlite-jdbc library used in Room](https://issuetracker.google.com/issues/174695268?pli=1) on an M1 Mac
```
 error: cannot find symbol
    private final XMLRPC mMethod;
```

As more people are starting use M1 Macs, it would be good to do this at the earliest possible.

There's also a work around [PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2272) put on hold to fix this.  Please either merge it if required or close it when this upgrade happens.

wzieba:
Closes #2279 

This PR bumps Room dependency to `2.4.2` and Kotlin to `1.6.10`.

### Tests issues
~There are failing tests but I believe all of them are caused by the same reflection-related issue. It seems that `deliverResponse.invoke(request, o.result)` stopped working as it was. In the end `XMLRPCRequestBuilder#buildGetRequest` receives `null` in `Response.Listener` (lambda) when it should received a value declared by `deliverResponse.invoke(request, o.result)` invocation.~

~I have no ideas why it's happening. @wordpress-mobile/raven could you please take a look at this?~

See comments below and internal communication: paqN3M-ti-p2


wzieba:
I've rebased the branch on top of #2258 as Kotlin's 1.6.10 Lint [conflicts](https://issuetracker.google.com/issues/185418482#comment14) with AGP 4.2

zwarm:
@wzieba - Pinging @develric for you re: failing tests

develric:
Just sharing I was looking into this a bit last week and in the end the issue is a cast exception when doing `listener.onResponse(response);` [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/1436f46fc296ede0e65e117bbfced38fa34cec6d/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequest.java#L158) that then calls a `listener.onResponse(null);` with the null argument in the catcher exception [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/1436f46fc296ede0e65e117bbfced38fa34cec6d/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequest.java#L166).

For the record and in case it rings any bell, when debugging the [testFetchSite](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/1436f46fc296ede0e65e117bbfced38fa34cec6d/example/src/test/java/org/wordpress/android/fluxc/site/SiteXMLRPCClientTest.kt#L81) test, the exception I get there is:

```
java.lang.ClassCastException: class java.util.HashMap cannot be cast to class [Ljava.lang.Object; (java.util.HashMap and [Ljava.lang.Object; are in module java.base of loader 'bootstrap')
```

Don't have an actionable idea tbh, trying to spend another bit of time along this week 🤔 . cc @wzieba

develric:
Was looking into this a bit more during Hack Week. I used the example fluxc app to check the fetchSite call (button `Fetch First Site` in the example app):
- Just to be sure, I was checking the xmlrpc request at network layer and as expected confirmed no changes from using the code in this PR. 
- I also confirmed the cast exception is present at runtime with the example app as well (so the unit test failures are actually capturing a real runtime failure we will get; unit testing ftw :) )

The cast exception seems related to the fact that the `XMLRPCSerializer -> deserialize` [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/1436f46fc296ede0e65e117bbfced38fa34cec6d/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCSerializer.java#L200) returns an Object but the underline code can return an Object, an Object[] or an [HashMap](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/1436f46fc296ede0e65e117bbfced38fa34cec6d/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCSerializer.java#L261).

Since the `XMLRPCRequest  -> mListener` is declared as a generic on `? super Object[]`,  the cast applied [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/1436f46fc296ede0e65e117bbfced38fa34cec6d/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequest.java#L158) fails on HashMap data.

Was not able to dig to the end of why that cast was allowed and working before the code in this PR, but tried to get a  dirty code [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2311) moving the Client listeners to using a sealed class. 

The code there is more sort of exploration poc and it's not clean code for a PR, also maybe we could find a way to make the cast working again 🤔 (maybe there is some trivial way I was not thinking at 🙃). This said, I'm not sure if I will be able to look into this soon, so I thought it could be worth sharing that branch as a possible path to further dig. 

I did a few smoke testing (so I expect there can be issues I didn't see) but at least the unit tests seem happy 😄 ! cc @wzieba, let's further align on this one 🙇

wzieba:
Thanks @develric for update on that!

> I also confirmed the cast exception is present at runtime with the example app as well

! good that you found it, initially I thought it's only test-related issue. Thank you also for sharing POC of solution, I can't work on that right now but I'll try to prioritise it and find some capacity as eventually it'll be blocker not only for Room.

cc: @ParaskP7 as you're might be interested as well 🙌.

wzieba:
@develric 👋 

> I also confirmed the cast exception is present at runtime with the example app as well (...)

Do you remember the reproduction steps for this?

wzieba:
<details><summary>Dependency diff (click me)</summary>
<p>

```diff
-+--- androidx.room:room-ktx:2.3.0
-|    +--- androidx.room:room-common:2.3.0
-|    |    \--- androidx.annotation:annotation:1.1.0
-|    +--- androidx.room:room-runtime:2.3.0
-|    |    +--- androidx.room:room-common:2.3.0 (*)
-|    |    +--- androidx.sqlite:sqlite-framework:2.1.0
-|    |    |    +--- androidx.annotation:annotation:1.0.0 -> 1.1.0
-|    |    |    \--- androidx.sqlite:sqlite:2.1.0
-|    |    |         \--- androidx.annotation:annotation:1.0.0 -> 1.1.0
-|    |    +--- androidx.sqlite:sqlite:2.1.0 (*)
-|    |    +--- androidx.arch.core:core-runtime:2.0.1
-|    |    |    +--- androidx.annotation:annotation:1.0.2 -> 1.1.0
-|    |    |    \--- androidx.arch.core:core-common:2.0.1 -> 2.1.0
-|    |    |         \--- androidx.annotation:annotation:1.1.0
-|    |    \--- androidx.annotation:annotation-experimental:1.1.0
-|    +--- org.jetbrains.kotlin:kotlin-stdlib:1.4.31
-|    |    +--- org.jetbrains.kotlin:kotlin-stdlib-common:1.4.31
-|    |    \--- org.jetbrains:annotations:13.0
-|    \--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.4.1
-|         +--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1
-|         |    \--- org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.4.1
-|         |         +--- org.jetbrains.kotlin:kotlin-stdlib:1.4.0 -> 1.4.31 (*)
-|         |         \--- org.jetbrains.kotlin:kotlin-stdlib-common:1.4.0 -> 1.4.31
-|         \--- org.jetbrains.kotlin:kotlin-stdlib:1.4.0 -> 1.4.31 (*)
++--- androidx.room:room-ktx:2.4.1
+|    +--- androidx.room:room-common:2.4.1
+|    |    \--- androidx.annotation:annotation:1.1.0
+|    +--- androidx.room:room-runtime:2.4.1
+|    |    +--- androidx.room:room-common:2.4.1 (*)
+|    |    +--- androidx.sqlite:sqlite-framework:2.2.0
+|    |    |    +--- androidx.annotation:annotation:1.0.0 -> 1.1.0
+|    |    |    \--- androidx.sqlite:sqlite:2.2.0
+|    |    |         \--- androidx.annotation:annotation:1.0.0 -> 1.1.0
+|    |    +--- androidx.sqlite:sqlite:2.2.0 (*)
+|    |    +--- androidx.arch.core:core-runtime:2.0.1
+|    |    |    +--- androidx.annotation:annotation:1.0.2 -> 1.1.0
+|    |    |    \--- androidx.arch.core:core-common:2.0.1 -> 2.1.0
+|    |    |         \--- androidx.annotation:annotation:1.1.0
+|    |    \--- androidx.annotation:annotation-experimental:1.1.0
+|    +--- org.jetbrains.kotlin:kotlin-stdlib:1.6.10
+|    |    +--- org.jetbrains:annotations:13.0
+|    |    \--- org.jetbrains.kotlin:kotlin-stdlib-common:1.6.10
+|    \--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.2
+|         +--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.2
+|         |    \--- org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.5.2
+|         |         +--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.5.30 -> 1.6.10
+|         |         |    +--- org.jetbrains.kotlin:kotlin-stdlib:1.6.10 (*)
+|         |         |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.6.10
+|         |         |         \--- org.jetbrains.kotlin:kotlin-stdlib:1.6.10 (*)
+|         |         \--- org.jetbrains.kotlin:kotlin-stdlib-common:1.5.30 -> 1.6.10
+|         \--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.5.30 -> 1.6.10 (*)
-+--- org.jetbrains.kotlin:kotlin-stdlib:1.4.10 -> 1.4.31 (*)
++--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.6.10 (*)
 +--- project :fluxc
 |    +--- com.squareup.okhttp3:okhttp:4.9.0
 |    |    +--- com.squareup.okio:okio:2.8.0
-|    |    |    +--- org.jetbrains.kotlin:kotlin-stdlib:1.4.0 -> 1.4.31 (*)
+|    |    |    +--- org.jetbrains.kotlin:kotlin-stdlib:1.4.0 -> 1.6.10 (*)
-|    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib-common:1.4.0 -> 1.4.31
+|    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib-common:1.4.0 -> 1.6.10
-|    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.4.10 -> 1.4.31 (*)
+|    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.4.10 -> 1.6.10 (*)
-|    +--- org.jetbrains.kotlin:kotlin-stdlib:1.4.10 -> 1.4.31 (*)
-|    +--- org.jetbrains.kotlin:kotlin-android-extensions-runtime:1.4.10
-|    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.4.10 -> 1.4.31 (*)
+|    +--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.6.10 (*)
+|    +--- org.jetbrains.kotlin:kotlin-android-extensions-runtime:1.6.10
+|    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.6.10 (*)
 |    +--- org.wordpress:utils:develop-eebc5d8e91a1d90190919f900f924b39c861a528
-|    |    +--- org.jetbrains.kotlin:kotlin-android-extensions-runtime:1.4.10 (*)
+|    |    +--- org.jetbrains.kotlin:kotlin-android-extensions-runtime:1.4.10 -> 1.6.10 (*)
 |    |    +--- androidx.core:core-ktx:1.2.0
-|    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.3.41 -> 1.4.31 (*)
+|    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.3.41 -> 1.6.10 (*)
-|    |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.4.10
-|    |         \--- org.jetbrains.kotlin:kotlin-stdlib:1.4.10 -> 1.4.31 (*)
+|    |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.4.10 -> 1.6.10 (*)
 |    +--- com.squareup.okhttp3:okhttp-urlconnection:4.9.0
-|    |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.4.10
-|    |         +--- org.jetbrains.kotlin:kotlin-stdlib:1.4.10 -> 1.4.31 (*)
-|    |         \--- org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.4.10 (*)
+|    |    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.4.10 -> 1.6.10 (*)
-|    +--- androidx.room:room-runtime:2.3.0 (*)
+|    +--- androidx.room:room-runtime:2.4.1 (*)
-|    +--- androidx.room:room-ktx:2.3.0 (*)
+|    +--- androidx.room:room-ktx:2.4.1 (*)
-|    +--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9 -> 1.4.1 (*)
+|    +--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9 -> 1.5.2 (*)
-|    \--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.9 -> 1.4.1 (*)
+|    \--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.9 -> 1.5.2 (*)
-+--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9 -> 1.4.1 (*)
++--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9 -> 1.5.2 (*)
-+--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.9 -> 1.4.1 (*)
++--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.9 -> 1.5.2 (*)
-\--- androidx.room:room-runtime:2.3.0 (*)
+\--- androidx.room:room-runtime:2.4.1 (*)

```

</p>
</details>

develric:
Hey @wzieba 👋 , sorry that I didn't document it last time 🤦 , here are the steps I was using from the FluxC example app:
- Use the `Sign In & Fetch Sites` button to login to a self-hosted site
- Use the `SITES` button to access to the Sites related actions buttons
- Use the `FETCH FIRST SITE` button to fetch the added self-hosted site (this should trigger the SiteXMLRPCClient.fetchSite->XMLRPCRequest.deliverResponse execution path)
- Check you get the `java.lang.ClassCastException` coming from the catch branch [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/1436f46fc296ede0e65e117bbfced38fa34cec6d/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequest.java#L158)

Hope this helps to reproduce 🙇

wzieba:
@develric thanks, I'll try it in a sec!

But in the meantime, as I was trying to make code more readable, I think I've fixed the issue!

See 70e5232f39147a2b8be096d42b4fc7b5dc3a6987

And this is how `com.android.volley.Listener` type in `XMLRPCRequest` constructor is interpreted in different scenarios: 

| scenario | interpretation |
| --- | --- |
| trunk | Screenshot 2022-04-04 at 12 57 11 |
| on this branch but before my fix | Screenshot 2022-04-04 at 12 52 24 | 
| on this branch after my fix | Screenshot 2022-04-04 at 12 51 18 |

☝️ this seems like a good fix as it aligns with current `trunk` behaviour. I'll try to understand what has changed between Kotlin versions but in a meantime - can I ask you to reproduce the issue with this new code?

wzieba:
The same goes for the SAM interface. So for everyone lurking in, the issue basically looks like this:

https://user-images.githubusercontent.com/5845095/161539619-d4cf907e-b53b-45c8-9e5c-216d65e88c41.mov

wzieba:
Some notes:
- just bumping to recently released Kotlin 1.6.20 doesn't itself resolve the issue.
- `invokedynamics` is something new (introduced in Kotlin 1.5.0) regarding lambda and SAM resolutions but reverting this behaviour with compilation argument, doesn't change anything ([source](https://kotlinlang.org/docs/whatsnew15.html#sam-adapters-via-invokedynamic))

develric:
> But in the meantime, as I was trying to make code more readable, I think I've fixed the issue!

Nice finding @wzieba 🚀 

> can I ask you to reproduce the issue with this new code?

Yep I can confirm that following the steps with the example app, no exception now 👍 

> this seems like a good fix as it aligns with current trunk behaviour

I still don't fully get how that works but agree that seems a good path 👍

wzieba:
I've managed to create a small reproduction repository with exactly the same issue happening and with [proof](https://github.com/wzieba/KotlinLambdaResolvingBug150/actions/runs/2090803364) that it starts to appear in Kotlin 1.5.0. Still looking for the root cause and some reasonable fix. If someone would like to contribute, here's the reproduction repository: https://github.com/wzieba/KotlinLambdaResolvingBug150

---

Calling it the day. I'm starting to lean forward to #2311 solution, even if we understand exactly what is causing the issue as this is connected closely with how Kotlin cooperates with Java's `Object` so I believe it's prone to errors/changes.

wzieba:
I think what we experienced here is a Kotlin compiler bug. In Kotlin 1.5.0 SAM interfaces (and apparently lambdas too) started to be compiled using `invokedynamic` and, by looking at byte code, it changed a lot of things.

In our case, the lambda for `com.android.volley.Response` is always parametrized via `Object[]` and not, as it should and was before Kotlin 1.5.0, via `Object`. Hence the crash.

See attached above reproduction repository. Here's the issue for Kotlin: https://youtrack.jetbrains.com/issue/KT-51868.

---

I'm working now on a solution based on your PR @develric .

ParaskP7 (approved):
👋 @wzieba and @develric !

Great investigation work to both of you and thank you so much for working on this `Kotlin & Room` update right till the very end, you are both awesome! ❤️ 🥇 💯 

I reviewed and tested this PR in isolation and with both, the `WPAndroid` & `WCAndroid` clients, through composite builds. Everything works as expected for both, the example and clients apps. I didn't encounter any breakage. I also reviewed the build output for before (`Kotlin 1.4.10 & Room 2.3.0`) and after (`Kotlin 1.6.10 & Room 2.4.2`) the update and I didn't see anything worrying. Everything looks as expected.

Suggestion (💡): The only thing I found is that the Lint task was failing when I run that after enabling `FluxC` composite builds. Lint was failing on this [WellSqlConfig.onConfigure(...)](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/0ca63247edb0506e40341866541c10e3aec72288/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt#L1877-L1884) function due to the fact that this check is now redundant and could be simplified. As such, I can only recommend applying this long overdue change to keep Lint happy.

I have nothing else to suggest and it is thus a ✅  from my side!

🎉

wzieba:
Thanks a lot @ParaskP7 ! 

> Suggestion (💡): The only thing I found is that the Lint task was failing when I run that after enabling FluxC composite builds.

I see the line and agree that it's redundant but I wonder why isn't this PR failing then? 🤔 

I'd opt to merge this PR as it is now and work on the issue you mention in next PR if that's okay 🙏 .

ParaskP7:
👋 @wzieba !

> I see the line and agree that it's redundant but I wonder why isn't this PR failing then? 🤔

It is only failing if you run Lint with `WPAndroid` and composite `FluxC` build together, as `ObsoleteSdkInt`'s severity is explicitly set to `error` (see `lint.xml` file). However, this will not fail if you run Lint with `WCAndroid` as you don't have the same `lint.xml` configuration on your side as well.

> I'd opt to merge this PR as it is now and work on the issue you mention in next PR if that's okay 🙏 .

Yeap, please progress as you see fit, this is merely a suggestion. 💯

wzieba:
> However, this will not fail if you run Lint with WCAndroid as you don't have the same lint.xml configuration on your side as well.

Oh, gotcha!  Thanks. 

Pushing this big green button then 😄 .

develric:
>Great investigation work to both of you...

Fair to say, my major kudos to @wzieba ...thanks for all the work here 🎉 !

wzieba:
Thank you all, it was teamwork!

PR #2288, opened 2022-02-11
Issue/5382 Use of site id in order to uniquely identify WCOrderNoteModel

kidinov:
I found only 1 place where we didn't use site id in order to query either `WCOrderNoteModel` or `WCOrderShipmentTrackingModel`. This PR adds site id to the query


wzieba:
In #2299 @hichamboushaba migrated Order Notes to Room with the same improvement ([here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2299/files#diff-9efb8ae1e30d3ef4142278b310a96526ddf1b8fbb1cf60cd2ba2d974c92ebfc3R16)). I'll close this PR then.

PR #2289, opened 2022-02-11
Change ShippingLine to a data class with all properties part of the constructor

hichamboushaba:
This simple PR just updates the `ShippingLine` class to move all the properties to the primary constructor, which would make it possible to instantiate with valid values in WCAndroid.
I made it as a data class, it's not really needed not, but I think having a proper equality check is already useful to avoid hidden bugs.

This is not a breaking change for WCAndroid, so it can be merged right away.


wzieba (approved):
> I made it as a data class, it's not really needed not, but I think having a proper equality check is already useful to avoid hidden bugs.

Agree 👍 looks good, merging!

PR #2290, opened 2022-02-14
[Woo] ability to serialize null values for shipping and fee lines to support deletion

hichamboushaba:
To delete a shipping line or a fee line from an Order, the API expects the following:
1. Passing `null` in the `method_id` field for the shipping line.
2. Passing `null` in the `name` field of the fee line.

So we need to be able to keep the `null` values when serializing those fields, this PR achieves this.

@ThomazFB please check the code, and if you have a better solution, please don't hesitate on suggesting it, I wanted to avoid dealing with Json manually, so I tried to make use of the `TypeAdapter` API.


ThomazFB (approved):
Tested with both shipping fee line and fee line, works as expected! Great work here, @hichamboushaba. Thanks!

PR #2294, opened 2022-02-16
POC for exposing a Flow to observe the NotificationModel table

hichamboushaba:
@JorgeMucientes as discussed, this is a POC for exposing a Flow to observe the NotificationModel table changes.

My first plan when I spoke with you, was to try to use the native [update hook](https://www.sqlite.org/c3ref/update_hook.html) from SQLite, but it turns out it's not exposed on Android, and even Room uses its own logic for keeping track of the changes through the [`InvalidationTracker`](https://developer.android.com/reference/androidx/room/InvalidationTracker) which keeps a separate table for logging all changes, then they [use](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/room-ktx/src/main/java/androidx/room/CoroutinesRoom.kt;l=104-139) those events to compose the Flow.

So I followed a similar approach, but just for the `NotificationModel`, I added a `SharedFlow` that emits a value after each DB update, it's then used as a base for any function that wants to observe the DB, I added an implementation for `observeNotificationsForSite` which we currently need, but this can be extended to all other functions.

I think the advantage of this over the previous solution is the following:
1. It's similar to what Room does, and the logic is kept inside the storage layer, so migrating to Room later won't cause any unwanted changes.
2. It's extensible, and can be used for all other functions.
3. It observes the DB changes, which is always the best approach for "DB first" patterns.
4. It simplifies `UnseenReviewsCountHandler` further, since `observeNotificationsForSite` returns the list of notifications directly.

This PR goes with the following WCAndroid [changes](https://github.com/woocommerce/woocommerce-android/compare/issue/5752-refactor-unseen-reviews-count...review-notifications-observe-db?expand=1).

As always, this is just a suggestion, if you feel this doesn't bring any added value, let's close this PR, it was a fun exercise after all 😄.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/NotificationSqlUtils.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/NotificationSqlUtils.kt
@@ -10,7 +10,6 @@ import com.yarolegovich.wellsql.core.annotation.Column
 import com.yarolegovich.wellsql.core.annotation.PrimaryKey
 import com.yarolegovich.wellsql.core.annotation.Table
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.flowOn
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/NotificationSqlUtils.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/NotificationSqlUtils.kt
@@ -141,7 +141,7 @@ class NotificationSqlUtils @Inject constructor(private val formattableContentMap
                 withContext(Dispatchers.IO) {
                     getNotificationsForSite(site, order, filterByType, filterBySubtype)
                 }
-        }
+            }
     }
 
     fun hasUnreadNotificationsForSite(
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/NotificationSqlUtils.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/NotificationSqlUtils.kt
@@ -9,6 +9,12 @@ import com.yarolegovich.wellsql.core.Identifiable
 import com.yarolegovich.wellsql.core.annotation.Column
 import com.yarolegovich.wellsql.core.annotation.PrimaryKey
 import com.yarolegovich.wellsql.core.annotation.Table
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.mapLatest
+import kotlinx.coroutines.flow.onStart
+import kotlinx.coroutines.withContext
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.notification.NoteIdSet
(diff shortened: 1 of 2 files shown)


hichamboushaba:
The remark I mentioned before about the worry for someone forgetting to emit to the SharedFlow is also present for this PR, but I don't think there is anything we can do with what we have 😄

JorgeMucientes:
First, thanks for spending time on this @hichamboushaba Changes look really nice!! 

I have a minor concern with the amount of extra times we are hitting the DB just for refreshing the unseen count. With this approach, for every update to the DB, we do an extra query to that same DB to `getNotificationsForSite`. Whereas before this queries to `getNotificationsForSite` DB,  I'd say were more "controlled" and only done for very specific events. Although in the quick tests I've done I have not seen anything wrong.

On the other hand regarding this concern
>  someone forgetting to emit to the SharedFlow is also present for this PR, ...

That's true, but I think that `NotificationSqlUtils` functions are less likely to change (we just have to look at the Git history to see it changes way less than NotificationStore). So I'd say this implementation reduces the risk of forgetting amissions significantly. 

Also one minor change we would need to add to the SharedFlow from `UnseenReviewsCountHandler` is `replay=1` policy otherwise we won't refresh the badge count for the More Menu button when we open the app for the first time and there are unread reviews.

One nitpick, I'd like to use a more "streamlined" syntax. I'd change `NotificationSqlUtils.observeNotificationsForSite` flow to this: 

```kotlin
    fun observeNotificationsForSite(
        site: SiteModel,
        @SelectQuery.Order order: Int = ORDER_DESCENDING,
        filterByType: List<String>? = null,
        filterBySubtype: List<String>? = null
    ): Flow<List<NotificationModel>> {
        return dataUpdatesTrigger
            .onStart { emit(Unit) }
            .mapLatest {
                val foo = getNotificationsForSite(site, order, filterByType, filterBySubtype)
                Log.i("TEST_UNSEEN", "NotificationSqlUtils.getNotificationsForSite -> hitting the DB")
                foo
            }.flowOn(Dispatchers.IO)
    }
```

I think it should be equivalent but we make use of a more "flowish" api.

Other than that I'd say we can apply this changes. I think they are safer in terms of future code changes from other devs.

hichamboushaba:

> I have a minor concern with the amount of extra times we are hitting the DB just for refreshing the unseen count. With this approach, for every update to the DB, we do an extra query to that same DB to `getNotificationsForSite`. Whereas before this queries to `getNotificationsForSite` DB, I'd say were more "controlled" and only done for very specific events. Although in the quick tests I've done I have not seen anything wrong.

This is true, but this is also the behavior of Room and other ORMs that offer the reactive APIs (SQLDelight for example), I think it should still be OK, since it's a simple operation, and done properly on a background thread.

> That's true, but I think that NotificationSqlUtils functions are less likely to change (we just have to look at the Git history to see it changes way less than NotificationStore). So I'd say this implementation reduces the risk of forgetting amissions significantly.

👍 

> Also one minor change we would need to add to the SharedFlow from UnseenReviewsCountHandler is replay=1 policy otherwise we won't refresh the badge count for the More Menu button when we open the app for the first time and there are unread reviews.

Since we are using the `onStart { emit(Unit) }`, we don't need the `replay=1`, since it will emit each time we start collecting `observeNotificationsForSite`.
And just as a note, here `replay=1` can't replace the `onStart { emit(Unit) }`, since we need also to emit even if there was no changes, meaning the `SharedFlow` is still empty.

> One nitpick, I'd to use a more "streamlined" syntax. I'd change NotificationSqlUtils.observeNotificationsForSite flow to this

Sure, done in 1f07ea8b44e56ce33317ba31228b12c546a3f9b3

JorgeMucientes (approved):

JorgeMucientes (approved):

Issue #2282, opened 2022-02-09
Provide `@ColumnInfo` annotations for Room entities


wzieba:
As discussed here https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2274#discussion_r801497452 we should provide `@ColumnInfo` annotation with desired default values in order to omit any discrepancies between created vs migrated databases.

PR #2295, opened 2022-02-17
Fetch coupons

### LINKED ISSUE #2228: [Coupons]: add endpoint to fetch a list of coupons for a store
This issue is part of the coupon management project and adds a new endpoint to fetch/display list of coupons for a store

0nko:
This PR implements the endpoint for fetching coupons.

Since it involves new functionality, I added the new tables to the Room DB. Coupons can reference products and product categories and there is many-to-many relationship between the entities, so there are also new cross-reference tables. Because I wanted to construct a full model tree when serving the result, I also added a new `Products` and `ProductCategories` tables to Room. When products and categories are added, updated or deleted, the parallel Room tables are also updated.

The `Products` and `ProductCategories` tables are incomplete (I left out the referenced entities, which can be added later on when we migrate the tables fully to Room).

This PR has a lot of line changes because the schema json has 1200+ lines and the table definitions are also very spread out. If you'd like me to split the PR up, let me know.

@wzieba I added you as a reviewer, because you've started adding Room tables, sorry :). 

I'll be adding unit tests in a separate PR.

**To test:**
1. Start the example app 
2. Log in and go to Woo -> Select site (make sure you have some coupons ready)
3. Fetch coupons
4. Notice the coupons are fetched (you can also verify this in the DB inspector)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
```diff
@@ -0,0 +1,42 @@
...
+@Dao
+abstract class CouponsDao {
+    @Transaction
+    @Query("SELECT * FROM Coupons WHERE siteId = :siteId ORDER BY id")
+    abstract fun observeCoupons(siteId: Long): Flow<List<CouponWithEmails>>
+
+    @Query("SELECT * FROM CouponsAndProductCategories WHERE isExcluded = :areExcluded")
+    abstract fun getCouponProductCategories(
+        areExcluded: Boolean
+    ): List<CouponAndProductCategoryEntity>
```
wzieba: 🔧  Can we also make it a `Flow`, like `observeCoupons`? This way it's more difficult to run this query on the main thread.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    abstract fun insertCoupon(entity: CouponEntity): Long
```
wzieba: 🔧 Can we make this `suspend fun` ? For the same reason - it's possible to use this call outside of `CouponsDao#transaction` and execute this on main thread.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
```diff
@@ -0,0 +1,42 @@
...
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    abstract fun insertCouponAndProduct(entity: CouponAndProductEntity): Long
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    abstract fun insertCouponEmail(entity: CouponEmailEntity): Long
+
+    @Transaction
+    open suspend fun transaction(block: suspend () -> Unit) {
+        block()
+    }
```
wzieba: 🔧 WDYT about not adding `transaction` function? I'm worried that this idea moves responsibility for handling database insertion logic outside of Dao object, when it's its responsibility.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
wzieba:
```suggestion
    abstract fun insertCoupon(entity: CouponEntity)
```
⛏️ A small thing but I see that we don't use generated id (or we don't have generated id at all e.g. `CouponEntity`). Maybe let's remove this return type to not cause confusion.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
```diff
@@ -11,23 +11,40 @@ import org.wordpress.android.fluxc.persistence.converters.LocalIdConverter
...
             GlobalAddonGroupEntity::class,
-            WCOrderModel::class
-        ]
+            WCOrderModel::class,
+            CouponEntity::class,
+            CouponEmailEntity::class,
+            CouponAndProductEntity::class,
+            CouponAndProductCategoryEntity::class,
+            ProductEntity::class,
+            ProductCategoryEntity::class
```
wzieba: ⛏️ Nitpick but can we keep alphabetical order here?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
```diff
@@ -11,23 +11,40 @@ import org.wordpress.android.fluxc.persistence.converters.LocalIdConverter
...
+        ],
+        exportSchema = true
```
wzieba: 🔧 I think it's not needed, this option is enabled in build arguments. Maybe let's remove it to not duplicate.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/coupons/CouponDto.kt
```diff
@@ -0,0 +1,56 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.coupons
+
+import com.google.gson.annotations.SerializedName
+import org.wordpress.android.fluxc.persistence.entity.CouponEntity
+
+data class CouponDto(
+    @SerializedName("id") val id: Long,
+    @SerializedName("code") val code: String?,
```
wzieba, line 8: 💯 I like how you added `SerializedName` annotation to every field (also those which are named the same) and you use `Dto` name for remote object.

example/build.gradle
```diff
@@ -107,7 +111,7 @@ dependencies {
     implementation 'androidx.annotation:annotation:1.0.2'
     implementation "androidx.arch.core:core-common:$arch_core_version"
     implementation "androidx.arch.core:core-runtime:$arch_core_version"
-    implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.2.0"
+    implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.4.0"
```
wzieba, line 114: ❓ Just out of curiosity: what was the reason for this bump?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponAndProductCategoryEntity.kt
```diff
@@ -0,0 +1,31 @@
+package org.wordpress.android.fluxc.persistence.entity
+
+import androidx.room.Entity
+import androidx.room.ForeignKey
+import androidx.room.Index
+
+@Entity(
+    tableName = "CouponsAndProductCategories",
```
wzieba, line 8: 🔧 Can we keep original name `CouponAndProductCategoryEntity` here? We already have a few entities named in a convention `*Entity` so maybe we should rather stick to this (I now see it's not perfect).

Or would you prefer to rename all other tables? WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponAndProductCategoryEntity.kt
```diff
@@ -0,0 +1,31 @@
...
+    foreignKeys = [
+        ForeignKey(
+                entity = CouponEntity::class,
+                parentColumns = arrayOf("id", "siteId"),
+                childColumns = arrayOf("couponId", "siteId"),
```
wzieba: ⛏️ Minor, but you can use also a shorter, to me more readable form 
```suggestion
                parentColumns = ["id", "siteId"],
                childColumns = ["couponId", "siteId"],
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEmailEntity.kt
```diff
@@ -0,0 +1,25 @@
...
+                    entity = CouponEntity::class,
+                    parentColumns = arrayOf("id", "siteId"),
+                    childColumns = arrayOf("couponId", "siteId"),
+                    onDelete = ForeignKey.CASCADE
+            )
+        ],
+        indices = [Index("couponId", "siteId")]
+)
+data class CouponEmailEntity(
+    @PrimaryKey(autoGenerate = true) val id: Long = 0,
```
wzieba: ❓ Is this autogenerated id used anywhere?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEntity.kt
```diff
@@ -0,0 +1,49 @@
...
+
+@Entity(
+    tableName = "Coupons",
+    primaryKeys = ["id", "siteId"],
+    indices = [Index("id", "siteId")]
+)
+data class CouponEntity(
+    val id: Long,
+    val siteId: Long,
+    val code: String? = null,
```
wzieba, line 16: 🔧 For newly created entities, let's use `@ColumnInfo` which might make our lives easier in the future. More context in #2282 and linked discussion.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/ProductEntity.kt
```diff
@@ -0,0 +1,125 @@
...
+import org.wordpress.android.fluxc.model.WCProductModel
+
+@Entity(
+    tableName = "Products",
+    primaryKeys = ["id", "siteId"]
+)
+data class ProductEntity(
+    val id: Long,
+    val siteId: Long,
+    val name: String? = null,
```
wzieba, line 13: ❓ I wonder about default values, should it be empty string or null? In `WCProductModel` it's empty string rather than null, should we keep that? I'm not sure what are a consequences.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -919,6 +924,7 @@ class WCProductStore @Inject constructor(
                 }
             } else {
                 val rowsAffected = ProductSqlUtils.insertOrUpdateProduct(result.product)
+                productsDao.insertOrUpdate(result.product.toDataModel(result.site.siteId))
```
wzieba: 🔧 Can we provide some mechanism (helper class?) that will join adding products to WellSql *and* Room?

I'm worried that someone working on products later will be confused of choosing database or will forget to add entities to one of them.

wzieba (commented):
Thanks @0nko for working on that, I like how you followed new approaches with naming and of course I love usage of Room. 

I left quite a few comments but also this PR adds a few features - I hope that's fine! If you prefer to divide this PR and work with a feature branch I'm okay with that but I'm also okay with keeping one big PR.

I left some fix requests (🔧 ), nitpicks (⛏️ ) and questions (❓ ). To not duplicate comments, I just want to mention that comments about entities or daos are about all of entities and daos in this PR - not only those with comments.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
0nko: This method is just a leftover from testing and it's not being used. I'll get rid of it.

wzieba:
@0nko just a heads up - next week I'm AFK and won't be able to follow up on this review then.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
0nko: I made all inserts and deletes suspended. This was a bit more involved, because as per your other suggestion, I put both WellSql and Room operations inside a helper class. See the [commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2295/commits/a99d06f4a94be890aa5fe3d89c5e83edb0a1aecb).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEmailEntity.kt
0nko: Well it's a unique primary key. Even though it's not required, it is a good practice to have one, AFAIK.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/ProductEntity.kt
0nko: Me neither. But `null` can indicate a missing value while an empty string could either be a missing value or it could also be a valid "regular" value. I'm not aware of any specific reason the original `WCProductModel` used empty strings.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
0nko: See the [commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/1011ab61328fb89f204cf91e294e056da8589044).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEntity.kt
0nko: Removing the Kotlin default values would prevent us from creating an instance without specifying every parameter. WDYT?

0nko:
Thanks so much for your review, @wzieba! I believe I addressed all of your comments.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
wzieba: Yeah I see it required quite a few changes but I believe it'll repay. Thanks!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEmailEntity.kt
wzieba: Gotcha, but as it's not used anywhere, I'm not sure there will be any use for uniqueness of it.

WDYT about instead of it, create a composed key of (`couponId` x `siteId` x `email`) ? This way we won't have a duplicates in this table but also we won't have to add unused field.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/ProductEntity.kt
wzieba: Gotcha - you're right, now when I look nulls seem more correct. This made me think that we probably shouldn't add empty Strings to `OrderEntity` as default parameters, but rather nulls, like here.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEntity.kt
wzieba: In my original comment I was thinking about using Kotlin's default values but *also* `@ColumnInfo`, but as I agree with you that `null` values are defaults, this makes `@ColumnInfo` completely optional/redundant - `NULL` will be a default value if not specified and it seems that there's no need for specifying it ([source](https://stackoverflow.com/questions/7533532/is-there-a-reason-to-specify-default-null-on-a-nullable-column))

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -867,44 +871,44 @@ class WCProductStore @Inject constructor(
...
-                        ?.model?.asProductVariationModel()
-                        ?.apply { insertOrUpdateProductVariation(this) }
-                        ?.let { WooResult(it) }
-                        ?: WooResult(WooError(INVALID_RESPONSE, GenericErrorType.INVALID_RESPONSE))
-            } ?: WooResult(WooError(WooErrorType.GENERIC_ERROR, UNKNOWN))
+        coroutineEngine.withDefaultContext(API, this, "generateEmptyVariation") {
+            product.attributeList
+                .filter { it.variation }
+                .map { ProductVariantOption(it.id, it.name, "") }
+                .let { Gson().toJson(it) }
```
wzieba, line 890: 🌱   I see it's not your part but maybe we could inject `Gson` into constructor and use it here instead of creating instance every time, WDYT? Maybe not necessarily in this PR, just commenting for awareness.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -458,35 +457,95 @@ class ProductRestClient @Inject constructor(
             Array<ProductApiResponse>::class.java
     )
 
+    @JvmName("handleResultFromProductApiResponse")
```
wzieba: ❓  I can't find usage of `handleResultFrom` anywhere else except this class - is `@JvmName` used anywhere?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/ProductCategoriesDao.kt
```diff
@@ -0,0 +1,34 @@
...
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    abstract suspend fun insertOrUpdateProductCategories(entities: List<ProductCategoryEntity>)
+
+    @Query("SELECT * FROM ProductCategories p JOIN CouponsAndProductCategories c " +
+        "ON p.id = c.productCategoryId WHERE c.isExcluded = :areExcluded " +
+        "AND c.couponId = :couponId ORDER BY p.id")
+    abstract fun getCouponProductCategories(
+        couponId: Long,
+        areExcluded: Boolean
+    ): List<ProductCategoryEntity>
```
wzieba: 🔧 I'd have similar suggestion as in `CouponsDao` - the API that returns `List` is easy to execute on main thread, it happened a lot with WellSql `*SqlUtils` classes. WDYT about changing them to return `Flow`s? It'll be of course still possible to run on main thread but harder.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -0,0 +1,214 @@
...
+            couponsDao.insertOrUpdateCouponAndProductCategory(
+                CouponAndProductCategoryEntity(
+                    couponId = dto.id,
+                    siteId = site.siteId,
+                    productCategoryId = categoryId,
+                    isExcluded = true
+                )
+            )
+        }
+    }
```
wzieba, line 117: 💯 I love this how nicely this code looks, like from some coroutines tutorial 😄

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -0,0 +1,214 @@
...
+                }
+            }
+            .flowOn(Dispatchers.IO)
+            .distinctUntilChanged()
+
+    private suspend fun fetchMissingProducts(productIds: List<Long>?, site: SiteModel) {
+        if (!productIds.isNullOrEmpty()) {
+            val products = productsDao.getProductsByIds(site.siteId, productIds).map { it.id }
+
+            Log.d("COUPONS", "Found products in DB: ${products.size}")
```
wzieba: 🔧  Maybe we could go with our wrapper (`AppLogWrapper`) around logging so we won't add not necessary Android dependency to `CouponStore`?

wzieba (commented):
Thank you @0nko, all improvements looks good! 

Some comments were left without answer but I'd say it's Github's fault - it hidden some comments with "See more" kind of button, I also missed this at first. 

Overall it looks all good, I'm leaving a few more comments that appeared to me in second review.

----

On the testing side it all looks good. I really like how you designed relations in the database. For easier understanding, I've exported it and generated a graph. Leaving this for future reference. 
Screenshot 2022-02-18 at 22 38 30

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
0nko: Oh yeah, this is a leftover from when I overloaded the function. Gone now.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
0nko: Good call. I agree but `Gson()` used in a lot of places so I'd keep it to another PR, though.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEmailEntity.kt
0nko: Sure, that sounds good.

example/build.gradle
0nko: It was mainly for the `repeatOnLifecycle()` extension function for collecting flows in the fragments.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponAndProductCategoryEntity.kt
0nko: This is more of a philosophical decision. An "entity" is a representation of a single table record and it didn't sound right to me since a table contains a _collection_ of records/entities. Also in a query, it just looks nicer: `SELECT * FROM Products`. That's why I chose the plural form without the word "entity". But I also agree that consistency is important and I don't really feel that strongly about it :). So I'll just change it to your way to avoid problems with renaming and migrations for the existing users.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
0nko: This was a forgotten leftover from my own testing. Thanks for spotting it. 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/ProductCategoriesDao.kt
0nko: I understand the main thread argument and but I feel like it would unnecessarily complicate the code in this case. The queries above are used to [construct the relationships](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2295/files#diff-007b178a183185fa597f2dd0feccd8c9b7f864ae4a8800d82ddba9498db6bb9dR150) for a final coupon data model. These calls are already inside a flow collector and converting this query to a flow means that multiple flows would have to be merged/combined. Besides, as side effect, an update to any of the touched tables would trigger an unnecessary update on the final combined flow. I'm not against adding another `observeProductCategories()` method for a case when we actually want to display the categories in a reactive manner. I just think we should also keep a regular query since the data being observed is not the objective here. What do you think?

0nko:
@wzieba Thanks for another thorough review. Sorry I missed some of your comments from the 1st round. I'm not sure why some weren't shown under the "Conversation" tab. I guess I'll have to start reading the "Files changed" tab as it seems to show all of them.

I believe I addressed all of the comments, either by fixing the problem or discussing it :).

wzieba:
hi @0nko - just a heads up that I'm after my AFK and I'll be prioritising this review today 🙂.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/ProductCategoriesDao.kt
wzieba: Gotcha, yes after second thought I agree it'd complicate code. Eventually, the better way to assert not main thread will be removing `allowMainThreadQueries` from database configuration so probably we shouldn't focus so much on using `Flow` instead of plain `List`. 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponAndProductCategoryEntity.kt
wzieba: Yes, you're right. I like naming you proposed with no `Entity` word and plural form - it's the correct one from my experience with SQL. Thanks for changing it to assert consistency. 

My suggestion will be to change every table to correct one when we'll bump Room in #2287 and we'll have `AutoMigration` feature. It will be super easy then and without any data loss. See [@RenameTable](https://developer.android.com/reference/kotlin/androidx/room/RenameTable). WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEmailEntity.kt
wzieba: @0nko I think this improvement has been missed

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponAndProductCategoryEntity.kt
wzieba: So, after second thought, maybe it's worth to keep the names you proposed? @0nko

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEmailEntity.kt
0nko: Weird, I could swear I changed it! :) Thanks!

0nko: Okay, so I'm not crazy. You'll need to refresh the PR. I looked at the code and it's done already :).

wzieba: Oh true, it is. I was checking the state from writing the comment, sorry for confusion!

wzieba (approved):
Thanks for working on this @0nko , code looks good to me now and when I was testing it I haven't spotted anything wrong + I've learned a domain of coupons (partially at least).

One question: how would you roughly estimate migrating `Products` completely to Room?

.idea/codeStyles/Project.xml
```diff
@@ -9,7 +9,7 @@
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value>
-          <package name="kotlinx.android.synthetic" withSubpackages="true" static="false" />
+          <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
```
wzieba, line 12: @0nko is edit of this file intentional? 

btw, when working with FluxC I often also have `Project.xml` changed by AS but I never really managed to take care of it. I wonder if it happens when working with FluxC as included build (from context of WooCommerce app)

0nko: It wasn't and I just found your comment 🤦. I ignore the changes in this file, this one slipped through somehow.

PR #2296, opened 2022-02-25
Woo: Add possibility to specify locale when formatting prices

hichamboushaba:
This is a simple change that does the following:
1. Allows passing a custom `Locale` for formatting.
2. Adds an overloaded version that's not coupled to the FluxC model `WCSettingsModel`

This is needed for the WCAndroid PR: https://github.com/woocommerce/woocommerce-android/pull/5929


0nko (approved):
LGTM 👍

PR #2286, opened 2022-02-10
Feature branch for Order model improvements

kidinov:

PR #2299, opened 2022-03-02
Migrate Order Notes to Room

hichamboushaba:
⚠️ Please don't merge until the corresponding WCAndroid [PR](https://github.com/woocommerce/woocommerce-android/pull/5971) is approved.

This PR started as a fix for the [issue](https://github.com/woocommerce/woocommerce-android/issues/5970), but then I decided to take it as an opportunity to migrate the table to Room too.

Even though the number of changed lines is big, the big part of the change is just the autogenerated Room schema file, which you can ignore.

#### Testing
Please test the Order Note feature in the example app (you need to make sure orders are fetched before you run the tests)
Or preferably refer to the WCAndroid PR https://github.com/woocommerce/woocommerce-android/pull/5971 for test steps.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
@@ -1,5 +1,7 @@
 package org.wordpress.android.fluxc.model;
 
+import static java.lang.annotation.RetentionPolicy.SOURCE;
+
 import androidx.annotation.IntDef;
 import androidx.annotation.NonNull;
 
@@ -11,6 +13,7 @@ import com.yarolegovich.wellsql.core.annotation.Table;
 
 import org.wordpress.android.fluxc.Payload;
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId;
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId;
 import org.wordpress.android.fluxc.network.BaseRequest.BaseNetworkError;
 import org.wordpress.android.util.AppLog;
 import org.wordpress.android.util.AppLog.T;
@@ -20,8 +23,6 @@ import java.lang.annotation.Retention;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
-
-import static java.lang.annotation.RetentionPolicy.SOURCE;
 
 @Table
 @RawConstraints({"UNIQUE (SITE_ID, URL)"})
@@ -152,6 +153,14 @@ public class SiteModel extends Payload<BaseNetworkError> implements Identifiable
 
     public LocalId localId() {
         return new LocalOrRemoteId.LocalId(mId);
+    }
+
+    public RemoteId remoteId() {
+        if (mSiteId != 0L) {
+            return new RemoteId(mSiteId);
+        } else {
+            return new RemoteId(mSelfHostedSiteId);
+        }
     }
 
     public SiteModel() {
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
@@ -1,7 +1,5 @@
 package org.wordpress.android.fluxc.model;
 
-import static java.lang.annotation.RetentionPolicy.SOURCE;
-
 import androidx.annotation.IntDef;
 import androidx.annotation.NonNull;
 
@@ -23,6 +21,8 @@ import java.lang.annotation.Retention;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
+
+import static java.lang.annotation.RetentionPolicy.SOURCE;
 
 @Table
 @RawConstraints({"UNIQUE (SITE_ID, URL)"})
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
@@ -20,6 +20,7 @@ import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_3_4
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_4_5
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_5_6
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_6_7
+import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_7_8
 
 @Database(
         version = 8,
@@ -55,6 +56,7 @@ abstract class WCAndroidDatabase : RoomDatabase() {
                 .addMigrations(MIGRATION_4_5)
                 .addMigrations(MIGRATION_5_6)
                 .addMigrations(MIGRATION_6_7)
+                .addMigrations(MIGRATION_7_8)
                 .build()
     }
 }
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
@@ -6,6 +6,7 @@ import androidx.room.Room
 import androidx.room.RoomDatabase
 import androidx.room.TypeConverters
 import org.wordpress.android.fluxc.model.WCOrderModel
+import org.wordpress.android.fluxc.model.WCOrderNoteModel
 import org.wordpress.android.fluxc.persistence.converters.BigDecimalConverter
 import org.wordpress.android.fluxc.persistence.converters.LocalIdConverter
 import org.wordpress.android.fluxc.persistence.converters.LongListConverter
@@ -21,12 +22,13 @@ import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_5_6
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_6_7
 
 @Database(
-        version = 7,
+        version = 8,
         entities = [
             AddonEntity::class,
             AddonOptionEntity::class,
             GlobalAddonGroupEntity::class,
-            WCOrderModel::class
+            WCOrderModel::class,
+            WCOrderNoteModel::class
         ]
 )
 @TypeConverters(
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
@@ -6,7 +6,7 @@ import androidx.room.Room
 import androidx.room.RoomDatabase
 import androidx.room.TypeConverters
 import org.wordpress.android.fluxc.model.WCOrderModel
-import org.wordpress.android.fluxc.model.OrderNoteEntity
+import org.wordpress.android.fluxc.persistence.entity.OrderNoteEntity
 import org.wordpress.android.fluxc.persistence.converters.BigDecimalConverter
 import org.wordpress.android.fluxc.persistence.converters.LocalIdConverter
 import org.wordpress.android.fluxc.persistence.converters.LongListConverter
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
@@ -6,7 +6,7 @@ import androidx.room.Room
 import androidx.room.RoomDatabase
 import androidx.room.TypeConverters
 import org.wordpress.android.fluxc.model.WCOrderModel
-import org.wordpress.android.fluxc.model.WCOrderNoteModel
+import org.wordpress.android.fluxc.model.OrderNoteEntity
 import org.wordpress.android.fluxc.persistence.converters.BigDecimalConverter
 import org.wordpress.android.fluxc.persistence.converters.LocalIdConverter
 import org.wordpress.android.fluxc.persistence.converters.LongListConverter
@@ -30,7 +30,7 @@ import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_7_8
             AddonOptionEntity::class,
             GlobalAddonGroupEntity::class,
             WCOrderModel::class,
-            WCOrderNoteModel::class
+            OrderNoteEntity::class
         ]
 )
 @TypeConverters(
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
@@ -12,6 +12,7 @@ import org.wordpress.android.fluxc.persistence.converters.LocalIdConverter
 import org.wordpress.android.fluxc.persistence.converters.LongListConverter
 import org.wordpress.android.fluxc.persistence.converters.RemoteIdConverter
 import org.wordpress.android.fluxc.persistence.dao.AddonsDao
+import org.wordpress.android.fluxc.persistence.dao.OrderNotesDao
 import org.wordpress.android.fluxc.persistence.dao.OrdersDao
 import org.wordpress.android.fluxc.persistence.entity.AddonEntity
 import org.wordpress.android.fluxc.persistence.entity.AddonOptionEntity
@@ -43,6 +44,7 @@ import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_7_8
 abstract class WCAndroidDatabase : RoomDatabase() {
     internal abstract fun addonsDao(): AddonsDao
     abstract fun ordersDao(): OrdersDao
+    abstract fun orderNotesDao(): OrderNotesDao
 
     companion object {
         fun buildDb(applicationContext: Context) = Room.databaseBuilder(
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
@@ -6,7 +6,6 @@ import androidx.room.Room
 import androidx.room.RoomDatabase
 import androidx.room.TypeConverters
 import org.wordpress.android.fluxc.model.WCOrderModel
-import org.wordpress.android.fluxc.persistence.entity.OrderNoteEntity
 import org.wordpress.android.fluxc.persistence.converters.BigDecimalConverter
 import org.wordpress.android.fluxc.persistence.converters.LocalIdConverter
 import org.wordpress.android.fluxc.persistence.converters.LongListConverter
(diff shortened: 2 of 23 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
```diff
@@ -0,0 +1,23 @@
+package org.wordpress.android.fluxc.persistence.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
+
+@Entity(tableName = "OrderNoteEntity", primaryKeys = ["localSiteId", "noteId"])
+data class OrderNoteEntity(
+    val localSiteId: LocalId,
```
hichamboushaba: The usage of `localSiteId` here is problematic too, same as using the `localOrderId` before: since it's a generated `id`, it's subject to changes if the list of sites gets removed then re-inserted, which would cause a mix-up on the order notes.
This scenario doesn't seem to occur in the app currently (there is a low chance of this occurring when migrating the SiteModel table), so I kept it to be consistent with the approach we follow with other tables.

Preferably we should depend on the `siteId`, which is a remote ID that's not subject to change, please let me know if you think we should use the `siteId` here even if it's not consistent with the other tables 🙏.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
```diff
@@ -0,0 +1,23 @@
+package org.wordpress.android.fluxc.persistence.entity
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
+
+@Entity(tableName = "OrderNoteEntity", primaryKeys = ["localSiteId", "noteId"])
```
wzieba: 👋 We have discussion with @0nko on a different PR about naming tables. WDYT about naming it `OrderNotes` ? I believe that name system suggested by @0nko is more correct in terms of SQL.

We can change `OrderEntity` and rest names of tables with `AutoMigration` later to keep everything consistent. Here's a link to discussion: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2295#discussion_r809240496

hichamboushaba: Sure, I have no strong opinion about this, so it's OK for me to use whatever name we prefer.

I'll update it ASAP.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
wzieba: Good point, I'd vote to use `siteId` (remote one). I see that `SiteModel` is using either `mSiteId` or `mSelfHostedSiteId` - so if I'm not mistaken we should have some if there to always get the correct remote id, depending if site is or isn't self hosted one.

This is why I always thought `localSiteId` is a safer option but, as you described in issue, it's not.

So I think that eventually we should also update `OrderEntity`, if we decide to use `siteId` (remote one) here.

hichamboushaba: >  I see that SiteModel is using either mSiteId or mSelfHostedSiteId - so if I'm not mistaken we should have some if there to always get the correct remote id, depending if site is or isn't self hosted one.

We don't support self-hosted in the Woo app, so this should'nt be an issue, even for the long term.

wzieba: 🤦 thank you, yes.

wzieba:
Just FYI this is conflicting with #2288 and probably with #2286 but as it's smaller, I'd vote to merge it first.

@kidinov I'm taking on me resolving all conflicts that will occur!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
hichamboushaba: Just FYI, I'm working right now on using the remote siteId for the OrderNotes table.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
hichamboushaba: Done in edbaddc

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
hichamboushaba: Done in adf1d19

fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
```diff
@@ -154,6 +155,14 @@ public LocalId localId() {
         return new LocalOrRemoteId.LocalId(mId);
     }
 
+    public RemoteId remoteId() {
+        if (mSiteId != 0L) {
+            return new RemoteId(mSiteId);
+        } else {
+            return new RemoteId(mSelfHostedSiteId);
+        }
```
hichamboushaba, line 163: @wzieba I ended doing the if/else you mentioned, I know we don't need it for our case, but as I wanted to use the name `remoteId` for the function, I wanted to make the function work for both cases.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderNotesDao.kt
```diff
@@ -0,0 +1,20 @@
...
+@Dao
+interface OrderNotesDao {
+    @Insert(onConflict = REPLACE)
+    suspend fun insertNotes(vararg notes: OrderNoteEntity)
+
+    @Query("SELECT * FROM OrderNotes WHERE siteId = :siteId AND orderId = :orderId")
+    suspend fun queryNotesOfOrder(siteId: RemoteId, orderId: RemoteId): List<OrderNoteEntity>
+
+    @Query("DELETE FROM OrderNotes WHERE siteId = :siteId")
+    fun deleteOrderNotesForSite(siteId: RemoteId)
```
kidinov, line 19: Is this intentional that this is not `suspend`?

hichamboushaba: Yes, it's intentional, this functional is [used](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/adf1d199fd894d7df579909196d6d53f726d0280/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt#L681) only in the legacy order's fetching logic (the action `FETCH_ORDERS`), which the app doesn't use, and which wasn't migrated to coroutines, so we can't make it suspendable easily, and it's not worth the effort since it's not used by the app.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
```diff
@@ -0,0 +1,22 @@
...
+import androidx.room.Entity
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
+
+@Entity(tableName = "OrderNotes", primaryKeys = ["siteId", "noteId"])
+data class OrderNoteEntity(
+    val siteId: RemoteId,
+    val noteId: RemoteId,
+    val orderId: RemoteId,
+    @ColumnInfo(defaultValue = "")
+    val dateCreated: String = "", // ISO 8601-formatted date in UTC, e.g. 1955-11-05T14:15:00Z
```
kidinov: Have you considered using TypeConverters and nullable Date as type here?

hichamboushaba: My goal was to fix the data inconsistency issue as soon as possible, and I was even torn out between migrating to room or just fixing it in WellSql, but I agree with you, a TypeConverter would be cleaner.
I can't promise I'll find time to do it today, so if can't find time by the end of the day, let's keep it as a future improvement, since it won't even require a DB migration.

wzieba: Another thing to consider is making fields nullable with `null` as a default value. I know empty Strings were used in `WCOrderNoteModel` but eventually I think it makes more sense to have it null if it's null.

I'm okay to do this in a different PR but doing this in this PR might reduce overhead of any migration later.

WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
```diff
@@ -19,14 +21,16 @@ import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_3_4
...
 @Database(
-        version = 7,
+        version = 8,
         entities = [
             AddonEntity::class,
             AddonOptionEntity::class,
             GlobalAddonGroupEntity::class,
-            WCOrderModel::class
+            WCOrderModel::class,
+            OrderNoteEntity::class
```
wzieba, line 33: Super minor but if you'll be making any changes to this PR, can we adjust to alphabetical order here? Thanks

hichamboushaba: I can do it, but just to see if I'm missing something, what's the added value of alphabetical order here vs adding new entities to the bottom of the list? 🙂

wzieba: Yeah, originally I thought that it'll be easier to navigate and to identify lacks of items (as it did help in [`ORDER_FIELDS`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/20ac15c3dfefb8b7497f75771359e92a438e8f81#diff-1faa106d0316334dc71e2013c675ab9009954e267691d14d16bf4d84b8c61fe9R810)) but in case of Room it's indeed redundant.

hichamboushaba: Yes, I think since those are actual types, it won't bring any added value, instead, it'll add a small overhead for anyone adding new entities to make sure it's in the right spot.
> as it did help in [ORDER_FIELDS](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/20ac15c3dfefb8b7497f75771359e92a438e8f81#diff-1faa106d0316334dc71e2013c675ab9009954e267691d14d16bf4d84b8c61fe9R810)

Yes, I agree about the help it brought there, since it was a list of Strings.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
hichamboushaba: I agree, `null` values makes more sense, I didn't think about it yesterday since I was focusing on fixing the issue, I'll try to add it with the TypeConverter today 🤞

hichamboushaba: All implemented:

5eda53b makes fields nullable
de81845 adds a typeconverter for making the `dateCreated` field as a Date object

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
+import org.wordpress.android.fluxc.persistence.converters.ISO8601DateConverter
+import java.util.Date
+
+@Entity(tableName = "OrderNotes", primaryKeys = ["siteId", "noteId"])
+data class OrderNoteEntity(
+    val siteId: RemoteId,
+    val noteId: RemoteId,
+    val orderId: RemoteId,
+    @field:TypeConverters(ISO8601DateConverter::class) val dateCreated: Date? = null,
```
hichamboushaba, line 14: I decided to add the `TypeConverter` here instead of adding to the Database itself to not force each `Date` object to the ISO8601 formatting, each Table can decide how to save the Date objects, depending on the use case.
Let me know what you think.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/converters/ISO8601DateConverter.kt
```diff
@@ -0,0 +1,16 @@
+package org.wordpress.android.fluxc.persistence.converters
+
+import androidx.room.TypeConverter
+import org.wordpress.android.util.DateTimeUtils
+import java.util.Date
+
+/**
+ * A Room type converter for ISO 8601-formatted dates
+ */
+class ISO8601DateConverter {
```
hichamboushaba, line 10: I kept the same format as before, which is `ISO 8601`, we can also save the date using the UNIX epoch time, which means just a Long number, I don't have a strong preference over one or the other, if you have a strong opinion about Unix epoch time, please let me know to discuss it.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderNotesDao.kt
malinajirka: > so we can't make it suspendable easily

It's not important and I agree that might not be worth the effort. But have you considered using `suspend` here and `runBlocking` in the non-coroutine code? I think this is a perfect example of a correct usage of `runBlocking` in production code -> it serves as a bridge between blocking and suspending code. Wdyt?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
```diff
@@ -0,0 +1,22 @@
...
+    val note: String? = null,
+    val author: String? = null,
+    @ColumnInfo(defaultValue = "0")
+    val isSystemNote: Boolean = false, // True if the note is 'system-created', else created by a site user
+    @ColumnInfo(defaultValue = "0")
+    val isCustomerNote: Boolean = false // False if private, else customer-facing. Default is false
```
wzieba: Shouldn't we also make those nullable? It's kind of adding a new interpretation to data gathered from API (we only know defaults because of a comment, I don't know if up-to-date).

If we know it's "false" if it's not provided, maybe we shouldn't add this in database layer but rather in quasi-domain?

But as it was like this before, it's not a blocker from my end.

wzieba (approved):
Thank you, good improvement 👍 Not merging as Woo PR isn't yet checked and there's a comment from @malinajirka

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderNotesDao.kt
hichamboushaba: Thanks @malinajirka for chiming in, good suggestion, I didn't consider it first, but now after thinking about, I think it's not worth it, because:
1. It'll be slower than the plain non-suspendable function, as the thread switching will have an impact, even if it's not noticeable, and we'll be ending up blocking two threads instead of just the calling one.
2. Starting from the principle of YAGNI, the suspendability of this function is not useful now, so it shouldn't have it. And even if we need it in the future, it's a pretty straightforward change.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/converters/ISO8601DateConverter.kt
wzieba: I do not have **strong** opinion about Unix epoch time, but I would lean forward to it more than any other format as I believe Unix epoch time is much more robust. But I don't have any proofs to prove this thesis, it's not a blocker from me.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
hichamboushaba: I didn't make them nullable, because:
1. isSystemNote will never be null, as it's calculated in the network-toDB [mapper](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/582e1a06461fb8369a5ff586810cc3de060f7167/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderNoteApiResponse.kt#L25)
2. There is a note about `isCustomer` being false by default, stated in the API response [file](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/582e1a06461fb8369a5ff586810cc3de060f7167/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderNoteApiResponse.kt#L14-L16), so I think there is historical reason that even if it's missing from the API, it should be considered as false.

hichamboushaba: > If we know it's "false" if it's not provided, maybe we shouldn't add this in database layer but rather in quasi-domain?

Can you elaborate on why this is better?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/converters/ISO8601DateConverter.kt
hichamboushaba: I gave this an additional thought yesterday by night, and I'm now a bit convinced the `ISO 8601` is better for our case, as it's human readable, it's easier for debugging purposes, as simple DB inspection can show it's value.
I know it's not a big factor, but it's a factor 😄 

> I believe Unix epoch time is much more robust

I don't think there is any different on this part, ISO 8601 is a robust format too, there is maybe just difference in storage size, but we shouldn't care about it too much.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt
wzieba: > Can you elaborate on why this is better?

Sure, so in this case we have a leak of domain layer to data layer. Database starts to be a source of business logic which shouldn't really be the case - it's not its responsibility. Otherwise we have a domain logic which is blurred across many layers which makes things difficult to test and follow.

If `isCustomer` is false by default (due to anything, either remote or our domain), maybe we don't have to add additional logic to cover that in local database?

Same with other field: if `isSystemNote` always computed before introducing to database, maybe we don't have to assign a default value for that?

hichamboushaba: I agree with this, I'll remove the default values, I'll keep them non-nullable, but without any default values.

hichamboushaba: Done in [1b30f23](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2299/commits/1b30f2395a00c83c06857c4bea659b1397f13823)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderNotesDao.kt
hichamboushaba: I'll merge this to make the fix available on next release, but we can always update this piece of code if we agree on something different 🙏

PR #2301, opened 2022-03-03
Stats update default insights list

ravishanker:
Update the default insights from previous defaults to new defaults as below:

Fixes: [#16029](https://github.com/wordpress-mobile/WordPress-Android/issues/16029)

Previous defaults

1. LATEST_POST_SUMMARY
2. TODAY_STATS
3. ALL_TIME_STATS
4. FOLLOWER_TOTALS

New defaults

1. MOST_POPULAR_DAY_AND_HOUR
2. ALL_TIME_STATS
3. TODAY_STATS
4. FOLLOWERS
5. COMMENTS

PR #2303, opened 2022-03-04
Fix Stats tests to account for new default Insights

renanferrari:
The default Insights list was updated as part of #2301, and even though we updated all relevant unit tests, we ended up missing an optional connected test. This PR updates that test to account for the new default Insights list.

To test:

Run the optional UI/connected tests and make sure the `ReleaseStack_ManageInsightsTestJetpack.testAddAndRemoveInsights` test passes successfully. Feel free to ignore other flaky tests, as that is a known issue unrelated to this.


ravishanker (approved):
Thank you for fixing this. 🙇🏻‍♂️

PR #2304, opened 2022-03-10
Products Room migration #1: Data columns updates

0nko:
This is the 1st step of the Products table migration from WellSql to Room.

This PR adds specific types to certain columns and adds the type converters to the DB.


wzieba:
@0nko just a heads up that I merged #2286 today so it'll for sure will be conflicting with this PR when merging to trunk.

PR #2307, opened 2022-03-10
Woo: Pass blank billing email as null to avoid order creation failure

### LINKED ISSUE #2306: WCAndroid: order creation fails with empty billing email
If I don’t enter a billing email address for a customer, the order fails to be created (backend message: Invalid parameter(s): billing).

nbradbury:
Closes #2306 - If we pass a blank billing email address when creating an order, the server responds with "Invalid parameter(s): billing" and the order fails to be created. This PR works around this by passing a blank email as null.

Only one review is required, but I tagged two reviewers because something about this fix seems hacky and I wanted feedback. It does, however, have the benefit of working :)

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
@@ -10,9 +10,9 @@ import org.wordpress.android.fluxc.action.WCOrderAction
 import org.wordpress.android.fluxc.generated.WCOrderActionBuilder
 import org.wordpress.android.fluxc.generated.endpoint.WOOCOMMERCE
 import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
-import org.wordpress.android.fluxc.model.OrderEntity
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.WCOrderListDescriptor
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt
```diff
@@ -825,15 +825,25 @@ class OrderRestClient @Inject constructor(
...
 
     private fun UpdateOrderRequest.toNetworkRequest(): Map<String, Any> {
-        return mutableMapOf<String, Any>().apply {
+        // the backend will fail to create the order if the billing email is an empty string,
+        // so we pass it as null to avoid this situation
+        val billingDto = if (billingAddress?.email?.isEmpty() == true) {
+            billingAddress.toDto().copy(email = null)
+        } else {
+            billingAddress?.toDto()
+        }
```
ThomazFB: I believe that maybe we could leave the `toDto` function to handle this operation for us as a rule of the data transfer object itself, something like this:

image

WDYT? This way we leave the `toNetworkRequest` function here untouched just building the request parameters.

nbradbury: Works for me! Changed in [02d8aa4](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2307/commits/02d8aa4848167cfe1413df68ba0a57edb5eb0528).

ThomazFB (approved):
Thanks for the changes and improvements @nbradbury! LGTM :shipit:

PR #2308, opened 2022-03-11
Reduce `OrderEntity` payload before saving it

### LINKED ISSUE #2305: Filter out redundant data of `Order` before saving it into database
In order to fix issues with database queries exceeding Android's `Cursor` size, we want to reduce payload of the Order before saving it in local database. 

Some plugins add exceptionally large portion of, not needed in mobile app, data. For more details, please see internal communication at p91TBi-7qQ-p2.

wzieba:
Closes: #2305 

**Description**

This PR reduces payload of OrderEntities, raw `JSON` fields by filtering them before saving in database. For details, please see comments.

**Test cases**
See corresponding Woo PR for area of tests: https://github.com/woocommerce/woocommerce-android/pull/6020

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapper.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapper.kt
@@ -1,32 +1,105 @@
 package org.wordpress.android.fluxc.network.rest.wpcom.wc.order
 
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.model.OrderEntity
 import org.wordpress.android.fluxc.model.order.OrderAddress
+import org.wordpress.android.fluxc.utils.DateUtils
+import java.math.BigDecimal
+import javax.inject.Inject
 
-object OrderDtoMapper {
-    fun OrderAddress.Billing.toDto() = OrderDto.Billing(
-            first_name = this.firstName,
-            last_name = this.lastName,
-            company = this.company,
-            address_1 = this.address1,
-            address_2 = this.address2,
-            city = this.city,
-            state = this.state,
-            postcode = this.postcode,
-            country = this.country,
-            email = this.email,
-            phone = this.phone
-    )
+internal class OrderDtoMapper @Inject constructor(
+    private val stripOrder: StripOrder
+) {
+    fun toDatabaseEntity(orderDto: OrderDto, localSiteId: LocalId): OrderEntity {
+        fun convertDateToUTCString(date: String?): String =
+                date?.let { DateUtils.formatGmtAsUtcDateString(it) } ?: "" // Store the date in UTC format
 
-    fun OrderAddress.Shipping.toDto() = OrderDto.Shipping(
-            first_name = this.firstName,
-            last_name = this.lastName,
-            company = this.company,
-            address_1 = this.address1,
-            address_2 = this.address2,
-            city = this.city,
-            state = this.state,
-            postcode = this.postcode,
-            country = this.country,
-            phone = this.phone
-    )
+        val rawRemoteDataEntity = with(orderDto) {
+            OrderEntity(
+                    orderId = this.id ?: 0,
+                    localSiteId = localSiteId,
+                    number = this.number ?: (this.id ?: 0).toString(),
+                    status = this.status ?: "",
+                    currency = this.currency ?: "",
+                    orderKey = this.order_key ?: "",
+                    dateCreated = convertDateToUTCString(this.date_created_gmt),
+                    dateModified = convertDateToUTCString(this.date_modified_gmt),
+                    total = this.total ?: "",
+                    totalTax = this.total_tax ?: "",
+                    shippingTotal = this.shipping_total ?: "",
+                    paymentMethod = this.payment_method ?: "",
+                    paymentMethodTitle = this.payment_method_title ?: "",
+                    datePaid = this.date_paid_gmt?.let { "${it}Z" } ?: "",
+                    pricesIncludeTax = this.prices_include_tax,
+                    customerNote = this.customer_note ?: "",
+                    discountTotal = this.discount_total ?: "",
+                    discountCodes = this.coupon_lines?.let { couponLines ->
+                        // Extract the discount codes from the coupon_lines list and store them as a comma-delimited String
+                        couponLines
+                                .filter { !it.code.isNullOrEmpty() }
+                                .joinToString { it.code!! }
+                    }.orEmpty(),
+                    refundTotal = this.refunds?.let { refunds ->
+                        // Extract the individual refund totals from the refunds list and store their sum as a Double,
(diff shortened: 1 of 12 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrder.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.OrderMappingConst.SHIPPING_PHONE_KEY
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.OrderMappingConst.isNotInternalAttributeData
+import javax.inject.Inject
+
+internal class StripOrder @Inject constructor(private val gson: Gson) {
+    operator fun invoke(fatModel: OrderEntity): OrderEntity {
+        return fatModel.copy(
+                lineItems = gson.toJson(fatModel.getLineItemList().map { lineItemDto: LineItem ->
+                    lineItemDto.copy(
+                            metaData = lineItemDto.metaData?.filter { it.isNotInternalAttributeData }
```
wzieba, line 16: Logic taken from https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/model/Order.kt#L114

It was only applied to `attributesDescription` and not to `attributesList` but I can't find any attribute that could start with `_` and being used in app. Add-ons don't start with `_`, e.g:
image

Does it make sense to you? Or am I missing some case?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrder.kt
```diff
@@ -0,0 +1,30 @@
...
+                    )
+                }),
+                shippingLines = gson.toJson(fatModel.getShippingLineList()),
+                feeLines = gson.toJson(fatModel.getFeeLineList()),
+                taxLines = gson.toJson(fatModel.getTaxLineList()),
+                metaData = gson.toJson(
+                        fatModel.getMetaDataList()
+                                .filter {
+                                            it.key == CHARGE_ID_KEY || it.key == SHIPPING_PHONE_KEY
```
wzieba, line 25: Logic taken from Woo codebase:

- `CHARGE_ID_KEY`: https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/model/Order.kt#L60-L61
- `SHIPPING_PHONE_KEY`: https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/shippinglabels/creation/CreateShippingLabelViewModel.kt#L433

Supplementary question: what's different with shipping phone passed in `OrderDto#Shipping#phone` vs the one from `MetaData`? @hichamboushaba do you remember maybe this case?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrder.kt
```diff
@@ -0,0 +1,30 @@
...
+    operator fun invoke(fatModel: OrderEntity): OrderEntity {
+        return fatModel.copy(
+                lineItems = gson.toJson(fatModel.getLineItemList().map { lineItemDto: LineItem ->
+                    lineItemDto.copy(
+                            metaData = lineItemDto.metaData?.filter { it.isNotInternalAttributeData }
+                    )
+                }),
+                shippingLines = gson.toJson(fatModel.getShippingLineList()),
+                feeLines = gson.toJson(fatModel.getFeeLineList()),
+                taxLines = gson.toJson(fatModel.getTaxLineList()),
```
wzieba, line 21: The whole logic for reducing payload is to first pass it through our models so during this process, all unneeded data will be left behind. As we don't have domain models, this operations are performed between networking and database layers.
reduce_order

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/utils/JsonObjectBuilder.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.google.gson.JsonElement
+import com.google.gson.JsonObject
+import java.util.ArrayDeque
+import java.util.Deque
+
+internal fun json(build: JsonObjectBuilder.() -> Unit): JsonObject {
+    return JsonObjectBuilder().json(build)
+}
+
+internal class JsonObjectBuilder {
```
wzieba, line 13: Internal, test only class for easier json DSL building.

wzieba:
hi @hichamboushaba @nbradbury @ThomazFB 👋 . Github suggested your review as you recently edited files edited in this PR, I hope you don't mind 🙂. I'll request review also for corresponding PR on Woo side. Thanks!

nbradbury:
@wzieba I did a first pass at this PR and all looks good, but given the size of the changes I'd definitely be more comfortable with another set of eyes on this. @hichamboushaba @ThomazFB

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrder.kt
ThomazFB: I'm not quite sure if we're missing something by doing this filter rule, but if that's the case, we're missing it from the start since this is the logic we're doing so far. And the `attributeList` property only gets attention from the Add-ons logic, so I guess it's safe.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapper.kt
```diff
@@ -1,34 +1,107 @@
...
+                company = this.company,
+                address_1 = this.address1,
+                address_2 = this.address2,
+                city = this.city,
+                state = this.state,
+                postcode = this.postcode,
+                country = this.country,
+                phone = this.phone
+        )
+    }
```
ThomazFB, line 106: Not a change request, but more of a question: I have the feeling that these two extension methods could be in a standalone kt file instead of necessarily being part of a companion object since they're not accessed through the `OrderDtoMapper` but the receiving object instead. Do you think they should be defined like this? I'm just asking because these two methods described as it is can't even be called through a direct `OrderDtoMapper` access, so I don't see the benefit of them exist inside an object.

ThomazFB (approved):
Changes are looking good to me, it's well tested and the improvements look self-contained. Thanks for introducing this enhancement, @wzieba! I'll move to the Woo counterpart of this PR now.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapper.kt
wzieba: True, we can move it to a different file. I'll handle this in next PR!

PR #2309, opened 2022-03-11
Fix robolectric test running locally on M1 mac

JorgeMucientes:
This PR intends to fix the issues with running Robolectric tests on Apple silicon chips. Several crashes needed to be fixed to get the test running: 

1. Fix the original problem. The version we were using of Robolectric did not support Mac M1 chips. Support [was added recently](https://github.com/robolectric/robolectric/issues/6311). Updating to the latest version of robolectric `4.7.3` fixed it.
2. After updating Robolectric version a new problem appeared with this new version. The issue is [described here](https://github.com/robolectric/robolectric/issues/6521). Problems with android jetifier. The solution was making jetifier ignore bouncy castle lib through adding to `gradle.properties` -> `android.jetifier.ignorelist=bcprov`. In this case this has to be added locally, but I've updated `gradle.properties-example` to fix this issue in CI enviroment. 
3. Next issue is that with the newer version of Robolectric we need to do a few tweaking in gradle `testOptions` to make android resources available and [fix this problem](https://github.com/touchlab/KaMPKit/issues/225). Solution was to add this to build.gradle of modules that have robolectric tests: `unitTests.includeAndroidResources  = true` 
4. There is an issue when running robolectric tests and trying to initialize Flipper. More on this [here](https://github.com/facebook/flipper/issues/2075). The solution was to catch with the `NullPointerException` that occurs in that context, so that the tests run successfully. This won't have an impact on initializing Flipper when launching the Example App. 

This PR contains all those changes that enabled fro Robolectric tests to work again in Apple Silicon chips

### DIFF
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -55,7 +55,7 @@ android {
         // pressure on the API (authentication/fetch profile/fetch sites).
         // Uncomment the next line to enable the Orchestrator.
         // execution 'ANDROID_TEST_ORCHESTRATOR'
-
+        unitTests.includeAndroidResources  = true
         unitTests.all {
             if (project.hasProperty('testsMaxHeapSize')) {
                 // maxHeapSize for tests is not limited unless we give an explicit value
@@ -137,7 +137,7 @@ dependencies {
 
     testImplementation 'junit:junit:4.13'
     testImplementation "org.jetbrains.kotlin:kotlin-test-junit:$gradle.ext.kotlinVersion"
-    testImplementation 'org.robolectric:robolectric:4.3.1'
+    testImplementation 'org.robolectric:robolectric:4.7.3'
     testImplementation 'androidx.test:core:1.3.0'
     testImplementation "org.mockito:mockito-core:$mockitoVersion"
     testImplementation 'com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0'
--- a/gradle.properties-example
+++ b/gradle.properties-example
@@ -1,3 +1,4 @@
 org.gradle.jvmargs=-Xmx1536m -XX:+HeapDumpOnOutOfMemoryError
 android.useAndroidX=true
 android.enableJetifier=true
+android.jetifier.blacklist=bcprov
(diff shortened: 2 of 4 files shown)


example/build.gradle
```diff
@@ -55,7 +55,7 @@ android {
         // pressure on the API (authentication/fetch profile/fetch sites).
         // Uncomment the next line to enable the Orchestrator.
         // execution 'ANDROID_TEST_ORCHESTRATOR'
-
+        unitTests.includeAndroidResources  = true
```
anitaa1990, line 58: np: it would be nice to add a comment here maybe to better understand why this is needed?

gradle.properties-example
```diff
@@ -1,3 +1,4 @@
 org.gradle.jvmargs=-Xmx1536m -XX:+HeapDumpOnOutOfMemoryError
 android.useAndroidX=true
 android.enableJetifier=true
+android.jetifier.ignorelist=bcprov
```
anitaa1990, line 4: np: similar to the comment above. People would need to update their gradle.properties with this value and it would be nice to add a comment here stating why this is needed 😊

anitaa1990 (approved):
Thank you for working on this @JorgeMucientes! I can confirm that this fixes the issue mentioned in the PR description 😊 Really great work! 

I've left some nitpicks in the code but will leave the merge to @wzieba since it's been a while since I've coded 🚢

wzieba (approved):
I can't test it but I browsed issues you linked in description @JorgeMucientes , looks good. This change won't affect production code so I'm safe with merging this without testing. Thank you for tweaking this - everyone who moves to M1 will benefit by it :).

wzieba:
btw I've re-run `Mocked Connected Tests` job. The test is failing but it's marked already as "flaky" so let's see if second run will resolve it.

JorgeMucientes:
Oops! Was merged before I was able to address those nitpicks! No worries, I'll P2 about this in the Apps division P2 to complete the info about building FluxC on Apple silicon chips. I'll also ping platform9, as we should update mobile-secrets repo with the change in `gradle.properties`

Thanks @anitaa1990 and @wzieba for checking this out :)

wzieba:
oh sorry @JorgeMucientes @anitaa1990 , I completely missed that!

PR #2310, opened 2022-03-12
Add getRevisionById method to PostStore

RenanLukas:
Implements `getRevisionById` method to `PostStore`. Currently we have a method that returns a single revision (`getLocalRevision`), but always the index 0 element of the returned DB list. There is currently no way to get a revision by ID.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java
@@ -341,6 +341,19 @@ public class PostSqlUtils {
                 .endGroup().endWhere().getAsModel();
     }
 
+    @Nullable
+    public LocalRevisionModel getRevisionById(@NonNull final String revisionId) {
+        final List<LocalRevisionModel> localRevisionModels = WellSql.select(LocalRevisionModel.class)
+                      .where().beginGroup()
+                      .equals(LocalRevisionModelTable.REVISION_ID, revisionId)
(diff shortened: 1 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PostSqlUtils.java
```diff
@@ -341,6 +341,19 @@ public void insertOrUpdateLocalRevision(LocalRevisionModel revision, List<LocalD
                 .endGroup().endWhere().getAsModel();
     }
 
+    @Nullable
+    public LocalRevisionModel getRevisionById(@NonNull final String revisionId) {
```
khaykov: I'm not sure if revision IDs are unique, to be honest. Maybe we should include site and post ID when getting them?

khaykov (commented):
Thanks for the changes! I left some comments here and in the client PR.

Apart from this, I think we also should delete all the revisions, when posts are removed [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java#L1153) and [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/fluxc/src/main/java/org/wordpress/android/fluxc/store/PostStore.java#L1139).

RenanLukas:
Hey @khaykov , thanks for the review! I've implemented your suggestions 👍

khaykov (approved):
Thanks for the changes, @RenanLukas ! Looks good 👍

PR #2315, opened 2022-03-16
Add new endpoints and fetch inbox notes from API

JorgeMucientes:
Closes: #5870

### Description
Adds new endpoints to load Inbox notes from the API. This first PR is just about loading the notes list and returning it. 

In following PRs I will implement saving the notes into Room DB and exposing a flow to observe changes over the persisted data. 
Also, tests will be added for the new WCInboxStore in following PRs as logic grows for saving the loaded notes

### Testing instructions
This should be tested along with this [PR](https://github.com/woocommerce/woocommerce-android/pull/6030) that makes use of the new `WCInboxStore` to load the notes list.


ThomazFB (approved):
Changes look good, I'm just approving so you can merge alongside the Woo changes.

PR #2319, opened 2022-03-22
Convert Update Variation action into suspendable

malinajirka:
Implements  https://github.com/woocommerce/woocommerce-android/issues/6064 partially

**Merge when [the WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/6065) is approved.**

This PR refactors updateVariation into a suspendable function - removes usage of the event bus.

## How to test
1. Run ReleaseStack_WCProductTest - testUpdateVariation()
2. Test `Woo -> Product -> Update Variation` action in the example app

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -1095,7 +1095,6 @@ class WCProductStore @Inject constructor(
                     storedVariation,
                     variation
                 )
-//                onVariationUpdated.causeOfChange = WCProductAction.UPDATED_VARIATION
                 return@withDefaultContext if (result.isError) {
                     OnVariationUpdated(
                         0,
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1086,10 +1082,35 @@ class WCProductStore @Inject constructor(
...
+                    site,
+                    variation.remoteProductId,
+                    variation.remoteVariationId
+                )
+                val result: RemoteUpdateVariationPayload = wcProductRestClient.updateVariation(
+                    site,
+                    storedVariation,
+                    variation
+                )
+//                onVariationUpdated.causeOfChange = WCProductAction.UPDATED_VARIATION
```
AnirudhBhat: ❓ Do you think we can remove this commented line of code?

malinajirka: Ohh, I kept it there just to not forget checking that it's not used in WCAndroid. I didn't forget to check it, but forgot to remove it :D. Thanks!
```suggestion
```

AnirudhBhat (approved):
@malinajirka Thanks for working on this! LGTM! :shipit:

PR #2322, opened 2022-03-23
Fix crash on empty item attribute key

### LINKED ISSUE #2321: NoSuchElementException: Char sequence is empty.
Sentry Issue: [WOOCOMMERCE-ANDROID-3JM](https://sentry.io/organizations/a8c/issues/3123787363/?referrer=github_integration)

```
NoSuchElementException: Char sequence is empty.
    at kotlin.text.StringsKt___StringsKt.first(_Strings.kt:71)
    at kotlin.text.StringsKt.first
    at org.wordpress.android.fluxc.network.rest.wpcom.wc.order.OrderMappingConst.isNotInternalAttributeData$woocommerce_release(OrderMappingConst.kt:9)
    at org.wordpress.android.fluxc.network.rest.wpcom.wc.order.StripOrder.invoke(StripOrder.kt:16)
    at org.wordpress.android.fluxc.network.rest.wpcom.wc.order.StripOrder.invoke(StripOrder.kt:16)
...
(15 additional frame(s) were not displayed)
```

wzieba:
Closes: #2321 

This PR fixes a bug which occurs when `Item`'s `Attribute` (taken from `meta_data` remote key) has an empty `key`.

Possible area of impact:
- Add ons (tested already)
- Variations (I don't suspect but just for double check) 

**If you think there's some valid case when `key` can be empty and used in the app, please share this.**

### Testing instructions

1. Checkout WooCommerce `c85e393` commit (it's `8.8-rc1` commit)
2. Use FluxC from this PR - either by updating `fluxCVersion` or using composite build
3. Smoke tests variations and/or add-ons (focus on possible cases for a valid empty `key` for Attributes)

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrder.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrder.kt
@@ -5,7 +5,7 @@ import org.wordpress.android.fluxc.model.OrderEntity
 import org.wordpress.android.fluxc.model.order.LineItem
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.OrderMappingConst.CHARGE_ID_KEY
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.OrderMappingConst.SHIPPING_PHONE_KEY
-import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.OrderMappingConst.isNotInternalAttributeData
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.OrderMappingConst.isInternalAttribute
 import javax.inject.Inject
 
 internal class StripOrder @Inject constructor(private val gson: Gson) {
@@ -13,7 +13,7 @@ internal class StripOrder @Inject constructor(private val gson: Gson) {
         return fatModel.copy(
                 lineItems = gson.toJson(fatModel.getLineItemList().map { lineItemDto: LineItem ->
                     lineItemDto.copy(
-                        metaData = lineItemDto.metaData?.filter { it.isNotInternalAttributeData }
+                        metaData = lineItemDto.metaData?.filter { it.isInternalAttribute.not() }
                     )
                 }),
                 shippingLines = gson.toJson(fatModel.getShippingLineList()),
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrder.kt
```diff
@@ -13,7 +13,9 @@ internal class StripOrder @Inject constructor(private val gson: Gson) {
         return fatModel.copy(
                 lineItems = gson.toJson(fatModel.getLineItemList().map { lineItemDto: LineItem ->
                     lineItemDto.copy(
-                            metaData = lineItemDto.metaData?.filter { it.isNotInternalAttributeData }
+                            metaData = lineItemDto.metaData?.filter {
+                                it.key.isNotBlank() && it.isNotInternalAttributeData
```
hichamboushaba: np, WDYT about moving the condition to be part of `isNotInternalAttributeData`, I mean:
```kotlin
    internal val WCMetaData.isNotInternalAttributeData
        get() = key.isNotEmpty() && key.first() != '_'
```
I don't have strong opinion about this, but I feel like the condition should be contained by itself, as here the order of conditions is what avoids the crash.

hichamboushaba (approved):
Looks good, and I'm pretty confident it will not have any impact, since we don't read the `key` by itself, instead we rely on the `display_key` https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/86be64d41484772a5a4f580ca34b38a4245de589/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/LineItem.kt#L31

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrder.kt
wzieba: I was thinking about that as well but my contr-argument would be that empty key shouldn't determine if attribute is `internal` or not. But now when I'm putting a second thought to that, it's probably extremely low chance to happen 🤷 . I'll add your suggestion as I agree that it's more pragmatic.

hichamboushaba: > but my contr-argument would be that empty key shouldn't determine if attribute is internal or not

That's a good reason actually, so should we maybe use `key.startsWith('_')` instead of `key.first()`, as it should work with empty strings and avoid the need of `isNotEmpty` 🤔 ?

Just thinking aloud, you are not required to make any more changes 🙂.

wzieba: I like `startsWith` pretty much, thanks! In 5ba7c8e I also inverted `isNotInternalAttribute` to `isInternalAttribute` as the first one contained more negations which I think reduces readability. WDYT about current solution? 🙂

hichamboushaba: Looks nice, thanks so much

PR #2325, opened 2022-03-24
Convert fetch single variation action into suspendable

malinajirka:
Implements  https://github.com/woocommerce/woocommerce-android/issues/6064 partially

**Merge when [the WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/6090) is approved.**

This PR refactors updateVariation into a suspendable function - removes usage of the event bus.

## How to test
1. Run ReleaseStack_WCProductTest - testFetchSingleVariation()
2. Test `Woo -> Product -> Fetch Single Variation` action in the example app


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -936,8 +932,25 @@ class WCProductStore @Inject constructor(
...
 
-    private fun fetchSingleVariation(payload: FetchSingleVariationPayload) {
-        with(payload) { wcProductRestClient.fetchSingleVariation(site, remoteProductId, remoteVariationId) }
+    suspend fun fetchSingleVariation(payload: FetchSingleVariationPayload): OnVariationChanged {
+        return coroutineEngine.withDefaultContext(T.API, this, "fetchSingleVariation") {
+            val result = with(payload) {
+                wcProductRestClient.fetchSingleVariation(site, remoteProductId, remoteVariationId)
+            }
+            return@withDefaultContext if (result.isError) {
+                OnVariationChanged(0).also {
```
kidinov: It looks like that we don't need `OnVariationChanged`, only what we need is the either error or success (without any payload). Maybe to use `Result<Unit>` as return type from this method and rename it to something `fetchSingleVariationAndStoreLocally`?

kidinov (approved):
The code looks and works fine. The only thing I would advise changing is the API we expose here. 

I think something like:


```
suspend fun fetchSingleVariation(
         site: SiteModel,
         remoteProductId: Long,
         remoteVariationId: Long
     ): Result<Unit>
```
Would look like better

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
malinajirka: This is tricky, I think this would require WordPress and WooCommerce wide discussion. [The current architecture documentation](https://github.com/wordpress-mobile/WordPress-FluxC-Android/wiki/Architecture) says `5. Store returns the OnChanged event back to the app.`. I'd personally opt for not changing it as I don't see much benefit and in this case I favor consistency over a slightly better API. Wdyt @kidinov?

malinajirka:
Thanks for the review @kidinov! I've removed the Payload as I agree it's not required since we don't use EventBus anymore. I've left a comment regarding the type the method returns. Let me know what you think, thanks!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
kidinov: @malinajirka 

The page you are referencing to seems to be mixing suspend function approach and event bus (check the picture), also 

>  OnChanged **_event_** 

Why is this an event?

Also, I see that there are other approaches in the same Store:

```
  suspend fun fetchProductCategoryListSynced(
      site: SiteModel,
      categoryIds: List<Long>
  ): List<WCProductCategoryModel>?

    suspend fun generateEmptyVariation(
        site: SiteModel,
        product: WCProductModel
    ): WooResult<WCProductVariationModel>
```

Looking into that I would agree that introducing the use of the kotlin's `Result` here would be wrong. Wdyt about use typeallise here:


```
typealias OnVariationChanged = WooResult<Unit>
```

Or if it feels too complicated I would at least remove unused data from `OnVariationChanged` ?

malinajirka: > The page you are referencing to seems to be mixing suspend function approach and event bus (check the picture), also

My understanding is that the page describes the current approach. If you are referring to "dispatcher" on the picture, I believe it's correct there -> events which rely on event bus in the client apps can still mix usage of both suspend + event bus, so they remain backward compatible.

One thing I agree is different on the picture vs what is implemented in this PR is that we don't pass Action into the suspendable function => we stopped doing that for events which are not relying on event bus at all => read as events which don't need to be backward compatible. Might be worth updating the documentation.

> Also, I see that there are other approaches in the same Store:

I agree but it feels a bit off to use since it isn't following the currently documented architecture. More importantly, I don't see any real benefit there but I see some downsides:

WooResult - AFAIK it doesn't support specifying different error types for different endpoints. I'm not sure if it's usage is documented somewhere but I haven't found it. OnVariationChanged supports a different set of errors - ProductErrorType -> imho WooResult can't be reused for this scenario.

List<WCProductCategoryModel> - this imho breaks the idea FluxC is built around => the idea being, that DB is the only source of truth. My understanding of FluxC is that events should be send to FluxC so FluxC can take care of syncing the data with the remote. Imo these events should not return the data itself - eg. WCProductCategoryModel -> they should just return db identifiers so the client can load the data from the single source of truth = db. The only grey zone being endpoints which are not cached locally - I think it's fine to return the data itself there since there is no other source.

> Or if it feels too complicated I would at least remove unused data from OnVariationChanged ?

I think it wouldn't be complicated to change, but due to the reasons shared above, I'd prefer not to implement the changes. I agree that removing the unused data from the OnChanged events is a good idea, thanks - fixed in 512fd91f22b30b8841330aacf4a7608ef66bc0b2.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -936,8 +922,29 @@ class WCProductStore @Inject constructor(
...
+        site: SiteModel,
+        remoteProductId: Long,
+        remoteVariationId: Long
+    ): OnVariationChanged {
+        return coroutineEngine.withDefaultContext(T.API, this, "fetchSingleVariation") {
+            val result = wcProductRestClient
+                .fetchSingleVariation(site, remoteProductId, remoteVariationId)
+
+            return@withDefaultContext if (result.isError) {
+                OnVariationChanged(0).also {
```
kidinov: You pass 0 as `remoteProductId`, I think it;'s not what you want?

malinajirka: Good call, this was copy pasted from the previous implementation. I think it doesn't make a difference at runtime, since the id gets overridden in `also`, but it's confusing. Fixed in 7e9f4bc61c71bf75ca68078496830c9bcd0704b2.

malinajirka:
> It looks like both remoteProductId and remoteVariationId are not used.

I agree, but I'd vote for keeping them there. They identify what is the "OnChanged" event related to. I agree the caller knows what it's related to, however, if the caller passed the result into another method that method might not know the id. Perhaps, this is one of the downsides of using OnChanged events with suspendable functions. But I think that requires a separate team wide discussion before we start making changes to the architecture - there is a pretty high chance we might be missing the bigger picture.

PR #2330, opened 2022-03-28
Included/excluded coupon categories fix

0nko:
This PR fixes the wrong _excluded_ flag for categories, which would result in the incorrect data being returned for the observed coupons (included categories being returned as excluded and vice-versa). I also added coupon store unit tests that verify the fix & also the data being inserted by the DAO.

**To test:**
Run the unit tests and make sure they pass.


hafizrahman (approved):
Nice work, with the unit tests as well. :shipit:

PR #2331, opened 2022-03-29
Refactored fetch supported woo action into suspendable

anitaa1990:
This PR updates the fetch supported Woo version methods into a suspendable function and removes the usage of the event bus.

## Description
- [The event bus architecture is considered as legacy/deprecated](https://github.com/wordpress-mobile/WordPress-FluxC-Android/wiki/%5BDeprecated%5D-Architecture)
- This PR is in conjecture with the refactoring of the Site picker activity in the WCAndroid app. The site picker screen relies heavily on event bus which need to be registered - this leads to memory leaks when the presenter doesn't unregister from the event bus.
- I also updated the `WooCommerceFragment` to utilise the new `StoreSelectingFragment`. 

## Testing
- Run `WooCommerceStoreTest`.
- Test Woo -> Select Site -> Log Woo api versions.

**DO NOT MERGE this PR until the corresponding Woo PR can be reviewed and merged since this has some breaking changes.**


example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
```diff
@@ -203,6 +221,44 @@ class WooCommerceStoreTest {
         }
     }
 
+    @Test
+    fun `when fetching supported api version succeeds, then success returned`() {
```
atorresveiga, line 225: np: Not something that needs to be addressed but I'm wondering why not use the test block (backdoors is doing the same). Something like
```
    @Test
    fun `when fetching ssr succeeds, then success returned`() = test {
        val result = fetchSSR(isError = false)

        Assertions.assertThat(result.isError).isFalse
        Assertions.assertThat(result.model).isNotNull
    }
```

atorresveiga (approved):
The code looks great, Anitaa! 🎉 
I have tested the changes, and everything is working as expected. I only found a nitpick in using the test block, but as this is not something that needs addressing, I will pre-approve the PR :shipit: 
✅ Run WooCommerceStoreTest.
✅ Test Woo -> Select Site -> Log Woo api versions.
✅ PR on Woo

atorresveiga (approved):

anitaa1990:
Thanks for the review @atorresveiga 🙇‍♀️ Those test cases were already present in the code and not related to the changes in this PR. That being said, I'll update as suggested in the next PR 😊 Thanks!

PR #2332, opened 2022-03-29
Refactored fetch site settings & site product settings into suspendable

anitaa1990:
This PR completed step 2 of woocommerce/woocommerce-android#6111. It converts the fetch site settings and product settings actions into a suspendable function and removes the usage of the event bus.

## Description
- [The event bus architecture is considered as legacy/deprecated](https://github.com/wordpress-mobile/WordPress-FluxC-Android/wiki/%5BDeprecated%5D-Architecture)
- I added a mapper class that converts the parsing logic in `WooCommerceRestClient` into it's own class. 

## Testing
- Run `WooCommerceStoreTest`.
- Run `WCSettingsMapperTest`.
- Test Woo -> Select Site -> Fetch WC Site settings.
- Test Woo -> Select Site -> Fetch WC Site Product settings.

**DO NOT MERGE this PR until the corresponding Woo PR can be reviewed and merged since this has some breaking changes.**


example/src/test/java/org/wordpress/android/fluxc/wc/settings/WCSettingsMapperTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import org.junit.Test
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.model.WCProductSettingsModel
+import org.wordpress.android.fluxc.model.WCSettingsModel
+import org.wordpress.android.fluxc.model.WCSettingsModel.CurrencyPosition
+import org.wordpress.android.fluxc.model.settings.WCSettingsMapper
+import kotlin.test.assertEquals
+
+class WCSettingsMapperTest {
```
atorresveiga, line 11: Great job creating a util class for parsing the WCSettingsModel.  Chapeau! 🎩 🎉

example/src/test/java/org/wordpress/android/fluxc/wc/settings/WCSettingsTestUtils.kt
```diff
@@ -0,0 +1,34 @@
+package org.wordpress.android.fluxc.wc.settings
+
+import org.wordpress.android.fluxc.JsonLoaderUtils.jsonFileAs
+import org.wordpress.android.fluxc.network.discovery.RootWPAPIRestResponse
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.SiteSettingsResponse
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.system.WooSystemRestClient.SSRResponse
+
+object WCSettingsTestUtils {
```
atorresveiga, line 8: Np: Not something to address but maybe the class name could be something like `WCSettingsTestData` because it only provides data for the tests. My picture of a Util class is more of a class with general extension functions, transformations, etc. WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/settings/WCSettingsMapper.kt
```diff
@@ -0,0 +1,57 @@
...
+        val dimensionUnit = getValueForSettingsField(response, "woocommerce_dimension_unit")
+
+        return WCProductSettingsModel().apply {
+            localSiteId = site.id
+            this.dimensionUnit = dimensionUnit ?: ""
+            this.weightUnit = weightUnit ?: ""
+        }
+    }
+
+    private fun getValueForSettingsField(settingsResponse: List<SiteSettingsResponse>, field: String): String? {
```
atorresveiga, line 54: Np: Thinking out loud, maybe this is a good opportunity for an extension function. 🤔 
```
    fun List<SiteSettingsResponse>.getValueForSettingsField(field: String): String? {
        return this.find { it.id != null && it.id == field }?.value?.asString
    }
```

atorresveiga (approved):
Everything looks good to me, Anitaa. Awesome job encapsulating the mapping logic into a separate class, cleaning up, and adding unit tests 🎉.

✅ Run WooCommerceStoreTest.
✅ Run WCSettingsMapperTest.
✅ Test Woo -> Select Site -> Fetch WC Site settings.
✅ Test Woo -> Select Site -> Fetch WC Site Product settings.

example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
```diff
@@ -221,11 +190,57 @@ class WooCommerceStoreTest {
         }
     }
 
+    @Test
+    fun `when fetch site settings succeeds, then success returned`() {
+        runBlocking {
```
atorresveiga: Np: I made the same observation on this [PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2331) I'm wondering if we should use the test block here too
> np: Not something that needs to be addressed but I'm wondering why not use the test block (backdoors is doing the same). Something like
> ```
>     @Test
>     fun `when fetching ssr succeeds, then success returned`() = test {
>         val result = fetchSSR(isError = false)
> 
>         Assertions.assertThat(result.isError).isFalse
>         Assertions.assertThat(result.model).isNotNull
>     }
> ```

anitaa1990: Fixed in [9612813](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2332/commits/9612813e2c2a361671625e0c2ba69c495c8eba33) 👍

PR #2333, opened 2022-03-29
Issue/add replace comment remote parent id column with user

khaykov:
This PR replaces unused `remoteParentCommentId` with `authorId` in Comments table.

At some point we introduced `remoteParentCommentId` to the Comment model, with serves same function as `parentId` and was used like that with self hosted sites. With wpcom/jp sites it was populated with `autorId` for some reason, but luckily it was not used anywhere.

~I removed the `remoteParentCommentId`, and added `autorId` column to Comments table, added migration script~*, updated all the models and switched from using `remoteParentCommentId` to `parentId` in XmlRpc context.

\* we decided to drop and recreate table instead of migration for a better user experience, since `authorId` will be missing immediately after migration.

Companion wpandroid PR is [here](https://github.com/wordpress-mobile/WordPress-Android/pull/16214).


develric (approved):
Thanks for looking into this @khaykov 🙇 ! Code LGTM, also tested it with the companion WPAndroid PR and worked for me 👍

PR #2335, opened 2022-03-30
Issue/5870 save inbox notes

JorgeMucientes:
### Description
Save inbox notes into Room database and expose an observable to the database state so we can observe its changes from WooCommerce app.

Note: This code is only used in WooCommerce Inbox that has not been released yet as is hidden under feature flag. So we have absolute freedom to apply any modifications to the DB

### Testing instructions
The best way to test this is to download the apk  from this [PR](https://github.com/woocommerce/woocommerce-android/pull/6122) and go to More Menu -> Inbox and check that inbox notes are displayed


wzieba:
Just FYI @JorgeMucientes I'm prioritizing this review today.

wzieba:
Just FYI @JorgeMucientes I'm prioritizing this review today.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
```diff
@@ -1,41 +1,80 @@
...
     }
+
+    fun observeInboxNotes(siteId: Long): Flow<List<InboxNoteWithActions>> =
+        inboxNotesDao.observeInboxNotes(siteId)
+            .flowOn(Dispatchers.IO)
+            .distinctUntilChanged()
+
+    private suspend fun saveInboxNotes(result: Array<InboxNoteDto>, siteId: Long) {
+        result.forEach { dto ->
+            database.runInTransaction {
```
wzieba: 🔧 WDYT about moving this logic into `InboxNotesDao` and not add `WCAndroidDatabase` dependency to this class?

My argument would be that this exposes some logic that `*Dao` classes meant to encapsulate. Moving this to Dao will make e.g. tests easier (no need to mock database behaviour [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2335/files#diff-912040930691dc12a90d410efa4f42517c5d4e7a1e502f6aad277289eafbdec5R43-R45)).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
```diff
@@ -1,41 +1,80 @@
...
+                    WooResult(Unit)
                 }
                 else -> WooResult(WooError(GENERIC_ERROR, UNKNOWN))
             }
         }
     }
+
+    fun observeInboxNotes(siteId: Long): Flow<List<InboxNoteWithActions>> =
+        inboxNotesDao.observeInboxNotes(siteId)
+            .flowOn(Dispatchers.IO)
```
wzieba: ❓ Is this needed? I believe by default Room will handle threading: e.g. `CoroutinesRoom#createFlow` selects `transactionDispatcher` for that ([source](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/room-ktx/src/main/java/androidx/room/CoroutinesRoom.kt;l=119?q=CoroutinesRoom)) which I believe is not main by default. Maybe we can check if this operation is on main thread without this line?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
```diff
@@ -1,41 +1,80 @@
...
-                    WooResult(response.result.toList())
+                    saveInboxNotes(response.result, site.siteId)
+                    WooResult(Unit)
                 }
                 else -> WooResult(WooError(GENERIC_ERROR, UNKNOWN))
             }
         }
     }
+
+    fun observeInboxNotes(siteId: Long): Flow<List<InboxNoteWithActions>> =
```
wzieba, line 48: 💯 I like CQRS-like approach for `fetch` and `observe` here!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/InboxNoteEntity.kt
```diff
@@ -0,0 +1,39 @@
...
+    val id: Long,
+    val siteId: Long,
+    val name: String,
+    val title: String,
+    val content: String,
+    val dateCreated: String,
+    val status: LocalInboxNoteStatus,
+    val source: String? = null,
+    val type: String? = null,
+    val dateReminder: String? = null
```
wzieba: ❓ I _think_ we don't experience discrepancy issues related to default values as described [here](https://developer.android.com/training/data-storage/room/migrating-db-versions#handle-default-values-migrations) but I'd love it if you could confirm @JorgeMucientes 🙏 .

wzieba (commented):
I see that `10.json` and `11.json` files are edited - is that a mistake or intended behaviour? If intended, I'm not sure if we should do this, won't it cause validation schema error for existing users?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
```diff
@@ -1,41 +1,80 @@
...
-        return coroutineEngine.withDefaultContext(API, this, "fetchCoupons") {
-            val response = restClient.fetchNotes(site, page, pageSize)
+    ): WooResult<Unit> {
+        return coroutineEngine.withDefaultContext(API, this, "fetchInboxNotes") {
+            val response = restClient.fetchInboxNotes(site, page, pageSize)
             when {
                 response.isError -> WooResult(response.error)
                 response.result != null -> {
-                    WooResult(response.result.toList())
+                    saveInboxNotes(response.result, site.siteId)
```
hichamboushaba, line 40: Sorry for lurking around, I was just checking this PR as part of the Woo PR.
Two questions:
1. Shouldn't you handle data invalidation here? when fetching the first page, normally you would need to clear the DB to avoid keeping reference to items that were deleted in the backend.
2. What's the plan for pagination? I see the `fetch` function returns `Unit`, which means we can't know if we fetched all pages or not.

hichamboushaba:
@JorgeMucientes just adding to what @wzieba's mentioned above about the schema files, I think a merge conflict resulted in making changes to `10.json` and `11.json` schemas, normally they shouldn't be touched at all, you will just introduce the new version.
Also the fact that you have `DROP` statements in the migration is causing this crash:
```no such table: InboxNotes (code 1 SQLITE_ERROR[1]): , while compiling: DROP TABLE InboxNotes```
Since the tables are added for the first time, I don't think there is any need for the `DROP` statements.

File: plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt [hunk 6]
```diff
@@ -1,41 +1,80 @@
...
         }
     }
+
+    fun observeInboxNotes(siteId: Long): Flow<List<InboxNoteWithActions>> =
+        inboxNotesDao.observeInboxNotes(siteId)
+            .flowOn(Dispatchers.IO)
+            .distinctUntilChanged()
+
+    private suspend fun saveInboxNotes(result: Array<InboxNoteDto>, siteId: Long) {
+        result.forEach { dto ->
```
hichamboushaba: This forEach is what's causing the UI glitch you mentioned @JorgeMucientes in the WCAndroid's PR, I wrote a P2 about it paqN3M-sT-p2 to let everyone know about this hidden cost/impact of Room.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
hichamboushaba: Just a tip, if you decide to leave it here (I don't have a strong opinion about either solutions), please use `withTransaction` instead of `runInTransaction`, this would allow removing the `coroutineEngine.launch` below, as `withTransaction` is already suspendable.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
hichamboushaba: I agree, we don't need to specify the dispatcher when using Room.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
JorgeMucientes: Oh! just seen this comment after reading your post. Thanks so much for sharing that insight. I was driving nuts to see so many "emission" through the flow when updating the DB. 
Thank you!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/InboxNoteEntity.kt
JorgeMucientes: hmm, interesting didn't know about that. I'd say [this](https://developer.android.com/training/data-storage/room/migrating-db-versions#handle-default-values-migrations) does not have an impact when setting the default value as null on an nullable field. 
Anyway, I think this is redundant, fields are nullable and if there is no present value in the DB they are going to equal null anyways. So, I'm removing this default asigments. Thanks for pointing it out.

JorgeMucientes:
Thank you so much for all the great insight @wzieba and @hichamboushaba. 

I think I've addressed all the issues and suggestions, and this is ready for another round of review. 

Regarding pagination. I'm still unsure if in the end we are going to need it. From what I see in my different accounts, that have been accumulating "a lot" of inbox notes, the list is never too long (around 50-60 elements). Also, the items are pretty lightweight, with not much data in them, and the request is super fast. For those reasons I kind of set the pagination task to low priority and will get back to it once the rest of the tasks are done. 

One thing I noticed while testing the latest changes is there is a bug on how data is retrieved from the DB. I'm still loking into it but I'll share the issue in case you may know an already existing solution for it. 

So, `InboxNotes` are stored per each `siteId` and `InboxNoteActions` are always associated to an `InboxNoteId` and `siteId` 
The problem comes when switching sites multiple times and saving Inbox notes for all of them. When retrieving the InboxNotes with their `noteActions` I'm using the `@Relation` approach: 
```kotlin
data class InboxNoteWithActions(
    @Embedded val inboxNote: InboxNoteEntity,
    @Relation(
        parentColumn = "id",
        entityColumn = "inboxNoteId"
    )
    val noteActions: List<InboxNoteActionEntity>
)
```
The problem with this is that this approach is ignoring the `siteId` for the `noteAction`. I've checked if Room allows something like composite primary keys, but it didn't: 
```kotlin
data class InboxNoteWithActions(
    @Embedded val inboxNote: InboxNoteEntity,
    @Relation(
        parentColumn =  ["id", "siteId"],
        entityColumn = ["inboxNoteId", "siteId"]
    )
    val noteActions: List<InboxNoteActionEntity>
)
```
I could also build my own custom query in a transactional way like [suggested here](https://stackoverflow.com/a/57729242/2979970) but I think that won't allow me to benefit from the `observer` pattern I'm using with Flows. 

There is could also be easily solved with Room `v2.4.0` using [multimap return types @Dao methods](https://developer.android.com/jetpack/androidx/releases/room#:~:text=contains%20these%20commits.-,New%20Features,-Room%20now%20supports). But still need to solve the blocking issue with Room. 

I'm missing something? Will keep looking into it to find a solution, but let me know if you've faced something similar in the past and know about existing solution :)

hichamboushaba:
I'm on my phone, so I can't go into details, but from what you explained @JorgeMucientes, I think this can be easily solved if you add a local generated primary id 🤔, you can keep the constraint of [siteId + remote Id] same as @wzieba did with the intermediate step for OrderEntity.

JorgeMucientes:
Thanks for the suggestion @hichamboushaba although I'm not 100% percent I followed your suggested approach. I implemented a solution using the `localId`. It works fine the only thing I don't like too much is that whenever I fetch inbox notes from the API to refresh the local DB I have cannot simply do "upserts" but I have to remove all the inbox notes for the given `siteId` first and then insert everything again. Why? because now `InboxNoteEntity` has a unique autogenerated primary key. I was not able to keep the constraint of [siteId + remote Id] for `InboxNoteEntity` as you mentioned. Unsure what I'm missing there.

Specific changes for this solution can be found in this [commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2335/commits/d511de554ea09ee47f1c84daa135e7af722def1f). Let me know your thoughts 🙂

hichamboushaba:
> It works fine the only thing I don't like too much is that whenever I fetch inbox notes from the API to refresh the local DB I have cannot simply do "upserts" but I have to remove all the inbox notes for the given siteId first and then insert everything again.

I think this should be the preferred behavior no matter what solution you use, you need to invalidate the cached data after fetching new one to make sure deleted notes from the server are deleted in the app as well.

> I was not able to keep the constraint of [siteId + remote Id] for InboxNoteEntity as you mentioned

I didn't mean to use it in the `@Relation`, I meant adding a unique constraint like what we had for `OrderEntity` before removing the `local_id` https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/c794717110fe67912f92311cefc03d49ba0602e3/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCOrderModel.kt#L18-L21 It's not mandatory, but just as a way to make sure the DB respects the constrait of not duplicating notes per site id, and also for allowing queriying using [`remote_id` + `site_id`] without any performance hit.
For the `@Relation`, you need to use `local_id`, as it's the primary and unique ID of each note.

JorgeMucientes:
> you need to invalidate the cached data after fetching new one to make sure deleted notes from the server are deleted in the app as well

You are right, invalidation is needed anyway. 

> I didn't mean to use it in the @Relation, I meant adding a unique constraint like what we had for OrderEntity before removing the local_id

Yeah, I wasn't thinking about `@Relation` either. I was thinking about how to keep that uniqueness constraint for  `[siteId + remote Id] ` on `InboxNote` table once the combination of `[siteId + remote Id] ` was removed from being the primary key. Didn't know defining an index as `unique = true` would enforce that constraint. I thought that was only for improving performance on queries. TIL. Thank you! Added it.

JorgeMucientes:
Hey folks 👋🏼  @hichamboushaba @wzieba. Would any of you be able to take a look at this PR tomorrow? 🙏  Would be great so that can unblock this other [PR](https://github.com/woocommerce/woocommerce-android/pull/6122#issuecomment-1088943790)

example/src/test/java/org/wordpress/android/fluxc/store/WCInboxStoreTest.kt
```diff
@@ -0,0 +1,139 @@
...
+    @Test
+    fun `Given notes are fetched successfully, when notes fetched, then save them into DB`() =
+        test {
+            givenNotesAreFetchedSuccesfully()
+
+            sut.fetchInboxNotes(ANY_SITE)
+
+            verify(inboxNotesDao).insertInboxNotesAndActions(
+                ANY_SITE.siteId, *INBOX_NOTES_WITH_ACTIONS_ENTITY.toTypedArray()
+            )
```
hichamboushaba: np, shouldn't we check if the table is cleared as well.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/InboxNotesDao.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    abstract suspend fun insertOrUpdateInboxNoteAction(entity: InboxNoteActionEntity)
+
+    @Query("DELETE FROM InboxNotes WHERE siteId = :siteId")
+    abstract fun deleteInboxNotesForSite(siteId: Long)
+
+    @Transaction
+    open suspend fun insertInboxNotesAndActions(siteId: Long, vararg notes: InboxNoteWithActions) {
+        deleteInboxNotesForSite(siteId)
```
hichamboushaba, line 30: np, this function does two things, and it's not clear from its signature, if I see this function, I wouldn't assume it will clear the table before inserting the new notes, I think it's better to be either:
1. Rename the function.
2. Call the `deleteInboxNotesForSite` in the store directly.

example/src/test/java/org/wordpress/android/fluxc/store/WCInboxStoreTest.kt
hichamboushaba: I see that that the `delete` function is called from the insert one, so this is not valid (depending on what you decide for the comment below)

hichamboushaba (approved):
Looks good @JorgeMucientes, I left two np comments, but nothing blocking.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/InboxNotesDao.kt
JorgeMucientes: Agree! Renamed it (although not sure if that is the best naming but at least is clearer that is doing 2 things). Option 2 will require adding `database` dependency back to `WcInboxStore` to run everything within the same transaction. I rather avoid that. 

Thanks for reviewing this and all the suggestions!

PR #2337, opened 2022-03-30
[Coupons] Fetch and observe a coupon

### LINKED ISSUE #2336: [Coupons]: add endpoint to fetch a single coupon for a store
This issue is part of the coupon management project and adds a new endpoint to fetch/display single coupon for a store

0nko:
Fixes #2336, which adds a method to fetch and observe a single coupon.

**To test:**
Run the `CouponDaoTests` tests.
+
1. Start the example app
2. Sign in & go to Woo -> Coupons -> Select site
3. Tap on Fetch single coupon button
4. Enter a random string
5. Notice an error message is printed
6. Tap on the button again and enter a random number
7. Notice a message is displayed that the coupon wasn't found
8. Tap on the button again and enter a valid coupon ID
9. Notice the coupon entity is printed


hichamboushaba (approved):
Nice work @0nko, this looks great.

Just one nickpick comment regarding the name of the functions, IMHO the `Single` part is not needed and is just a noise, I think calling the functions: `fetchCoupon` and `observeCoupon` should be enough to explain what they do, this in addition to their signature, and it would also match the other variant: `insertCoupon` and `insertOrUpdateCoupon`, but this is a subjective matter, and it's not blocking 🙂.

0nko:
> Just one nickpick comment regarding the name of the functions, IMHO the Single part is not needed and is just a noise, I think calling the functions: fetchCoupon and observeCoupon should be enough to explain what they do, this in addition to their signature, and it would also match the other variant

Thanks for the review, @hichamboushaba! I totally agree, I was just being consistent with products and orders. I'll change it though.

hichamboushaba:
Thanks @0nko for making the changes, :shipit:

PR #2338, opened 2022-03-31
WC Coupons: fetching performance reports

hichamboushaba:
This work is needed for https://github.com/woocommerce/woocommerce-android/issues/5530, it adds support for fetching performance report for given coupons using `wc-analytics` endpoint.

For information, I decided to not cache this data since for the following reasons:
1. As it's about analytics, it needs to be up-to-date always.
2. It's accessed only inside the coupons details screen, so whenever the user opens this screen, we will fetch it no matter what.
3. If the device is offline, it doesn't hurt to not have it (maybe 🤔).

If you don't agree with those arguments, please let me know, and just for the record, iOS seem to be doing the same.

#### Testing
1. Open the example app.
2. Click on Woo.
3. Click on Coupons.
4. Select a site.
5. Click on "Fetch coupon report"
6. Enter a coupon id.
7. Confirm the report result is printed to the log.


0nko (approved):
Looking great! 🎖️

PR #2339, opened 2022-04-01
Issue/5875 handle inbox note actions

JorgeMucientes:
FluxC changes needed for:  [#5875](https://github.com/woocommerce/woocommerce-android/issues/5875)

This PR is ready to be reviewed but please do not merge until the base branch of this PR is automatically updated to `trunk`

### Description 
Add missing functions and API endpoints to support:

- Marking an Inbox note as actioned when the user clicks on any of the actions (except "Dismiss")
- Dismiss an inbox note. This is basically marking the note as "Deleted" in backend side, so that the note does not appear in the list results anymore. 

### Demo
Changes can be tested with [this apk ](https://github.com/woocommerce/woocommerce-android/pull/6158#issuecomment-1090019384) generated from the WooComerce Android changes [here](https://github.com/woocommerce/woocommerce-android/pull/6158)

https://user-images.githubusercontent.com/2663464/162023638-72b5e164-c86b-4051-83f7-7fd15ac59424.mp4


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/inbox/InboxRestClient.kt
```diff
@@ -52,4 +52,47 @@ class InboxRestClient @Inject constructor(
...
+            .action.item(inboxNoteActionId).pathV4Analytics
+
+        val response = jetpackTunnelGsonRequestBuilder.syncPostRequest(
+            this,
+            site,
+            url,
+            emptyMap(),
+            InboxNoteDto::class.java
+        )
+        return when (response) {
```
ovitrif, line 67: np: This could be written like the `return when` of `deleteNote`, if we're expecting nothing to change then we don't necessarily need the curly braces for the `when` cases

ovitrif (approved):
Great work @JorgeMucientes!

The code LGTM, unfortunately I can't test its integration in the WooCommerce Android app because I don't know how to enable the `Inbox` card on the Menu tab of the app.

I downloaded the app by scanning the QR code in this PR: woocommerce/woocommerce-android/pull/6158 but I get what's in the screenshot and didn't find the option on how to enable `Inbox` like you showed in the clip of the PR 🤷‍♂️

<details>
<summary>App → Menu</summary>


</details>

I'd really love if you can guide me how to do this so that I could "complete" the review with testing, that will also be great for me to find out more about the WooCommerce App.

Thanks in advance 🙇‍♂️

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/inbox/InboxRestClient.kt
JorgeMucientes: True! I've removed the braces 👍🏼

JorgeMucientes:
Thanks for reviewing this @ovitrif 🙂

> I downloaded the app by scanning the QR code in this PR: https://github.com/woocommerce/woocommerce-android/pull/6158 but I get what's in the screenshot and didn't find the option on how to enable Inbox like you showed in the clip of the PR 🤷‍♂️

Ouch! That is right! I forgot this feature is hidden under a feature flag. This flag will only make the Inbox button visible in `debug` builds. So, the only way to test this now would be to compile from AS WooCommerce app from `issue/5875-handle-inbox-note-action` which is far from convenient, but right now seems the only way to do it.

ovitrif:
> Ouch! That is right! I forgot this feature is hidden under a feature flag. This flag will only make the Inbox button visible in `debug` builds. So, the only way to test this now would be to compile from AS WooCommerce app from `issue/5875-handle-inbox-note-action` which is a understand is far from convenient, but right seem the only way to do it.

Thank you! Next time I'll know 🥳

PR #2343, opened 2022-04-06
Fixed failing testInstantiateAndCreateReplyComment test

khaykov:
Test condition got messed up in #2333. This PR fixes is.

To test:
 - I will run connected tests for this PR. Make sure the `testInstantiateAndCreateReplyComment` is not failing (there are some toher flaky tests that might fail tho).


wzieba (approved):
Looks good! 

I'm jumping here as I'd love to check if #2287 is not failing any test cases after merging this fix.

PR #2345, opened 2022-04-07
Bump Kotlin to `1.5.32`

wzieba:
Part of: #2279

### Description

This PR bumps Kotlin to the last `1.5.x` stable version (`1.5.32`). It also contains a fix for changed Kotlin behaviour. For details please see this post in internal communication: paqN3M-ti-p2

**Why not bump to Kotlin `1.6.x`?**

The currently used Room version is not supporting Kotlin `1.6.x` - there's a kapt crash during compilation. I think we can bump Kotlin to `1.5.x` here and then in #2287 both Kotlin and Room.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
@@ -37,6 +37,7 @@ class XMLRPCRequestBuilder
             method,
             params,
             // **Do not** convert it to lambda! See https://youtrack.jetbrains.com/issue/KT-51868
+            @Suppress("RedundantSamConstructor")
             Listener<Any> { obj: Any? ->
                 if (obj == null) {
                     errorListener.invoke(BaseNetworkError(INVALID_RESPONSE))
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
@@ -1,5 +1,6 @@
 package org.wordpress.android.fluxc.network.xmlrpc
 
+import com.android.volley.Response.Listener
 import kotlinx.coroutines.suspendCancellableCoroutine
 import org.wordpress.android.fluxc.generated.endpoint.XMLRPC
 import org.wordpress.android.fluxc.network.BaseRequest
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
```diff
@@ -31,16 +32,28 @@ class XMLRPCRequestBuilder
         listener: (T) -> Unit,
         errorListener: (BaseNetworkError) -> Unit
     ): XMLRPCRequest {
-        return XMLRPCRequest(url, method, params, { obj: Any? ->
```
ParaskP7, line 34: Suggestion (💡): Consider adding the `@Suppress("RedundantSamConstructor")` annotation for the statement here to make it explicity that we know about this problem and chose to explicitly suppress it. This way, this converting recomendationdue to the redundant-SAM constructor will no longer appear both in the IDE or any build logs.

ParaskP7 (approved):
👋 @wzieba !

Thank you so much for opening this PR, I review and tested it, it LGTM! 🥇 

> This PR bumps Kotlin to the last 1.5.x stable version (1.5.32). It also contains a fix for changed Kotlin behaviour. For details please see this post in internal communication: paqN3M-ti-p2

Once more thank you for getting deep into that and figuring out the root cause of why this update was failing our tests. I also replied to you via paqN3M-ti-p2. 🌟

> The currently used Room version is not supporting Kotlin 1.6.x - there's a kapt crash during compilation. I think we can bump Kotlin to 1.5.x here and then in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2287 both Kotlin and Room.

Yeap, I did try that myself while trying to update Kotlin to `1.6.10` last week and got this crash as well. I was followign closely this #2287 PR on bumping Room to `2.4.1` and I too think we should bump those two, together, in one go, it would be indeed easier. We did the same with `Kotlin 1.6.10` and `Compose 1.1.1` for Woo (see [here](https://github.com/woocommerce/woocommerce-android/pull/6086)). I think is situation like this we need to be doing two bumps at once and only hope to avoid doing such in the future. 🤞 

PS.1: If you too agree, I would recommend updating this #2287 Room PR's description to `Bump Kotlin to 1.6.10 & Room to 2.4.1`.
PS.2: I also triggered the `Optional Connected Tests` CI job just to get an extra 👍 from our CI, just in case something else get discovered. I really hope we won't discover any more problems with this update. 🤞

fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
wzieba: Nice, good idea!

PR #2346, opened 2022-04-07
Revert "Bump Kotlin to `1.5.32`"

wzieba:
Reverts wordpress-mobile/WordPress-FluxC-Android#2345

Sorry @ParaskP7 , I think I've missed that `Optional Connected Tests` failed. I'm investigating it now this can take some time - let's revert the bump.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/XMLRPCRequestBuilder.kt
@@ -1,6 +1,5 @@
 package org.wordpress.android.fluxc.network.xmlrpc
 
-import com.android.volley.Response.Listener
 import kotlinx.coroutines.suspendCancellableCoroutine
 import org.wordpress.android.fluxc.generated.endpoint.XMLRPC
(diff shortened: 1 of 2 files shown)


ParaskP7:
👋 @wzieba !

> Sorry @ParaskP7 , I think I've missed that Optional Connected Tests failed.

Yea, I triggered those but they never completed. Then I thought you merged because you got a 🟢 light from CI, weird... So the FluxC PRs can be merged even if this job fails? 🤔 

>  I'm investigating it now this can take some time - let's revert the bump.

I think the failure you see existed even before your merge, please double check that, maybe we shouldn't revert after all. 🤔 

PS: Even if we end up reverting, maybe just reverting Kotlin to `1.4.10` is enough. We don't need to revert the `XMLRPCRequestBuilder.kt` fix, do we? Just to keep it simple for this revert and next bump PR. Wdyt?

wzieba:
Thanks @ParaskP7! I've seen broken tests and immediately associated them with the Kotlin bump (especially as they were XMLRPC-related). But it seems that the bump was not a cause and fixes for failing tests are already in review:
- #2343
- #2347

ParaskP7:
👋 @wzieba !

Great, thank you for coming back on that and... 😅

PR #2347, opened 2022-04-07
Fix Woo 11 to 12 database migration

wzieba:
Connected PR: #2335 

I think that we ended up with a broken 11 to 12 migration, I've noticed this as it broke some Optional Tests 😅 .

**To test**
Run `testMigrate11to12` test locally.

### DIFF
--- a/plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
+++ b/plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationTests.kt
@@ -8,6 +8,7 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_10_11
+import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_11_12
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_3_4
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_4_5
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_5_6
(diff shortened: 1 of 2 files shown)


JorgeMucientes:
Ouch, dammit!! Forgot to manually test the migration after applying the last changes in #2335 

Thanks so much for this quick fix 🙇🏼‍♂️  and for adding proper tests. Should’ve checked there was a `MigrationTest.kt` class and added the test myself 🤦🏼 . The good side is I learned quite a bit regarding Room usage with these last PRs. 

Btw, I couldn't run the test locally because I got this compile error (either in this branch or in trunk) probably to the recent AGP or Kotlin updates. But I see the test ran successfully in CI so there must be something wrong in my local environment. Don't want to block this fix because of that. So approving and merging! 
```
> Task :fluxc:kaptDebugKotlin FAILED
> Task :example:mergeExtDexDebug

* What went wrong:
Execution failed for task ':fluxc:kaptDebugKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask$KaptExecutionWorkAction
   > java.lang.reflect.InvocationTargetException (no error message)

and then at the end of the stacktrace there is the culprit...

Caused by: java.lang.Exception: No native library is found for os.name=Mac and os.arch=aarch64. path=/org/sqlite/native/Mac/aarch64
	at org.sqlite.SQLiteJDBCLoader.loadSQLiteNativeLibrary(SQLiteJDBCLoader.java:333)
	at org.sqlite.SQLiteJDBCLoader.initialize(SQLiteJDBCLoader.java:64)
	at androidx.room.verifier.DatabaseVerifier.<clinit>(DatabaseVerifier.kt:71)
	... 53 more
```

JorgeMucientes (approved):

wzieba:
oh yes @JorgeMucientes - please see this internal communication: p1649266627673559/1521528167.000206-slack-C0QB79UV8

I'll also put this somewhere globally!

PR #2348, opened 2022-04-08
Coupons - Paging support

0nko:
This PR adds paging support to for fetching coupons. 

**To test:**
1. Make sure your test store has > 3 coupons
2. Start the example app and go to Woo -> Coupons
3. Tap on the Fetch coupons button
4. Notice the 1st page of 3 coupons is fetched
5. Keep tapping the button, the list of fetched coupons will grow
6. When all coupons are fetched, notice the button gets disabled


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -50,18 +50,27 @@ class CouponStore @Inject constructor(
...
             when {
                 response.isError -> WooResult(response.error)
                 response.result != null -> {
-                    response.result.forEach { addCouponToDatabase(it, site) }
-                    WooResult(Unit)
+                    database.withTransaction {
+                        // clear the table if the 1st page is requested
+                        if (page == 1) {
+                            couponsDao.deleteAllCoupons(site.siteId)
+                        }
```
hafizrahman, line 65: Nice 👍🏼

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -47,6 +47,7 @@ class CouponStore @Inject constructor(
         const val DEFAULT_PAGE = 1
     }
 
+    // Returns a boolean indicating whether more coupons can be fetched
```
hafizrahman, line 50: Thanks for adding this!

hafizrahman (dismissed):
Code looks good to me! There's now 5 failed tests in `CouponStoreTest`, I think we can merge once those are fixed.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -193,15 +193,17 @@ class CouponStore @Inject constructor(
                     assembleCouponDataModel(site, it)
                 }
             }
+            .distinctUntilChanged()
+            .flowOn(Dispatchers.IO)
```
hichamboushaba: Room uses its own internal dispatcher, there is no need on specifying a different dispatcher here.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -193,15 +193,17 @@ class CouponStore @Inject constructor(
...
 
     @ExperimentalCoroutinesApi
     fun observeCoupons(site: SiteModel): Flow<List<CouponDataModel>> =
         couponsDao.observeCoupons(site.siteId)
             .mapLatest { list ->
                 list.map { assembleCouponDataModel(site, it) }
             }
-            .flowOn(Dispatchers.IO)
             .distinctUntilChanged()
+            .flowOn(Dispatchers.IO)
```
hichamboushaba: Same as above.

hichamboushaba (commented):
Works well, I left just two comments about the usage of `Dispatchers.IO`.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
0nko: Good call.

example/src/test/java/org/wordpress/android/fluxc/store/CouponStoreTest.kt
```diff
@@ -219,12 +221,16 @@ class CouponStoreTest {
...
             productStore,
-            database
+            transactionExecutor
         )
+    }
 
-        val blockArg1 = argumentCaptor<Runnable>()
-        whenever(database.runInTransaction(blockArg1.capture())).then {
-            blockArg1.firstValue.run()
+    private suspend fun mockTransaction() {
```
hichamboushaba: np, another way instead of having a separate function here, would be to use the `mock` function for mocking instead of the annotation, then replace the code with:
```kotlin
    private val transactionExecutor: TransactionExecutor = mock {
        val blockArg1 = argumentCaptor<suspend () -> Unit>()
        onBlocking { executeInTransaction(blockArg1.capture()) }.then {
            TEST_SCOPE.launch {
                blockArg1.lastValue.invoke()
            }
        }
    }
```

example/src/test/java/org/wordpress/android/fluxc/store/CouponStoreTest.kt
```diff
@@ -219,12 +221,16 @@ class CouponStoreTest {
...
+        val blockArg1 = argumentCaptor<suspend () -> Unit>()
+        whenever(transactionExecutor.executeInTransaction(blockArg1.capture())).then {
+            TEST_SCOPE.launch {
```
hichamboushaba: np, I would prefer to use runBlocking here to make sure the result of the blockArg1 is returned in the mocked version, which would match the behavior of executeInTransaction.
I wouldn't worry about any delays to be caused by runBlocking, since the functions we call in tests here are just not really blocking.

hichamboushaba (approved):
Looks good @0nko, I left two np comments, but nothing blocking.

example/src/test/java/org/wordpress/android/fluxc/store/CouponStoreTest.kt
0nko: This is good, thanks for the tip!

hichamboushaba (approved):
Looks good, thanks @0nko

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
hafizrahman: TIL!

PR #2349, opened 2022-04-08
[Gutenberg] EditorTheme - Add Quote block V2 flag

geriux:
- **Gutenberg Mobile PR**: https://github.com/wordpress-mobile/gutenberg-mobile/pull/4744
- **Gutenberg PR**: https://github.com/WordPress/gutenberg/pull/40133
- **WordPress Android PR**: https://github.com/wordpress-mobile/WordPress-Android/pull/16286

### Description

Adds a Gutenberg feature flag for the Quote V2 block, for more information check the Gutenberg PR description.

### To test

Check the Gutenberg PR description.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/EditorTheme.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/EditorTheme.kt
@@ -21,6 +21,7 @@ const val MAP_KEY_ELEMENT_STYLES: String = "rawStyles"
 const val MAP_KEY_ELEMENT_FEATURES: String = "rawFeatures"
 const val MAP_KEY_IS_FSETHEME: String = "isFSETheme"
 const val MAP_KEY_GALLERY_WITH_IMAGE_BLOCKS: String = "galleryWithImageBlocks"
+const val MAP_KEY_QUOTE_BLOCK_V2: String = "quoteBlockV2"
 
 data class EditorTheme(
     @SerializedName("theme_supports") val themeSupport: EditorThemeSupport,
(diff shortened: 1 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/EditorTheme.kt
```diff
@@ -34,7 +35,8 @@ data class EditorTheme(
                     blockEditorSettings.styles?.toString(),
                     blockEditorSettings.features?.toString(),
                     blockEditorSettings.isFSETheme,
-                    blockEditorSettings.galleryWithImageBlocks
+                    blockEditorSettings.galleryWithImageBlocks,
+                    blockEditorSettings.quoteBlockV2
```
mchowning, line 39: I was a little surprised to see `quoteBlockV2` (and `galleryWithImageBlocks`) in this `themeSupports` object. I assume this is because of the endpoint this information is coming through?

mchowning (approved):
Code looks good. Tested and approved via https://github.com/wordpress-mobile/WordPress-Android/pull/16286#pullrequestreview-940047192

fluxc/src/main/java/org/wordpress/android/fluxc/model/EditorTheme.kt
geriux: > I assume this is because of the endpoint this information is coming through?

That's correct, I think initially it made sense because we were only using the Global styles data, but lately, it has been useful for feature flags and maybe more features for the block editor itself. Maybe we should think about having the block editor settings separate.

PR #2350, opened 2022-04-08
Coupons - Search functionality

0nko:
This PR adds the search (by string) functionality.

**To test:**
1. Go to Woo -> Coupons
2. Select a site
3. Tap on the Search coupons button
4. Enter a query that is present in some of the coupons
5. Notice the query results in coupons being fetched & saved in the database
6. Enter a query that isn't present
7. Notice no data is returned and database isn't changed

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
@@ -74,7 +74,6 @@ class CouponStore @Inject constructor(
         }
     }
 
-    // Returns a boolean indicating whether more coupons can be fetched
     suspend fun searchCoupons(
         site: SiteModel,
         searchString: String,
@@ -89,8 +88,6 @@ class CouponStore @Inject constructor(
                     database.executeInTransaction {
                         response.result.forEach { addCouponToDatabase(it, site) }
                     }
-
-                    database
 
                     val couponIds = response.result.map { it.id }
                     val coupons = couponsDao.getCoupons(site.siteId, couponIds)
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -74,10 +74,40 @@ class CouponStore @Inject constructor(
...
+        return coroutineEngine.withDefaultContext(API, this, "searchCoupons") {
+            val response = restClient.fetchCoupons(site, page, pageSize, searchString)
+            when {
+                response.isError -> WooResult(response.error)
+                response.result != null -> {
+                    database.executeInTransaction {
+                        response.result.forEach { addCouponToDatabase(it, site) }
+                    }
+
+                    database
```
hichamboushaba: I think this dangling mention of the argument "database" here is not needed/useful.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -74,10 +74,40 @@ class CouponStore @Inject constructor(
         }
     }
 
-    suspend fun fetchCoupon(
+    // Returns a boolean indicating whether more coupons can be fetched
```
hichamboushaba: This comment is not accurate I'm afraid, I think it was just mistakenly copied from `fetchCoupons`.

hichamboushaba (approved):
Nice work @0nko, left two np comments, but nothing is blocking.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
0nko: It was true but then I decided to return the actual search results, so the `Boolean` was replaced by `CouponSearchResult`. Thanks for noticing!

PR #2352, opened 2022-04-08
woo/2351-add-searched-products-to-db

### LINKED ISSUE #2351: Woo: Add searched products to the database
When a product search finishes, [handleSearchProductsCompleted](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/23e8fe682e21e4e045cf365a06ca8be46f55d425/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt) doesn't store the matching products in the database. This impacts us in order creation because if the user searches for products and adds a matching product to their order, the product won't be in the database and fails to be added to the order.

[Related WCAndroid issue](https://github.com/woocommerce/woocommerce-android/issues/6217)

nbradbury:
Fixes #2351 - in WCAndroid's order creation, if the user searches for a product and chooses one that's not already in the database, it fails to be added to the order because product search doesn't store its results in the database. This PR corrects this by storing the searched products.

Related Slack thread: p1649443131016309-slack-CGPNUU63E

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -1215,14 +1215,14 @@ class WCProductStore @Inject constructor(
                     payload.site,
                     payload.products
                 )
-                emitChange(
-                    OnProductsSearched(
-                        payload.searchQuery,
-                        payload.products,
-                        payload.canLoadMore
-                    )
-                )
             }
+            emitChange(
+                OnProductsSearched(
+                    payload.searchQuery,
+                    payload.products,
+                    payload.canLoadMore
+                )
+            )
         }
     }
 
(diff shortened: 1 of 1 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1207,12 +1207,23 @@ class WCProductStore @Inject constructor(
...
+                    payload.products
+                )
+            }
+            emitChange(
+                OnProductsSearched(
+                    payload.searchQuery,
+                    payload.products,
+                    payload.canLoadMore
+                )
+            )
```
hichamboushaba: I would move this inside the `launch` block, because otherwise we might let the observers know about the search result before finishing saving the results in the DB.

hichamboushaba (approved):
Works well @nbradbury, thanks for the fix.
I left one comment to consider.

And am I right in assuming you will create a companion WC PR to integrate the change?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
nbradbury: Good call. Changed in [0491dc9](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2352/commits/0491dc9863788613ad859b0775a36cc57748e193).

nbradbury:
> And am I right in assuming you will create a companion WC PR to integrate the change?

Yes.

PR #2355, opened 2022-04-12
Issue/5876 dismiss all notes

### LINKED ISSUE #5876: [Home Screen updates - Inbox] Add toolbar action to dismiss all Inbox notes
Add overflow menu in Toolbar with an action to dismiss all the notes at once.

JorgeMucientes:
Closes: [#5876 ](https://github.com/woocommerce/woocommerce-android/issues/5876)

### Description
This PR adds the missing endpoint and database methods to delete all existing Inbox notes all at once. Related to this [PR](https://github.com/woocommerce/woocommerce-android/pull/6236) where the "Dismiss All" button is added. 

How the API endpoint: `/wc-analytics/admin/notes/delete/all` works is kind of "special" to say the least 😅 . Basically, you can't just make a request to that endpoint and expect all notes to be deleted. You have to delete notes in a paginated way. This pagination is limited to delete notes at maximum rate of 25 at the time. So if you have more than 25 notes in your Inbox, and user hits the "Dismiss All" notes button we will have to make as many requests as necessary to remove all the notes in blocks of 25 or less notes. 

### Testing instructions
- Go to More Menu -> Inbox notes. 
- Click overflow menu -> Dismiss all. 
- You should see an empty Inbox
- Navigate back 
- Enter again in the Inbox and wait until notes are fetched from the API. The screen should still be empty. 
- You can also check that the inbox in `wc-core` is empty

### Images/gif

https://user-images.githubusercontent.com/2663464/163374226-30330c43-8340-4bcf-a9db-80d09714c31b.mp4


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
```diff
@@ -23,17 +23,27 @@ class WCInboxStore @Inject constructor(
...
+        // for site we have to calculate the number of "pages" to be deleted. This only applies for deletion.
+        const val MAX_PAGE_SIZE_FOR_DELETING_NOTES = 25
         const val DEFAULT_PAGE_SIZE = 100
         const val DEFAULT_PAGE = 1
+        val INBOX_NOTE_TYPES_FOR_APPS = arrayOf(
+            "info",
+            "survey",
+            "marketing",
+            "warning"
+        )
```
JorgeMucientes, line 36: Added this as we are only interested in those kind of types for the Apps

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
```diff
@@ -23,17 +23,27 @@ class WCInboxStore @Inject constructor(
     private val inboxNotesDao: InboxNotesDao
 ) {
     companion object {
+        // API has a restriction were no more than 25 notes can be deleted at once so when deleting all notes
+        // for site we have to calculate the number of "pages" to be deleted. This only applies for deletion.
+        const val MAX_PAGE_SIZE_FOR_DELETING_NOTES = 25
```
JorgeMucientes, line 28: This is the weird part I was mentioning in the main comments. We can only remove notes in chunks of 25 or less notes. So I've added a comment to try to explain this. 
This limitation only affects DELETION. For GET requests, page size can be anything.

0nko (approved):
LGTM, nice job! 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/InboxNotesDao.kt
```diff
@@ -13,9 +13,13 @@ import org.wordpress.android.fluxc.persistence.entity.InboxNoteWithActions
 @Dao
 abstract class InboxNotesDao {
     @Transaction
-    @Query("SELECT * FROM InboxNotes WHERE siteId = :siteId ORDER BY dateCreated DESC")
+    @Query("SELECT * FROM InboxNotes WHERE siteId = :siteId ORDER BY dateCreated DESC, remoteId DESC")
```
wzieba, line 16: 👋 The PR looks good, I have a minor concern here: why do we sort (order) the items here? If it's for presentation purposes, I'd suggest moving this logic to the presentation layer, WDYT?

JorgeMucientes: Thanks for the suggestion @wzieba . The reason I did it here is: 

- Simplicity. It's easier and less error prone to sort this here rather than at viewModel layer with custom sort algorithm. 
- Performance. I though sorting on database would be faster but this was more a gut feeling than a proven fact 😅 . A quick search on google revealed that that might not be true specially when sorting using non indexed columns. So, this is not a valid point anymore. 

If we are being strict about architecture, I agree it belongs in the presentation layer. On the other side, the simplicity of doing it here makes me doubt I should refactor this and move it to presentation layer where I would manually have to parse each string to a date and then apply sorting function to it. 

Wdyt? Do you think it's worth it?

wzieba: okay I see!

> Simplicity. It's easier and less error prone to sort this here rather than at viewModel layer with custom sort algorithm.

I think we could do this without hustle with `List#sortedBy` method in VM, no need for custom sort algorithm. 

> the simplicity of doing it here makes me doubt I should refactor this and move it to presentation layer where I would manually have to parse each string to a data and then apply sorting function to it.

I'm just wondering what if we'll need to `observeInboxNotes` with different order, but that's something we can deal with when it'll be needed (maybe never?). Thanks for response!

JorgeMucientes: If you need to observe the list in a different order you can just apply your custom sorting elsewhere without any problem. We'd be doing duplicated sorting, yes, but the impact would be minimal as the list of notes is very short. But as you said, we can deal with that in the future if that use case is ever needed. YAGNI for the win 😄

PR #2357, opened 2022-04-12
Issue/handle non numerical ID in notification range

khaykov:
This PR adds a handling logic for non numerical ID's in the forgettable notification ranges.
We can safely ignore the non-numerical values, so instead of adding a special parsing logic, I just concealed the string ID behind custom getter, that will return null if it fails to parse String into Long.

This PR can be tested with [this](https://github.com/wordpress-mobile/WordPress-Android/pull/16314) client-side PR.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/tools/FormattableContentMapper.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/tools/FormattableContentMapper.kt
@@ -82,7 +82,7 @@ data class FormattableRange(
     @SerializedName("value") val value: String? = null,
     @SerializedName("indices") val indices: List<Int>? = null
 ) {
-    // IS in json response is string, and can be non numerical.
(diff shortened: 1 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/tools/FormattableContentMapper.kt
```diff
@@ -82,6 +82,15 @@ data class FormattableRange(
     @SerializedName("value") val value: String? = null,
     @SerializedName("indices") val indices: List<Int>? = null
 ) {
+    // IS in json response is string, and can be non numerical.
```
develric: np: I think this was intended to be // ID in json, maybe?

develric (approved):
Thanks for this PR @khaykov 🙇 ! LGTM. I only left a minor np for what I think could be a typo in a comment. Feel free to merge this though after you see it the np makes sense 🙇 .

PR #2361, opened 2022-04-14
Coupons: Fix coupons ordering

0nko:
This PR removes the custom ordering of coupons when observing the DB. The default ordering is by date.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
@@ -16,11 +16,11 @@ import org.wordpress.android.fluxc.persistence.entity.CouponWithEmails
 @Dao
 abstract class CouponsDao {
     @Transaction
-    @Query("SELECT * FROM Coupons WHERE siteId = :siteId ORDER BY id")
+    @Query("SELECT * FROM Coupons WHERE siteId = :siteId")
     abstract fun observeCoupons(siteId: Long): Flow<List<CouponWithEmails>>
 
     @Transaction
-    @Query("SELECT * FROM Coupons WHERE siteId = :siteId AND id IN (:couponIds) ORDER BY id")
+    @Query("SELECT * FROM Coupons WHERE siteId = :siteId AND id IN (:couponIds)")
     abstract fun getCoupons(siteId: Long, couponIds: List<Long>): List<CouponWithEmails>
 
     @Transaction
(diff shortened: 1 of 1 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
```diff
@@ -16,11 +16,11 @@ import org.wordpress.android.fluxc.persistence.entity.CouponWithEmails
 @Dao
 abstract class CouponsDao {
     @Transaction
-    @Query("SELECT * FROM Coupons WHERE siteId = :siteId ORDER BY id")
+    @Query("SELECT * FROM Coupons WHERE siteId = :siteId")
     abstract fun observeCoupons(siteId: Long): Flow<List<CouponWithEmails>>
 
     @Transaction
-    @Query("SELECT * FROM Coupons WHERE siteId = :siteId AND id IN (:couponIds) ORDER BY id")
```
hichamboushaba, line 23: @0nko I would still keep an `ORDER BY` clause, but just one that matches the order of the API, this is because we insert coupons for some cases other than the regular fetch call, which might mean the coupons will be inserted in an order different than `date`.
So if the API returns result ordered by `dateCreated` by default, we should use it here too.

WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
```diff
@@ -20,11 +20,13 @@ abstract class CouponsDao {
...
     @Transaction
-    @Query("SELECT * FROM Coupons WHERE siteId = :siteId AND id IN (:couponIds)")
+    @Query("SELECT * FROM Coupons " +
+        "WHERE siteId = :siteId AND id IN (:couponIds) ORDER BY dateCreated")
     abstract fun getCoupons(siteId: Long, couponIds: List<Long>): List<CouponWithEmails>
 
     @Transaction
-    @Query("SELECT * FROM Coupons WHERE siteId = :siteId AND id = :couponId")
+    @Query("SELECT * FROM Coupons " +
+        "WHERE siteId = :siteId AND id = :couponId ORDER BY dateCreated")
```
hichamboushaba: @0nko the ORDER BY should be inside `observeCoupons`, not `observeCoupon` 😄

0nko: Ehm, of course, I was just testing you... 😅

hichamboushaba (approved):
Looks good, thanks for the fixes @0nko

PR #2364, opened 2022-04-19
Issue/woo 6119 fetch variations list

malinajirka:
Implements  https://github.com/woocommerce/woocommerce-android/issues/6119 partially

**Merge when [the WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/6294) is approved.**

This PR refactors fetchProductVariations into a suspendable function - removes usage of the event bus.

## How to test
1. Run ReleaseStack_WCProductTest - testFetchProductVariations()
2. Test `Woo -> Product -> Fetch Product Variations` action in the example app


AnirudhBhat (approved):
Thanks for working on this! LGTM! :shipit:

PR #2365, opened 2022-04-20
Wait for api response success before deleting all notes from DB

JorgeMucientes:
Small fixes and refactor to: 
- Ensure API request was successful before clearing the inbox notes from DB
- Fix json parsing error in `InboxRestClient.deleteAllNotesForSite()` as the response was expecting an `array` instead of an `object` and previous implementation was making the function return error always even if the request was succesful.
- Minor refactor to simplify `deleteNotesForSite` function

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
@@ -95,26 +95,29 @@ class WCInboxStore @Inject constructor(
         site: SiteModel,
         pageSize: Int = MAX_PAGE_SIZE_FOR_DELETING_NOTES,
         inboxNoteTypes: Array<String> = INBOX_NOTE_TYPES_FOR_APPS
-    ): List<WooResult<Unit>> =
+    ): WooResult<Unit> =
         coroutineEngine.withDefaultContext(API, this, "fetchInboxNotes") {
-            val sizeOfCachedNotesForSite = inboxNotesDao.getInboxNotesForSite(site.siteId).size
-            var numberOfPagesToDelete = sizeOfCachedNotesForSite / MAX_PAGE_SIZE_FOR_DELETING_NOTES
-            if (sizeOfCachedNotesForSite % MAX_PAGE_SIZE_FOR_DELETING_NOTES > 0) {
-                numberOfPagesToDelete++
-            }
-            val results = mutableListOf<WooResult<Unit>>()
-            for (page in 1..numberOfPagesToDelete) {
-                val response = restClient
+            var latestResult = WooResult(Unit)
+            for (page in 1..getNumberOfPagesToDelete(site)) {
+                latestResult = restClient
                     .deleteAllNotesForSite(site, page, pageSize, inboxNoteTypes)
-                results.add(response.asWooResult())
-
-                if (response.isError) break
+                    .asWooResult()
+                if (latestResult.isError) break
             }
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
```diff
@@ -102,13 +102,16 @@ class WCInboxStore @Inject constructor(
...
             for (page in 1..numberOfPagesToDelete) {
-                results.add(
-                    restClient.deleteAllNotesForSite(site, page, pageSize, inboxNoteTypes)
-                        .asWooResult()
-                )
+                val response = restClient
+                    .deleteAllNotesForSite(site, page, pageSize, inboxNoteTypes)
+                results.add(response.asWooResult())
+
+                if (response.isError) break
```
JorgeMucientes, line 111: Added this improvement to avoid making more request than needed. If any request fails the we stop trying with the following pages.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
```diff
@@ -102,13 +102,16 @@ class WCInboxStore @Inject constructor(
...
+            }
+            if (!results.any { it.isError }) {
+                inboxNotesDao.deleteInboxNotesForSite(site.siteId)
```
JorgeMucientes, line 114: This is the actual fix. Clear DB only if all requests are successful.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCInboxStore.kt
```diff
@@ -95,23 +95,29 @@ class WCInboxStore @Inject constructor(
...
-            val sizeOfCachedNotesForSite = inboxNotesDao.getInboxNotesForSite(site.siteId).size
-            var numberOfPagesToDelete = sizeOfCachedNotesForSite / MAX_PAGE_SIZE_FOR_DELETING_NOTES
-            if (sizeOfCachedNotesForSite % MAX_PAGE_SIZE_FOR_DELETING_NOTES > 0) {
-                numberOfPagesToDelete++
+            var latestResult = WooResult(Unit)
+            for (page in 1..getNumberOfPagesToDelete(site)) {
+                latestResult = restClient
+                    .deleteAllNotesForSite(site, page, pageSize, inboxNoteTypes)
+                    .asWooResult()
+                if (latestResult.isError) break
```
JorgeMucientes, line 105: Small optimization. If one request fails we stop doing any more requests and propagate the error.

0nko:
This PR essentially removes the optimistic "dismiss all" behavior, since API call to dismiss inbox notes must be successful before the DB is cleared. Previously, the notes were deleted in the DB first. Is this correct? Just want to make sure that it is what we want.

JorgeMucientes:
Yes! That is the behavior we are seeking with these changes. This might not be the best solution from UX point of view but from what we've discussed in this other [PR](https://github.com/woocommerce/woocommerce-android/pull/6280) I think is a good balance between UX and effort. Keep in mind we don't know yet how many users are going to make use of "Dismiss all" action. I rather wait to see stats for this before investing more time into it, as [solutions](https://github.com/woocommerce/woocommerce-android/pull/6280#issuecomment-1104804915) to improve the experience for error handling in this case, imo are not straightforward or out of scope for the task.

0nko (approved):
Sounds good! 👍

PR #2366, opened 2022-04-21
[Coupons] some fixes

hichamboushaba:
This PR attempts to fix two issues: https://github.com/woocommerce/woocommerce-android/issues/6329 and https://github.com/woocommerce/woocommerce-android/issues/6330, with the following changes:

1. Making the sort logic in the DB the same as the API, meaning it's using a descending order using the "created_date".
2. Reverts to using a destructive update logic, meaning if an item exists, then it will be deleted before inserting the new item.

@0nko regarding point two, I know this was the behavior before, and you updated it in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2337, do you remember what was the reason? I checked the code, and it seems all of the tables that has a foreign key with the "Coupons" table are tables that normally should be cleared on each update, so it should be the behavior we want.

#### Testing
You can use the WCAndroid PR https://github.com/woocommerce/woocommerce-android/pull/6331 to test changes of this PR, and confirm the following:
1. Sorting matches what the API returns now (and matches iOS).
2. The steps explained in https://github.com/woocommerce/woocommerce-android/issues/6330 are not reproducible anymore.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/CouponsDao.kt
@@ -16,12 +16,12 @@ import org.wordpress.android.fluxc.persistence.entity.CouponWithEmails
 @Dao
 abstract class CouponsDao {
     @Transaction
-    @Query("SELECT * FROM Coupons WHERE siteId = :siteId ORDER BY dateCreated")
+    @Query("SELECT * FROM Coupons WHERE siteId = :siteId ORDER BY dateCreated DESC")
     abstract fun observeCoupons(siteId: Long): Flow<List<CouponWithEmails>>
 
     @Transaction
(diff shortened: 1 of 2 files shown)


0nko (approved):
> regarding point two, I know this was the behavior before, and you updated it in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2337, do you remember what was the reason? I checked the code, and it seems all of the tables that has a foreign key with the "Coupons" table are tables that normally should be cleared on each update, so it should be the behavior we want.

While writing unit tests for `CouponsDao` I noticed that when the `CouponEntity` was updated, all the entities with Coupons FK got deleted (the test you updated began to fail). I found this behavior surprising because a I thought a mere change to the parent entity shouldn't affect the relationships with children (I wrote a post about it: wp.me/paqN3M-sH). However, due to the way coupons are being inserted and updated (and not replaced), it broke the products/categories updates and the previous approach was, in fact, correct.

Anyways, thanks for fixing this! 🙇

PR #2368, opened 2022-04-26
Adds: Vault press active on site reason

AjeshRPai:
Part of https://github.com/wordpress-mobile/WordPress-Android/issues/14168

Parent https://github.com/wordpress-mobile/WordPress-Android/pull/16400

This PR adds a reason(VaultPress active). This will be used in the Use case to propagate the correct message to be shown to the user. 

**To test**
There's nothing to test in this PR as this change will be consumed in the Parent PR. Make sure that the reason has the same value as the response for a VP-powered site, which is
```
{
    "state": "unavailable",
    "threats": null,
    "has_cloud": true,
    "reason": "vp_active_on_site"
}
```

[Reference](https://github.com/wordpress-mobile/WordPress-iOS/issues/17146#issuecomment-1108787780)


ashiagr (approved):
LGTM 👍

PR #2371, opened 2022-04-27
[Coupons] Update coupons

0nko:
This PR adds the update coupon functionality & the example app test UI. 

Sorry about the large PR, there are many properties in coupons so that made the UI code lengthy. I will add unit tests later on separately.

image
image
image

**To test:**
1. Go to Woo -> Products
2. Click on Fetch products
3. Click on Fetch product categories
4. Navigate back to Woo -> Coupons
5. Select a site
6. Click on Update coupon
7. Click on Enter coupon ID, enter a valid coupon ID number and wait for a coupon to load
8. **Verify that all data is correctly loaded**
9. Modify different properties
10. **Tap on the check mark button and verify the data is correctly updated**
11. Tap on the Select expiry date
12. **Verify a date selection dialog appears**
13. Select a date
14. **Verify a correct date appears**
15. Tap on the Incl. and Excl. products
16. **Verify a dialog appears with a list of products**
17. Select different items and tap OK
18. **Verify the IDs are displayed correctly**
19. Do the same for incl. and excl. categories
20. **Update the coupon again and verify the products and categories are updated correctly**

**Note:**
After this PR is merged, the corresponding [Woo PR](https://github.com/woocommerce/woocommerce-android/pull/6419) must be also updated and merged to fix the build.


example/src/main/java/org/wordpress/android/fluxc/example/ui/WooCommerceFragment.kt
```diff
@@ -203,9 +203,7 @@ class WooCommerceFragment : StoreSelectingFragment() {
         }
 
         coupons.setOnClickListener {
-            selectedSite?.let {
-                replaceFragment(WooCouponsFragment())
-            } ?: showNoWCSitesToast()
+            replaceFragment(WooCouponsFragment())
```
0nko, line 206: The site needs to be selected in the coupons fragment, so no need to force it here.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -160,23 +161,29 @@ class CouponStore @Inject constructor(
         fetchMissingProductCategories(dto.productCategoryIds, site)
         fetchMissingProductCategories(dto.excludedProductCategoryIds, site)
 
-        dto.productCategoryIds?.forEach { categoryId ->
+        productCategoriesDao.getProductCategoriesByIds(
```
0nko: In case the required categories are not fetched, only insert those that are in the DB to avoid a missing FK exception.

example/src/main/java/org/wordpress/android/fluxc/example/ui/ListSelectorDialog.kt
```diff
@@ -26,11 +28,25 @@ class ListSelectorDialog : DialogFragment() {
             this.listItems = listItems
             this.selectedListItem = selectedListItem
         }
+
+        fun newInstance(
+            fragment: Fragment,
+            items: List<Triple<Long, String, Boolean>>,
```
hichamboushaba: np, WDYT about creating a simple data class for this to avoid the ambiguous Triple usage here.

example/src/main/java/org/wordpress/android/fluxc/example/ui/coupons/WooUpdateCouponFragment.kt
```diff
@@ -0,0 +1,382 @@
...
+            enableCouponDependentButtons()
+
+            entered_coupon_id.visibility = View.GONE
+            enter_coupon_id.visibility = View.GONE
+        }
+
+        product_ids.isEnabled = false
+        excluded_product_ids.isEnabled = false
+        category_ids.isEnabled = false
+        excluded_category_ids.isEnabled = false
```
hichamboushaba, line 121: Just a question, why do we need to disable those text views here? and if they should disabled always, we should be able to do it in XML.

example/src/main/java/org/wordpress/android/fluxc/example/ui/coupons/WooUpdateCouponFragment.kt
```diff
@@ -0,0 +1,382 @@
...
+import org.wordpress.android.fluxc.persistence.entity.CouponEntity.DiscountType
+import org.wordpress.android.fluxc.store.CouponStore
+import org.wordpress.android.fluxc.store.ProductCategoryStore
+import org.wordpress.android.fluxc.store.ProductStore
+import org.wordpress.android.fluxc.store.WooCommerceStore
+import org.wordpress.android.fluxc.utils.DateUtils
+import java.util.Calendar
+import javax.inject.Inject
+
+class WooUpdateCouponFragment : Fragment() {
```
hichamboushaba, line 40: np, I wish this was inhereting from `StoreSelectingFragment`, I know it's not a small change, so feel free to leave it as it's

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/coupons/CouponDto.kt
```diff
@@ -31,17 +32,17 @@ data class CouponDto(
...
             siteId = siteId,
             code = code,
-            amount = amount,
+            amount = amount?.toBigDecimalOrNull(),
             dateCreated = dateCreated,
             dateCreatedGmt = dateCreatedGmt,
             dateModified = dateModified,
             dateModifiedGmt = dateModifiedGmt,
-            discountType = discountType,
+            discountType = Companion.fromString(discountType),
```
hichamboushaba: This import makes this harder to understand, I think `DiscountType.fromString()` should work fine.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
```diff
@@ -67,7 +69,9 @@ import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_9_10
             LocalIdConverter::class,
             LongListConverter::class,
             RemoteIdConverter::class,
-            BigDecimalConverter::class
+            BigDecimalConverter::class,
+            DiscountTypeConverter::class,
+            ISO8601DateConverter::class
```
hichamboushaba: IMO these two converters should not be applicable to all tables:
1. DiscountTypeConverter is specific to coupons.
2. ISO8601DateConverter forces a choice on how the date is parsed, and how it's saved, which might not work for all cases (for example GMT vs non-GMT dates).

WDYT about moving the converter declaration to the Entity or even the property itself (like [this](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/889b815f5bfeddfc877f1bc6820fa72129b422c7/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderNoteEntity.kt#L14))?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEntity.kt
```diff
@@ -14,12 +15,12 @@ data class CouponEntity(
     val id: Long,
     val siteId: Long,
     val code: String? = null,
-    val amount: String? = null,
+    val amount: BigDecimal? = null,
```
hichamboushaba, line 20: Just a reminder for both of us 🙂, this and the below changes are breaking changes, so we need a small WCAndroid PR to accompany this before the merge.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
hichamboushaba: This would result in data loss, I think it's better to report the whole coupon fetching as failed in this case instead of showing an inaccurate coupon?

hichamboushaba (commented):
Really nice work @0nko, I didn't test the feature yet, I just took a look at the code, and I left some comments for you to consider.

I will continue the review later today.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
0nko: Hmm, yeah, I guess that's safer.

example/src/main/java/org/wordpress/android/fluxc/example/ui/coupons/WooUpdateCouponFragment.kt
0nko: I based this class on a copy of `WooUpdateProductFragment`, which also inherits from `Fragment`. Personally, I find it annoying having to select a site on every single screen. If you need to change it, you can always do it in the parent fragment. WDYT?

example/src/main/java/org/wordpress/android/fluxc/example/ui/coupons/WooUpdateCouponFragment.kt
0nko: They are disabled in XML, but once a coupon is loaded, all views are enabled by calling `enableCouponDependentButtons()`.

0nko:
Thank you for the review, @hichamboushaba! All feedback should be either addressed or I responded with a comment. Ready for another round. 🙇‍♂️

example/src/main/java/org/wordpress/android/fluxc/example/ui/coupons/WooUpdateCouponFragment.kt
hichamboushaba: I see, I just missed the fact that the `siteId` is being passed from the parent fragment, so I take back what I said, this works very well.

example/src/main/java/org/wordpress/android/fluxc/example/ui/coupons/WooUpdateCouponFragment.kt
hichamboushaba: I see, then why not using some textviews instead of EditText? I know it's like this on other screens, but just wondering why it's the case.

And please don't feel any obligation to change this, I'm just discussing, and it's a very minor thing.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
```diff
@@ -157,66 +161,110 @@ class CouponStore @Inject constructor(
...
+        categoryIds: List<Long>,
+        areExcluded: Boolean
+    ) {
+        fetchMissingProductCategories(categoryIds, site)
+
+        val dbCategoryIds = productCategoriesDao
+            .getProductCategoriesByIds(site.siteId, categoryIds)
+            .map { it.id }
+
+        if (dbCategoryIds.sorted() != categoryIds.sorted()) {
```
hichamboushaba: super np, I think comparing the lists' sizes should be enough, and maybe even easier to read (IMHO).

example/src/main/java/org/wordpress/android/fluxc/example/ui/coupons/WooUpdateCouponFragment.kt
```diff
@@ -0,0 +1,383 @@
...
+        outState.putLong(ARG_SELECTED_SITE_ID, siteId)
+        couponId?.let { outState.putLong(ARG_SELECTED_COUPON_ID, it) }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        if (!isAddNewCoupon) {
+            enter_coupon_id.setOnClickListener {
+                showSingleLineDialog(activity, "Enter the ID of coupon to fetch:") { editText ->
```
hichamboushaba: np, passing `isNumeric = true` would make the experience better.

hichamboushaba (approved):
This looks super nice @0nko, I left two np comments, but it's ready for merge whenever you want.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/CouponStore.kt
0nko: Yeah, you're right 👍

example/src/main/java/org/wordpress/android/fluxc/example/ui/coupons/WooUpdateCouponFragment.kt
0nko: Yeah, it's like this because I just copied it from another fragment 🙂. I guess one benefit is that you can simply change the IDs manually if you want to. Also, you can see what it is based on the label above. Since it's already working I think I'll just keep it.

PR #2367, opened 2022-04-22
[Issue-2320] Add Product variation batch update support

samiuelson:
## 🎯  Summary
This PR adds support for product variations batch update. 

## 🛠  Implementation details
* Added `/products/<id>/variations/batch endpoint`
* Implemented POST call to batch update variations `ProductRestClient::batchUpdateVariations`. Also introduced `BatchProductVariationsUpdateApiResponse` class wich models backend response.
* Implemented `WCProductStore::batchUpdateVariations` method responsible for variations batch update on the backend. It updates variations locally after a successful backend response.
* Created `BatchUpdateVariationsPayload` class consumed by `::batchUpdateVariations` and introduced `BatchUpdateVariationsPayload.Builder` to simplify creating variations updates map required by the REST API.

## 🧪  Testing

For now, you can run `WCProductStoreTest`.  New test methods were added here. 

## 🧭  Roadmap

Next PR to this branch will add `example` app update allowing to test batch update operation on the device.

Issue #2320, opened 2022-03-22
Woo: Add product variation batch update support


0nko:
Implement the endpoint client to support batch updates of product variations (https://woocommerce.github.io/woocommerce-rest-api-docs/#batch-update-product-variations).

PR #2374, opened 2022-04-30
[Issue-2320] Variations batch update: example app

samiuelson:
**This PR updates `example` to allow testing variations batch update. It's based on the other [PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2367) which introduced batch variations update feature in `WcProductStore`. It is the last implementation part of the https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/2320 issue.** 

#### 🎯 Summary
This PR updates `example` app to allow testing variations batch update feature.

#### 🛠 Implementation details
* Added `WooBatchUpdateVariationsFragment`, accessible from `WooProductsFragment`.

#### 🧪 Testing
Run the example app and:
1. Navigate to `Batch Update Variations` screen.  
2. Enter product id, and ids of variations (comma-separated).
3. Click `Set up product id and variations ids` button to validate ids and check if they exist in DB.
4. Fill in the form below with the data you want to batch update the variations.
5. Click the button at the bottom.

In-app logs should inform about the product/variations validation errors and the batch update operation result.

#### 🧭 Roadmap
Once this PR is merged, there will be an aggregate PR raised containing the previous [PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2367) and closing the https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/2320 issue.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooBatchUpdateVariationsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooBatchUpdateVariationsFragment.kt
@@ -158,9 +158,6 @@ class WooBatchUpdateVariationsFragment : Fragment() {
             with(stock_quantity.getText()) {
                 if (isNotEmpty()) variationsUpdatePayloadBuilder.stockQuantity(this.toInt())
             }
-            with(stock_quantity.getText()) {
-                if (isNotEmpty()) variationsUpdatePayloadBuilder.stockQuantity(this.toInt())
-            }
             with(weight.getText()) {
                 if (isNotEmpty()) variationsUpdatePayloadBuilder.weight(this)
             }
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooBatchUpdateVariationsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooBatchUpdateVariationsFragment.kt
@@ -1,0 +1,252 @@
+package org.wordpress.android.fluxc.example.ui.products
+
+import android.app.DatePickerDialog
+import android.app.DatePickerDialog.OnDateSetListener
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.lifecycleScope
+import dagger.android.support.AndroidSupportInjection
+import kotlinx.android.synthetic.main.fragment_woo_batch_update_variations.*
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.Dispatchers.Main
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import org.wordpress.android.fluxc.example.R.layout
+import org.wordpress.android.fluxc.example.prependToLog
+import org.wordpress.android.fluxc.example.ui.ListSelectorDialog
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooResult
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.product.BatchProductVariationsUpdateApiResponse
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.product.CoreProductStockStatus
+import org.wordpress.android.fluxc.store.WCProductStore
+import org.wordpress.android.fluxc.store.WCProductStore.BatchUpdateVariationsPayload
+import org.wordpress.android.fluxc.store.WooCommerceStore
+import org.wordpress.android.fluxc.utils.DateUtils
+import java.lang.NumberFormatException
+import java.util.Calendar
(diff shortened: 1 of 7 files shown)


example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooBatchUpdateVariationsFragment.kt
```diff
@@ -0,0 +1,252 @@
...
+            with(date_on_sale_from_gmt.getText()) {
+                if (isNotEmpty()) variationsUpdatePayloadBuilder.startOfSaleGmt(this)
+            }
+            with(date_on_sale_to_gmt.getText()) {
+                if (isNotEmpty()) variationsUpdatePayloadBuilder.endOfSaleGmt(this)
+            }
+            with(stock_quantity.getText()) {
+                if (isNotEmpty()) variationsUpdatePayloadBuilder.stockQuantity(this.toInt())
+            }
+            with(stock_quantity.getText()) {
```
0nko, line 151: This is a duplicate of the previous change. I think you meant to update the stock status here.

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooBatchUpdateVariationsFragment.kt
```diff
@@ -0,0 +1,252 @@
...
+                if (isNotEmpty()) variationsUpdatePayloadBuilder.stockQuantity(this.toInt())
+            }
+            with(weight.getText()) {
+                if (isNotEmpty()) variationsUpdatePayloadBuilder.weight(this)
+            }
+            with(dimensions.getText()) {
+                if (isNotEmpty()) variationsUpdatePayloadBuilder.dimensions(this)
```
0nko: This won't work unless you manually construct a JSON string with the right properties. See the [Dimensions](https://woocommerce.github.io/woocommerce-rest-api-docs/#product-properties) property definition for more details.

0nko (changes requested):
I've played around with the app and the batch update working well (apart for some small things) :+1:. 

I left some comments left in the code and I noticed something about the sale dates -- the layout contains a date selector for non-GMT date but only an edit text for GMT date. I don't think it's necessary to have both here. Recently there was a discussion about what date to use, we concluded that the non-GMT (store date) probably makes most sense for the store owner. The GMT is then automatically offset and updated.

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooBatchUpdateVariationsFragment.kt
samiuelson: It's just a duplicated code. Stock status is updated using a ListSelectorDialog ([here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2374/files#diff-9bfe9cd7927b14a1c1ad164783ea1bcb3212ca35c60c015d7e2b541d2901eb27R217)). Fixed with [4055ce9](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2374/commits/4055ce9b2653540c59e0c72c076e57dc80ac2dd0).

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooBatchUpdateVariationsFragment.kt
samiuelson: Good catch. I think the right approach will be to update `BatchUpdateVariationsPayload.Builder::dimensions` method to take `length`, `width`, and `height` args instead of a single String and build `dimensions` object internally. What do you think @0nko?

samiuelson:
> I've played around with the app and the batch update working well (apart for some small things) 👍.
> 
> I left some comments left in the code and I noticed something about the sale dates -- the layout contains a date selector for non-GMT date but only an edit text for GMT date. I don't think it's necessary to have both here. Recently there was a discussion about what date to use, we concluded that the non-GMT (store date) probably makes most sense for the store owner. The GMT is then automatically offset and updated.

@0nko  thanks for review!
Regarding GMT date, do you think we should remove it from the `BatchUpdateVariationsPayload.Builder` too or just from the example app?

0nko:
> Regarding GMT date, do you think we should remove it from the BatchUpdateVariationsPayload.Builder too or just from the example app?

I think you can remove it there too.

example/src/main/java/org/wordpress/android/fluxc/example/ui/products/WooBatchUpdateVariationsFragment.kt
0nko: Yeah, that sounds reasonable.

samiuelson:
Alright, I've added the fixes:
* Dimensions property is revised to the correct model. I've tested this and the dimensions update works fine now 👍🏼 
* GMT sale dates handling is removed both from the payload and example app.

0nko (approved):
Nice job!

samiuelson:
> Nice job!

🙌 Thanks for review!

PR #2378, opened 2022-05-04
Issue/blogging prompts dao

khaykov:
This PR adds a blogging prompts DAO, provider, related utils and tests.

There is not much to test yet - just make sure the CI passes.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/bloggingprompts/BloggingPromptModel.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/bloggingprompts/BloggingPromptModel.kt
@@ -9,6 +9,7 @@ data class BloggingPromptModel(
     val content: String,
     val date: Date,
     val isAnswered: Boolean,
+    val attribution: String,
     val respondentsCount: Int,
     val respondentsAvatars: List<String>
 )
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/bloggingprompts/BloggingPromptModel.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/bloggingprompts/BloggingPromptModel.kt
@@ -11,5 +11,5 @@ data class BloggingPromptModel(
     val isAnswered: Boolean,
     val attribution: String,
     val respondentsCount: Int,
-    val respondentsAvatars: List<String>
+    val respondentsAvatarUrls: List<String>
 )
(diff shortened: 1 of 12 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/bloggingprompts/BloggingPromptModel.kt
```diff
@@ -0,0 +1,15 @@
...
+data class BloggingPromptModel(
+    val id: Int,
+    val text: String,
+    val title: String,
+    val content: String,
+    val date: Date,
+    val isAnswered: Boolean,
+    val attribution: String,
+    val respondentsCount: Int,
+    val respondentsAvatars: List<String>
```
RenanLukas: Do you think it would be clearer to have `Url` in this property name? e.g. `respondentsAvatarsUrls`

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/bloggingprompts/BloggingPromptsUtils.kt
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Locale
+
+object BloggingPromptsUtils {
+    private val DATE_FORMAT: ThreadLocal<DateFormat> = object : ThreadLocal<DateFormat>() {
+        override fun initialValue(): DateFormat {
+            return SimpleDateFormat("yyyy-MM-dd", Locale.US)
+        }
+    }
+
+    fun dateToString(date: Date): String {
```
RenanLukas, line 16: Do you think it would make sense to have both functions as extension functions instead? I'm thinking about maybe making them reusable in other flows.
If you think it's not worth it, please disregard.

e.g. (not tested)
```
fun Date.toString(formatter: DateFormat = DATE_FORMAT.get()): String = formatter.format(this)
```

and

```
fun String.toDate(formatter: DateFormat = DATE_FORMAT.get()): Date = try {
        formatter.parse(this) ?: Date()
    } catch (exception: ParseException) {
        Date()
    }
}
```

RenanLukas (commented):
LGTM ✅ :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/bloggingprompts/BloggingPromptsUtils.kt
khaykov: I like extensions, but in this particular case, I do not want to default to `DATE_FORMAT` which is very specific use, so I'll leave them as is for now :)

RenanLukas (approved):
LGTM :shipit:

PR #2381, opened 2022-05-06
Issue/blogging prompts rest client

khaykov:
This PR adds a blogging prompts Rest Client and tests.

There is not much to test yet - just make sure the CI passes.


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/bloggingprompts/BloggingPromptsRestClient.kt
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    data class BloggingPromptsListResponse(
+        @SerializedName("prompts") val prompts: List<BloggingPromptResponse>?
+    ) {
+        fun toBloggingPrompts() = prompts?.map { it.toBloggingPromptModel() } ?: emptyList()
+    }
+
+    data class BloggingPromptResponse(
```
RenanLukas, line 62: If any property that is not nullable returns `null` from the server, will it throw an exception? If it is the case, would it make sense to make all the properties nullable?

RenanLukas (approved):
LGTM :shipit: 
Left an optional suggestion

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/bloggingprompts/BloggingPromptsRestClient.kt
khaykov: I would rather we crash early and fix the endpoint instead of not noticing potentially half-broken functionality :)

PR #2383, opened 2022-05-07
Added blogging prompt store and test.

khaykov:
This PR adds a blogging prompts Store and test.

I modeled it after the CardStore.

`fetchPrompts` fetches and store the prompts in the DB, but does not return then.

Currently, I think we will need following methods:
`getPromptForDate`
`getPromptById`
`getPrompts`

Let me know you think something else might come in handy (eg. maybe adding a default Date value to `getPromptForDate`?)

You can test this branch in [this](https://github.com/wordpress-mobile/WordPress-Android/pull/16504) Client side PR.


fluxc/src/main/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStore.kt
```diff
@@ -0,0 +1,87 @@
...
+        error: BloggingPromptsError
+    ): BloggingPromptsResult<List<CardModel>> = when (error.type) {
+        AUTHORIZATION_REQUIRED -> {
+            promptsDao.clear()
+            BloggingPromptsResult()
+        }
+        else -> BloggingPromptsResult(error)
+    }
+
+    fun getPromptForDate(site: SiteModel, date: Date) =
```
RenanLukas: Would it make sense to add explicit return types for the public functions, so that the client knows exactly what to expect?

RenanLukas (approved):
LGTM :shipit: 
Left an optional comment

fluxc/src/main/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStore.kt
khaykov: Yep, agree! Added the return types 👍

khaykov:
Thanks for suggestion, @RenanLukas ! Ready for another round :)

RenanLukas (approved):
LGTM :shipit:

PR #2387, opened 2022-05-11
Quick Start for Existing Users V2 - Add existing site quick start tasks

ashiagr:
Parent https://github.com/wordpress-mobile/WordPress-Android/issues/16350

This PR only adds tasks for the existing site quick start type.

PS: Currently only common tasks (`CHECK_STATS`, `VIEW_SITE`, `FOLLOW_SITE`) are enabled. The remaining tasks will be enabled when they're supported in the client app.

## How to Test?

See https://github.com/wordpress-mobile/WordPress-Android/pull/16530


AjeshRPai (approved):
I have looked at the code changes and tested the Parent PR as per the instructions. Everything looks good to me. 👍 🙌

PR #2388, opened 2022-05-11
[Coupons] Add suspended functions and observed DB for products, variations and categories

0nko:
This PR adds the product, variation and category search & fetch suspended functions required for the product/category selection dialog in coupon editing. It also adds the observation functions that return `Flow`s.

**To test:**
Tests the **observation**:
1. Go to Woo -> Products
2. Select Fetch product categories
3. Then select Observe product categories
4. Notice a list of product categories is shown (uses the new observe function)

Tests the new product fetching using parameters:
1. Go to Woo -> Products
2. Select fetch specific products
3. Enter valid product IDs, separated by comma
4. Notice the products with valid IDs get fetched

Also, run the unit tests and make sure they pass.

**Note:** Please merge after #2407.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/ProductsDao.kt
```diff
@@ -21,6 +22,11 @@ abstract class ProductsDao {
     @Query("SELECT * FROM Products WHERE siteId = :siteId ORDER BY id")
```
hichamboushaba: This function should support the same ordering as we receive from the API.

hichamboushaba (approved):
Looks good, just leaving two remarks to consider for this or a separate PR:
1. The product selector for coupons would need to support filtering (as you can see in the image below), so maybe we should add support for it, we can reuse the `ProductFilterOption` for this.
Screen Shot 2022-05-12 at 08 58 25

2. You may already have a plan for this, just pointing it to not miss it 🙂, but shouldn't we aim for having the same functionality as legacy WellSql table to make the migration easier and sooner? I mean since we added fetching, we should add also support for sorting similar to what we had before (`ProductSorting` enum)?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/ProductsDao.kt
0nko: Thanks for catching this 👍

0nko:
Thanks for the review, @hichamboushaba!

> The product selector for coupons would need to support filtering (as you can see in the image below), so maybe we should add support for it, we can reuse the ProductFilterOption for this.

I was planning to do the selector screen in 2 steps since it's a bit more complex -- 1. product selection + search and 2. filtering in another PR. But you're right, it makes sense to have the FluxC ready. I'll add it.

> but shouldn't we aim for having the same functionality as legacy WellSql table to make the migration easier and sooner

I didn't think of this really, I was just adding the functionality needed for coupons. I agree with you, though. I'll include the sorting options, as well. Thank you! 🙇

hichamboushaba:
> I didn't think of this really, I was just adding the functionality needed for coupons. I agree with you, though. I'll include the sorting options, as well. Thank you! 🙇

Thanks @0nko, It doesn't have to be on the same PR though, if you want to prioritize the coupons task, feel free to leave this for another PR.

0nko:
@hichamboushaba Thank you for the review. I've made updates based on the removal of Products and ProductCategories Room tables and I added functions for observing changes. I included more parameters for filtering products, variations and categories. 

Ready for another round.

example/src/test/java/org/wordpress/android/fluxc/wc/product/ProductTestUtils.kt
```diff
@@ -105,6 +105,17 @@ object ProductTestUtils {
         }
     }
 
+    fun generateProductList(siteId: Int = 6): List<WCProductModel> {
+        with(ArrayList<WCProductModel>()) {
```
hichamboushaba: np, I think we can use `listOf` which is shorter, or `List(size) {}` like below.

hichamboushaba (approved):
Nice work @0nko, the changes looks good, thanks so much for making them.

I don't see any breaking changes, so this should be ready for merge, but I'll leave it for you to double check and merge it.

PR #2397, opened 2022-05-13
Remove Coupons Cross-ref tables

0nko:
This PR addresses an issue when a product variation is included/excluded in a coupon.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEntity.kt
```diff
@@ -34,7 +34,11 @@ data class CouponEntity(
     val isShippingFree: Boolean? = null,
     val areSaleItemsExcluded: Boolean? = null,
     val minimumAmount: BigDecimal? = null,
-    val maximumAmount: BigDecimal? = null
+    val maximumAmount: BigDecimal? = null,
+    val includedProductIds: String? = null,
+    val excludedProductIds: String? = null,
+    val includedCategoryIds: String? = null,
+    val excludedCategoryIds: String? = null
```
hichamboushaba: This can be done in a separate PR, but I think using a TypeConverter would be useful here to avoid exposing the data as String.

hichamboushaba: I added this in [10712ec](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2397/commits/10712ecf0f33e9ce49e575a01b5f4d04a95e29bc)

hichamboushaba (commented):
This looks good, but since I made changes too, please review them before merging.

And something else we need to discuss: the usefulness of the Room Products and Categories tables after those changes? I'm not saying we should delete them, but maybe we should stop saving data to them after fetching until we finish the Room migration, as it will have a performance and disk size impact.
If we do so, we can continue using `ProductListRepository` in the WCAndroid app for the product selector, which should save us some time, WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/converters/LongListConverter.kt
```diff
@@ -8,12 +8,15 @@ class LongListConverter {
     }
 
     @TypeConverter
-    fun listToString(value: List<Long>): String = value.joinToString(separator = SEPARATOR)
+    fun listToString(value: List<Long>?): String? =
```
hichamboushaba, line 11: TypeConverters should account for nullability always.

example/src/test/java/org/wordpress/android/fluxc/store/CouponStoreTest.kt
```diff
@@ -100,7 +100,7 @@ class CouponStoreTest {
         minimumAmount = couponDto.minimumAmount?.toBigDecimal(),
         maximumAmount = couponDto.maximumAmount?.toBigDecimal(),
         includedProductIds = couponDto.productIds?.joinToString(","),
-        excludedProductIds = couponDto.productIds?.joinToString(","),
+        excludedProductIds = couponDto.excludedProductIds?.joinToString(","),
```
0nko: 🤦

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/converters/LongListConverter.kt
0nko: Nice! I should have thought of that!

0nko:
>  the usefulness of the Room Products and Categories tables after those changes? I'm not saying we should delete them, but maybe we should stop saving data to them after fetching until we finish the Room migration, as it will have a performance and disk size impact.

I'm using them in the product/category selector dialog. I'd use the old tables but I want to observe the DB changes.

hichamboushaba (approved):
Let's ship this 🚢

hichamboushaba:
> I'm using them in the product/category selector dialog. I'd use the old tables but I want to observe the DB changes.

Just sharing my opinion @0nko, but it's not a blocker by any mean, so feel free to ignore it.
I think that having a duplicate table/store for those entities brings more issues than the advantages:
- Disk size and performance as I pointed above.
- Makes things harder for anyone that isn't familiar with the Coupons project, as it will be less clear where to make changes if needed.

I would love to see a Room migration for those tables, but I would love to see it done properly, by migrating the SQL layer and leaving the usage of the current Store, then we can update the APIs when needed (similar to what was done with Orders, and this is not a minor project, it would be a good Hack Week project maybe).

For observability, if it's something we really need, we can add it without Room, we did [something similar](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2294) with the NotificationModel table when we want it to make the notification badge reactive.

PR #2399, opened 2022-05-13
Convert register device into suspendable. Allow to check for FCM registration success.

### LINKED ISSUE #2398: Convert register device into suspendable

wzieba:
Closes #2398 

Corresponding Woo PR: https://github.com/woocommerce/woocommerce-android/pull/6535

***Description***
This PR adds a coroutine-based API for registering devices on WPCOM. It also adds the feature of checking if the device is registered.

❗️Please wait with merging this PR until https://github.com/woocommerce/woocommerce-android/pull/6535 is approved!


fluxc/src/main/java/org/wordpress/android/fluxc/store/GetDeviceRegistrationStatus.kt
```diff
@@ -0,0 +1,21 @@
+package org.wordpress.android.fluxc.store
+
+import javax.inject.Inject
+import org.wordpress.android.fluxc.utils.PreferenceUtils
+
+class GetDeviceRegistrationStatus @Inject constructor(
+    private val prefsWrapper: PreferenceUtils.PreferenceUtilsWrapper
+) {
+    operator fun invoke(): Status {
+        val deviceId = prefsWrapper.getFluxCPreferences().getString(NotificationStore.WPCOM_PUSH_DEVICE_SERVER_ID, null)
```
wzieba, line 10: When a device gets registered on WPCOM, we receive a "server id" and save it in SharedPreferences.

I think that checking for the presence of this "server id" is the most robust way to check if device registration was indeed successful.

wzieba, line 10: When a device gets registered on WPCOM, we receive a "server id" and save it in SharedPreferences.

I think that checking for the presence of this "server id" is the most robust way to check if device registration was indeed successful.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
```diff
@@ -55,7 +55,49 @@ class NotificationRestClient @Inject constructor(
...
+                        "device_token" to fcmToken,
+                        "device_family" to "android",
+                        "app_secret_key" to appKey.value,
+                        "device_name" to deviceName,
+                        "device_model" to "${Build.MANUFACTURER} ${Build.MODEL}",
+                        "app_version" to PackageUtils.getVersionName(appContext),
+                        "version_code" to PackageUtils.getVersionCode(appContext).toString(),
+                        "os_version" to Build.VERSION.RELEASE,
+                        "device_uuid" to uuid
+                ).toMap(),
```
atorresveiga: Not sure if I'm missing something, but if `mapOf` is returning a map, why do we need to call `toMap()`

fluxc/src/main/java/org/wordpress/android/fluxc/store/NotificationStore.kt
```diff
@@ -273,6 +273,36 @@ class NotificationStore @Inject constructor(
     fun getNotificationByLocalId(noteId: Int) =
             notificationSqlUtils.getNotificationByIdSet(NoteIdSet(noteId, 0, 0))
 
+    suspend fun registerDevice(token: String, appKey: NotificationAppKey): RegisterDeviceResponsePayload {
```
atorresveiga, line 276: I noticed that we have `WooPayload` and `WooResult` classes. I'm wondering if this refactor is a good opportunity to update the function response? (use WooPayload and WooResult). Is it in our plans to update all results to use these classes? I'm asking to learn because I was in doubt in the past about how and when to use these classes 😁

example/src/test/java/org/wordpress/android/fluxc/store/GetDeviceRegistrationStatusTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.nhaarman.mockitokotlin2.whenever
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.junit.MockitoJUnitRunner
+import org.wordpress.android.fluxc.store.NotificationStore.Companion.WPCOM_PUSH_DEVICE_SERVER_ID
+import org.wordpress.android.fluxc.utils.PreferenceUtils
+
+@RunWith(MockitoJUnitRunner::class)
+class GetDeviceRegistrationStatusTest {
```
atorresveiga, line 15: Excellent work adding tests! 🙌

fluxc/src/main/java/org/wordpress/android/fluxc/store/NotificationStore.kt
```diff
@@ -273,6 +273,36 @@ class NotificationStore @Inject constructor(
...
+                    }
+                } else {
+                    preferences.edit().putString(WPCOM_PUSH_DEVICE_SERVER_ID, deviceId).apply()
+                    AppLog.i(T.NOTIFS, "Server response OK. Device ID: $deviceId")
+                }
+            }
+        }
+    }
+
+    @Deprecated("EventBus is deprecated.", ReplaceWith("registerDevice(token, appKey)"))
```
atorresveiga, line 305: Nice!

fluxc/src/main/java/org/wordpress/android/fluxc/store/NotificationStore.kt
wzieba: That'd be a good idea to migrate but the device registration code is used by both WordPress and WooCommerce.

This particular method is not yet used, but I think that in future it will be, so I'd leave the `RegisterDeviceResponsePayload` return type, WDYT?

fluxc/src/main/java/org/wordpress/android/fluxc/action/NotificationAction.java
```diff
@@ -19,6 +19,7 @@
 public enum NotificationAction implements IAction {
     // Remote actions
     @Action(payloadType = RegisterDevicePayload.class)
+    @Deprecated
```
atorresveiga, line 22: Nice!

atorresveiga (approved):
Great work Wojtek! I had some questions out of curiosity but the changes looks good to me and the tests are all green so I will pre-approve this 👍

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt
wzieba: You're right - it's not needed, thanks! Fixed in 85664cffd1333e67da3646b0e9ffd9ca33d125fc

wzieba:
Thanks @atorresveiga ! I've addressed the comments. I'll wait with the merge of this PR until Woo PR will be ready to merge as it might require FluxC changes.

PR #2407, opened 2022-05-17
[Coupons] Temporarily remove Products and ProductCategories Room tables

0nko:
This PR removes the Products and ProductCategories Room tables, which are not needed and will be readded later, when we eventually migrate to Room for all the functionality.

**To test:**
1. Make sure the coupons can be fetched and updated 
2. Make sure the unit tests pass


hichamboushaba (approved):
Looks good, thanks for making these changes @0nko

PR #2412, opened 2022-05-19
Use mapOf instead of toMap to fix failing test

malinajirka:
`testUpdateProductPassword` has been failing for the last several days with the following exception:

```
2022-05-17 19:30:28.936 26844-26941/? E/AndroidRuntime: FATAL EXCEPTION: pool-2-thread-1
    Process: org.wordpress.android.fluxc.example, PID: 26844
        at org.greenrobot.eventbus.EventBus.handleSubscriberException(EventBus.java:537)
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:519)
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:511)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:920)
     Caused by: com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_ARRAY but was BEGIN_OBJECT at line 1 column 3 path $[0]
        at com.google.gson.Gson.fromJson(Gson.java:939)
        at com.google.gson.Gson.fromJson(Gson.java:892)
        at com.google.gson.Gson.fromJson(Gson.java:841)
        at org.wordpress.android.fluxc.network.utils.JsonExtKt.toMap(JsonExt.kt:20)
        at org.wordpress.android.fluxc.network.rest.wpcom.wc.product.ProductRestClient.updateProductPassword(ProductRestClient.kt:635)
        at org.wordpress.android.fluxc.store.WCProductStore.updateProductPassword(WCProductStore.kt:1144)
        at org.wordpress.android.fluxc.store.WCProductStore.onAction(WCProductStore.kt:860)
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:517)
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:511) 
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167) 
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641) 
        at java.lang.Thread.run(Thread.java:920) 
     Caused by: java.lang.IllegalStateException: Expected BEGIN_ARRAY but was BEGIN_OBJECT at line 1 column 3 path $[0]
        at com.google.gson.stream.JsonReader.beginArray(JsonReader.java:350)
        at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:172)
        at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:145)
        at com.google.gson.Gson.fromJson(Gson.java:927)
        at com.google.gson.Gson.fromJson(Gson.java:892) 
        at com.google.gson.Gson.fromJson(Gson.java:841) 
        at org.wordpress.android.fluxc.network.utils.JsonExtKt.toMap(JsonExt.kt:20) 
        at org.wordpress.android.fluxc.network.rest.wpcom.wc.product.ProductRestClient.updateProductPassword(ProductRestClient.kt:635) 
        at org.wordpress.android.fluxc.store.WCProductStore.updateProductPassword(WCProductStore.kt:1144) 
        at org.wordpress.android.fluxc.store.WCProductStore.onAction(WCProductStore.kt:860) 
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:517) 
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:511) 
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167) 
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641) 
        at java.lang.Thread.run(Thread.java:920) 
```

I noticed it's not failing due to a server response as usual with this type of exception. It's failing when we attempt to create a body of the request. 

This PR fixing the failing tests. `However, it's still a draft since I plan to continue investigating what is the root cause of the issue and what change in our codebase caused the test to start failing.` @hichamboushaba Correctly pointed out that the issue was caused by a wrong import to our custom version of `toMap` function which doesn't work in certain cases. The PR is now ready for review - I've left a couple in code comments explaining my train of thoughts.

~@wzieba This issue slightly reminds me the investigation you did regarding lambdas in the new version of Kotlin. Pinging you in case you'd know from the top of your head what is causing this issue. If nothing comes to your mind I'll investigate it tomorrow. Thanks!~

To test:
1. Run `testUpdateProductPassword` and verify it succeeds

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -630,9 +630,9 @@ class ProductRestClient @Inject constructor(
      */
     fun updateProductPassword(site: SiteModel, remoteProductId: Long, password: String) {
         val url = WPCOMREST.sites.site(site.siteId).posts.post(remoteProductId).urlV1_2
-        val body = listOfNotNull(
+        val body = mapOf(
                 "password" to password
-        ).toMap()
+        )
 
         val request = WPComGsonRequest.buildPostRequest(url,
                 body,
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -34,7 +34,6 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooError
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooErrorType
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.toWooError
-import org.wordpress.android.fluxc.network.utils.toMap
 import org.wordpress.android.fluxc.store.WCProductStore
 import org.wordpress.android.fluxc.store.WCProductStore.Companion.DEFAULT_CATEGORY_SORTING
 import org.wordpress.android.fluxc.store.WCProductStore.Companion.DEFAULT_PRODUCT_CATEGORY_PAGE_SIZE
@@ -822,7 +821,6 @@ class ProductRestClient @Inject constructor(
             val variationsUpdates: List<Map<String, Any>> = variationsIds.map { variationId ->
                 modifiedProperties.toMutableMap()
                     .also { properties -> properties["id"] = variationId }
-                    .toMap()
             }
             jetpackTunnelGsonRequestBuilder.syncPostRequest(
                 this@ProductRestClient,


wzieba:
Thanks for the ping @malinajirka ! From the top of my head, I don't have a clear idea, those kinds of stack traces remind me only of JSON parsing errors.

Maybe this will be helpful regarding the draft: as I was working recently on notifications, I've also seen `... = listOfNotNull(...).toMap()` pattern there, which is working fine ([here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt#L66-L77)).

hichamboushaba:
Sharing something too, the cause of the failing test (and a crash in the app too for the same scenario) is this commit: https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/34df84be4eb53d7741fc7d86c56ce7770e195806, we introduced an import to the our `toMap` extension, which replaced the default `kotlin.collections.toMap` that was used before.

So I believe the fix using `mapOf` is the correct one for this case, as there is no need for creating a list before converting to `toMap`.
Then we should revisit the usage of `toMap` extension, we can probably start by renaming it to something else to avoid those confusions in the future, then we need to confirm its usefulness and its impact on performance, it uses Gson for the conversion, which is not super performant.

hichamboushaba:
> I've also seen ... = listOfNotNull(...).toMap() pattern t[here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/notifications/NotificationRestClient.kt#L66-L77), which is working fine (here).

@wzieba this is working fine because it it's using the correct `toMap` function 🙂.

malinajirka:
Thank you both!

> Sharing something too, the cause of the failing test (and a crash in the app too) is this commit: https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/34df84be4eb53d7741fc7d86c56ce7770e195806, we introduced an import to the our toMap extension, which replaced the default kotlin.collections.toMap that was used before.

Wow nice find @hichamboushaba ! I checked a bunch of recent code changes, but I didn't think of checking the imports 😞.

> we introduced an import to the our toMap extension

I'm wondering why does the code in [the commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/34df84be4eb53d7741fc7d86c56ce7770e195806) need to use our own version of `toMap`? I don't see anything special there 😕. Tbh I don't see any reason why we even need to call `toMap` at all - it's already a map, isn't it? Wdyt?

> So I believe the fix using mapOf is the correct one for this case, as there is no need for creating a list before converting to toMap.

I agree creating the list is unnecessary. However, depending on the answer to the above, we might want to consinder removing the import to our implementation of `toMap` too, wdyt?

hichamboushaba:
> I'm wondering why does the code in [the commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/34df84be4eb53d7741fc7d86c56ce7770e195806) need to use our own version of toMap? I don't see anything special there 😕. Tbh I don't see any reason why we even need to call toMap at all - it's already a map, isn't it? Wdyt?

I didn't spend much time trying to understand the code 😄, but I agree with your reasoning.
@samiuelson WDYT about the above? was it just an accidental import?

samiuelson:
> > I'm wondering why does the code in [the commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/34df84be4eb53d7741fc7d86c56ce7770e195806) need to use our own version of toMap? I don't see anything special there 😕. Tbh I don't see any reason why we even need to call toMap at all - it's already a map, isn't it? Wdyt?
> 
> I didn't spend much time trying to understand the code 😄, but I agree with your reasoning. @samiuelson WDYT about the above? was it just an accidental import?

@hichamboushaba Well, indeed, this was supposed to be the standard `kotlin.collections::toMap` function. Sorry, my bad! The import can just be deleted [there](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt#L37) and the correct `::toMap` will be engaged automatically.

hichamboushaba:
Thanks @samiuelson for the confirmation, and good luck with the trial 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -34,7 +34,6 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooError
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooErrorType
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.toWooError
-import org.wordpress.android.fluxc.network.utils.toMap
```
malinajirka, line 37: This was the root cause of the issue - the code was using our custom `toMap` function which apparently doesn't work in certain cases. H/T to @hichamboushaba for noticing the issue 🙇

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -630,9 +629,9 @@ class ProductRestClient @Inject constructor(
      */
     fun updateProductPassword(site: SiteModel, remoteProductId: Long, password: String) {
         val url = WPCOMREST.sites.site(site.siteId).posts.post(remoteProductId).urlV1_2
-        val body = listOfNotNull(
+        val body = mapOf(
```
malinajirka, line 632: This is an unnecessary improvement: The app was creating a list and converting it to a map right after that. The updated code creates a map without any need for conversion.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -822,7 +821,6 @@ class ProductRestClient @Inject constructor(
             val variationsUpdates: List<Map<String, Any>> = variationsIds.map { variationId ->
                 modifiedProperties.toMutableMap()
                     .also { properties -> properties["id"] = variationId }
-                    .toMap()
```
malinajirka, line 825: Also unnecessary improvement: Removes unnecessary `toMap` invocation since the result of the expression gets saved into a `Map` not `MutableMap` anyway so the `toMap` call was just adding an overhead.

Samuel tested this change in the example app and it worked as expected 🥳. Thanks!

samiuelson: Good catch! Thanks 🙌

wzieba:
> (...) Then we should revisit the usage of toMap extension, we can probably start by renaming it to something else to avoid those confusions in the future, then we need to confirm its usefulness and its impact on performance, it uses Gson for the conversion, which is not super performant.

Exactly - I've created an issue for that #2416 

@malinajirka , does `testUpdateProductPassword` pass on your local env? For me it doesn't:
```
onProductPasswordChanged has unexpected error: GENERIC_ERROR, Unknown post
```

I have credentials applied via `applyConfiguration` - should I do something more?

malinajirka:
@wzieba It works for me locally

ProductId: `TEST_WC_PRODUCT_ID = 779`
SiteName: `woo-jetpack`

Seems like the product is not found on your site - have you perhaps manually modified the properties and added a different site for testing?

I tried re-running applyConfiguration and it still works for me.

hichamboushaba (approved):
The test works fine for me, and I confirmed it also by running the circleci task.

@malinajirka it seems the CI task in buildkite is failing with the error: `'org.wordpress.fluxc.fluxc-annotations:2412-21f9cb0b7bf3acba4e3092888fec98290f1289ae' is already published to S3`, I don't know what caused the same build to be reuploaded twice.
We can probably ignore the error, or to be sure, you can force the regeneration of last commit (for example by performing a soft reset to `HEAD~1` then re-commiting and force pushing the branch)

wzieba (approved):
The test worked fine today without any changes 🤔 🤷 .

I'll amend the last commit to regenerate hash.

PR #2414, opened 2022-05-20
Issue/fix product parent id exceeds int range

malinajirka:
This PR converts WCProductModel.parentId from integer to long. It hopefully fixes the following exception `com.google.gson.JsonSyntaxException: java.lang.NumberFormatException: Expected an int but was 5060147738939 at line 1 column 1399 path $.data.parent_id` which we've been seeing in Sentry for some time. It also makes sense since `WCProductModel.remoteId` is long so `parentId` should be long too.


example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCProductsTest.kt
```diff
@@ -138,6 +138,14 @@ class MockedStack_WCProductsTest : MockedStack_Base() {
         assertEquals(payload.error.type, ProductErrorType.INVALID_PRODUCT_ID)
     }
 
+    @Test
```
nbradbury, line 144: @malinajirka WDYT about adding a small comment here stating why this test was added? For historical reasons it might be good to "document" that `parentId` was initially an integer.

malinajirka: 👍  Fixed in 38d1b64a03448b07a5332c7a66b78c3056f7f679

nbradbury (approved):
Looks good! :shipit:

PR #2419, opened 2022-05-23
Add the variation ID list deserialization to WCProductModel

0nko:
This PR adds the remote variation ID list deserialization. The list is being fetched but only the count has been available to the client.

**To test:**
1. Make sure the unit tests pass

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
@@ -409,6 +409,8 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
         return productIds
     }
 
+    fun getNumVariations() = getVariationIdList().size
+
     fun getVariationIdList() = parseJson(variations)
 
     fun getGroupedProductIdList() = parseJson(groupedProductIds)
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -409,25 +409,7 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
         return productIds
     }
 
-    fun getNumVariations(): Int {
```
hichamboushaba, line 412: WDYT about keeping `getNumVariations` with an implementation similar to the below one?
```kotlin
fun getNumVariations() = getVariationIdList().size
```

This will keep the PR as non-breaking one, and we can merge it right away, otherwise we need an accompanying WCAndroid PR before merging this one.

hichamboushaba (approved):
Looks good, I left one question, and I'll leave the decision on to make the change or not up to you.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
0nko: That's true. I'll add it back 👍

PR #2421, opened 2022-05-24
Add support for crash logger

malinajirka:
This PR 
- Introduces Crash Logger into FluxC
- Records "Request path" when an error occurs while parsing a server response

This is a pretty native but imo a good enough implementation of Crash Logger. The client apps need to initialize the logger using the `FluxCCrashLoggerProvider.initLogger()` method.

This PR also records `request path` when an error occurs while parsing a server response. If the app is using JetpackTunnel, the actual path of the endpoint is stored in `path` parameter. Otherwise, the actual uri path is used. The client apps can for example add a breadcrumb into Sentry so the logs in Sentry contain information about the endpoint which resulted in the parsing error.

To test:
1. Add `AppLog.e(AppLog.T.API, "Testing: path is " + getPath());` above [this line](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/f132e83b5d8d9fb6872ecfec65179b3141ed6f89/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java#L96).
2. Start the example app.
3. Try running different queries - ideally some used by WPAndroid (not using jetpack tunnel) and some used by WCAndroid (using jetpack tunnel).
4. Check the logs and notice the correct path is recorded.

Screenshot 2022-05-24 at 10 07 27

The actual CrashLogging will be tested in a PR in WCAndroid.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
@@ -1,0 +1,19 @@
+package org.wordpress.android.fluxc.logging
+
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.AppLog.T
+
+object FluxCCrashLoggerProvider {
+    var crashLogger: FluxCCrashLogger? = null
+        get() {
+            if (field == null) {
+                AppLog.w(T.MAIN, "FluxCCrashLogger is not initialized.")
+            }
+            return field
+        }
+        private set
+
+    fun initLogger(logger: FluxCCrashLogger) {
+        crashLogger = logger
+    }
+}
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
@@ -4,6 +4,7 @@ import org.wordpress.android.util.AppLog
 import org.wordpress.android.util.AppLog.T
 
 object FluxCCrashLoggerProvider {
+    @Volatile
     var crashLogger: FluxCCrashLogger? = null
         get() {
             if (field == null) {
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
@@ -2,6 +2,7 @@ package org.wordpress.android.fluxc.logging
 
 import org.wordpress.android.util.AppLog
 import org.wordpress.android.util.AppLog.T
+import org.wordpress.android.util.AppLog.T.MAIN
 
 object FluxCCrashLoggerProvider {
     @Volatile
@@ -15,7 +16,20 @@ object FluxCCrashLoggerProvider {
         private set
 
     fun initLogger(logger: FluxCCrashLogger) {
-        if (crashLogger != null) throw IllegalStateException("FLuxCCrashLogger is already initialized")
-        crashLogger = logger
+        if (crashLogger == null) {
+            synchronized(FluxCCrashLoggerProvider.javaClass) {
+                if (crashLogger == null) {
+                    crashLogger = logger
+                } else {
+                    logAlreadyInitialized()
+                }
+            }
+        } else {
+            logAlreadyInitialized()
+        }
+    }
+
+    private fun logAlreadyInitialized() {
+        AppLog.w(MAIN, "FluxCCrashLoggerProvider already initialized.")
     }
 }
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
@@ -15,6 +15,7 @@ object FluxCCrashLoggerProvider {
         private set
 
     fun initLogger(logger: FluxCCrashLogger) {
+        if (crashLogger != null) throw IllegalStateException("FLuxCCrashLogger is already initialized")
         crashLogger = logger
     }
 }
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
@@ -11,6 +11,8 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.JsonSyntaxException;
 
 import org.jetbrains.annotations.Nullable;
+import org.wordpress.android.fluxc.logging.FluxCCrashLogger;
+import org.wordpress.android.fluxc.logging.FluxCCrashLoggerProvider;
 import org.wordpress.android.fluxc.network.BaseRequest;
 
 import java.io.UnsupportedEncodingException;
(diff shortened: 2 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
```diff
@@ -0,0 +1,19 @@
+package org.wordpress.android.fluxc.logging
+
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.AppLog.T
+
+object FluxCCrashLoggerProvider {
+    var crashLogger: FluxCCrashLogger? = null
```
kidinov, line 9: Shell we make access to this prop thread-safe? 

Also, you use "private set" but `initLogger` is the same as public, any reason behind that?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
```diff
@@ -107,4 +111,26 @@ private GsonBuilder setupGsonBuilder() {
...
+        FluxCCrashLogger logger = FluxCCrashLoggerProvider.INSTANCE.getCrashLogger();
+        if (logger != null) {
+            String path = getPath();
+            if (path != null) {
+                logger.recordEvent(path, "Request path");
+            }
+        }
+    }
+
+    private String getPath() {
```
kidinov, line 122: Not sure if this can work here. What if we'll add the abstract `getPath` method here and implement it in different implementations? This way you won't need to make a guess:

```
        // If JetpackTunnel is being used the actual path is stored in `path` parameter
        if (params != null && params.get("path") != null) {
```

Also, you will be able to write a test for every case

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
```diff
@@ -93,8 +95,10 @@ protected Response<T> parseNetworkResponse(NetworkResponse response) {
             }
             return Response.success(res, createCacheEntry(response));
         } catch (UnsupportedEncodingException e) {
```
kidinov:
```
        } catch (UnsupportedEncodingException|JsonSyntaxException e) {
            logRequestPath();
            return Response.error(new ParseError(e));
        }
```
Shell work too?

kidinov (commented):
Thanks for this improvement! 🙏

 I left a few things you might consider to think about

fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
malinajirka: > Shell we make access to this prop thread-safe?

That's what I meant by the naive implementation - I don't think it's worth making it thread safe. It should be called only once on application startup so it'll be called from a UI thread. The access doesn't matter as it's thread safe by nature. Wdyt?

> Also, you use "private set" but initLogger is the same as public, any reason behind that?

Yes, if we decide to do more stuff in the init, we won't need to change the clients. If it was a pure setter, doing more stuff in the init would felt weird. Wdyt?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
malinajirka: Ohh, I didn't realize GsonRequest had multiple implementations. We can do that.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
malinajirka: Are you sure it'll work? I remember encountering issues on older android version when this syntax was used. Perhaps, it's not causing issues anymore, but we should be better sure, since this code is run very often in all client apps.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
malinajirka: Hmm, so it seems `JetpackTunnelGsonRequest` doesn't extend GsonRequest, so unless I'm missing something it's not possible. Wdyt @kidinov ?

fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
kidinov: > That's what I meant by the naive implementation - I don't think it's worth making it thread safe. It should be called only once on application startup so it'll be called from a UI thread. The access doesn't matter as it's thread safe by nature. Wdyt?

If you write from one thread, there is no guarantee that this change will be visible from another thread. So I would at least mark it `volatile`

And strictly speaking 

> It should be called only once on application startup

This is not ensured or declared anyhow in the current implementation. If you think that it has to be initiated only once, maybe throw an exception if someone tries to init it a second time?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
kidinov: Android studio complains about that. Probably just running this on Android 5 will ensure that it doesn't break anything. Most probably the issue back then was in the compiler that was not JDK 1.8+

fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
malinajirka: > If you write from one thread, there is no guarantee that this change will be visible from another thread. So I would at least mark it volatile

I can't imagine a scenario in which the OS/CPU would decide to put this variable into a cache before it's used for the first time. All subsequent usages are not changing the value of the variable - and imo even then the OS/CPU won't put this particular variable into the cache. `Volatile` is imo an overkill but I don't mind adding it - this is not a performance hungry code anyway -> updated in b4de11e6d6e515df71336f294a0dfa41495de54d

> And strictly speaking
> > It should be called only once on application startup

Again, same as with the Thread safety - I personally think that adding this adds unnecessary complexity which won't bring us anything at all. If the clients want to invoke the method multiple times for some very weird reason it's their decision - the code is clear enough that it should be used just once imo, the method is called `init = initialize`. Having said that I don't mind adding it -  updated in d95c8b5cf59c2a796dee39b7c7ce0f6eb035d0ed.

P.S. Introducing a double locking mechanism feels too much so unless you really think it's worth it, I'd vote for keeping the code simple over absolutely safe.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
malinajirka: Updated in b40e9c214fc20537f960d7fb8fa611cd6dd34812

malinajirka:
Thanks for the review and for the suggestions @kidinov. I've updated the code and replied to your comments. It's ready for another round, thanks!

fluxc/src/main/java/org/wordpress/android/fluxc/logging/FluxCCrashLoggerProvider.kt
kidinov: This is a library which can be used by other clients, that's why I think it better to keep it on the safe side, because you never know how people will use this, so better to be safe than sorry =)

malinajirka: I still think this is not worth the time or the code complexity. The chances of the original implementation going wrong were imo close to 0, moreover, even if it'd go wrong this is not a critical code. Having said that, I don't mind updating it - done in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2421/commits/0b8e8594d94f9307c0480a56484ae85612ee0fdd.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java
kidinov: Yep, won't work =(

kidinov (approved):
LGTM!

PR #2422, opened 2022-05-24
Issue/add prompt metadata to post

khaykov:
This PR adds answered prompt ID to the post model and supports storing it as a metadata when retrieving or submitting posts.

Before that we only used metadata for geolocation information (not missing in the apps, and left for backward-compatibility reasons) and I added additional handling for blogging prompt metadata when submitting and retrieving post. Metadata is converted into blogging prompt ID which is stored in DB.

Companion WP Android PR is [here](https://github.com/wordpress-mobile/WordPress-Android/pull/16676).


RenanLukas (approved):
LGTM :shipit:

PR #2423, opened 2022-05-26
[Coupons] Set the missing remote product ID to WCProductVariation table

0nko:
This PR fixes an issue where the `remoteProductId` value was not being set after fetching variations. This resulted in variations not being loaded from the DB. 

I also added a test to verify that the variations are being fetched and stored properly now.

**To test:**
Make sure all unit tests pass.


hichamboushaba (approved):
LGTM.

PR #2426, opened 2022-05-27
Quick Start: Expose Quick Start Status Set

ashiagr:
This PR adds a function to `QuickStartStore` to return if `Quick Start` status is set in DB.

## How to Test?

See instructions in https://github.com/wordpress-mobile/WordPress-Android/pull/16651


AjeshRPai (approved):
I have looked at the code changes and tested the Parent PR as per the instructions, all looks good to me 👍🏼 .

PR #2427, opened 2022-05-27
Fix blogging reminder db migration test

khaykov:
This PR fixes blogging reminder db migration test.

Looks like the test was broken initially, and for some reason was not picked up by CI.

To test is you can run the offending test locally and make sure it passes.


RenanLukas (approved):
LGTM :shipit:

PR #2429, opened 2022-05-28
Observe orders count

### LINKED ISSUE #2428: Add a function to observe the total orders in the database
We want to rely on small queries to fix issues with database queries that exceed the size of the Android cursor and improve app performance. It was necessary to observe the total number of orders in the database as a first step to improve how we can display this total. Now we can rely on SQLite to count the orders instead of processing all the orders to know the total number of orders.

More info here p91TBi-8Nj-p2

atorresveiga:
Closes: #2428

### Description
This PR adds a new function for counting the total number of orders in the DB. It does that by relying on an SQLite query using `COUNT` and Room for observing the changes.

### Testing
Run tests in WCOrderStoreTest
See corresponding Woo PR for tests https://github.com/woocommerce/woocommerce-android/pull/6659

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
@@ -58,9 +58,6 @@ abstract class OrdersDao {
     @Query("SELECT COUNT(*) FROM OrderEntity WHERE localSiteId = :localSiteId")
     abstract fun getOrderCountForSite(localSiteId: LocalId): Int
 
-    @Query("SELECT COUNT(*) FROM OrderEntity WHERE localSiteId = :localSiteId")
-    abstract fun observeOrderCountForSite(localSiteId: LocalId): Flow<Int>
-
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
```diff
@@ -58,6 +58,12 @@ abstract class OrdersDao {
     @Query("SELECT COUNT(*) FROM OrderEntity WHERE localSiteId = :localSiteId")
     abstract fun getOrderCountForSite(localSiteId: LocalId): Int
 
+    @Query("SELECT COUNT(*) FROM OrderEntity WHERE localSiteId = :localSiteId")
+    abstract fun observeOrderCountForSite(localSiteId: LocalId): Flow<Int>
```
wzieba: I think there's no use case for observing all orders. I'd vote to not add this and force API consumers to provide a list of statuses (YAGNI). 

Or even more: just expose `observeProcessedOrderCountForSite` as it's the only use case.

wzieba (approved):
Codewise looks good! I'm checking the Woo part now.

PR #2433, opened 2022-05-31
Woo: Coupons: Allow empty values on restriction fields when updating Coupons.

### LINKED ISSUE #2431: Woo: Coupons: Allow null value as update parameters for certain fields.
These are some of the available restrictions on a Coupon:
- Usage Limit Per Coupon (`usage_limit`)
- Limit Usage To X Items (`limit_usage_to_x_items`)
- Usage Limit Per User (`usage_limit_per_user`)

These restrictions are optional, and the values can be empty. When fetching through REST API, they are represented by `null` value:

Screen Shot 2022-05-30 at 16 34 39

Because of this, if we want to update a Coupon, we also need to allow those values to be null in the request parameter.

### Issue

Currently, the part of the code that builds the request parameter for updating only allows sending data if the data is not null:

https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/bee7c2597ba565dea0e9b8ccf57edb2c64c24d48/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/coupons/CouponRestClient.kt#L223-L225

The code needs to be updated to allow entering `null` values for those fields, so that in case merchants want to edit a restriction value from non-empty to empty, it can be saved properly.

hafizrahman:
Fixes: #2431 

### Description:
This PR allows the update Coupons REST API call to have `0` value for these fields:
- `usage_limit`
- `usage_limit_per_user`
- `limit_usage_to_x_items`

The purpose of the `0` value is to allow merchants to set those optional fields as empty from the app.

**Note:** It's undocumented in the API but it seems to only accept `0` as empty, not empty string or other stuff I've tried.

### To test:
1. Go to Example app > Woo > Coupons,
2. Select a site that contains a Coupon,
3. Find a Coupon ID for testing, either using Fetch Coupons button or via wp-admin,
4. Click "Update Coupon",
5. Clik "Enter Coupon ID", and enter a Coupon ID from step 3. Wait until the data loads,
6. Scroll down to "Usage limit", "Usage limit per user", "Limit usage to X items" fields,
7. Set all of those values as empty,
8. Scroll up and find the check mark on top right to save.
9. Make sure Coupon is saved properly. Check the Coupon in wp-admin and make sure the values for those are empty.

Optional: back in Example app, try entering some value for those fields, save, check, then try emptying it out again. Make sure the data is shown the same in wp-admin.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/coupons/CouponRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/coupons/CouponRestClient.kt
@@ -220,12 +220,15 @@ class CouponRestClient @Inject constructor(
             isShippingFree?.let { put("free_shipping", it) }
             productCategoryIds?.let { put("product_categories", it) }
             excludedProductCategoryIds?.let { put("excluded_product_categories", it) }
-            usageLimit?.let { put("usage_limit", it) }
-            usageLimitPerUser?.let { put("usage_limit_per_user", it) }
-            limitUsageToXItems?.let { put("limit_usage_to_x_items", it) }
             restrictedEmails?.let { put("email_restrictions", it) }
             isForIndividualUse?.let { put("individual_use", it) }
             areSaleItemsExcluded?.let { put("exclude_sale_items", it) }
+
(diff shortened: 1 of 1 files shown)


JorgeMucientes:
Looking good! Testted it and works as expected 🚢

JorgeMucientes (approved):

PR #2436, opened 2022-06-06
Fetch order summaries after adding a new order

hichamboushaba:
This is needed to fix the issue https://github.com/woocommerce/woocommerce-android/issues/6693

Check the PR https://github.com/woocommerce/woocommerce-android/pull/6706 for details.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
@@ -4,9 +4,12 @@ import com.google.gson.JsonArray
 import com.google.gson.JsonObject
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.FlowCollector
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -234,6 +239,11 @@ class OrderUpdateStore @Inject internal constructor(
             } else {
                 val model = result.result!!
                 ordersDao.insertOrUpdateOrder(model)
+                dispatcher.dispatch(
```
nbradbury, line 244: This change is good, I simply recommend adding a comment explaining why we fetch the order list here.

hichamboushaba: Done in 1af3ce9b4f30bece545acc7b021c7d90cfc2c680

nbradbury (approved):
Thanks for this change! :shipit:

PR #2437, opened 2022-06-09
[Woo] add support for searching product categories

hichamboushaba:
This simply adds an additional function to search for product categories, it reuses the same network implementation with just an additional parameter for the search query.

Check https://github.com/woocommerce/woocommerce-android/pull/6722 for testing instructions.


0nko (approved):
LGTM :shipit:

PR #2432, opened 2022-05-30
Revert "Merge pull request #2413 from wordpress-mobile/Stats-Revamp-v…

ashiagr:
This PR reverts commit 3435ea3b94ee0409486c7ae32041dfb37258db18, reversing changes made to 34f0870e70e5e8ccb2c14f7f3dd7d26ad9240b0d.

This is done to resolve the `WPAndroid` `release/19.9` merge conflict (see https://github.com/wordpress-mobile/WordPress-Android/pull/16661#issuecomment-1141211381).

/cc @ravishanker, @irfano

PR #2439, opened 2022-06-14
[Woo] Fix crash when parsing variable products that include details of variations

hichamboushaba:
This PR fixes the crash https://github.com/woocommerce/woocommerce-android/issues/6737, it updates the parsing logic to handle an Array of objects as well, and tries to extract the object id from the attribute `id` if it exists, otherwise it'll fail silently similar to what we were doing for other parsing errors.

I added a product with the `variations` set to an array of objects to the `wc-fetch-products-response-success.json` file to test this scenario too.

I created the branch `release/1.44.1` to use as target branch for the merge, WC 9.3 uses `1.44.0`, but I noticed the `1.44.1` was created after this [PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2432), and decided to use it instead, @oguzkocer please let me know if my thinking is correct or not.


0nko (approved):
Looks good! 👍

PR #2449, opened 2022-06-22
[Woo] Add Enable Coupons Function

hafizrahman:
### Description

This PR allows for enabling Coupons directly via API. This uses the WooCommerce Site Settings Option in https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#update-a-setting-option

### To Test
1. In wp-admin, disable Coupons first in WooCommerce > Settings > General tab > "Enable coupons"
3. Run Example app,
4. Go to Woo > Coupons 
5. Click "Select Site" and pick the site from step 1,
6. Back to Example app, click "Enable Coupons Feature" and check the log for success message,
7. Check back in wp-admin and make sure Coupons is now enabled by going to Marketing > Coupons


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -294,6 +296,28 @@ open class WooCommerceStore @Inject constructor(
...
+    ): WooResult<SiteSettingOptionResponse> {
+        return coroutineEngine.withDefaultContext(T.API, this, "updateSiteSetting") {
+            val response = wcCoreRestClient.updateSiteSettingOption(site, request, groupId, optionId)
+            return@withDefaultContext when {
+                response.isError -> {
+                    AppLog.w(T.API, "Failed to update site setting for ${site.siteId}")
+                    WooResult(response.error)
+                }
+                response.result?.value != null -> {
+                    fetchSiteGeneralSettings(site)
```
hafizrahman: After getting a result, we want to save the change to the database to keep it in sync. At first I considered updating just the individual field that's updated. However this turns out to require quite a lot of code since each `optionId` and the returned value from the API response will have to be mapped differently to different fields of `WCSettingsModel`:

```
- woocommerce_enable_coupons from API response has to be mapped to couponsEnabled in WCSettingsModel
- woocommerce_currency to be mapped to currencyCode
- woocommerce_currency_pos to be mapped to currencyPosition

...and so on.
```

For simplicity's sake, I opted for the more simple option of fetching everything (which then updates the DB as well).

example/src/main/java/org/wordpress/android/fluxc/example/ui/coupons/WooCouponsFragment.kt
```diff
@@ -29,6 +30,7 @@ class WooCouponsFragment : StoreSelectingFragment() {
 
     private val coroutineScope = CoroutineScope(Dispatchers.Main)
     private var couponPage = DEFAULT_PAGE
+    private var isCouponsEnabled = false
```
wzieba: I think this variable is not used

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -294,6 +296,37 @@ open class WooCommerceStore @Inject constructor(
...
+     * https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#list-all-settings-groups
+     * @param optionId The ID of the option to be updated.  List of IDs are available in
+     * https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#list-all-setting-options
+     */
+    suspend fun updateSiteSettingOption(
+        site: SiteModel,
+        request: UpdateSettingRequest,
+        groupId: String,
+        optionId: String
+    ): WooResult<SiteSettingOptionResponse> {
```
wzieba: I'd recommend changing the API we'll provide to the WooCommerce app. It'd be best IMO, to not leak details like `groupId` or `optionId` to the WooCommerce app - they're implementation details of WooCommerce Core API. Also, `UpdateSettingRequest` is quite specific. I wouldn't know how to use it, not looking into the source code of FluxC.

I'd suggest providing methods like `disableCoupons/enableCoupons` and don't bother the WooCommerce app with API-specifics. WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/settings/UpdateSettingRequest.kt
```diff
@@ -0,0 +1,5 @@
+package org.wordpress.android.fluxc.model.settings
+
+data class UpdateSettingRequest(
+    val value: String = ""
+)
```
wzieba: As currently `UpdateSettingRequest` can accept only two valid values, I'd suggest using `enum` here and removing the default value, so it'll be easier to understand how to use it. I'd also omit the words "yes/no" as it is an implementation detail.

I'd also specify it for our case, if it'll be needed in the future, we can think about abstraction.

So overall, my suggestion would be:

```kotlin
internal enum class CouponsSettingRequest {
ENABLE, DISABLE
}
```
and some mapper that will create network requests with "yes/no" values. WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
wzieba: Could you please provide a little more context about enabling/disabling coupons? 

Will it be a feature in settings that users will use very rarely or some frequent action?

wzieba (commented):
Good job @hafizrahman ! All works as expected.

I'm leaving a few comments about API modeling, so we can reduce the number of things that WooCommerce needs to know to make it work. WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
hafizrahman: in WooCommerce app, if a site has Coupons disabled and a merchant navigates to the Coupons area, then we display a message about it with a button to enable Coupons directly from there.

This is not available in any settings area so merchants will not be toggling Coupons on or off frequently in the app. I imagine it will be convenient for those already in the Coupons area in the app and want to enable it, though.

wzieba: Ok, I see, thanks!

Comment relating to other parts of the PR: so maybe it makes sense to not add API for disabling coupons at all now, as I understand we won't offer this feature? WDYT?

hafizrahman:
Thanks for your review @wzieba ! I really like the idea of simplifying this to just be for enabling Coupons, instead of the more generic site setting option updating (which is not yet used anywhere). The PR just becomes so much simpler as well.

This is ready for another review.

(I've updated the testing instruction in the original post as well)

wzieba (approved):
Awesome, great improvements! I've re-checked if enabling coupons still work - it does.

Merging :shipit: !

PR #2450, opened 2022-06-22
Woo/search products by SKU

nbradbury:
This PR enables searching products by SKU. The example app has been updated to enable searching by SKU, and a test was added. You can use [this draft WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/6838) to test further.

A few things to note:

* The backend treats searching by SKU as separate from the _normal_ product search. We expect this to change so that SKU is automatically included in the search, but for now we have to accept this limitation.
* Prior to core v6.6, searching by SKU relied on the `sku` parameter which only matches on the full SKU. In v6.6 a `search_sku` parameter was added which performs a partial match.
* If both the above parameters are included, the latter takes priority. So we can add both parameters without needing to check for the core version.


nbradbury:
Just a note to say that we've decided to hold off implementing this until the API can include SKU when searching. The current API searches _either_ the product fields or the SKU, which requires the user to choose what to search, which isn't ideal.

nbradbury:
@wzieba @atorresveiga Are either of you able to review this?

wzieba:
sorry @nbradbury I've missed this. Thank you for the ping, I'll prioritize it tomorrow (it's EOD for me now).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -432,13 +448,21 @@ class ProductRestClient @Inject constructor(
 
     fun searchProducts(
         site: SiteModel,
-        searchQuery: String,
+        searchQuery: String?,
```
wzieba: I think we don't have to make this parameter nullable, WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -367,11 +368,20 @@ class ProductRestClient @Inject constructor(
...
+            "per_page" to pageSize.toString(),
+            "orderby" to orderBy,
+            "order" to sortOrder,
+            "offset" to offset.toString()
+        )
+
+        if (searchQuery?.isNotEmpty() == true) {
+            if (isSkuSearch) {
+                params["sku"] = searchQuery // full match
+                params["search_sku"] = searchQuery // partial match, added in WC core 6.6
```
wzieba, line 380: Minor but IDE keeps nudging for 
image
image

can we follow it? It makes navigation via "F2" less cluttered.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -600,10 +627,16 @@ class ProductRestClient @Inject constructor(
...
-            .putIfNotEmpty("include" to ids.map { it }.joinToString())
+        ).putIfNotEmpty("include" to ids.map { it }.joinToString())
             .putIfNotEmpty("exclude" to excludedProductIds.map { it }.joinToString())
 
+        if (isSkuSearch) {
+            params.putIfNotEmpty("sku" to searchQuery)
+            params.putIfNotEmpty("search_sku" to searchQuery)
+        } else {
+            params.putIfNotEmpty("search" to searchQuery)
+        }
```
wzieba, line 638: Can we extract this logic and share it with `fetchProducts`?

I think this will reduce the risk of having discrepancies in the future.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/StringExt.kt
```diff
@@ -0,0 +1,20 @@
+package org.wordpress.android.fluxc.utils
+
+fun String.semverCompareTo(otherVersion: String): Int {
```
wzieba: I think we don't need this method

wzieba (commented):
Works good, leaving some comments, WDYT?

I see that you also included a change to `checkstyle-idea.xml` - can we revert it?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
nbradbury: I started to do this then realized we should reuse `buildProductParametersMap` in `fetchProducts`. Rather than do that in this PR, I'd like to assign a separate issue to myself to tackle that separately. Sound good?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/StringExt.kt
nbradbury: Oops, I added that early on when I thought I needed to check the core version. It's removed now.

nbradbury:
@wzieba Thanks for the review! I tried to revert the checkstyle changes by editing the file directly in GitHub, and it keeps adding a new line at the end. I think this is harmless. WDYT?

Anyway...ready for another round!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
wzieba: Sure, sounds good!

wzieba (approved):
> I think this is harmless. WDYT?

Yes - I've doubled checked if AS doesn't remove this line and creates not needed diff but it's all good. 

:shipit: !

PR #2455, opened 2022-06-30
Fix stats cache buster

ThomazFB:
Summary
==========
Fix issue https://github.com/woocommerce/woocommerce-android/issues/6595 by removing the random page size acquisition logic to calculate the precise needed page size into the stats request, since now the API is fixed and doesn't require us to keep doing that, just like iOS did [here](https://github.com/woocommerce/woocommerce-ios/pull/6640)

How to Test
==========
This PR is better tested using directly the Woo app, so I recommend configuring this code into the `fluxCVersion` on the Woo app and testing the four types (day, week, month, and year) of stats are displaying exactly as before the changes introduced here.

Update release notes:


nbradbury (approved):
Fix is good! :shipit:

PR #2459, opened 2022-07-05
Nick/metadata test

nbradbury:
This is a test PR to experiment with storing order metadata in a separate table. Right now when an order is fetched we strip most of the metadata before storing it. We do this in several places:

* `fetchSingleOrder`
* `fetchOrders`
* `fetchOrdersByIds`
* `updateOrder`

The approach taken here is to create a function that's used in all these places to store the metadata before its stripped. Another option would be to fetch the metadata separately when needed, but I didn't like the idea of adding another network request to order detail.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
@@ -32,6 +32,8 @@ abstract class WCDatabaseModule {
 
         @Provides fun provideOrderNotesDao(database: WCAndroidDatabase) = database.orderNotesDao
 
+        @Provides fun provideOrderMetaDataDao(database: WCAndroidDatabase) = database.orderMetaDataDao
+
         @Provides fun provideInboxNotesDao(database: WCAndroidDatabase) = database.inboxNotesDao
     }
     @Binds abstract fun bindTransactionExecutor(database: WCAndroidDatabase): TransactionExecutor
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/OrderMetaDataEntity.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/OrderMetaDataEntity.kt
@@ -1,0 +1,22 @@
+package org.wordpress.android.fluxc.model
+
+import androidx.room.Entity
+import androidx.room.Index
+import com.google.gson.annotations.SerializedName
+
+@Entity(
+    tableName = "OrderEntity",
+    indices = [Index(
+        value = ["localSiteId", "orderId"]
+    )],
+    primaryKeys = ["localSiteId", "orderId", "id"]
+)
+data class OrderMetaDataEntity(
+    @SerializedName("id") val id: Long,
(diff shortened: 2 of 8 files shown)


ThomazFB:
Hey @nbradbury, thanks for starting experimenting with this solution! I believe that's the way to go for us to use the separate table for the metadata, I just notice that you decided to update the metadata of a given order for every network action we make, even when the order itself is not updated in the database, like searching. 

Is this because of the metadata strip that happens in the `OrderDtoMapper`? I ask this because I was wondering if we shouldn't inject the `orderMetaDataDao` into the `WCOrderStore` instead of the `OrderRestClient`, which is where all DAOs are injected and the insert operation is called (`ordersDao.insertOrUpdateOrder`), this way we would only update the order metadata when the order itself is being updated too. I believe this happens in five places:

- `WCOrderStore.fetchSingleOrder`
- `WCOrderStore.updateOrderStatus`
- `WCOrderStore.updateOrderStatusLocally`
- `WCOrderStore.handleFetchOrdersCompleted`
- `WCOrderSTore.handleFetchOrderByIdsCompleted`

 Or do you think we should update the metadata constantly?

nbradbury:
> Is this because of the metadata strip that happens in the `OrderDtoMapper`?

That's exactly why. By the time `WCOrderStore` is involved, the metadata has already been stripped. 

> I just notice that you decided to update the metadata of a given order for every network action we make, even when the order itself is not updated in the database, like searching.

Yeah, we can cut back on the number of places we update the metadata.

I do wonder if we should consider fetching metadata separately. I don't like the separate request, but that would be cleaner. If we build a UI that collapses the metadata (and I'm leaning towards that), then we could defer the request until the user expands it. That would also be a nice way to track usage.

ThomazFB:
>I do wonder if we should consider fetching metadata separately. I don't like the separate request, but that would be cleaner. If we build a UI that collapses the metadata (and I'm leaning towards that), then we could defer the request until the user expands it. That would also be a nice way to track usage.

I like the idea of using a dedicated request to fetch the metadata! This way we can apply more control of what we want to strip or keep, without "hitchhiking" the main Order requests to acquire the data we need and leave what the `OrderDtoMapper` stripping does to be just the essential metadata that needs to be acquired with the Order.

rachelmcr:
> I do wonder if we should consider fetching metadata separately. I don't like the separate request, but that would be cleaner. If we build a UI that collapses the metadata (and I'm leaning towards that), then we could defer the request until the user expands it. That would also be a nice way to track usage.

How would that work for orders that don't have any meta data to display? My expectation was that we'd only show a section for meta data if the order has some to display, but in this case would that section always be there and just sometimes be empty?

nbradbury:
> How would that work for orders that don't have any meta data to display?

I was thinking that when we enter order detail we already fetch the order, and we should be able to tell from that if it has any useful metadata. So even though we don't store the metadata with the order, we could store whether the order has useful metadata. It's not a perfect solution, but it should (?) work.

nbradbury:
@ThomazFB It just occurred to me that we could simply parse out the metadata in `OrderRestClient.fetchSingleOrder`, before its stripped. That's called by order detail, so a separate request for metadata isn't really necessary. I don't think we don't need to update metadata anywhere else. WDYT?

ThomazFB:
@nbradbury awesome idea! I think we should go with this!

PR #2460, opened 2022-07-06
Wc/order metadata table

nbradbury:
This PR implements the order metadata table and populates it following a request to `fetchSingleOrder`.


nbradbury:
~~Marking this "not ready for merge"  pending the decision [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2459#issuecomment-1176450624).~~

nbradbury:
@ThomazFB I've updated this to save order metadata following `fetchSingleOrder`, and I've added code to the database layer that will enable the UI to determine if metadata exists for an order.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
```diff
@@ -515,6 +515,8 @@ internal class AutoMigration13to14 : AutoMigrationSpec
 @DeleteTable(tableName = "ProductCategories")
 internal class AutoMigration14to15 : AutoMigrationSpec
 
+internal class AutoMigration16to17 : AutoMigrationSpec
```
ThomazFB, line 518: np: Should this declaration be at the end of the file? Correct me if I'm wrong, but I believe the migrations follow a timeline order, so in theory, this one should be after the `MIGRATION_15_16`.

ThomazFB (approved):
Migration, Entity, and DAO declarations look good! Just left a small np, but nothing important.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderMetaDataDao.kt
```diff
@@ -0,0 +1,61 @@
...
+    open fun updateOrderMetaData(
+        orderDto: OrderDto,
+        localSiteId: LocalId
+    ) {
+        val orderId = orderDto.id ?: 0
+        deleteOrderMetaData(localSiteId, orderId)
+
+        val responseType = object : TypeToken<List<WCMetaData>>() {}.type
+        val metaData = Gson().fromJson(orderDto.meta_data, responseType) as? List<WCMetaData>
+            ?: emptyList()
```
wzieba: In this case, for every order we'll create a new Gson object - having a list of the object, we might introduce some visible performance/memory flaws as Gson is a pretty heavy, reflection-based library.

Also, I think it shouldn't be the responsibility of `*Dao` class to parse JSON. 

My suggestion would be to move this logic to a different class and perform deserialization before sending data to `*Dao`. It'd be nice to inject Gson object there, like in `StripOrder` class.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderMetaDataDao.kt
```diff
@@ -0,0 +1,61 @@
...
+    @Transaction
+    @Query("DELETE FROM OrderMetaDataEntity WHERE localSiteId = :localSiteId AND orderId = :orderId")
+    abstract fun deleteOrderMetaData(localSiteId: LocalId, orderId: Long)
+
+    @Transaction
+    open fun updateOrderMetaData(
+        orderDto: OrderDto,
+        localSiteId: LocalId
+    ) {
+        val orderId = orderDto.id ?: 0
```
wzieba: In case when core returns an Order with a null id, I think we shouldn't save its metadata - it's very likely that the Order is malformed.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderMetaDataDao.kt
```diff
@@ -0,0 +1,61 @@
...
+    @Query("SELECT COUNT(*) FROM OrderMetaDataEntity WHERE orderId = :orderId AND localSiteId = :localSiteId")
+    abstract suspend fun getOrderMetaDataCount(orderId: Long, localSiteId: LocalId): Int
+
+    @Transaction
+    @Query("DELETE FROM OrderMetaDataEntity WHERE localSiteId = :localSiteId AND orderId = :orderId")
+    abstract fun deleteOrderMetaData(localSiteId: LocalId, orderId: Long)
+
+    @Transaction
+    open fun updateOrderMetaData(
+        orderDto: OrderDto,
```
wzieba: I'd suggest not using `OrderDto` directly in `*Dao`. 

The `*Dto` object is one from the networking layer; ideally, the `*Dao` object from the persistence layer shouldn't know anything about the networking.

My suggestion would be to send just `localSiteId` and `orderId` instead as parameters.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderMetaDataEntity.kt
```diff
@@ -0,0 +1,27 @@
...
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.Index
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+
+/**
+ * The OrderMetaDataEntity table is used to store viewable order meta data separately from the order
+ */
+@Entity(
+    tableName = "OrderMetaDataEntity",
```
wzieba: Minor: I've introduced a bad convention with `OrderEntity` - we should name tables with SQL standard, like we do with `CouponEntity` - `Coupons` or `OrderNoteEntity` - `OrderNotes`.

wzieba (commented):
Hi @nbradbury 👋 as I was reviewing other FluxC PRs, I've chimed in here. I left some comments, I think those might be valuable improvements - I hope you don't mind!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
nbradbury: @ThomazFB  Does the declaration order determine the migration timeline? I assumed that's controlled by [Room.databaseBuilder](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ae5410d016884c3b2b4faed8600b44ee4e6ad8ad/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt#L76).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrderMetaDataDao.kt
nbradbury: @wzieba That's a good point, especially since we need to prune out useless metadata and that shouldn't be handled in either place. I'll create a separate class for this.

nbradbury:
@ThomazFB @wzieba I've made the suggested changes, so ready for another round!

Note that I renamed the table, so you'll need to clear storage to reset the database.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
ThomazFB: it does not! I believe it's just the pattern adopted in the file migration declarations, not something that affects how Room works, that's why I believe this is a np 😬

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderMetaDataEntity.kt
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * The OrderMetaDataEntity table is used to store viewable order meta data separately from the order
+ */
+@Entity(
+    tableName = "OrderMetaData",
+    indices = [Index(
+        value = ["localSiteId", "orderId"]
+    )],
+    primaryKeys = ["localSiteId", "orderId", "id"]
```
wzieba, line 17: I think we should also declare a foreign key, so `OrderMetaDataEntity` will be removed from the database with the `Order`. Maybe we could also add a test for that?

I don't know if we want to add this now but that might be a good idea so we'll not have to prepare a migration later. But I'm leaving this for you to decide 👍 !

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderMetaDataHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.model.WCMetaData
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.OrderMappingConst.isInternalAttribute
+import org.wordpress.android.fluxc.persistence.dao.OrderMetaDataDao
+import org.wordpress.android.fluxc.persistence.entity.OrderMetaDataEntity
+import javax.inject.Inject
+
+class OrderMetaDataHandler @Inject constructor(
```
wzieba, line 12: nit: this could be `internal` so it won't ever leak to Woo codebase.

wzieba (approved):
Looks good, thanks for the improvements @nbradbury ! Leaving basically one comment about ForeignKey constraint.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderMetaDataEntity.kt
nbradbury: @wzieba I added a foreign key, but for some reason that was causing order tests to fail. I've not created a foreign key with Room before so I'm assuming I'm doing something wrong and need to read up on the topic. For now, let's just skip the foreign key. We know we're going to have to drop the table and recreate it at some point, so I think we're okay to do that.

nbradbury:
@ThomazFB Would it be possible to get this merged today?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderMetaDataEntity.kt
wzieba: I see. The problem is that in the commit [e7d7369](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2460/commits/e7d736966276f4fb5d754a7ef21b7c4f514936ce) we're saving metadata to the database before saving orders, so foreign key constraint is failing.

I just realized another issue: we save metadata in `OrderRestClient`, which should handle only network communication. We should move this logic to `WCOrderStore`, which orchestrates all networking and persistence logic. The best way to resolve this would be to see where `OrdersDao#insertOrUpdateOrder` is used, and in those places save OrdersMetaData.

ThomazFB:
Hey @nbradbury, sure! I already left the PR approved, so now that we have Wojtek's green light I think we're good to go!

wzieba:
I just noted that we don't have `17.json` generated file in this PR. @nbradbury could you please prepare PR for that? 🙏

nbradbury:
@wzieba Sorry about that! For some reason I had it my head that we should exclude the JSON from our PRs. I'll prepare a new PR shortly.

wzieba:
We both missed that! Thank you.

btw: I'm thinking if there's some way in Room to have some check for not having generated JSON schema so the build will fail in CI 🤔 .

ThomazFB:
We all missed that 😅, I'm a Room newbie and end up just reviewing the basics. Maybe we should add a document to our repo stating what's expected when doing changes to Room so anyone can be on the same page when reviewing or doing something around Room migrations? 🤔

PR #2462, opened 2022-07-07
Wc/reuse build parameters map

nbradbury:
This PR updates `ProductRestClient` to reuse `buildProductParametersMap` in `fetchProducts` rather than have two places where the parameters are created.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -476,11 +476,11 @@ class ProductRestClient @Inject constructor(
         pageSize: Int = DEFAULT_PRODUCT_PAGE_SIZE,
         offset: Int = 0,
         sortType: ProductSorting = DEFAULT_PRODUCT_SORTING,
-        includedProductIds: List<Long> = emptyList(),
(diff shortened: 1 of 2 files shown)


wzieba:
Sorry for the late review @nbradbury ! I'll prioritize this PR early next week.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -476,19 +449,19 @@ class ProductRestClient @Inject constructor(
         pageSize: Int = DEFAULT_PRODUCT_PAGE_SIZE,
         offset: Int = 0,
         sortType: ProductSorting = DEFAULT_PRODUCT_SORTING,
-        includedProductIds: List<Long> = emptyList(),
-        excludedProductIds: List<Long> = emptyList(),
+        includedProductIds: List<Long>? = null,
+        excludedProductIds: List<Long>? = null,
```
JorgeMucientes, line 453: Curious why better nullable than non-nullable and empty. 
Is it to simplify code in `buildProductParametersMap` and use `let` instead of `if`? Like:
```kotlin 
includedProductIds?.let { ... } 
```
Instead of something like this?
```kotlin
if (includedProductIds.isNullOrEmpty().not()) { ... }
```

JorgeMucientes (approved):

JorgeMucientes:
Code looks good! :shipit:

PR #2466, opened 2022-07-13
Woo/Allow displayValue to be null

nbradbury:
This tiny (one character!) PR fixes an issue with the `OrderMetaData` table that caused a null `displayValue` to be stored as the string "NULL".

**Before**
before

**After**
after


nbradbury:
Closing this as we decided on Slack to drop `displayKey` and `displayValue`. ref: p1657734792470349-slack-CGPNUU63E

PR #2468, opened 2022-07-13
Woo/order metadata drop display fields

### LINKED ISSUE #2467: Woo: Drop displayKey and displayValue from order metadata
As per this Slack discussion: p1657734792470349-slack-CGPNUU63E

We don't need the `displayKey` and `displayValue` fields in order metadata and they should be dropped.

nbradbury:
Closes #2467 - as discussed on Slack the `displayKey` and `displayValue` fields are unused in the order metadata, so this PR drops them from the entity and table.

p1657734792470349-slack-CGPNUU63E


ThomazFB (approved):
A new  schema file size always gives me a surprise when entering the PR 😆, anyway, LGTM :shipit: 

I'm leaving this PR just approved in case you want to sync anything with the Woo side before merging everything.

PR #2475, opened 2022-07-31
Send remote payload error when updateOrderStatus fails

atorresveiga:
This small PR ensures sending the remote payload error as part of the OnOrderChanged event so we can know when something was wrong with the request on the client side. 

To test, try to update the order status in airplane mode and check that the OnOrderChanged result object contains the remote payload error.


wzieba (commented):
hi @atorresveiga 👋 

Could you help me to understand the goal of this PR? From my point of view, it doesn't change any behavior 🤔 . If it does, I'm sorry maybe it's not my day 😅 - could you please add unit tests that cover this new behavior (and which will show the change)?

ThomazFB:
I'm stepping forward with this PR review since I'm reviewing [this one](https://github.com/woocommerce/woocommerce-android/pull/7082), which depends on the code introduced here. I hope that's ok!

atorresveiga:
Hi @wzieba and @ThomazFB!
The problem this PR is trying to fix is that in the previous version of the code when using `.copy` to update the `causeOfChange` in both `OnOrderChanged` events, the new _copy_ was missing the error property the event from `revertOrderStatus` should keep. This caused that when we received the OnOrderChanged in Woo, we treated it as a Success event even when it was not, because error property was `null`.

> could you please add unit tests that cover this new behavior

Sure, I will add one  😁

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -857,7 +857,8 @@ class WCOrderStore @Inject constructor(
 
     private suspend fun revertOrderStatus(payload: RemoteOrderPayload.Updating): OnOrderChanged {
         updateOrderStatusLocally(payload.order.orderId, payload.order.localSiteId, payload.order.status)
-        return OnOrderChanged().also { it.error = payload.error }
+        return OnOrderChanged(causeOfChange = WCOrderAction.UPDATE_ORDER_STATUS)
+            .also { it.error = payload.error }
```
wzieba: Oh ok @atorresveiga I see. That's because we update here `OnChanged#error`, which is not part of `data class OnOrderChanged`. That's quite a tricky API. I've introduced it :( but it's still better than before IMO. I think we're prisoners of `OnChanged` super class here.

My suggestion would be to use following constructor:
```suggestion
        return OnOrderChanged(orderError = payload.error)
```

and remove change from line 575-576. I think that this way the code is more readable.

Either way I'd appreciate unit tests as you agreed, so we'll make sure that no one falls into this trap again.

wzieba (approved):
Thanks @atorresveiga ! Last suggestion - can we keep the `copy` as it was before? This way, it's visible right away that any event, success or fail, has the same `causeOfChange`. 

It's not a blocking change though.

wzieba (approved):
Looks good, thank you!

PR #2171, opened 2021-11-02
Encode email delimiter chars defined in RFC 3986

jd-alexander:
Fixes #2170 
Closes https://github.com/wordpress-mobile/WordPress-Android/issues/15480

Related PRs:
- https://github.com/wordpress-mobile/WordPress-Android/pull/15526
- https://github.com/wordpress-mobile/WordPress-Login-Flow-Android/pull/73

## Description 

Encodes delimiters in email addresses when calling the `auth-options` endpoint, based on the [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) specification.

See: [Similar fix in iOS](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/457).

More info in this issue comment: https://github.com/wordpress-mobile/WordPress-Android/issues/15480#issuecomment-1202342661.

**To Test:**
Follow the testing steps from the integrating PR: https://github.com/wordpress-mobile/WordPress-Android/pull/15526.

## Merging Strategy

**Notice** ✅ One review should be enough to merge these PRs.
Check the [merging strategy](https://github.com/wordpress-mobile/WordPress-Android/pull/15526) in the `WordPress-Android` PR.

PR #2480, opened 2022-08-02
[DO NOT MERGE] Encode email delimiter chars defined in RFC 3986

### LINKED ISSUE #2170: Update login and sign up email to use RFC 3986 encoding
Two API paths that utilize the user's email to fetch auth options and perform signups fail when the email contains any of these characters `!’()*` therefore the RFC 3986 standard is being utilize to restrict these values as seen here https://datatracker.ietf.org/doc/html/rfc3986

**API actions being enhanced with sanitized emails**

- `newFetchAuthOptionsAction`
-  `newSendAuthEmailAction`

ovitrif:
Fixes #2170 

## Description

This PR encodes the delimiter characters based on the [RFC 3986 standard](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) because our backend code makes multiple references to RFC 3986.

**To Test:**
Follow the testing steps from the integrating PR: https://github.com/wordpress-mobile/WordPress-Android/pull/15526.

**Note:**
This PR was created to investigate why the CI checks don't succeed on the original one: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2171.


ovitrif:
Not needed, concluded the issue was another call to `WPUrlUtils.escapeUrlPathWithRFC3986`.

PR #2489, opened 2022-08-08
[Tooling/CI] Update CI pipeline to group steps + add missing step dependencies

AliSoftware:
Up until now, the various Lint and Unit Test steps that [have been recently added to the pipeline](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2472) were not set up as prerequisites for the various PublishToS3 steps. Which means that nothing prevented us to publish a version to S3 before realizing that the lint found violations or the Unit Tests failed.

This PR:
 - Groups the steps per topic, using [Buildkite's Group steps](https://buildkite.com/docs/pipelines/group-step), both to improve the way those are presented in the Buildkite dashboard and summary header, and to make the dependency logic easier to write
 - Add step dependencies between the `Publish to S3` group and all the linters and unit test steps/groups
 - Move the "Publish to S3" group / steps at the end of the pipeline, so that they appear at the end of the list in the Buildkite UI / dashboard too for consistency (since they'll run last, after Lint and Tests)

image

> **Note**: We deliberately decided not to make the Publish to S3 group depend on the Connected Tests step for now, because:
> - Those Connected Tests as still a bit flaky and prone to randomly fail, and preventing the publication to S3 because of this flakiness would be quite impractical.
> - Those Connected Tests take ≈40mn to run, which would delay the publication of new versions of FluxC by quite a lot compared to the speed we have now between creating a new tag or commit and having it published.

<details><summary>As you can see, the four <em>Publish</em> steps don't start while the Lint and Unit tests are still running and haven't passed yet.</summary>
image
</details>

<details><summary>And a bit later, you can see that the Publish steps start once the Lint and Unit Test steps have finished — even if the Connected Tests step is still running (because we decided not to depend on that one, see above)</summary>
image
</details>

> **Note**: This PR might be easier to review with "Ignore Whitespace" turned on, since a large portion of the diff is change in indentation to move existing-and-unmodified-steps under a `group:` parent step.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -12,7 +12,61 @@ agents:
   queue: "android"
 
 steps:
+  ############################
+  # Linters
+  ############################
+  - group: "🕵️ Linters"
+    key: all-linters
+    steps:
+      - label: "🕵️ checkstyle"
+        command: |
+          cp gradle.properties-example gradle.properties && cp -a example/properties-example/ example/properties/
+          ./gradlew checkstyle
+        artifact_paths: *artifact_paths
+
+      - label: "🕵️ detekt"
+        command: |
+          cp gradle.properties-example gradle.properties && cp -a example/properties-example/ example/properties/
+          ./gradlew detekt
+        artifact_paths: *artifact_paths
+
+      - label: "🕵️ Lint"
+        command: |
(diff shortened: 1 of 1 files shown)


ParaskP7 (approved):
👋 @AliSoftware !

I have reviewed and checked CI, everything LGTM. Thank you so much for making this changes and reordering, this now looks so much more organized! 🥇 🌟 🚀

PR #2492, opened 2022-08-10
Add PUT/POST endpoint in fluxc to support enabling/disabling Cash on Delivery as payment method

### LINKED ISSUE #7141: Add PUT/POST endpoint in fluxc  to support enabling/disabling Cash on Delivery  as payment method
The goal of this task is to add an endpoint that supports enabling/disabling Cash on Delivery in the FluxC module. The endpoint will be PUT/POST
`/wp-json/wc/v3/payment_gateways/<id>`

rossanafmenezes:
Closes [woocommerce-android/issues/7141](https://github.com/woocommerce/woocommerce-android/issues/7141) and [woocommerce-android/issues/7144](https://github.com/woocommerce/woocommerce-android/issues/7144)

This PR adds the endpoint functions to enable/disable cash on delivery from the Woo Mobile app and give the user the ability to change the Cash on Delivery title. 

Context: pdfdoF-167-p2

It also adds a new test file related to the gatewayRestClient.kt file and the equivalent tests for it.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt
```diff
@@ -50,6 +50,55 @@ class GatewayRestClient @Inject constructor(
...
+                WooPayload(response.error.toWooError())
+            }
+        }
+    }
+
+    suspend fun postCashOnDeliveryTitle(
+        site: SiteModel,
+        title: String
+    ): WooPayload<GatewayResponse> {
+        val url = WOOCOMMERCE.payment_gateways.gateway(title).pathV3
```
malinajirka: ⚠️ This line is responsible for the construction of the URL. However, it seems we are including the user-facing title as part of the url.

For example, when the user sets the title to `Pay with a credit/debit card or cash upon delivery.`, the url we construct will be
`.../wc/v3/payment_gateways/Pay with a credit/debit card or cash in person./`. 


Since, the endpoint for disabling/enabling COD and changing the title is one and the same, we can re-use the code from line 58 - `val url = WOOCOMMERCE.payment_gateways.gateway(gatewayId.toString()).pathV3`. 

We could even go as far as merging `postPaymentGateway` and `postCashOnDeliveryTitle` into one method, perhaps called something like `updatePaymentGateway` - it's one endpoint, so having two separate methods for it feels slightly misleading. We'd just need to make sure to make the `title` parameter as optional so we post it to the server only when the user changes it. Wdyt?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt
```diff
@@ -50,6 +50,55 @@ class GatewayRestClient @Inject constructor(
...
+        val params = mapOf("Cash on Delivery" to title)
```
malinajirka: ⚠️ The `key` in the map is currently set to `Cash on Delivery`, however [according to the documentation](https://woocommerce.github.io/woocommerce-rest-api-docs/#update-a-payment-gateway) it seems the correct name for the `key` is `title`. I believe we might need to change this line to `mapOf("title" to title)`, wdyt?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt
```diff
@@ -82,4 +131,8 @@ class GatewayRestClient @Inject constructor(
         @SerializedName("method_description") val methodDescription: String?,
         @SerializedName("method_supports") val features: List<String>?
     )
+
+    enum class GatewayId {
+        COD
```
malinajirka: 🔍 Nitpick: Most devs who will be reading this code will realize, that COD means Cash On Delivery. However, probably not all of them. I think we might consider making it a bit more explicit/readable by calling the enum CASH_ON_DELIVERY and setting a key to `cod`. Wdyt?

```
enum class GatewayId(val apiKey: String) {
        CASH_ON_DELIVERY("cod")
    }
```

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/GatewayRestClientTest.kt
```diff
@@ -0,0 +1,298 @@
...
+            jetpackTunnelGsonRequestBuilder,
+            mock(),
+            requestQueue,
+            accessToken,
+            userAgent
+        )
+    }
+
+    @Test
+    fun `given success response, when post gateway, return success`() {
```
malinajirka: 💡 The tests in this class feel quite complicated considering what they actually test. Unless I'm missing something we are mocking essentially everything. The tests only test that when a success is returned the response is stored into `WooPayload.result` and when a failure is returned the response is stored into `WooPayload.error`.

Perhaps, we could simplify them to the following format to make it more explicit what they actually test:

```
@Test
    fun `given success response, when post gateway, return success`() {
        runBlocking {
            whenever(
                jetpackTunnelGsonRequestBuilder.syncPostRequest(
                    any(),
                    any(),
                    any(),
                    any(),
                    any<Class<GatewayRestClient.GatewayResponse>>()
                )
            ).thenReturn(
                JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackSuccess(mock())
            )

            val actualResponse = gatewayRestClient.postPaymentGateway(
                SiteModel(),
                COD,
                enabled = false
            )

            Assertions.assertThat(actualResponse.isError).isFalse
            Assertions.assertThat(actualResponse.result).isNotNull
        }
    }

    @Test
    fun `given error response, when post gateway, return error`() {
        runBlocking {
            val expectedError = mock<WPComGsonNetworkError>().apply {
                type = mock()
            }
            whenever(
                jetpackTunnelGsonRequestBuilder.syncPostRequest(
                    any(),
                    any(),
                    any(),
                    any(),
                    any<Class<GatewayRestClient.GatewayResponse>>(),
                )
            ).thenReturn(
                JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackError(expectedError)
            )

            val actualResponse = gatewayRestClient.postPaymentGateway(
                SiteModel(),
                COD,
                enabled = true
            )

            Assertions.assertThat(actualResponse.isError).isTrue
            Assertions.assertThat(actualResponse.error).isNotNull
        }
    }
```

Wdyt?


P.S. In general, I personally prefer the Mocked tests over unit tests for rest clients. The mocked tests actually test the code works and test the json parsing works. Moreover, they serve as a documentation for possible responses from the endpoint. Something similar to what we have in `MockedStack_InPersonPaymentsTest`. But we can do that in another PR later, it's not something we need to do. Wdyt?

malinajirka (changes requested):
Thanks @rossanafmenezes! You are on the right path, but I noticed a couple of issues (⚠️). Moreover, I left some thoughts/nitpicks (🔍 💡 ) for you to consider - those don't require code changes, just think about them and try to apply them next time if they make sense to you. Thanks :)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt
rossanafmenezes: 💯  Changed

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt
rossanafmenezes: @malinajirka  thanks!

[I had originally done only one method](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2492/commits/ecc30907f43ca62518ea2f8f66447ba2245a30d3) (and renamed it after) but had not made the title optional (sorry, it did not cross my mind) so it would not work the way I did and ended up splitting it into two. 

cc: @AnirudhBhat 

When you say "make it optional" do you mean making the title nullable?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt
rossanafmenezes: Absolutely. Sorry about that.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt
malinajirka: No reason to apologize :).

> When you say "make it optional" do you mean making the title nullable?

Make it nullable and default value `null` => when the value is `null` do not send the `enabled`/`title` parameter to the server so the server doesn't change the setting. Something like this:

```
 suspend fun ...(
        site: SiteModel,
        gatewayId: GatewayId,
        enabled: Boolean? = null,
        title: String? = null
    ) ...
    ...
      mutableMapOf<String, Any>().apply {
            enabled?.let { put("enabled", enabled) }
            title?.let { put("title", title) }
        }
```

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/GatewayRestClientTest.kt
rossanafmenezes: I had also started with a [mocked test at first ](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2492/commits/8a7a3a960165ea9ebae886e911ad90dfd92af114) but I got a bit lost. I will simplify the existing test and I can add an issue to work on the mock tests to replace these ones

AnirudhBhat:
@rossanafmenezes Thank you so much for working on this! I've got a small comment.

I see that you have added the endpoint implementation in `GatewayRestClient` class. AFAIU, the WooCommerce module doesn't touch the client classes directly; all communication happens via the store class which takes care of thread switching. In this case, it would be the `WCGatewayStore` class. 

What do you think about adding methods in `WCGatewayStore` class that in turn calls the methods from the `GatewayRestClient` class? 

This can be a separate PR as well. Just wanted to bring this to your attention. Sorry I missed noticing this during our initial discussion.

rossanafmenezes:
Hi @AnirudhBhat 

Thanks for the input :) 

> What do you think about adding methods in WCGatewayStore class that in turn calls the methods from the GatewayRestClient class?

While  I understand this is usually addressed on the same PR, the plan was to do that on a separate PR this time. :)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/gateways/GatewayRestClient.kt
rossanafmenezes: Great. Thank you. I have consolidated both methods into one

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/GatewayRestClientTest.kt
malinajirka: Thanks! I wouldn't replace them, they both add some value - but adding the mocked tests in a follow up PR would be awesome. Thank you 🙇

malinajirka (approved):
LGTM, great job, thank you 🙇 ! 🚢

PR #2503, opened 2022-08-17
Use apiKey from the GatewayId enum for constructing the URL.

AnirudhBhat:
### NOTE: Please ensure to review [this](https://github.com/woocommerce/woocommerce-android/pull/7205) PR soon after merging this one.

This PR basically removes `gatewayId.toString()` and instead uses `gatewayId.apiKey` for constructing the payment gateway URL.

**Why?**
Because, using  `gatewayId.toString()` returns `CASH_ON_DELIVERY`. Whereas for payment gateway URL to work we need `cod` which is returned when `gatewayId.apiKey` is used.


rossanafmenezes (approved):
LGTM

PR #2509, opened 2022-08-18
Fetch list plugins by name

### LINKED ISSUE #2508: Add a function to retrieve the plugin list information
As part of the orders performance project, we need to add a new function to retrieve the information, if available, of a list of plugins, so we can be more accurate on whether to refresh the plugin data or not.

atorresveiga:
Closes #2508 
This PR adds a new function to retrieve the local plugin information from a list of plugin names. This is needed as part of the orders performance project to reduce the number of calls made in the orders detail screen.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PluginSqlUtils.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PluginSqlUtils.java
@@ -28,11 +28,11 @@ public class PluginSqlUtils {
     public static @NonNull
     List<SitePluginModel> getSitePlugins(@NonNull SiteModel site) {
         return WellSql.select(SitePluginModel.class)
-                      .where()
-                      .equals(SitePluginModelTable.LOCAL_SITE_ID, site.getId())
-                      .endWhere()
-                      .orderBy(SitePluginModelTable.DISPLAY_NAME, ORDER_ASCENDING)
-                      .getAsModel();
+            .where()
+            .equals(SitePluginModelTable.LOCAL_SITE_ID, site.getId())
+            .endWhere()
+            .orderBy(SitePluginModelTable.DISPLAY_NAME, ORDER_ASCENDING)
+            .getAsModel();
     }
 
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/PluginSqlUtils.java
```diff
@@ -25,13 +25,14 @@
...
-    public static @NonNull List<SitePluginModel> getSitePlugins(@NonNull SiteModel site) {
+    public static @NonNull
+    List<SitePluginModel> getSitePlugins(@NonNull SiteModel site) {
         return WellSql.select(SitePluginModel.class)
-                .where()
-                .equals(SitePluginModelTable.LOCAL_SITE_ID, site.getId())
-                .endWhere()
-                .orderBy(SitePluginModelTable.DISPLAY_NAME, ORDER_ASCENDING)
-                .getAsModel();
+                      .where()
```
wzieba: Can we please revert those code style changes? 🙏

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -161,6 +162,11 @@ open class WooCommerceStore @Inject constructor(
         return PluginSqlUtils.getSitePluginByName(site, plugin.pluginName)
     }
 
+    fun getSitePlugins(site: SiteModel, plugins: List<WooPlugin>): List<SitePluginModel> {
+        val pluginNames = plugins.map { it.pluginName }
+        return PluginSqlUtils.getSitePluginByNames(site, pluginNames)
+    }
```
wzieba: This will execute the database operation on the main thread. WDYT about wrapping this in a coroutine/suspend fun?

atorresveiga:
Good catch @wzieba! This is ready for another round

wzieba (approved):
Looks good 👍

PR #2511, opened 2022-08-24
Woo/issue/7141 add mocked tests

rossanafmenezes:
This PR introduces Mocked tests for the GatewayRestClient.kt functions


rossanafmenezes:
Hi, @malinajirka  and @AnirudhBhat I'm drafting the PR just so you can take a look at the code and I can get some feedback on it. 

I will add comments to the code changes. Thank you in advance

example/src/androidTest/resources/wc-pay-fetch-all-gateways-response-error.json
```diff
@@ -0,0 +1,7 @@
+{
```
rossanafmenezes, line 1: For example, here, since I'm using basically the same error json response for all 3 tests, do I need 3 separate files? Or am I doing this wrong and the errors should be indeed different?

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCGatewayTest.kt
```diff
@@ -0,0 +1,108 @@
...
+            title = "Pay on Delivery",
+            description = "Pay by cash or card on delivery"
+        )
+
+        assertTrue(result.isError)
+        assertEquals(API_ERROR, result.error.type)
+    }
+
+    @Test
+    fun whenFetchAllGatewaysSucceedsReturnSuccess() = runBlocking {
```
rossanafmenezes, line 100: This test is failing and I need to work on it, I will get back to it tomorrow. 

I think it's failing because it expects an array and I'm giving it an object in the json file 🤔

example/src/androidTest/resources/wc-pay-fetch-all-gateways-response-error.json
malinajirka: Good question, I personally don't feel like one approach is significantly better than the other. Use whichever feels better to you.

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCGatewayTest.kt
malinajirka: I've tried calling the endpoint and as you correctly pointed out, it returns an array of objects. So yep, we need to update the `json`.

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCGatewayTest.kt
```diff
@@ -0,0 +1,108 @@
...
+            title = "Pay on Delivery",
+            description = "Pay by cash or card on delivery"
+        )
+
+        Assert.assertFalse(result.isError)
+        assertTrue(result.result != null)
+    }
+
+    @Test
+    fun whenInvalidDataProvidedUpdateGatewayThenSErrorReturned() = runBlocking {
```
malinajirka: ❓ I'm wondering which of the provided data are invalid? Seems like all of these are valid:
```
 CASH_ON_DELIVERY,
enabled = true,
title = "Pay on Delivery",
description = "Pay by cash or card on delivery"
```



I've tried calling the endpoint with `"enabled": "hello"` and it returned 
400
```
{
"error": "rest_invalid_param",
"message": "Invalid parameter(s): enabled"
}
```

not the WCPay error as in the test. I think having test(s) for wcpay_server_error is also useful as this error might occur when there is a bug on the server and we need to expect bugs. However, I'd consider having one test which mentions `invalid data` and expects `rest_invalid_param` as that's what the server returns and a separate test for unexpected server error. Wdyt

example/src/androidTest/resources/wc-pay-fetch-all-gateways-response-success.json
```diff
@@ -0,0 +1,6 @@
+{
```
malinajirka, line 1: ⚠️ I've tried calling the endpoint and it never returned `"status":"succeeded"` (same comment applies to all three endpoints which we are testing) -> any reason you expect it or was it just a copy-pasting mistake?

Example response on my side looks like this:
Screenshot 2022-08-24 at 11 39 07

Have you tried using curl (or similar) or checking the documentation ([get one](https://woocommerce.github.io/woocommerce-rest-api-docs/#retrieve-an-payment-gateway), [get all](https://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-payment-gateways), [update](https://woocommerce.github.io/woocommerce-rest-api-docs/#update-a-payment-gateway)) to see what the server actually returns?

example/src/androidTest/resources/wc-pay-fetch-all-gateways-response-error.json
rossanafmenezes: 👍

rossanafmenezes: since they are test files and won't impact in the apk (right?) I will leave it as 3 separate files because I have created them already but will consider the other approach for a future mocked test

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCGatewayTest.kt
rossanafmenezes: 👍 Thank you. I will work on this

rossanafmenezes: Ok, I have added an array to the json file and the test works now 😄

example/src/androidTest/resources/wc-pay-fetch-all-gateways-response-success.json
rossanafmenezes: >  any reason you expect it or was it just a copy-pasting mistake?

the tests files were copied and pasted. I am still working on the actual json responses 😄

example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCGatewayTest.kt
rossanafmenezes: As per slack convo, I have renamed the current test to `whenUpdateGatewayFailesDueToUnexpectedServerErrorThenServerErrorReturned` and added a new test to simulate invalid data added.

malinajirka (commented):
LGTM, great job @rossanafmenezes!! :shipit:

malinajirka (approved):

PR #2517, opened 2022-08-29
Save shippingPhone value in OrderEntity

### LINKED ISSUE #2516: Woo: Shipping phone information is never saved
While working on Address validation, I note that the phone shipping information is never updated/saved and always displays an empty value.
Expected behavior: The shipping phone information is displayed correctly and can be updated.

atorresveiga:
Closes #2516 
This small PR adds the shipping phone information into OrderEntity. Before this change, the shipping phone information was never saved, causing the shipping phone value always to be an empty string


ThomazFB (approved):
LGTM! :shipit:

PR #2519, opened 2022-08-30
Add product path to product leather board request when available

### LINKED ISSUE #6688: Support `/wc-analytics/leaderboards/products` endpoint
In https://github.com/woocommerce/woocommerce/pull/33185 we're adding `/wc-analytics/leaderboards/products` (along with individual endpoints for the other leaderboards. Since we only need the product leaderboard in the apps, we should switch to that new endpoint. This will be faster, lessen the load on the origin site, and transfer less data that we don't need.

These new endpoints should be in WC 6.7, so we'll need to check that they exist and fallback to the full `/wc-analytics/leaderboards` endpoint. The data is in pretty much the same format to make parsing the response easy.

JorgeMucientes:
Closes: [#6688 ](https://github.com/woocommerce/woocommerce-android/issues/6688)

### Description
`/wc-analytics/leaderboards/products` Was added along with individual endpoints for the other leaderboards. Since we only need the product leaderboard in the apps, we should switch to that new endpoint. This will be faster, lessen the load on the origin site, and transfer less data that we don't need.

In this PR I've added the minimal changes to support using the new endpoint easily from WC android repo. Changes in `wc-android` have been applied [here](https://github.com/woocommerce/woocommerce-android/pull/7308). 

### Testing instructions
This PR adds no breaking change. But to make sure that the request `/leaderboards` keeps working as expected, inside FluxC sample app: 
- Go to Woo -> Leaderboards
- Select any site that has product purchases
- Fetch leaderboards for different granularities and verify request is successful.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardsRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardsRestClient.kt
@@ -30,10 +30,12 @@ class LeaderboardsRestClient @Inject constructor(
         site: SiteModel,
         unit: StatsGranularity?,
         queryTimeRange: LongRange?,
-        quantity: Int?
-    ) = WOOCOMMERCE.leaderboards.pathV4Analytics
-            .requestTo(site, unit, queryTimeRange, quantity)
-            .handleResult()
+        quantity: Int?,
+        addProductsPath: Boolean = false
+    ) = when (addProductsPath) {
+        true -> WOOCOMMERCE.leaderboards.products.pathV4Analytics
+        else -> WOOCOMMERCE.leaderboards.pathV4Analytics
+    }.requestTo(site, unit, queryTimeRange, quantity).handleResult()
 
     private suspend fun String.requestTo(
         site: SiteModel,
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardsRestClient.kt
```diff
@@ -30,22 +30,24 @@ class LeaderboardsRestClient @Inject constructor(
...
-        quantity: Int?
-    ) = WOOCOMMERCE.leaderboards.pathV4Analytics
-            .requestTo(site, unit, queryTimeRange, quantity)
-            .handleResult()
+        quantity: Int?,
+        addProductsPath: Boolean = false
+    ) = when (addProductsPath) {
+        true -> WOOCOMMERCE.leaderboards.products.pathV4Analytics
+        else -> WOOCOMMERCE.leaderboards.pathV4Analytics
+    }.requestTo(site, unit, queryTimeRange, quantity).handleResult()
```
JorgeMucientes, line 38: Adding a new optional param `addProductsPath` that would determine if we should use the new `/leaderboards/products` endpoint or keep using the older/slower one `/leaderboards`

hafizrahman (approved):
"leather board" == "leaderboard"?

But otherwise this looks good to me :)

JorgeMucientes:
> "leather board" == "leaderboard"?

It's just a typo.  Luckily, I think I only applied that typo in the PR title and description and not in the code 😅 

Thanks for the review

PR #2512, opened 2022-08-24
Feature/7254 reduce infinte loading

JorgeMucientes:
Closes: #7254 

### Description
We want to mitigate the "endless loading" experience for users overall the app. We're going to start by:
- Reducing for all Jetpack tunnel requests the retry policy from 2 to 1
- Reducing the timeout time from 30s to 15s. Currently this time reduction will only apply to site verification request.

### Testing instructions
In order to test this WooCommerce app pointing to this FluxC version is needed. So I'll request testing of these changes in [this PR](https://github.com/woocommerce/woocommerce-android/pull/7271). Nothing to test here. 

Connected tests are failing due to unrelated reasons https://a8c.slack.com/archives/C02QANACA/p1661424977107179

PR #2520, opened 2022-08-31
New overrideRetryPolicy param to enable A/B testing

JorgeMucientes:
A/B test for : [#7254](https://github.com/woocommerce/woocommerce-android/issues/7254)

### Description

Adds new optional parameter so that we can use this in A/B tests in WooCommerce Android. Once A/B test is finished parameter will be removed. 

If the value is true we will apply the new `jetpackTunnelGsonRequestBuilder.buildDefaultTimeoutRetryPolicy()` that I added in [this PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2512#issue-1349299146) that basically reduces the default tiemout from 30s to 15s. 

How the A/B test is set in WooCommerce Android can be reviewed and [tested here. ](https://github.com/woocommerce/woocommerce-android/pull/7313)

### Testing

Nothing to test from this changes


hichamboushaba (approved):
Looks good, thanks @JorgeMucientes

PR #2521, opened 2022-09-02
Feature/6719 add support force refresh

JorgeMucientes:
Closes:[ #6719](https://github.com/woocommerce/woocommerce-android/issues/6719)

This FluxC changes work along with [this small changes](https://github.com/woocommerce/woocommerce-android/pull/7335) in WooCommerce-Android.

### Context 

https://github.com/woocommerce/woocommerce/pull/33325 will add a new parameter to `wc-analytics` API endpoints, `force_cache_refresh`, which will cause the current request to bypass the cache, re-run the queries for the requested data, and overwrite the previous cache entry with the new results. The motivation behind this change is to make it possible to refresh cached data from the apps without having to resort to [cache-busting parameter hacks](https://github.com/woocommerce/woocommerce-ios/issues/1183).

In Android realized, we are not applying that hacky workaround to bypass the cache[ like iOS does](https://github.com/woocommerce/woocommerce-ios/issues/1183). That means that stats are not refreshed in "real time" after new orders come in, even if you swipe to refresh multiple times. It will take a couple mins or less to refresh data. Don't really know what are the conditions to get fresh data, but its a bit random from what I've seen. 

We do have a `force` parameter that we sometimes pass all the way through to the Jetpack tunnel request, but from my tests this parameter is ignored for the `wc-analytics` stats requests. 

### Description 

This PR  adds the new parameter `force_cache_refresh` to two API requests: revenue stats and product leaderboards. The parameter doesn't make a difference right now, but will be necessary when the cache changes are implemented in a future WC version.

Minor: I also renamed the parameters from `force` to `forceRefresh` as I though better explains what the value is for

### Testing instructions

We can't verify that the new parameter works for now, as it hasn't been released yet. The only thing we can do is tests [in this PR](https://github.com/woocommerce/woocommerce-android/pull/7335) that stats for "My Store Screen" are loaded correctly for all 4 granularities.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCStatsTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCStatsTest.kt
@@ -485,9 +485,9 @@ class MockedStack_WCStatsTest : MockedStack_Base() {
         // Make the same stats request, but this time pass force=true to force a network request
         interceptor.respondWith("wc-revenue-stats-response-success.json")
         orderStatsRestClient.fetchRevenueStats(
-                site = siteModel, granularity = StatsGranularity.DAYS,
-                startDate = "2019-07-01T00:00:00", endDate = "2019-07-07T23:59:59",
-                perPage = 35, force = true
+            site = siteModel, granularity = StatsGranularity.DAYS,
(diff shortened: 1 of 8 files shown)


hafizrahman (approved):
This LGTM since the many lines of changes essentially are just parameter updating, and formatting changes. Good point about renaming it!

PR #2525, opened 2022-09-12
Assign correct causeOfChange when dispatching signout event

hichamboushaba:
When dispatching the event `OnAccountChanged` after signing out, we don't set the correct causeOfChange, and this caused recently an issue in WCAndroid.


nbradbury (approved):
Looks good :shipit:

PR #2526, opened 2022-09-17
Feature/6619 limit stats refresh

JorgeMucientes:
Part of: [#6618 ](https://github.com/woocommerce/woocommerce-android/issues/6618)

### Description
`wc-analytics` endpoint is one of the slowest endpoints we hit from the app, and is one that we hit a lot. We fetch site stats and top performers every time we open My Store tab and when we change between tabs. This translates into the "slow/endless loading" perception of the app. This PR alongside changes applied in this other [PR in woocommerce-android](https://github.com/woocommerce/woocommerce-android/pull/7411) aim to reduce that loading "slowness" for **top performer products data**  with the following changes: 

- Migrate top performers cache to Room to easily expose an observable
- Simplify top performers cached model to avoid having to parse multiple JSON or query different SQL tables to load top performers from cache
- Invalidate top performers cache when a new order is placed 
- Invalidate top performers cache after 1h. Maybe this is to often? please share your thoughts

Finally, I update all broken unit tests, remove some that did not apply anymore and added a couple more. 

### Testing instructions
This can be tested alongside the changes from WooCommerce Android. Please follow the test instructions from [this PR](https://github.com/woocommerce/woocommerce-android/pull/7411).


example/src/test/java/org/wordpress/android/fluxc/wc/leaderboards/WCLeaderboardsStoreTest.kt
```diff
@@ -20,142 +20,223 @@ import org.wordpress.android.fluxc.model.WCProductModel
...
-    private lateinit var storeUnderTest: WCLeaderboardsStore
-    private lateinit var restClient: LeaderboardsRestClient
-    private lateinit var productStore: WCProductStore
-    private lateinit var mapper: WCProductLeaderboardsMapper
+    private val restClient: LeaderboardsRestClient = mock()
+    private val productStore: WCProductStore = mock()
+    private var mapper: WCProductLeaderboardsMapper = mock()
+    private val topPerformersDao: TopPerformerProductsDao = mock()
+
+    private var storeUnderTest = WCLeaderboardsStore(
```
hichamboushaba: very np 😄, I noticed some tests call `createStoreUnderTest`, and some don't.

This is a little confusing, I think the reason is that sometimes we want to setup the mocks before creating the instance.

A small suggestion, but you don't have to implement it: remove the @Before from setup, and add a `mocks` builder argument to it, then call it always, and make sure it calls `createStoreUnderTest` (similar to [this](https://github.com/woocommerce/woocommerce-android/blob/d03d01894fdb1eb1cfb336664ed4f4ab3724f3f5/WooCommerce/src/test/kotlin/com/woocommerce/android/push/UnseenReviewsCountHandlerTests.kt#L22-L29) from the WC app), this would make things consistent, always start with `setup`, but without thinking about a second instantiation of `storeUnderTest`

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/TopPerformerProductsDao.kt
```diff
@@ -0,0 +1,65 @@
...
+    @Transaction
+    @Query("SELECT * FROM TopPerformerProducts WHERE granularity = :granularity AND siteId = :siteId")
+    abstract fun observeTopPerformerProducts(
+        siteId: Long,
+        granularity: String
+    ): Flow<List<TopPerformerProductEntity>>
+
+    @Transaction
+    @Query("SELECT * FROM TopPerformerProducts WHERE granularity = :granularity AND siteId = :siteId")
+    abstract fun getTopPerformerProductsFor(
```
hichamboushaba: Shouldn't this be a suspendable function to make sure it's using the correct background thread?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/TopPerformerProductsDao.kt
```diff
@@ -0,0 +1,65 @@
...
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import androidx.room.Transaction
+import kotlinx.coroutines.flow.Flow
+import org.wordpress.android.fluxc.persistence.entity.TopPerformerProductEntity
+
+@Dao
+abstract class TopPerformerProductsDao {
+    @Transaction
```
hichamboushaba: Is there a reason to mark this and the following functions as transactions? 🙂

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/TopPerformerProductsDao.kt
```diff
@@ -0,0 +1,65 @@
...
+    @Query("SELECT * FROM TopPerformerProducts WHERE granularity = :granularity AND siteId = :siteId")
+    abstract fun getTopPerformerProductsFor(
+        siteId: Long,
+        granularity: String
+    ): List<TopPerformerProductEntity>
+
+    @Transaction
+    @Query("SELECT * FROM TopPerformerProducts WHERE siteId = :siteId")
+    abstract fun getTopPerformerProductsForSite(
```
hichamboushaba: Same as above.

hichamboushaba (commented):
Thanks @JorgeMucientes for working on this, I left some questions about blocking functions that need to be answered before approving this.

In the meantime I'll test the functionality in WCAndroid.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/TopPerformerProductsDao.kt
JorgeMucientes: Yes, good catch!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/TopPerformerProductsDao.kt
JorgeMucientes: Not in these cases. Removed, good catch!
Those transactions are leftovers when my initial approach was to have 2 different tables for top performers' data, one for aggregated top performer data and one other for the product details. Similar to the approach we had prior to the refactor, although product details were stores as json string. Then I noticed I didn't need any of that.

example/src/test/java/org/wordpress/android/fluxc/wc/leaderboards/WCLeaderboardsStoreTest.kt
JorgeMucientes: Nice approach! I like it. I've applied it as I wanted to see it in action. In the end for this specific case it ended up being a bit more verbose, as I had to add an additional `setup()` call to all tests, when only one test needed to override the current instance of `storeUnderTest`

But I'll keep the changes as I think is consistent and I think is a cleaner way to defer SUT creation to override mocks.

Thanks for the tip 🙏🏼

hichamboushaba (approved):
Looks good, thanks for the fixes @JorgeMucientes.

I'll leave merging this to you, to be able to coordinate between the two PRs.

PR #2533, opened 2022-09-29
Updates the target sdk version to 31 for Android 12 migration 

### LINKED ISSUE #2529: Update `targetSdkVersion` to 31
Parent #https://github.com/wordpress-mobile/WordPress-Android/issues/16062

As part of the migration of WP Android to Android 12, the `targetSdkVersion` on the dependent libraries can be updated to 31. [WooCommerce](https://github.com/woocommerce/woocommerce-android) is already using `targetSdkVersion` = 31, This change wouldn't affect the app( [WooCommerce](https://github.com/woocommerce/woocommerce-android),[Wordpress Android](https://github.com/wordpress-mobile/WordPress-Android)) since the `targetSdkVersion` on the app has a [higher priority and the manifest merge wouldn't make any issues](https://stackoverflow.com/questions/35465502/should-android-libraries-define-targetsdkversion). The [Wordpress Android](https://github.com/wordpress-mobile/WordPress-Android) had a `targetSdkVersion` of 30.  now has been updated to 31 in this PR - https://github.com/wordpress-mobile/WordPress-Android/pull/17153. So in order to keep the `targetSdkVersion` to be consistent with the apps this change needs to be taken up.

AjeshRPai:
Part of #2529 

This PR updates the target SDK version to 31 for the Android 12 migration 

Test Instructions 
- Install the app from the Parent PR 
- Make sure the functionalities provided by the library work as expected

### DIFF
--- a/fluxc/build.gradle
+++ b/fluxc/build.gradle
@@ -13,7 +13,7 @@ android {
 
     defaultConfig {
         minSdkVersion 21
-        targetSdkVersion 30
+        targetSdkVersion 31
         javaCompileOptions {
             annotationProcessorOptions {
                 arguments += [
--- a/settings.gradle
+++ b/settings.gradle
@@ -50,7 +50,7 @@ dependencyResolutionManagement {
             version("facebook-soloader", "0.9.0")
             version("kotlinx-coroutines", "1.3.9")
             version("mockito", "3.3.3")
-            version("wordpress-utils", "develop-13b4f06680ed8c145cbd09d63c4998dfe8b3b9de")
+            version("wordpress-utils", "develop-eebc5d8e91a1d90190919f900f924b39c861a528")
         }
     }
 }
--- a/settings.gradle
+++ b/settings.gradle
@@ -50,7 +50,7 @@ dependencyResolutionManagement {
             version("facebook-soloader", "0.9.0")
             version("kotlinx-coroutines", "1.3.9")
             version("mockito", "3.3.3")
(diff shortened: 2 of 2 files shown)


fluxc/build.gradle
```diff
@@ -13,7 +13,7 @@ android {
 
     defaultConfig {
         minSdkVersion 21
-        targetSdkVersion 30
+        targetSdkVersion 31
```
ParaskP7, line 16: 👋 @AjeshRPai !

Maybe we can take this opportunity, in this PR, or, most probably, in a separate PR afterwards (as to not block this one) to do the following as well:
- Reaching out to the WCAndroid engineers and ask them about updating the `targetSdkVersion` for the `plugins/woocommerece` lib module.
- Try to update the `targetSdkVersion` on our `example` app module as well, I am seeing it being `22` and I wonder if we can easily update it to `31`, which will give us another additional safety net when testing such upgrades in the future.

Wdyt? 🤔

AjeshRPai: @ParaskP7
I agree. As discussed. I will add it to my to-do and take it up in a separate PR since I don't want to block this PR. 
Thanks for pointing it out.

ParaskP7: Awesome, I agree, thank you so much Ajesh! 🙇

ParaskP7 (approved):
LGTM! 💯 

As per the testing that was done as part of [WordPress-Android#17153](https://github.com/wordpress-mobile/WordPress-Android/pull/17153), which verifies correctness of this library working alongside the WPAndroid client app, I am approving this PR too.

PR #2537, opened 2022-10-14
Patch cookiemanager

hichamboushaba:
This PR implements the 4th solution from the suggested solutions in paqN3M-K4-p2, this allows solving the Android [bug](https://issuetracker.google.com/issues/174647435), while keeping the changes transparent to all of the clients.

#### Testing
To be safe, I took the testing cases that the OpenJDK have for this class, and applied them to our class too.

**As a non-regression test:**
1. Open the example app.
2. Click on Sign in.
3. Use site credentials with XMLRPC url.
4. Click on React Native.
5. Click on Execute Request.
6. Confirm that "Request succeeded" is logged to the console.

As the "React Native" APIs use the WPOrg API, this test confirms that we are still saving the Cookies we need.

**Optional**
And to confirm that the bug is solved, you can check the test steps of the Woo [PR](https://github.com/woocommerce/woocommerce-android/pull/7545) that depends on it.


hichamboushaba:
@ParaskP7 I requested your review too on this PR, just as someone who has visibility on the different apps, if you are busy, feel free to ignore it.

example/src/test/java/org/wordpress/android/fluxc/network/CookieManagerTest.kt
```diff
@@ -0,0 +1,223 @@
...
+import java.net.CookieManager
+import java.net.CookiePolicy
+import java.net.URI
+import java.util.LinkedList
+
+/**
+ * The test cases here are copied from OpenJdk's CookieManagerTest:
+ * https://github.com/openjdk/jdk/blob/20db7800a657b311eeac504a2bbae4adbc209dbf/test/jdk/java/net/CookieHandler/CookieManagerTest.java
+ */
+class CookieManagerTest {
```
hichamboushaba: I'm sure there are some optimizations that we can follow for a more idiomatic Kotlin in this class, but I didn't want to spend too much time on it, and I think it's better to stay close to the way the source code is written in Java.

0nko (approved):
The test works as described 👍. 

I glanced over the code but I didn't spend too much time on it, since it's just copied over.

ParaskP7:
> @ParaskP7 I requested your review too on this PR, just as someone who has visibility on the different apps, if you are busy, feel free to ignore it.

Thank you for the ping @hichamboushaba ! 🙇 

FYI: I'll also review this PR today, most probably in a couple of hours. 👍

example/src/test/java/org/wordpress/android/fluxc/network/CookieManagerTest.kt
ParaskP7: 👋 @hichamboushaba !

> ...I think it's better to stay close to the way the source code is written in Java.

💯 for that and to top it up, I think it might be even wiser for us to not even deal with the Kotlin conversion, not at all. As such, I would recommend that you consider reverting this Kotlin conversion and have the `PatchedCookieManager` as a plain old Java.

In addition to the above, and after me diffing the `Android` related `CookieManager` vs. the `OpenJdk` related `CookieManager` I am seeing that the diffing goes beyond the `put` public methods that you overrode in the patch class.

> ...all I had to do was to inherit from Android’s java.net.CookieManager, then override the put function using OpenJdk‘s implementation, and this allowed getting all of the cookies.

I understand that you did that to minimize the change, but what about the other changes in that version of `CookieManager` from `OpenJdk`, maybe those need to work together with the `put` change? 🤔

What I am trying to say is, what if we copy instead this `OpenJdk` version of `CookieManager`, calling it `OpenJdkCookieManager`, just as is and use that instead of the `PatchedCookieManager`? Then, this fix will become more straightforward, wouldn't it? We will be just using another version of `CookieManager` to overcome [this bug](https://issuetracker.google.com/issues/174647435).

Another thing that makes me think about having to extend from `CookieManager` and then override just the `put` method is what would happen if on a subsequent update of Android, this `CookieManager` gets updated, maybe even by having this bug fixed on Google's side, but by also updating their internal mechanism. Does this mean that us overriding the `put` method we would be reverting this work, even causing some incompatibility issues, thus us needing to keep an eye on that from now onwards? 🤔

Apologies if I am misunderstand something here, I read your P2 post multiple times and thought about all 4 options you provided. I am just trying to understand the situation we are in better in order to be helpful to you.

example/src/test/java/org/wordpress/android/fluxc/network/CookieManagerTest.kt
```diff
@@ -0,0 +1,223 @@
+package org.wordpress.android.fluxc.network
```
ParaskP7: Suggestion (💡): Consider renaming that to `PatchedCookieManagerTest` so that to associate that test with the newly created `PatchedCookieManager` class.

PS: I understand why you kept the `CookieManagerTest` to match that on the `OpenJDK` side. As such, maybe we should first deal with [this comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2537/files#r998058894) of mine first, and only then come back to this.

example/src/test/java/org/wordpress/android/fluxc/network/CookieManagerTest.kt
hichamboushaba: Thanks for the feedback @ParaskP7 🙏 

All good points, and I agree with you in the fact that taking the whole implementation as a Java class is simpler. Still, I'm not sure they will be straight forward to implement, as the version from OpenJdk is using API that's not available in Java 8, so if we want to import it as a whole, we might need to make several changes to get it to work with Android.
 
Another point, the `put` and `get` functions don't depend on each other, so having one implementation impact the other one is unlikely, and this is because the `CookieManager` uses `CookieStore` behind the scenes, and the `put` just parses then converts the cookies from the header to `HttpCookie`, while the `get` function just reads the list of `HttpCookies` from the store then converts them into a header.
My idea here, is that since we are not touching the `CookieStore` implementation, updating the logic of broken `put` is safe from any potential incompatibilities. 

I'll test importing the whole class, and see how much changes are needed, and if they are manageable, I'll implement it, I'll let you know how it goes.

ParaskP7: Thanks for your reply @hichamboushaba ! 🙇 

> ...as the version from OpenJdk is using API that's not available in Java 8...

Ah shot, I wasn't aware of that, is this why you converted the patched class to Kotlin in the first place? 🤔

> Another point, the put and get functions don't depend on each other, so having one implementation impact the other one is unlikely, and this is because the CookieManager uses CookieStore behind the scenes, and the put just parses then converts the cookies from the header to HttpCookie, while the get function just reads the list of HttpCookies from the store then converts them into a header.

👍 

> My idea here, is that since we are not touching the CookieStore implementation, updating the logic of broken put is safe from any potential incompatibilities.

👍 🤞 

> I'll test importing the whole class, and see how much changes are needed, and if they are manageable, I'll implement it, I'll let you know how it goes.

As I see you already thoughts about what I commented, please don't feel obliged to do this work, I really wasn't aware of the Java 8 obstacle with Android and this `OpenJDK` version of `CookieManager`.

ParaskP7: Btw, I tested the existing solution, as per your test instructions on this PR, and everything worked as expected. Though, FYI, I didn't go through the optional testing with WCAndroid. As such, please don't feel you are being blocked by me.

PS: I also believe that this change is straight forward when it comes to the WPAndroid side, as per the thoughts you shared on the 4th solution, it should act as an improvement there too.

hichamboushaba:
@ParaskP7 I applied your suggestions, I imported the complete implementation from OpenJdk, and updated it to be compatible with older APIs.
One significant change that I had to is: inheriting from `CookieManager` instead of `CookieHandler`, I had to do this because `InMemoryCookieStore` is hidden in the Android API, and it's needed for the implementation, so by using the default `CookieManager` as a base class, we can access it using `getCookieStore`.
I left the test class as a Kotlin class, it would take me some time to do the test in Java, and I don't think it's worth the effort.

Please take a look when you have time, and let me know what you think.

ParaskP7:
👋 @hichamboushaba !

> @ParaskP7 I applied your suggestions, I imported the complete implementation from OpenJdk, and updated it to be compatible with older APIs.

Thank you! 🥇 

> One significant change that I had to is: inheriting from CookieManager instead of CookieHandler, I had to do this because InMemoryCookieStore is hidden in the Android API, and it's needed for the implementation, so by using the default CookieManager as a base class, we can access it using getCookieStore.

Got it! 👍 

> I left the test class as a Kotlin class, it would take me some time to do the test in Java, and I don't think it's worth the effort.

Agreed! 💯 

> Please take a look when you have time, and let me know what you think.

I'll take a look at your changes first thing tomorrow morning, I hope I am not blocking you too much on your overall progress, apologies if I am in advance if I do... 🙏

ParaskP7:
👋 @hichamboushaba !

I did another review on the new `OpenJdkCookieManager` custom-ish implementation and seeing the result, especially due to the `InMemoryCookieStore` not being available limitation, but also some other changes like:
- `CookiePathComparator` vs. `CookieComparator`
- `Integer.parseInt(lst, 0, i, 10)` vs. `Integer.parseInt(lst.substring(0, i))`
- `if (cookie.isHttpOnly())` vs. `if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && cookie.isHttpOnly())`
- ...plus other minor changes like `Map.of()`

I now feel that it might not have been the best idea to go for copy-pasting the `CookieManager` implementation from `OpenJDK`, simply because it is not possible.

Apologies from driving you back and forward with that, I was honestly just trying to help and provide as much feedback as I could. As such, comparing this `OpenJdkCookieManager` vs. your previous `PatchedCookieManager` implementation, I feel I am more comfortable with the previous one, especially since this was also tested thoroughly alongside the WCAndroid client.

Let me know your thoughts? 🙏

hichamboushaba:
Thanks @ParaskP7 for the review, personally I still find having the whole `OpenJdkCookieManager` more consistent 😅.

Just for reference (I apologize for not clarifying this before), all the changes you mentioned are done by going through the history of the `CookieManager` class in the `OpenJdk` repo, so we are still using an OpenJdk implementation, just adapted for older APIs:
1. CookiePathComparator vs. CookieComparator -> the conversion was done by reverting https://github.com/openjdk/jdk/commit/b5fb6b3566f67ffd079bb88e178f5eace1443ed7, also the implementation is similar to what the Android JVM has https://cs.android.com/android/platform/superproject/+/master:libcore/ojluni/src/main/java/java/net/CookieManager.java;l=466?q=CookieManager
2. Integer.parseInt(lst, 0, i, 10) vs. Integer.parseInt(lst.substring(0, i)) -> reverts https://github.com/openjdk/jdk/commit/579e48fd608c6ffc2ed7a9b6c65061320bc7bc55
3. if (cookie.isHttpOnly()) vs. if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && cookie.isHttpOnly()) -> I think this one is straight forward

I can revert to the original implementation, as I don't have a strong opinion on either approach now 🙂, but I just wanted to clarify my approach and get your opinion afterwards 🙏.

ParaskP7:
👋 @hichamboushaba !

> Thanks @ParaskP7 for the review, personally I still find having the whole OpenJdkCookieManager more consistent 😅.

😅 👍 

> Just for reference (I apologize for not clarifying this before), all the changes you mentioned are done by going through the history of the CookieManager class in the OpenJdk repo, so we are still using an OpenJdk implementation, just adapted for older APIs

Ah, then this is great that you did that. This now makes me feel more confident in those changes I saw! 🥇 

> I can revert to the original implementation, as I don't have a strong opinion on either approach now 🙂, but I just wanted to clarify my approach and get your opinion afterwards 🙏.

As I here to just provide you with a second option, and given the fact that you did go through the history of the `CookieManager` and adapt to the old API, I would rather you choose which implementation to pick... But yea, maybe going for `OpenJdkCookieManager` is the better choice, you convinced me Sir... 😃 😊 😅

hichamboushaba:
Thanks for the confirmation @ParaskP7.
Additional point, I retested the new implementation with WCAndroid before pushing the changes, it works well.

I'm going to merge the PR now.

ParaskP7:
> Additional point, I retested the new implementation with WCAndroid before pushing the changes, it works well.

Awesome, thank you @hichamboushaba ! 🥇 🙇 🌟 

> I'm going to merge the PR now.

❤️ 🚀 🟢

PR #2538, opened 2022-10-16
Jetpack Connection: register site if needed

hichamboushaba:
Please don't merge until #2537 is merged.

This PR adds support for fetching the Jetpack connection URL even if the site doesn't have any connected accounts (we call it a blog without site level connection).
The flow is simple:
1. We call the API `/jetpack/v4/connection/url`
2. We check if the returned URL is pointing to `jetpack.wordpress.com`.
    i. if it does, then the site have a site-level connection, and we can continue as before.
    ii. if it doesn't, then this means the site is not registered yet, so we call the URL we received using the `no-redirect` OkHttp client, and then we catch the `Location` header from the redirection response. Then we return this location

Check the parent [issue](https://github.com/woocommerce/woocommerce-android/issues/7525) for more details, it explains the two possible approaches to solve this, I went with the second approach because it makes the WordPress.com authentication on the WebView easier.

To test, please check the Woo [PR](https://github.com/woocommerce/woocommerce-android/pull/7545).

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/RawRequest.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/RawRequest.kt
@@ -5,18 +5,25 @@ import com.android.volley.Response
 import com.android.volley.Response.Listener
 import com.android.volley.toolbox.HttpHeaderParser
 
+/**
+ * A request that allows returning the network response untouched, this is useful when
+ * we want to read the network response headers.
+ */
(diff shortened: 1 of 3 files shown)


JorgeMucientes (approved):

JorgeMucientes:
Great job @hichamboushaba I've tested it following instructions from the [PR](https://github.com/woocommerce/woocommerce-android/pull/7545/files) in WC Android and works great.

Issue #2536, opened 2022-10-10
Add support for the batch orders endpoint


atorresveiga:
Add support for the batch orders endpoint
https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#batch-update-orders
This endpoint is required for the Bulk update orders project because it will make it possible to update/delete a list of orders

For simplicity the order JSON representation is substituted for `{ "id" : _the id_ ... }`

The request will take the following form:
```
{
  "create" : [ { "id" : 745 ... } ] ,
  "update" : [ { "id" : 771 , "status" : "on-hold" }  ],
  "delete" : [ 723 ] ,
}
```

The response will take the following form:
```
{
  "data" : {
    "create"  : [ { "id" : 745 ... } ]
    "update" : [ { "id" : 771 ... } ] ,
    "delete" : [ { "id" : 723 ... } ] 
  }
}
```

![image](https://user-images.githubusercontent.com/18119390/194938351-f743c4b1-12ac-4a1b-82e8-08137c3cd4bb.png)

⚠️ Please note that historically in FluxC, the event bus was used to manage the communication between different components. With the advent of coroutines, we've been slowly migrating the requests to make use of their structured concurrency. For this task, favor the coroutines approach.

PR #2539, opened 2022-10-17
[Issue-2536] Add support to orders batch update

rodrigorods:
Closes [#2536](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/2536)
Initial P2 discussion [here](https://woomobilep2.wordpress.com/2022/10/14/trial-task-batch-orders-support/) 

This PR enables user to perform multiple actions, Create, Update and/or Delete Orders, to be performed on same request, to a maximum of 100.

Done - 
1. Error handling. ✅ [commit](b1bb2bd218d52926084b0263b8697033dbf50866)
2. UI of test app that allow creation of several Orders and Updates ✅ [commit](da52b97e2e116cad7ae1e5bde2945e1db72a616d)

First fetched orders | Fetching orders after batch update
--- | --- |
Screenshot_20221017_003203 | Screenshot_20221017_003250


example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
```diff
@@ -597,6 +597,88 @@ class WooOrdersFragment : StoreSelectingFragment(), WCAddOrderShipmentTrackingDi
                 }
             }
         }
+
+        update_order_batch.setOnClickListener {
```
atorresveiga, line 570: Nice! Great work adding the update order batch request to the FluxC example

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
```diff
@@ -597,6 +597,88 @@ class WooOrdersFragment : StoreSelectingFragment(), WCAddOrderShipmentTrackingDi
...
+                    val products = showSingleLineDialog(
+                        activity = requireActivity(),
+                        message = "Please type a comma separated list of product IDs",
+                        isNumeric = false
+                    )?.split(",")?.map { it.toLongOrNull() }
+
+                    if (products == null || products.any { it == null }) {
+                        prependToLog("Error while parsing the entered product IDs")
+                        return@launch
+                    }
```
atorresveiga: I noted that here the `create` request is mandatory. Taking into account that `create`, `update` and `delete` requests are optional in the order batch request. Does it make any sense to make this request also optional? This way, it will be easy to test only one of them.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
```diff
@@ -597,6 +597,88 @@ class WooOrdersFragment : StoreSelectingFragment(), WCAddOrderShipmentTrackingDi
...
+                        shippingAddress = shippingAddress,
+                        billingAddress = billingAddress,
+                        customerNote = customerNote
+                    )
+
+                    val result = orderUpdateStore.updateOrdersBatch(
+                        site,
+                        createRequest = mutableListOf<UpdateOrderRequest>().apply {
+                            repeat(duplicatesCount) { this.add(orderToCreate) }
+                        },
```
atorresveiga: np: I understand that here your goal is to create a list initialized with the same `orderToCreate` object `duplicatesCount` times. I feel that using the list constructor would be better suited for your goal here. 
Something like :
```List(duplicatesCount) { orderToCreate } ```.  WDYT?

Is the `mutableListOf` function used instead of the `listOf` function for any particular reason here?

File: plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClient.kt [hunk 4]
```diff
@@ -815,6 +862,13 @@ class OrderRestClient @Inject constructor(
         }
     }
 
+    private fun buildOrderEntity(
```
atorresveiga, line 865: Nice!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrdersBatchDto.kt
```diff
@@ -0,0 +1,29 @@
...
+
+object OrdersBatchUpdateRequest {
+    fun buildBody(
+        createRequest: List<Map<String, Any>>,
+        updateRequest: List<Map<String, Any>>,
+        deleteRequest: List<Long>
+    ) = mutableMapOf<String, Any>().apply {
+        put("create", createRequest)
+        put("update", updateRequest)
+        put("delete", deleteRequest)
```
atorresveiga: Maybe the request could be removed from the body when we know the list is empty so we ensure that only necessary data is sent. WDYT?

Is this map mutable for any particular reason?

Maybe the [buildMap](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/build-map.html) function could be handy here.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -231,6 +232,41 @@ class OrderUpdateStore @Inject internal constructor(
         }
     }
 
+    suspend fun updateOrdersBatch(
+        site: SiteModel,
+        createRequest: List<UpdateOrderRequest> = emptyList(),
+        updateRequest: List<UpdateOrderRequest> = emptyList(),
+        deleteRequest: List<Long> = emptyList()
```
atorresveiga, line 239: I'm curious about your decision-making process here. Why did you favor the default to `emptyList()` approach here rather than the default  to `null` one?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -231,6 +232,41 @@ class OrderUpdateStore @Inject internal constructor(
...
+
+                if (createRequest.isNotEmpty()) {
+                    model.createdEntities.map { ordersDao.insertOrUpdateOrder(it) }
+                    requestListRefresh(site)
+                }
+                if (updateRequest.isNotEmpty()) {
+                    model.updatedEntities.map { ordersDao.insertOrUpdateOrder(it) }
+                }
+                if (deleteRequest.isNotEmpty()) {
+                    deleteRequest.map { ordersDao.deleteOrder(site.localId(), it) }
```
atorresveiga: Shouldn't the orders be deleted only if they are present in the response (model.deletedEntities)?

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClientTest.kt
```diff
@@ -0,0 +1,145 @@
...
+    @Test
+    fun `updateOrdersBatch should call gson builder request with expected params`(): Unit = runBlocking {
+        val createRequest = listOf(
+            buildOrderRequest("create1"), buildOrderRequest("create2")
+        )
+        val updateRequest = listOf(
+            buildOrderRequest("update1"), buildOrderRequest("update2")
+        )
+        val deleteRequest = listOf(
+            1.toLong(), 2.toLong(), 3.toLong()
```
atorresveiga: np: A long value could be represented in Kotlin as `1L`  instead of `1.toLong()`. Using the `1L` representation IMO is short and more Kotlin-ish. Do you agree with me?

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/store/OrderUpdateStoreTest.kt
```diff
@@ -520,6 +522,137 @@ class OrderUpdateStoreTest {
...
+    @Test
+    fun `should delete local copy of order when bulky delete request succeeds`(): Unit = runBlocking {
+        setUp {
+            orderRestClient = mock {
+                onBlocking {
+                    updateOrdersBatch(
+                        site = site,
+                        createRequest = emptyList(),
+                        updateRequest = emptyList(),
+                        deleteRequest = listOf(1.toLong(), 2.toLong())
```
atorresveiga: [Same comments as here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2539/files#r998499561)

Would not be better to maintain the same ids in the requests and the responses? I see that the function is being called with ids `1` and `2`, but in the mock response, the ids are set as `123` and `456`

Maybe we can extract the list of order ids to a value and use it on the mock response when calling the function. WDYT?

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/store/OrderUpdateStoreTest.kt
```diff
@@ -520,6 +522,137 @@ class OrderUpdateStoreTest {
...
+                    )
+                }.doReturn(
+                    WooPayload(
+                        OrdersDatabaseBatch(
+                            createdEntities = emptyList(),
+                            updatedEntities = emptyList(),
+                            deletedEntities = listOf(
+                                OrderEntity(orderId = 123.toLong(), localSiteId = LocalId(1)),
+                                OrderEntity(orderId = 456.toLong(), localSiteId = LocalId(2))
```
atorresveiga: It would be great if we also test edge cases. For example: 
What would happen if we call the batch delete function with ids `1,2,3` but the response only includes ids `1,3`?

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClientTest.kt
```diff
@@ -0,0 +1,145 @@
...
+import org.wordpress.android.fluxc.model.order.OrderAddress.Shipping
+import org.wordpress.android.fluxc.model.order.UpdateOrderRequest
+import org.wordpress.android.fluxc.network.UserAgent
+import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackSuccess
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
+
+@RunWith(MockitoJUnitRunner::class)
+class OrderRestClientTest {
```
atorresveiga, line 32: Nice! Thanks for adding unit tests to this class.

I believe we could also include a unit test for the failure case as on [LeaderboardsRestClientTest](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/da52b97e2e116cad7ae1e5bde2945e1db72a616d/example/src/test/java/org/wordpress/android/fluxc/wc/leaderboards/LeaderboardsRestClientTest.kt). WDYT?

atorresveiga:
Great work Rodrigo! Thanks for adding more tests to the code and the batch update request to the FluxC example.
I left a few comments, and I'm looking forward to hearing your thoughts on those before merging the code.🙂

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClientTest.kt
```diff
@@ -0,0 +1,145 @@
...
+        lastName = "lastName",
+        company = "company",
+        address1 = "address1",
+        address2 = "address2",
+        city = "city",
+        state = "state",
+        postcode = "postcode",
+        country = "countryCode",
+        phone = "phone"
+    )
```
atorresveiga: I wonder if we really need all this information. I don't see that we are taking advantage of this information in the test

rodrigorods: The idea was get to get the code ready to check the generated body with mockito. I did not wanted to make big changes on something unrelated to my code, and to get this done I would have to move function `UpdateOrderRequest.toNetworkRequest`. 

I will create a class `OrderRequestBodyBuilder ` and move all body construction code to it.
this will allow me to test the `toNetworkRequest()` code and check the body param at OrderRestClientTest.

Ex:
```
object OrderRequestBodyBuilder {
    fun build() { ... } // This would be the older toNetworkRequest() fun
    fun buildFromOrdersBatch() { ... }
}
```

Would you say that this is a good aproach?

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClientTest.kt
rodrigorods: Sure (: I will add tests for the error scenarios.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
rodrigorods: Yes, it makes more sense to delete only whats is available at the response. I will work on that.

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderRestClientTest.kt
atorresveiga: My point above was that if substitute:
```
   val response = orderRestClient.updateOrdersBatch(
        site = site,
        createRequest = createRequest,
        updateRequest = updateRequest,
        deleteRequest = deleteRequest
    )
```
With 
```
   val response = orderRestClient.updateOrdersBatch(
        site = site,
        createRequest = emptyList(),
        updateRequest = emptyList(),
        deleteRequest = emptyList()
    )
```
The test will still pass, so maybe no need for the extra info.

After a second look, I don't think the test's main goal of ensuring **_updateOrdersBatch should call gson builder request with expected params_** can't be guaranteed. The test is checking that the response is a success and not checking the parameters used to call the gson builder. One suggestion here could be to use the `verify` function of mockito.

With this change, the test could look like this:
```
@Test
    fun `updateOrdersBatch should call gson builder request with expected params`(): Unit =
        runBlocking {
            val idsToDelete = listOf(1L, 2L, 3L)
            val body = OrdersBatchUpdateRequest.buildBody(
                emptyList(),
                emptyList(),
                idsToDelete
            )

            whenever(
                jetpackTunnelGsonRequestBuilder.syncPostRequest(
                    eq(orderRestClient),
                    site = eq(site),
                    url = eq(WOOCOMMERCE.orders.batch.pathV3),
                    body = any(),
                    clazz = eq(OrdersBatchDto::class.java)
                )
            ).thenReturn(
                JetpackSuccess(OrdersBatchDto())
            )

            orderRestClient.updateOrdersBatch(
                site = site,
                createRequest = emptyList(),
                updateRequest = emptyList(),
                deleteRequest = idsToDelete
            )

            verify(jetpackTunnelGsonRequestBuilder, times(1))
                .syncPostRequest(
                    restClient = eq(orderRestClient),
                    site = eq(site),
                    url = eq(WOOCOMMERCE.orders.batch.pathV3),
                    body = eq(body),
                    clazz = eq(OrdersBatchDto::class.java)
                )
        }
```
With this version, we don't need to modify `UpdateOrderRequest.toNetworkRequest`, and if we change the parameters we use to call the function without modifying the expected response, the test will fail.

example/src/main/java/org/wordpress/android/fluxc/example/ui/orders/WooOrdersFragment.kt
```diff
@@ -597,6 +566,105 @@ class WooOrdersFragment : StoreSelectingFragment(), WCAddOrderShipmentTrackingDi
...
+                        UpdateOrderRequest(
+                            id = it,
+                            customerNote = "$newCostumerNote"
+                        )
+                    } ?: emptyList()
+
+                    if (
+                        ordersToCreate.isEmpty() &&
+                        ordersToUpdate.isEmpty() &&
+                        ordersToUpdate.isEmpty()
```
atorresveiga: Note that here we are missing the `ordersToDelete` list and checking twice for the `ordersToUpdate`

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/store/OrderUpdateStoreTest.kt
```diff
@@ -574,8 +582,8 @@ class OrderUpdateStoreTest {
                         OrdersDatabaseBatch(
                             createdEntities = emptyList(),
                             updatedEntities = listOf(
-                                OrderEntity(orderId = 123.toLong(), localSiteId = LocalId(1)),
-                                OrderEntity(orderId = 456.toLong(), localSiteId = LocalId(2))
+                                OrderEntity(orderId = 123L, localSiteId = LocalId(1)),
+                                OrderEntity(orderId = 456L, localSiteId = LocalId(2))
```
atorresveiga: If we are using the `orderEntities` for the assertion below (line 600). Wouldn't make sense to use the same list in the mocked result?
Something like this:
```
.doReturn(
      WooPayload(
          OrdersDatabaseBatch(
              createdEntities = emptyList(),
              updatedEntities = orderEntities,
              deletedEntities = emptyList()
          )
      )
  )
```
This way, if someone needs to make changes to `orderEntities`, the tests won't be affected.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OrderUpdateStore.kt
```diff
@@ -260,7 +260,7 @@ class OrderUpdateStore @Inject internal constructor(
                     model.updatedEntities.map { ordersDao.insertOrUpdateOrder(it) }
                 }
                 if (deleteRequest.isNotEmpty()) {
-                    deleteRequest.map { ordersDao.deleteOrder(site.localId(), it) }
+                    model.deletedEntities.map { ordersDao.deleteOrder(it.localSiteId, it.orderId) }
```
atorresveiga, line 263: A suggestion for the future is to submit these changes in a separate commit, so it is easy for the reviewer to track those.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrdersBatchDto.kt
```diff
@@ -15,10 +15,14 @@ object OrdersBatchUpdateRequest {
...
         deleteRequest: List<Long>
-    ) = mutableMapOf<String, Any>().apply {
-        put("create", createRequest)
-        put("update", updateRequest)
-        put("delete", deleteRequest)
+    ) = buildMap {
+        putNotEmpty("create", createRequest)
+        putNotEmpty("update", updateRequest)
+        putNotEmpty("delete", deleteRequest)
+    }
```
atorresveiga: Nice!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrdersBatchDto.kt
```diff
@@ -15,10 +15,14 @@ object OrdersBatchUpdateRequest {
...
+
+    private fun MutableMap<String, Any>.putNotEmpty(key: String, value: List<*>) {
+        if (value.isNotEmpty()) this[key] = value
     }
```
atorresveiga: Checking the codebase, I found that we have a very similar extension function [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/MapExt.kt).
Maybe we could add this extension function to the same file with some small changes to make it more generic.
```
fun <K, V : List<*>> MutableMap<K, V>.putIfNotEmpty(key: K, value: V) {
  if (value.isNotEmpty()) this[key] = value
}
```
And make it available to everyone who may need it. WDYT?

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrdersBatchUpdateRequestTest.kt
```diff
@@ -0,0 +1,56 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.order
+
+import org.assertj.core.api.Assertions
+import org.junit.Test
+
+class OrdersBatchUpdateRequestTest {
```
atorresveiga, line 6: Nice!

atorresveiga:
Closed until we resume the work on bulk update orders

PR #2544, opened 2022-10-19
Feature/7510 account creation network

JorgeMucientes:
### Description

Adds new account creation request as well as new username suggestions request. This will enable native sign up using the latest version of WordPress.com API.

In this particular case I decided to decouple from `AccountStore` and create a new `SignUpStore` and `SignUpResClient` for the following reasons: 

- `AccountStore` has already too much logic in it
- `mAccountRestClient.newAccount()` doesn't use the latest WP API version to create a new account 
- `mAccountRestClient.newAccount` doesn't include the param `send_verification_email` we need to send the validation email
- Heavily relies on event bus
- It's java, so I wouldn't be able to use coroutines and I'd have to deal with event bus in `SignUpRepository` to wrap its calls and convert them to coroutines, making logic more complex ion the repository.
- Creating a new Store and RestClient was easier and more straightforward. Also results in cleaner code for our needs. 

### Testing
Testing instructions can be found in [the PR](https://github.com/woocommerce/woocommerce-android/pull/7565) for the WooCommerce app where this new requests are used.


hafizrahman (approved):
1000+ lines in `AccountStore` 😵  so I think it's a good decision to move this separately. Code looks clean and understandable to me.

fluxc/src/main/java/org/wordpress/android/fluxc/store/signup/SignUpStore.kt
```diff
@@ -0,0 +1,70 @@
...
+        return coroutineEngine.withDefaultContext(AppLog.T.API, this, "createWpAccount") {
+            val result = signUpRestClient.createWPAccount(email, password, username)
+            when {
+                result.isError -> CreateWpAccountResult(
+                    CreateWpAccountError(result.error?.apiError)
+                )
+                result.result == null -> CreateWpAccountResult(
+                    CreateWpAccountError(EMPTY_SUCCESSFUL_RESPONSE)
+                )
+                else -> CreateWpAccountResult(result.result.success, result.result.token)
```
hichamboushaba, line 53: I didn't review the whole code, and I know it's merged, but just a non-blocking question, wouldn't we want to save the received token at this step? something like what we do [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/dee3d2f1a33b7fc711bae342d16a32433a62d9ce/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java#L1261)?

hafizrahman: Oh ha! Thanks for mentioning that, I wasn't aware there's such requirement at this point. [The WC PR](https://github.com/woocommerce/woocommerce-android/pull/7565) seems to end just in an empty fragment so probably this is planned later for handling successful registration and such?

JorgeMucientes: Good question @hichamboushaba. TBH, I don't have a strong opinion on this. Currently I'm delegating saving/updating the token to the `SignUpRepository`  [here](https://github.com/woocommerce/woocommerce-android/blob/674f369af89e061dcb0e8e4c97e1f751cbfb0244/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/signup/SignUpRepository.kt#L34). I wanted to make this store as simple as possible and avoid having to update it in case the way we handle the token in the future changes. 
Is just a minor preference. But as said I wouldn't mind handling that in this store

hichamboushaba: I don't have a strong opinion either, it was just a question 🙂. I didn't review the other PR, so I didn't know it was handled there, so it's all good.

PR #2546, opened 2022-10-24
fix: crash on getting `UserAgent` on thread different than the main

wzieba:
Internal: p91TBi-9Zw-p2#comment-11208

Description:

This PR changes how we get a default `UserAgent` for HTTP requests. Instead of creating WebView and getting the default UserAgent from it, we use a native `WebSettings#getDefaultUserAgent` method which **does not crash when called from a background thread**.

For more details, please see the internal link above.

---

I've tested values of the default `UserAgent` and they are precisely the same before and after this change.

For safety, I ensured it did not crash on API version 21 when the call was executed on a background thread.

```kotlin
// does not crash on API 21. Returns a valid value.
        Thread {
            WebSettings.getDefaultUserAgent(this@MainActivity)
        }.start()
```

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/UserAgent.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/UserAgent.java
@@ -1,7 +1,6 @@
 package org.wordpress.android.fluxc.network;
 
 import android.content.Context;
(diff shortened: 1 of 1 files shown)


hichamboushaba (approved):
Nice work @wzieba, looks good to me.

PR #2547, opened 2022-10-24
Top performers with date period

atorresveiga:
⚠️ This PR can't be merged until https://github.com/woocommerce/woocommerce-android/pull/7626 is approved

This PR changes the `granularity` id on `TopPerformerProducts` Entity in favour of `datePeriod`.
```
granularity -> datePeriod
```
The `granularity` id was introduced on [this PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2526/files) to provide better offline support for top performers products and reduce the loading "slowness". 
As part of the new changes we are introducing to support Analytics hub, we were in the need to allow merchants to query top performance products not only by our defined granularities but also by customs date ranges. The main idea of this PR is to change the `granularity` field that would have the format of `granularity=today` for a new format that would allow us to maintain the improvements on top performers products and extend the range of the query to periods outside the defined granularities. This new format is `datePeriod=$startDate-$endDate`

### Testing instructions
This PR can be tested alongside the changes from WooCommerce Android. Please follow the test instructions from https://github.com/woocommerce/woocommerce-android/pull/7626.

TC 2
Test the new DB Migraction:

1. Install the Woo version from branch `feature/analytics`
2. Change between different MyStore Tabs
3. Open the Database Inspector on Android Studio or Flipper and check that the TopPerformerProducts table has data and the granularity column is present
4. Install this new version
5. Check that the app doesn't crash, the old cached data is wiped and the new datePeriod column is present

| Old TopPerformerProducts   | New TopPerformerProducts |
| ------------- | ------------- |
| |   |

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardsRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardsRestClient.kt
@@ -9,10 +9,7 @@ import org.wordpress.android.fluxc.network.UserAgent
 import org.wordpress.android.fluxc.network.rest.wpcom.BaseWPComRestClient
 import org.wordpress.android.fluxc.network.rest.wpcom.auth.AccessToken
 import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder
-import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder.JetpackResponse
-import org.wordpress.android.fluxc.network.rest.wpcom.wc.orderstats.OrderStatsRestClient.OrderStatsApiUnit
-import org.wordpress.android.fluxc.store.WCStatsStore.StatsGranularity
-import org.wordpress.android.fluxc.utils.DateUtils
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
 import org.wordpress.android.fluxc.utils.handleResult
 import javax.inject.Inject
 import javax.inject.Named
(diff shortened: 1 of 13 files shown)


wzieba (approved):
Looks good 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardsRestClient.kt
```diff
@@ -30,47 +27,41 @@ class LeaderboardsRestClient @Inject constructor(
     @Suppress("LongParameterList")
     suspend fun fetchLeaderboards(
         site: SiteModel,
-        unit: StatsGranularity?,
```
atorresveiga, line 33: Granularity was received here to calculate `start date`, `end date` and the `interval` in case this values were `null`. 
This logic was refactored and moved to the `WCLeaderboardStore`. Now the store expose two methods: one receiving the `granularity` as a parameter where all this calculations take place, an another one receiving the `start date`, `end date` and `interval` where no calculation is needed.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
```diff
@@ -603,3 +603,27 @@ internal class AutoMigration17to18 : AutoMigrationSpec
...
+                    `currency` TEXT NOT NULL,
+                    `total` REAL NOT NULL,
+                    `millisSinceLastUpdated` INTEGER NOT NULL,
+                    PRIMARY KEY(`datePeriod`,`productId`,`siteId`)
+                    )
+                    """.trimIndent()
+            )
+        }
+    }
+}
```
atorresveiga, line 629: An `AutoMigrationSpec` with the `@RenameColumn` annotation is another possible option for the database migration. The drawback of using `AutoMigrationSpec` is that old data using `granularity` as a key would be kept in the database. This was the main reason why I opted to this other version where all outdated data is deleted.

PR #2548, opened 2022-10-24
Add @Nullable annotation to AccountStore#getAccessToken 

RenanLukas:
Adds `@Nullable` to `AccountStore#getAccessToken` method, so we get a compilation error when calling it from Kotlin.

Before:
image

After:
image

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/AccountStore.java
@@ -1213,7 +1213,10 @@ public class AccountStore extends Store {
 
     /**
      * Should be used for very specific purpose (like forwarding the token to a Webview)
+     *
+     * @return the access token if it was already set, otherwise null
      */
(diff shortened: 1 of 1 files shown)


ParaskP7 (approved):
LGTM! 💯

irfano (approved):
LGTM! 👍🏻 
I wonder if consumer apps will get any compiler errors with this PR. It will be better to check it as we update FluxC versions on consumer projects.

ParaskP7:
> I wonder if consumer apps will get any compiler errors with this PR. It will be better to check it as we update FluxC versions on consumer projects.

I am sure that will be the case @irfano @RenanLukas ! 💯 

I recommend checking and updating the following repos:
- [WPAndroid](https://github.com/wordpress-mobile/WordPress-Android)
- [WCAndroid](https://github.com/woocommerce/woocommerce-android)
- [Login](https://github.com/wordpress-mobile/WordPress-Login-Flow-Android)

PR #2549, opened 2022-10-25
Release `1.56.1`

wzieba:
This release contains only a [hotfix](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2546), which has already been merged into the trunk.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/UserAgent.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/UserAgent.java
@@ -1,25 +1,20 @@
 package org.wordpress.android.fluxc.network;
 
 import android.content.Context;
-import android.webkit.WebView;
+import android.webkit.WebSettings;
 
 import org.wordpress.android.util.PackageUtils;
 
 public class UserAgent {
-    private String mUserAgent;
-    private String mDefaultUserAgent;
+    private final String mUserAgent;
+    private final String mDefaultUserAgent;
 
     public UserAgent(Context appContext, String appName) {
         // Device's default User-Agent string.
         // E.g.:
         //   "Mozilla/5.0 (Linux; Android 6.0; Android SDK built for x86_64 Build/MASTER; wv)
         //   AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/44.0.2403.119 Mobile Safari/537.36"
-        try {
-            mDefaultUserAgent = new WebView(appContext).getSettings().getUserAgentString();
-        } catch (RuntimeException re) {
-            mDefaultUserAgent = "Mozilla/5.0 (Linux; Android 6.0; default FluxC UA; wv) AppleWebKit/537.36"
-                    + "(KHTML, like Gecko) Version/4.0 Chrome/44.0.2403.119 Mobile Safari/537.36";
-        }
+        mDefaultUserAgent = WebSettings.getDefaultUserAgent(appContext);
         // User-Agent string when making HTTP connections, for both API traffic and WebViews.
         // Appends "wp-android/version" to WebView's default User-Agent string for the webservers
         // to get the full feature list of the browser and serve content accordingly, e.g.:


spencertransier (approved):
Confirmed that the changes in this release have been reviewed here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2546

AliSoftware:
@wzieba @spencertransier What's with the version number in this PR title? I'm super confused with a PR named `Release 1.56.1` being merged into `trunk` while `trunk` was already on `2.1.x`?

Especially, was a bit concerned when I [looked at the Changelog of release `2.2.0`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/releases/tag/2.2.0) and saw a mention of `Release 1.56.1` being included in that `2.2.0` release?!!

wzieba:
`1.56.1` is a hotfix for `1.56` that was needed to release Woo Android `10.8.1`. I've tried to follow the way documented here (internal: paqN3M-l3-p2).

`2.2.0` already contains the fix introduced for `1.56.1`. Is your concern raised about changelog entry or the process? If about the process - could you please send me a link to the documentation, so I can do better next time if this time was something wrong? Thanks!

AliSoftware:
Ahhh that explains it!

My concern was more that `1.56.1` felt quite "ancient" compared to `2.2.0`, so that made me think it was a very old release from months ago that we totally forgot to backmerge or whatnot. Should I realized that the jump from `1.56.x` to `2.0.0` was not _that_ long ago.

I think what threw me off as well is that usually, when there's a hotfix (e.g. for client apps like Woo or WP), we merge the hotfix in the immediate next version (e.g. if we do a `20.8.1` hotfix for WP, we merge it into `20.9`), so seeing that there had been multiple versions between `1.56.x` and `2.2.0` and that `1.56.1` was not merged in a (non-existent) `1.57.0` or more appropriately in `2.0.0` (which was the real, next version, I was just not used for us to do major bumps on FluxC given we haven't been very good at following SemVer with it so far) made me frown a bit.
So if anything, I definitively didn't expect a hotfix for multiple versions ago land in `2.2.0`… instead of `2.0.0`. If it had been a version `2.1.1` being merged into `2.2.0` I'd likely have not be as surprised, for example

In practice I think the process was totally right, and my concern was also misguided by the fact that FluxC moves faster than, e.g. WordPress Android (and especially with Woo doing weekly releases now, while WPAndroid still does his on 2 weeks basis, so we end up having bigger and faster version jumps on FluxC than I'm used to see on WPAndroid itself of course). And also by the fact that FluxC is used by both WC and WP and thus gets new tags sometimes for both client apps (if they're not in sync), hence more frequent tags there too. Which is probably why the `1.56.1` hotfix ended up only being backmerged into `2.2.0` — and not as soon as in `2.0.0` as I'd have expected for a more usual project — because by the time `1.56.1` was finalized and ready to be back-merged, we already had more than one new tag on the main FluxC release train?

wzieba:
Sorry for late reply @AliSoftware 🙇 

> My concern was more that 1.56.1 felt quite "ancient" compared to 2.2.0, so that made me think it was a very old release from months ago that we totally forgot to backmerge or whatnot.

Gotcha, I see it was quite confusing.

> with Woo doing weekly releases now, while WPAndroid still does his on 2 weeks basis, so we end up having bigger and faster version jumps on FluxC than I'm used to see on WPAndroid itself of course

I agree, the tempo of releasing the next versions increased significantly. And I was also caught off guard at some point seeing FluxC `2.x.x` 😄 .

> And also by the fact that FluxC is used by both WC and WP and thus gets new tags sometimes for both client apps (if they're not in sync), hence more frequent tags there too. Which is probably why the `1.56.1` hotfix ended up only being backmerged into `2.2.0` — and not as soon as in `2.0.0` as I'd have expected for a more usual project 

Good point again. I agree that usually, we should backmerge the fix to `2.0.0` as well, but as you correctly notice, the fix was needed only for WC.

PR #2558, opened 2022-11-02
Net revenue and average order value to interval subtotals

atorresveiga:
### Why
Calling the function  `getIntervalList()` from the `WCRevenueStatsModel` class deserializes the JSON revenues into a list of Interval objects. These Intervals objects contained only the `ordersCount` and `totalSales` information for the selected interval. 
As part of the new Analytics Hub feature, we also require the Intervals objects to retrieve the  `net revenue` and the `average order value` information to be able to display these values charts, as shown in the image below.

### Description
This small PR adds the net revenue and average order value to the SubTotal class, so these values can be retrieved by the Interval class and consumed by the Analytics Hub.

### Testing instructions
This PR can be tested alongside the Woo PR


ThomazFB (approved):
LGTM! Sorry for the waiting time, I assigned the PR and forgot to review it 🤦

PR #2559, opened 2022-11-03
Fix unit tests execution above JDK 11

wzieba:
Closes: #2540 
X: https://github.com/Automattic/android-dependency-catalog/pull/7

### Description

This PR fixes failing unit test execution on JDKs higher than 11 by updating `mockito-core` version. With this update, we needed to drop support for nhaarman's `mockito-kotlin` in favor of mockito's `mockito-kotlin`.

### Review instructions

Please review this PR commit by commit. Changes in commit marked with "automated" are changes made by IDE's find&replace + optimize imports tools. Those changes might not be worth reviewing.

### Test instructions

1. Firstly, make sure you can reproduce the issue. Install any JDK above 11 and update `JAVA_HOME` (if you use terminal) or settings in Android Studio (if you use AS). Hint: Android Studio Eel has embedded JDK version 17
2. Checkout `trunk`
3. Run unit tests
4. **Assert that unit tests fail** (`Mockito cannot mock this class (...)`)
5. Checkout `issue/2540_fix_unit_tests_execution_above_jdk11` branch
6. Run unit tests again
7. **Assert that unit tests pass**

### Merging instructions

Please do not merge this PR until a new version of the dependency catalog is released, and the tag in this PR is updated.

PR #2561, opened 2022-11-04
fix: remove nhaarman imports from JitmRestClientTest

wzieba:
Suplement to #2559 . Thanks @ParaskP7 for noticing 🙌 .


ParaskP7 (approved):
LGTM! 💯 

Thank you for the quick fix @wzieba ! 🥇

PR #2564, opened 2022-11-08
Update `minSdkVersion` (from `21` to `24`)

ParaskP7:
This PR upgrades FluxC to `minSdkVersion` to `24`. This has been unblocked because:
- `WPAndroid` was already on `minSdkVersion = 24`, for some time now (see [here](https://github.com/wordpress-mobile/WordPress-Android/pull/15135)).
- `WCAndroid` has been recently upgraded to `minSdkVersion = 24` as well (see [here](https://github.com/woocommerce/woocommerce-android/pull/7643)).

-----

Warning (⚠️): This update is currently blocked by the common upgrade of `targetSdkVersion` to `31` (see [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2563)).

-----

As part of this `minSdkVersion = 24` upgrade the below changes were also applied in order to fix any additional warnings that this change brought-up:

Warnings Resolution List:

1. [Resolve obsolete sdk int lint warning for well sql config.](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/986fc83303eb5472d38ff18a25982c17ceff64f5)
2. [Resolve obsolete sdk int lint warning for open jdk cookie mng.](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/3400d7b0a8a6e33a3ca2c943a47107376be896c2)
3. [Resolve obsolete sdk int lint warning for leaderboard prd item.](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/ea8f79dc9f1bd4a7977cbd0386f21f7caf4e0a2c)
4. [Resolve obsolete sdk int lint warning for mocked statck test.](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/ef933c2349c6d66cc545aba6f03d0dfee1271694)

-----

PS: @hichamboushaba @ovitrif I added you as the main reviewers, that is, in addition to the @wordpress-mobile/apps-infrastructure team itself, but randomly, since I just wanted someone from both, the `WordPress` and `Woo Commerce` teams, to sign-off on that change for WPAndroid and WCAndroid alike.

-----

### Testing instructions

- There is nothing much to test here.
- Verifying that all the CI checks are successful should be enough.
- However, if you want to be thorough about reviewing this change, you could:
    - Quickly smoke test the `FluxC Example` app and see if it works as expected.
    - Try and install the `FluxC Example` app on Android 5 (`API 21/22`) and Android 6 (`API 23`) devices and verify that you can't (see [API Levels](https://apilevels.com/)).

-----

### Merge instructions

- [x] Wait for #2563 to be approved and merged to `trunk`.
- [x] Make sure this PR's base is automatically updated to `trunk`.
- [x] Update PR with latest `trunk`.
- [x] Remove `[PR] Not Ready For Merge]` label.
- [ ] Merge PR to `trunk`.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
@@ -1,7 +1,5 @@
 package org.wordpress.android.fluxc.network.rest.wpcom.wc.leaderboards
 
-import android.annotation.TargetApi
-import android.os.Build
 import android.text.Html
 import android.text.SpannableStringBuilder
 import android.text.style.URLSpan
@@ -175,11 +173,9 @@ class LeaderboardProductItem(
             getSpans(0, length, URLSpan::class.java)
                     .toList()
 
-    @TargetApi(Build.VERSION_CODES.N)
-    private fun fromHtmlWithSafeApiCall(source: String?) = source
-            ?.takeIf { Build.VERSION.SDK_INT >= Build.VERSION_CODES.N }?.let {
-                Html.fromHtml(source, Html.FROM_HTML_MODE_LEGACY)
-            } ?: Html.fromHtml(source)
+    private fun fromHtmlWithSafeApiCall(source: String?) = source?.let {
+        Html.fromHtml(source, Html.FROM_HTML_MODE_LEGACY)
+    } ?: Html.fromHtml(source)
 
     /**
      * Returns the second object of the Top Performer Item Array if exists
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
@@ -3,7 +3,6 @@ package org.wordpress.android.fluxc.network.rest.wpcom.wc.leaderboards
 import android.text.Html
 import android.text.SpannableStringBuilder
 import android.text.style.URLSpan
-import androidx.core.text.HtmlCompat
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.leaderboards.LeaderboardsApiResponse.LeaderboardItemRow
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.leaderboards.LeaderboardsApiResponse.Type.PRODUCTS
 
@@ -85,7 +84,7 @@ class LeaderboardProductItem(
                 ?.split(";")
                 ?.filter { it.contains("&#") }
                 ?.reduce { total, new -> "$total$new" }
-                ?.run { HtmlCompat.fromHtml(this, HtmlCompat.FROM_HTML_MODE_LEGACY) }
+                ?.run { Html.fromHtml(this, Html.FROM_HTML_MODE_LEGACY) }
                 ?: plainTextCurrency
     }
 
@@ -106,7 +105,7 @@ class LeaderboardProductItem(
      *      Output: DKK
      */
     @Suppress("MaxLineLength") private val plainTextCurrency by lazy {
-        Html.fromHtml(priceAmountHtmlTag)
+        Html.fromHtml(priceAmountHtmlTag, Html.FROM_HTML_MODE_LEGACY)
                 .toString()
                 .replace(Regex("[0-9.,]"), "")
     }
@@ -151,7 +150,7 @@ class LeaderboardProductItem(
      * using the [SpannableStringBuilder] implementation in order to parse it
      */
     private val link by lazy {
-        Html.fromHtml(itemHtmlTag)
+        Html.fromHtml(itemHtmlTag, Html.FROM_HTML_MODE_LEGACY)
                 .run { this as? SpannableStringBuilder }
                 ?.spansAsList()
                 ?.firstOrNull()
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
@@ -84,7 +84,7 @@ class LeaderboardProductItem(
                 ?.split(";")
                 ?.filter { it.contains("&#") }
                 ?.reduce { total, new -> "$total$new" }
-                ?.run { fromHtmlWithSafeApiCall(this) }
+                ?.run { Html.fromHtml(this) }
                 ?: plainTextCurrency
     }
 
@@ -105,7 +105,7 @@ class LeaderboardProductItem(
      *      Output: DKK
      */
     @Suppress("MaxLineLength") private val plainTextCurrency by lazy {
-        fromHtmlWithSafeApiCall(priceAmountHtmlTag)
+        Html.fromHtml(priceAmountHtmlTag)
                 .toString()
                 .replace(Regex("[0-9.,]"), "")
     }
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
```diff
@@ -175,11 +173,9 @@ class LeaderboardProductItem(
...
                     .toList()
 
-    @TargetApi(Build.VERSION_CODES.N)
-    private fun fromHtmlWithSafeApiCall(source: String?) = source
-            ?.takeIf { Build.VERSION.SDK_INT >= Build.VERSION_CODES.N }?.let {
-                Html.fromHtml(source, Html.FROM_HTML_MODE_LEGACY)
-            } ?: Html.fromHtml(source)
+    private fun fromHtmlWithSafeApiCall(source: String?) = source?.let {
+        Html.fromHtml(source, Html.FROM_HTML_MODE_LEGACY)
+    } ?: Html.fromHtml(source)
```
hichamboushaba: I think we can simplify this function by removing the call to the parameterless `fromHtml`, as it's equivalent to the other one using `FROM_HTML_MODE_LEGACY`: https://developer.android.com/reference/android/text/Html#fromHtml(java.lang.String). This was necessary before because of the `takeIf`, but not anymore I think.

hichamboushaba (approved):
Looks good, thanks @ParaskP7 for working on this.
I left one comment, but it's not blocking, so I'm pre-approving.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
ParaskP7: Great, thanks for the comment @hichamboushaba ! I agree, the cleaner the better, I'll apply this change and let you know as soon as I push its commit. 🙇

ParaskP7:
Thank you so much for reviewing and testing this PR @hichamboushaba , you rock! 🙏

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
ParaskP7: 👋 @hichamboushaba !

This is now done, see ba0b9060f502d239d7a235ea8cbe8bc4667642c5 and 8637d1cb309b769edced2c03af6f77c5592e1449.

You will note that I went a bit further and tried to also resolve the `Deprecated` warning, but I only could go as far as doing so for 1/3 of those. You will find further detail in the commit description itself. Let me know what you think and thank you! 🙇

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
```diff
@@ -84,7 +85,7 @@ class LeaderboardProductItem(
                 ?.split(";")
                 ?.filter { it.contains("&#") }
                 ?.reduce { total, new -> "$total$new" }
-                ?.run { Html.fromHtml(this) }
+                ?.run { HtmlCompat.fromHtml(this, HtmlCompat.FROM_HTML_MODE_LEGACY) }
```
hichamboushaba: np, you don't have to change it.
`HtmlCompat` is useful only for apps with minSdk less than 24, for us, we can directly call the method that was introduced in API 24 https://developer.android.com/reference/android/text/Html#fromHtml(java.lang.String,%20int)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
```diff
@@ -150,7 +150,7 @@ class LeaderboardProductItem(
      * using the [SpannableStringBuilder] implementation in order to parse it
      */
     private val link by lazy {
-        fromHtmlWithSafeApiCall(itemHtmlTag)
+        Html.fromHtml(itemHtmlTag)
```
hichamboushaba: For using the new API instead of the deprecated one, WDYT about something like this?
```kotlin
itemHtmlTag?.let { Html.fromHtml(it, Html.FROM_HTML_MODE_LEGACY) }
                  ?.run { this as? SpannableStringBuilder }
                 ...
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
hichamboushaba: Thanks @ParaskP7 for making the changes 🙏 , I left some comments below.
All of this is minor, so no need for spending too much time on it 🙂.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
ParaskP7: Ah, thank you @hichamboushaba , you are totally right, I somehow got afraid of the fact that the `String source` is neither marked `@NonNull` nor `@Nullable` on the `Html.java` side and thus didn't want to make this change, but that doesn't matter, you have to trust that it works for `@Nullable` just like that as it anyways defaults to using the same mechanism underneath.

Thus, done, this update seems safe and apologies for the trouble with this review (😅): 5acc19a9206c17ed7d07498cd73f4cdb31134119

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/leaderboards/LeaderboardProductItem.kt
ParaskP7: 👋 @hichamboushaba !

I think maybe [this change](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2564#discussion_r1018976760) is enough for now, wdyt? 🤔

PS: I really don't want to change any of the functionality, even by that slight bit, not that it will change much, but you know, you can never be too careful, and then you need to really test that, even if you change that little bit.

hichamboushaba: Good point, and I think we are safe, when the argument is `null`, we were already calling `Html#fromHtml(String)` which proxies the call to the same function we are using now.

ParaskP7: Exactly, once more, thank YOU Hicham ! 🙇

ovitrif (approved):
Changes LGTM 🎉 

Though I wonder with regards to the existing discussion here, in WPAndroid we always use `HtmlCompat`, yet `minSdkVersion` is 24, isn't it then in fact not necessary? 😅

ParaskP7:
Thank you for the review and testing @ovitrif ! 🙇 ❤️ 🚀 

> Though I wonder with regards to the existing discussion here, in WPAndroid we always use HtmlCompat, yet minSdkVersion is 24, isn't it then in fact not necessary? 😅

Yeap, yeap, this is indeed the case, I just copy-pasted my solution from WCAndroid into WPAndroid, and since I used `HtmlCompat` there, I used it blindly as is on WPAndroid as well. But this is indeed not necessary and we could optimize on it. 💯 

PS: Feel free to do a quick PR on it if you like, else I'll get to that at some point myself. 👍

PR #2565, opened 2022-11-11
Add support for fetching plugins using Rest WPApi

hichamboushaba:
This PR adds the following changes to the PluginStore:

1. Adds support for fetching plugins using the WordPress REST API.
2. For plugin installation:
    i. Aligns the returned errors between the different network implementations.
    b. Expose the network error code in the error, to allow explaining to the user the cause of failure.

To keep the changes manageable, I decided to keep using the existing EventBus architecture.

This is needed for https://github.com/woocommerce/woocommerce-android/pull/7832

⚠️ Please don't merge until the above PR is approved.

#### Testing
1. Open the example app.
2. Sign in using the site credentials (enter the XMLRPC URL in the login form).
3. Open the Plugins screen.
4. Test installing a new plugin.
5. Confirm the installation works as expected.
6. Test re-installing the same plugin.
7. Confirm that the error message is correctly printed in the console.
8. Test fetching the plugin.
9. Confirm it works as expected.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/plugin/PluginWPAPIRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/plugin/PluginWPAPIRestClient.kt
@@ -52,6 +52,22 @@ class PluginWPAPIRestClient @Inject constructor(
         }
     }
 
+    suspend fun fetchPlugin(
+        site: SiteModel,
+        nonce: Nonce?,
+        pluginName: String
+    ): WPApiPluginsPayload<SitePluginModel> {
+        val url = buildUrl(site, pluginName)
+        val response =
+            wpApiGsonRequestBuilder.syncGetRequest(
+                restClient = this,
+                url = url,
+                clazz = PluginResponseModel::class.java,
+                nonce = nonce?.value
+            )
+        return handleResponse(response, site)
+    }
+
     suspend fun installPlugin(
         site: SiteModel,
         nonce: Nonce? = null,
--- a/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/PluginsFragment.kt
@@ -7,7 +7,6 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import dagger.android.support.AndroidSupportInjection
-import kotlinx.android.synthetic.main.fragment_notifications.*
 import kotlinx.android.synthetic.main.fragment_plugins.*
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
@@ -17,10 +16,10 @@ import org.wordpress.android.fluxc.example.utils.showSingleLineDialog
 import org.wordpress.android.fluxc.generated.PluginActionBuilder
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.store.PluginStore
-import org.wordpress.android.fluxc.store.PluginStore.FetchJetpackSitePluginPayload
+import org.wordpress.android.fluxc.store.PluginStore.FetchSitePluginPayload
 import org.wordpress.android.fluxc.store.PluginStore.InstallSitePluginPayload
-import org.wordpress.android.fluxc.store.PluginStore.OnJetpackSitePluginFetched
 import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginConfigured
+import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginFetched
 import org.wordpress.android.fluxc.store.PluginStore.OnSitePluginInstalled
 import javax.inject.Inject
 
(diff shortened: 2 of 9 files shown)


hafizrahman (commented):
The install and activate feature works nicely for me.

Fetching seems to always crash, even when fetching a plugin that I installed just a moment before. Also tried `woocommerce` on a site that has it, and still crashed:

```
    --------- beginning of crash
2022-11-15 06:22:14.382 2032-2032/org.wordpress.android.fluxc.example E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.wordpress.android.fluxc.example, PID: 2032
        at org.greenrobot.eventbus.EventBus.handleSubscriberException(EventBus.java:537)
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:519)
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:511)
        at org.greenrobot.eventbus.HandlerPoster.handleMessage(HandlerPoster.java:67)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
     Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.wordpress.android.fluxc.model.plugin.SitePluginModel.getDisplayName()' on a null object reference
        at org.wordpress.android.fluxc.example.PluginsFragment.onSitePluginFetched(PluginsFragment.kt:121)
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:517)
        at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:511) 
        at org.greenrobot.eventbus.HandlerPoster.handleMessage(HandlerPoster.java:67) 
        at android.os.Handler.dispatchMessage(Handler.java:107) 
        at android.os.Looper.loop(Looper.java:214) 
        at android.app.ActivityThread.main(ActivityThread.java:7356) 
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930) 
```

hichamboushaba:
> Fetching seems to always crash, even when fetching a plugin that I installed just a moment before. Also tried woocommerce on a site that has it, and still crashed:

Thanks for the review @hafizrahman, the issue should be fixed by e4464d9, and the reason it crashed is that I was missing setting the error correctly on the event in case there was an error, so the code assumed it was a success, and tried to read the plugin name for it.

And why it was a failure: it's because the fetch expects a full `name` (technically the `path`), so you should use: `woocommerce/woocommerce` to fetch WooCommerce plugin.

hafizrahman:
> And why it was a failure: it's because the fetch expects a full name (technically the path), so you should use: woocommerce/woocommerce to fetch WooCommerce plugin.

Oh I think we ran into this too when working on JP install initially, but I've completely forgotten about it.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/plugin/PluginWPAPIRestClient.kt
```diff
@@ -52,6 +52,22 @@ class PluginWPAPIRestClient @Inject constructor(
         }
     }
 
+    suspend fun fetchPlugin(
```
hafizrahman, line 55: > And why it was a failure: it's because the fetch expects a full name (technically the path), so you should use: woocommerce/woocommerce to fetch WooCommerce plugin.

I think it's good if we mention here what's the required format for `pluginName`, like in https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/72aaebeca02b525428092851a6a3eea5a063f6b7/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/plugin/PluginJetpackTunnelRestClient.kt#L45 

What do you think?

hafizrahman (approved):
All good and works well now 🥳 I left a small non-blocking question.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/plugin/PluginWPAPIRestClient.kt
hichamboushaba: IMO the meaning should be clear from the variable name: `name` vs `slug`.
Another point, anyone consuming the code will use it through the `PluginStore`, so they'll use a `FetchSitePluginPayload`, if we think this needs more clarification, we should add the comment to it instead 🙂, I don't think it's necessary, but let me know if you think we should add it 🙂.

And the issue with the example is not relevant to other usages, since we are use a text field to enter the name, which is not expected from final clients on a real app.

hichamboushaba: I'm going to merge the PR as is to be able to merge the WCAndroid PR, but please let me know if you think adding comments to the payloads would make understanding the code easier, then we can do it on a separate non-breaking PR.

PR #2566, opened 2022-11-11
Containment fix for My Site > Comments crash with standalone JP plugins.

develric:
This has a companion WordPress android PR [here](https://github.com/wordpress-mobile/WordPress-Android/pull/17456) 

This PR is part of a containment fix to avoid a crash happening in the Jetpack app for My Site > Comments on self-hosted sites not having the full Jetpack plugin but only one of the standalone plugins (like the jetpack backup plugin). This only avoids the crash allowing the relevant error to be displayed. For sites like those, the full rest api is not available but the username and password are actually null as well. This creates some not consistent behaviors in various areas of the app that need a more broad fix and review (more details in the internal p2 post and comments pe8j1f-V-p2); numbers of such cases are pretty low currently and this fix prioritizes the mentioned crash.

NOTE: the issue is only present in the JP app since the WP app was already filtering out the standalone JP plugin sites from the list of sites.

## To test

Prerequisite: install the JP backup plugin on a self-hosted site

### Replicate
- Install the JP app from WP Mobile trunk, and Login with the same user used to connect the backup plugin
- Select the backup plugin site from the list
- Go to My Site > Comments and notice the following crash in logcat `java.lang.NullPointerException: site.username must not be null`

### Check the fix
- Login in the WP app with the same user used to connect the backup plugin and notice the site IS NOT in the list of sites since it's filtered out
- Login in the JP app with the same user used to connect the backup plugin and notice the site IS in the list of sites
- Select the backup plugin site from the list
- Go to My Site > Comments, also move in between the different tabs and notice the crash is not happening and an error message for wrong username and password is displayed (since those fields are actually null/empty)

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/comments/CommentsXMLRPCClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/comments/CommentsXMLRPCClientTest.kt
@@ -150,6 +150,74 @@ class CommentsXMLRPCClientTest {
     }
 
     @Test
+    fun `fetchCommentsPage returns error without crashing when username is null`() = test {
+        mockedResponse = """<?xml version="1.0" encoding="UTF-8"?>
+                            <methodResponse>
+                              <fault>
+                                <value>
+                                  <struct>
+                                    <member>
+                                      <name>faultCode</name>
+                                      <value><int>403</int></value>
+                                    </member>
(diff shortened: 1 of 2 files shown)


RenanLukas (approved):
Thanks for the PR @develric . It works as expected :shipit:

PR #2567, opened 2022-11-15
POC: Add support for using WooCommerce REST API

hafizrahman:
Details on this POC is available internally at pe5sF9-PY-p2

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wcapi/WCAPIGsonRequestBuilder.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wcapi/WCAPIGsonRequestBuilder.kt
@@ -1,0 +1,132 @@
+package org.wordpress.android.fluxc.network.rest.wcapi
+
+import com.android.volley.Request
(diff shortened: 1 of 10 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wcapi/WCAPIGsonRequestBuilder.kt
```diff
@@ -0,0 +1,132 @@
...
+            request.cancel()
+        }
+
+        if (enableCaching) {
+            request.enableCaching(cacheTimeToLive)
+        }
+
+        if(authText != null) {
+            request.addHeader("Authorization", "Basic $authText")
+        }
```
hafizrahman, line 128: For now the code only supports the HTTP Basic Auth method of authentication: 

https://woocommerce.github.io/woocommerce-rest-api-docs/#authentication-over-https

`authText` needs to be the generated Woo REST API key and secret combined with semicolon(`<key>:<secret>`) , then encoded with Base64.

PR #2568, opened 2022-11-16
The fix that adopts the API changes

kidinov:
The API started to return 16 elements instead of 14, this fix adopts to the changes of the API


ParaskP7 (approved):
LGTM! 💯

PR #2569, opened 2022-11-16
Issue/7857 generate product variants

### LINKED ISSUE #7857: Support batched "create variations" endpoint
### Description

We have to add support for the networking layer to request creating many variations with a single endpoint.

We'll use this endpoint from Woo REST API: https://woocommerce.github.io/woocommerce-rest-api-docs/#batch-update-product-variations

atorresveiga:
Closes: https://github.com/woocommerce/woocommerce-android/issues/7857

### Why
We need to be able to create many variations for the Generate all combinations of variations project.
Creating all the variations in one request is the solution we agreed on after discussion, taking into account that it is fast and makes good use of resources. To achieve this solution, we'll use this endpoint from Woo REST API: https://woocommerce.github.io/woocommerce-rest-api-docs/#batch-update-product-variations.

### Description
This PR adds support to create product variations using the batch-update-product-variations endpoint. It accomplishes its goal by modifying the batchUpdateVariations function and adding the missing parameters to match the API request. It also adds a UI to the FluxC example app to make it easier to test the request.

### Testing instructions

1. Open the FluxC example app
2. Tap on Woo
3. Tap on Select Site
4. Select a Site
5. Tap on Products
6. Tap on Select Site
7. Select a Site
8. Scroll to Batch Generate Variations 
9. Tap on Batch Generate Variations 
10. Enter a valid variable product id
11. Tap on Set Product ID
12. Enter a valid Attribute Id, Name, and Option
13. Tap on Add Attribute
14. Tap on Add Variation
15. Repeat steps 12 to 14 (as many variations as you want to add)
16. Tap on Generate Variation
17. Check that the variations are created successfully

### Images

| Adding variants | Success response |
| ------------- | ------------- |
|   |   |

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/BatchProductVariationsApiResponse.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/BatchProductVariationsApiResponse.kt
@@ -6,7 +6,11 @@ import org.wordpress.android.fluxc.network.Response
 /**
  * Representation of Batch Update Product Variations API response.
  */
-class BatchProductVariationsUpdateApiResponse : Response {
+class BatchProductVariationsApiResponse : Response {
+    @SerializedName("create")
+    var createdVariations: List<ProductVariationApiResponse>? = null
     @SerializedName("update")
     var updatedVariations: List<ProductVariationApiResponse>? = null
+    @SerializedName("delete")
+    var deletedVariations: List<Long>? = null
 }
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/BatchProductVariationsApiResponse.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/BatchProductVariationsApiResponse.kt
@@ -6,11 +6,8 @@ import org.wordpress.android.fluxc.network.Response
 /**
  * Representation of Batch Update Product Variations API response.
  */
-class BatchProductVariationsApiResponse : Response {
-    @SerializedName("create")
-    var createdVariations: List<ProductVariationApiResponse>? = null
-    @SerializedName("update")
-    var updatedVariations: List<ProductVariationApiResponse>? = null
-    @SerializedName("delete")
-    var deletedVariations: List<Long>? = null
-}
+data class BatchProductVariationsApiResponse(
+    @SerializedName("create") val createdVariations: List<ProductVariationApiResponse>? = null,
+    @SerializedName("update") val updatedVariations: List<ProductVariationApiResponse>? = null,
+    @SerializedName("delete") val deletedVariations: List<Long>? = null
+) : Response
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -76,6 +76,7 @@ import org.wordpress.android.fluxc.store.WCProductStore.RemoteUpdatedProductPass
 import org.wordpress.android.fluxc.store.WCProductStore.RemoteVariationPayload
 import org.wordpress.android.fluxc.utils.handleResult
 import org.wordpress.android.fluxc.utils.putIfNotEmpty
+import org.wordpress.android.fluxc.utils.putIfNotNull
 import javax.inject.Inject
 import javax.inject.Named
 import javax.inject.Singleton
@@ -932,34 +933,50 @@ class ProductRestClient @Inject constructor(
 
     /**
      * Makes a POST request to `/wp-json/wc/v3/products/[WCProductModel.remoteProductId]/variations/batch`
-     * to batch update product variations.
+     * This API helps you to batch create, update and delete multiple product variations.
      *
+     * @param [site] The site containing the product
      * @param productId Id of the product.
-     * @param variationsIds Ids of variations that are going to be updated.
-     * @param modifiedProperties Map of the properties of variation that are going to be updated.
-     * Keys correspond to the names of variation properties. Values are the updated properties values.
+     * @param createVariations list of product variations to create.
+     * The Map keys correspond to the names of variation properties.
+     * The Map values are the variation properties values.
+     * @param updateVariations list of product variations to update.
+     * The Map keys correspond to the names of variation properties.
+     * The Map values are the variation properties values.
+     * @param deleteVariations list of product variations ids to delete.
      *
-     * @return Instance of [BatchProductVariationsUpdateApiResponse].
+     * @return Instance of [BatchProductVariationsApiResponse].
      */
     suspend fun batchUpdateVariations(
         site: SiteModel,
         productId: Long,
-        variationsIds: Collection<Long>,
-        modifiedProperties: Map<String, Any>
-    ): WooPayload<BatchProductVariationsUpdateApiResponse> = WOOCOMMERCE.products.id(productId).variations.batch.pathV3
-        .let { url ->
-            val variationsUpdates: List<Map<String, Any>> = variationsIds.map { variationId ->
-                modifiedProperties.toMutableMap()
-                    .also { properties -> properties["id"] = variationId }
+        createVariations: List<Map<String, Any>>? = null,
+        updateVariations: List<Map<String, Any>>? = null,
+        deleteVariations: List<Long>? = null
+    ): WooPayload<BatchProductVariationsApiResponse> =
+        WOOCOMMERCE.products.id(productId).variations.batch.pathV3
+            .let { url ->
+                val body = buildMap {
+                    putIfNotNull("create" to createVariations)
+                    putIfNotNull("update" to updateVariations)
+                    putIfNotNull("delete" to deleteVariations)
+                }
+
+                if (body.isEmpty()) {
+                    throw IllegalArgumentException(
+                        "At least one of createVariations, updateVariations or deleteVariations" +
+                            " should not be null"
+                    )
+                }
+
+                jetpackTunnelGsonRequestBuilder.syncPostRequest(
+                    this@ProductRestClient,
+                    site,
+                    url,
+                    body,
+                    BatchProductVariationsApiResponse::class.java
+                ).handleResult()
             }
-            jetpackTunnelGsonRequestBuilder.syncPostRequest(
-                this@ProductRestClient,
-                site,
-                url,
-                mapOf("update" to variationsUpdates),
-                BatchProductVariationsUpdateApiResponse::class.java
-            ).handleResult()
-        }
 
     /**
      * Makes a POST request to `/wp-json/wc/v3/products` to create
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -962,12 +962,7 @@ class ProductRestClient @Inject constructor(
                     putIfNotNull("delete" to deleteVariations)
                 }
 
-                if (body.isEmpty()) {
-                    throw IllegalArgumentException(
-                        "At least one of createVariations, updateVariations or deleteVariations" +
-                            " should not be null"
-                    )
-                }
+                require(body.isNotEmpty())
 
                 jetpackTunnelGsonRequestBuilder.syncPostRequest(
                     this@ProductRestClient,
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -172,6 +172,43 @@ class WCProductStore @Inject constructor(
         val variation: WCProductVariationModel
     ) : Payload<BaseNetworkError>()
 
+    class BatchGenerateVariationsPayload(
+        val site: SiteModel,
+        val remoteProductId: Long,
+        val variations: List<List<Map<String, Any>>>
+    ) : Payload<BaseNetworkError>() {
+        class Builder(
+            private val site: SiteModel,
+            private val remoteProductId: Long
+        ) {
+            private val variations = mutableListOf<List<Map<String, Any>>>()
+            fun getVariations() = variations.toList()
+
+            private val currentAttributes = mutableListOf<Map<String, Any>>()
+            fun getCurrentAttributes() = currentAttributes.toList()
+
+            fun addVariationAttribute(id: Long, name: String, option: String) {
+                currentAttributes.add( buildMap {
+                    put("id", id)
+                    put("name", name)
+                    put("option", option)
+                })
+            }
+
+            fun addVariation() {
+                if(currentAttributes.isEmpty()) return
+                variations.add(currentAttributes.toList())
+                currentAttributes.clear()
+            }
+
+            fun build() = BatchGenerateVariationsPayload(
+                site = site,
+                remoteProductId = remoteProductId,
+                variations = variations
+            )
+        }
+    }
+
     /**
      * Payload used by [batchUpdateVariations] function.
      *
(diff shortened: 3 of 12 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -932,34 +933,50 @@ class ProductRestClient @Inject constructor(
...
-        modifiedProperties: Map<String, Any>
-    ): WooPayload<BatchProductVariationsUpdateApiResponse> = WOOCOMMERCE.products.id(productId).variations.batch.pathV3
-        .let { url ->
-            val variationsUpdates: List<Map<String, Any>> = variationsIds.map { variationId ->
-                modifiedProperties.toMutableMap()
-                    .also { properties -> properties["id"] = variationId }
+        createVariations: List<Map<String, Any>>? = null,
+        updateVariations: List<Map<String, Any>>? = null,
+        deleteVariations: List<Long>? = null
+    ): WooPayload<BatchProductVariationsApiResponse> =
```
atorresveiga, line 956: Added the missing parameters to the batch update request. The`createVariations` was needed to create the new variations (our goal), and the `deleteVariations` was added to match the API request. Now the rest client call looks like API request and can be used as documentation.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -172,6 +172,43 @@ class WCProductStore @Inject constructor(
         val variation: WCProductVariationModel
     ) : Payload<BaseNetworkError>()
 
+    class BatchGenerateVariationsPayload(
```
atorresveiga, line 176: Named the Payload and the function `generate` instead of `create` with the intent to help future maintainers to realize that this function will only receive the `attributes` property. The rest of the variations properties will use their default values. The idea is to leave room for a future `create` function that will set the rest of the properties of the variations.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/BatchProductVariationsApiResponse.kt
```diff
@@ -6,7 +6,11 @@ import org.wordpress.android.fluxc.network.Response
...
  * Representation of Batch Update Product Variations API response.
  */
-class BatchProductVariationsUpdateApiResponse : Response {
+class BatchProductVariationsApiResponse : Response {
+    @SerializedName("create")
+    var createdVariations: List<ProductVariationApiResponse>? = null
     @SerializedName("update")
     var updatedVariations: List<ProductVariationApiResponse>? = null
+    @SerializedName("delete")
+    var deletedVariations: List<Long>? = null
```
wzieba: I know you followed the pattern here but I think we could safely change those fields to `val` as they shouldn't be mutable.

I also think this class could be a `data` one as it's a data container.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -172,6 +172,43 @@ class WCProductStore @Inject constructor(
...
+        val site: SiteModel,
+        val remoteProductId: Long,
+        val variations: List<List<Map<String, Any>>>
+    ) : Payload<BaseNetworkError>() {
+        class Builder(
+            private val site: SiteModel,
+            private val remoteProductId: Long
+        ) {
+            private val variations = mutableListOf<List<Map<String, Any>>>()
+            fun getVariations() = variations.toList()
```
wzieba: ❓ Why do we need this method?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -172,6 +172,43 @@ class WCProductStore @Inject constructor(
...
+
+            private val currentAttributes = mutableListOf<Map<String, Any>>()
+            fun getCurrentAttributes() = currentAttributes.toList()
```
wzieba: ❓ Why do we need this method?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -172,6 +172,43 @@ class WCProductStore @Inject constructor(
         val variation: WCProductVariationModel
     ) : Payload<BaseNetworkError>()
 
+    class BatchGenerateVariationsPayload(
+        val site: SiteModel,
+        val remoteProductId: Long,
+        val variations: List<List<Map<String, Any>>>
```
wzieba: Can we create a new class to help understand this nested structure for API consumers? Without deep knowledge of variations, I can't know what parameters I should put - where's the attribute name and where's the term value e.g.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -172,6 +172,43 @@ class WCProductStore @Inject constructor(
...
+            private val currentAttributes = mutableListOf<Map<String, Any>>()
+            fun getCurrentAttributes() = currentAttributes.toList()
+
+            fun addVariationAttribute(id: Long, name: String, option: String) {
+                currentAttributes.add( buildMap {
+                    put("id", id)
+                    put("name", name)
+                    put("option", option)
+                })
+            }
```
wzieba: This does not precisely add a variation attribute. "Attribute" is the whole thing: its name and all its options (terms).

In WooCommerce, the thing we add here is named "TermAssignment" (it's my proposal).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -172,6 +172,43 @@ class WCProductStore @Inject constructor(
...
+
+            fun addVariation() {
+                if(currentAttributes.isEmpty()) return
+                variations.add(currentAttributes.toList())
+                currentAttributes.clear()
+            }
```
wzieba: Why do we need this method? I mean I understand what it does, but can't we simplify the `BatchGenerateVariationsPayload` by just passing received `variations`? This reduces overhead for library consumers.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -172,6 +172,43 @@ class WCProductStore @Inject constructor(
...
+            private val site: SiteModel,
+            private val remoteProductId: Long
+        ) {
+            private val variations = mutableListOf<List<Map<String, Any>>>()
+            fun getVariations() = variations.toList()
+
+            private val currentAttributes = mutableListOf<Map<String, Any>>()
+            fun getCurrentAttributes() = currentAttributes.toList()
+
+            fun addVariationAttribute(id: Long, name: String, option: String) {
```
wzieba: ~⚙️ We shouldn't need to provide an `id` here. It sounds like a core responsibility - it'll be quite challenging to do this on the mobile side. Is it something that Core requests or can we skip it?~

wzieba: Sorry! I'm wrong. It's `id` of attribute right? So we don't have to generate it, we have it from API. 

So if we have `id` of the `attribute` do we need the `name`? I think we shouldn't.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: > This does not precisely add a variation attribute. "Attribute" is the whole thing: its name and all its options (terms)

I believe this is true for the attribute property of a **Product**, which is different than the attribute property of a **Variation**:

#### Product ` /wc/v3/products/847/&_method=get`
```
{
 id: 2
 name: "Size"
▶options: ["S", "M", "L"]
 position: 0
 variation: true
 visible: true
}
```
-------------------
#### Variation  `/wc/v3/products/847/variations/&_method=get`
```
{
 id: 2
 name: "Size"
 option: "L"
}
```


> In WooCommerce, the thing we add here is named "TermAssignment" (it's my proposal).

Not sure about this. IMO using a different name than the one used by the API could be hard to follow, and I feel that the name `option` fits well with what we are trying to describe here. Actually, _**option**_ is the name used by other apps to represent this concept. That said, if you are strong about using **_term_** instead of **_option_**. I will be happy to change it.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/BatchProductVariationsApiResponse.kt
atorresveiga: I was hesitant about using data classes because when I checked the product API response classes, they all used this pattern of a class with vars and an empty constructor. However, I agree that a data class here is a better representation.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: >  Is it something that Core requests or can we skip it?

Yes, we need to provide the attribute `id` here; otherwise, the variation won't be added accordingly to our plan.

Let me walk you through an example using the data from the attributes of the product above.
In my store, I have a variable product with the attribute Size. The attribute property looks like this:

```
{
 id: 2
 name: "Size"
▶options: ["S", "M", "L"]
 position: 0
 variation: true
 visible: true
}
```
*Note that the attribute id is 2*

If I sent a request without the id, Core will create an empty variation like in the image above:



Bonus test, If I change the name the variation would be created with the right name:

wzieba: Yes, thanks for the clarification. But can we then remove `name`?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
wzieba: > I believe this is true for the attribute property of a Product, which is different than the attribute property of a Variation:

So you suggest:

- Leave `com.woocommerce.android.model.ProductAttribute` as it list: with `name: String` and `terms: List<String>`
- Introduce `com.woocommerce.android.model.VariationAttribute` which would look like:
```kotlin
data class VariationAttribute(
    val productAttributeName: String,
    val productAttributeId: Long,
    val selectedProductTerm: String
)
```
right? I'm okay with that although I'd consider if we also need `productAttributeName` if we have `productAttributeId` (this [comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2569#discussion_r1025489174))

I'd keep naming the selection of attribute a "term" for reasons I've described [here](https://github.com/woocommerce/woocommerce-android/pull/7864#issuecomment-1319307375).

> Not sure about this.

I also didn't like `TermAssignment` 😅 .

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/BatchProductVariationsApiResponse.kt
wzieba: Yeah, let's maybe leave it with the same pattern as rest of the API responses to not worry about it 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
wzieba: I just found there's `com.woocommerce.android.model.VariantOption`, so I think we should stick to that instead of a new `VariationAttribute`, right?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: > Why do we need this method? I mean I understand what it does, but can't we simplify the BatchGenerateVariationsPayload by just passing received variations?

The idea was to provide a Builder to help with the creation of the Payload class.  I used those methods on the example tests and thought it could be helpful when we plan to _build_ the payload (step by step) and not only pass the list of variations. The `getCurrentAttributes()` and `getCurrentAttributes()` were helper methods to check the payload builder status and see how many variations were in the payload builder and, how many attributes those variations have. 
One thing is the Builder a class responsible for helping build the Payload, which abstracts the building phase, and the other is the Payload, which do receives the variations in the constructor in the form of `val variations: List<List<Map<String, Any>>>`. I will take a look at the `VariationAttribute` that you recommended and refactor this class to make it more readable.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: Answered [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2569#discussion_r1026309567)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: Answered [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2569#discussion_r1026309567)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: Thanks for the review @wzieba!🙏🏻 
I'll work on refactoring the Payload and the Builder class and let you know when the code is ready for another round.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
wzieba: Ok I see: I've used it as intended (I think) in this PR: https://github.com/woocommerce/woocommerce-android/pull/7893

If the consumer should use `Builder` can we make constructor of `BatchGenerateVariationsPayload` private (or internal) so there will be no doubt how to use it?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
wzieba: Added comment in different PR related to that: https://github.com/woocommerce/woocommerce-android/pull/7897#discussion_r1026319645

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: > But can we then remove name?

Is it possible to remove it? Yes, by the test, one can see that it is not used by Core at all. 
Still, I will vote to send the whole object because I don't have the context of why Core is not using it.
My concern is that they start using it in the future, and this breaks the app. This kind of change, breaking the API contract, will be hard to track in the future if you don't have the full context, and I don't see a huge impact on not sending the value. Not sure if I'm missing something here. WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
wzieba: > The getCurrentAttributes() and getCurrentAttributes() were helper methods to check the payload builder status and see how many variations were in the payload builder and, how many attributes those variations have.

I see, thanks for clarification. Ideally, those methods should be `internal`, but I understand this is challenging because of `example` module being just a separate module.

Can we annotate those two methods by `org.jetbrains.annotations.ApiStatus.Internal`?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
wzieba: I think we won't break API contract here; contrary, we'll satisfy it. Looking at the example in the [documentation](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2569#discussion_r1026331922):
image

the example shows we send only `id` and `option` values. Not sending `name` seems reasonable because `id` always points to one concrete `name` - it's functional dependant.

It's not a huge deal, I agree. But if we want to follow the API, I think we should drop `name` (and reduce payload size as a bonus). WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
wzieba: As you explained here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2569#discussion_r1026309567 I think we should rather make the constructor of the `BatchGenerateVariationsPayload` private.

If it'll be private, I think we don't need to provide new class - it's not exposed to consumers.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: I found `org.wordpress.android.fluxc.model.ProductVariantOption` and will stick to this instead of creating a new class. Here is the refactor using this class [b2770cc](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2569/commits/b2770cc083e764117715e8312501026ca5006796)

atorresveiga:
Hi @wzieba! This PR is ready for another round 😄

wzieba (approved):
Looks good! Thank you for all the changes and discussions that brought improvements in both FluxC and Woo codebases regarding this feature 🙏 .

Let's check what's wrong with CI and merge this PR 👍

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: Sorry, I missed these comments and refactored the class 😬

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
atorresveiga: Good catch. I missed that in the example we send only `id` and `option` values. I'll create a new issue to improve this, as we discussed internally.

atorresveiga: Issue created here https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/2570

PR #2576, opened 2022-11-24
Adds Jetpack migration API call

antonis:
Adds a call to the `wpcom/v2/mobile/migration` endpoint (ref `D91729-code`)

To test:
Check steps in https://github.com/wordpress-mobile/WordPress-Android/pull/17531


develric (approved):
Tested with the companion WP PR and works as expected. Also code LGTM 👍

PR #2577, opened 2022-11-24
Add batch product update endpoint

### LINKED ISSUE #7931: [FluxC] Support bulk update of product fields
### Description

This feature is about networking and persistence layers changes to allow the Woo mobile app to perform bulk updates of many products. We want to support editing:
- Status
- Price

Api reference: https://woocommerce.github.io/woocommerce-rest-api-docs/#batch-update-products

wzieba:
Closes: https://github.com/woocommerce/woocommerce-android/issues/7931

For testing instructions, please see Woo PR: https://github.com/woocommerce/woocommerce-android/pull/7965


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponse.kt
```diff
@@ -4,89 +4,76 @@ import com.google.gson.JsonArray
...
-    var variations: JsonElement? = null
-    var downloads: JsonElement? = null
-    var related_ids: JsonElement? = null
-    var cross_sell_ids: JsonElement? = null
-    var upsell_ids: JsonElement? = null
-    var grouped_products: JsonElement? = null
+typealias ProductDto = ProductApiResponse
+
+@Suppress("ConstructorParameterNaming")
+data class ProductApiResponse(
```
atorresveiga, line 12: Nice!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -931,6 +931,30 @@ class ProductRestClient @Inject constructor(
...
     }
 
+    suspend fun batchUpdateProducts(
+        site: SiteModel,
+        existingProducts: List<WCProductModel>,
+        updatedProducts: List<WCProductModel>
+    ): WooPayload<BatchProductApiResponse> = WOOCOMMERCE.products.batch.pathV3
+        .let { url ->
+            val body = buildMap {
+                putIfNotNull("update" to (existingProducts zip updatedProducts).map { (existing, updated) ->
```
atorresveiga: The `zip` operation will return a list of pairs taking into account the indexes of the two collections. I believe we must sort these two collections before applying the zip to ensure that the products have the same indexes and that, with the `zip` operation, the pair will be formed with the `same` product.

In the example above, the item at index `0` is different in both collections (updateProducts vs existingProducts)

Screen Shot 2022-12-01 at 08 39 59

Screen Shot 2022-12-01 at 08 41 34

wzieba: Thanks @atorresveiga , good point! Fixed in bb41344e2ff855061c8f8ddeb10160465c1501f8

atorresveiga:
Great work @wzieba! 
Kudos for refactoring the `ProductApiResponse` and converting it to a data class, and using `productModelToProductJsonBody` to send only the necessary field to update.
I left one comment about sorting the list to ensure `updateProducts` and `existingProducts` have the same order. Maybe we can unit-test this and add other tests to the `batchUpdateProducts`.🤔
It would be nice to add this batch update function to the example app on FluxC, so we can test the endpoint in isolation without needing the Woo app. Something that is not required for this PR, but a nice to have for the future.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
atorresveiga: Wow! That was fast 🚀. Thanks

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -931,6 +931,33 @@ class ProductRestClient @Inject constructor(
...
+        .let { url ->
+            val sortedExistingToUpdatedProducts = existingProducts
+                .sortedBy(WCProductModel::remoteProductId)
+                .zip(updatedProducts.sortedBy(WCProductModel::remoteProductId))
+
+            val body = buildMap {
+                putIfNotNull("update" to sortedExistingToUpdatedProducts.map { (existing, updated) ->
+                        productModelToProductJsonBody(
+                            productModel = existing, updatedProductModel = updated
+                        ).plus("id" to updated.remoteProductId)
```
atorresveiga: np: Adding the `plus` without checking if the map is empty will always send the product, even if it doesn't have any change. WDYT, if we check the map size first and send only products with changes?

Screen Shot 2022-12-01 at 10 21 44

wzieba:
> I left one comment about sorting the list to ensure updateProducts and existingProducts have the same order. Maybe we can unit-test this and add other tests to the batchUpdateProducts.🤔

Thanks. I've moved sorting from `RestClient` to `Store` (as I think it's rather `Store` responsibility to prepare correct data for `RestClient` to operate) and that allowed me to add a unit test.

I've also added your second improvement: 
>check the map size first and send only products with changes

Ready for the second round!

atorresveiga:
@wzieba I tested the code and found this:

Screen Shot 2022-12-02 at 13 16 03

Screen Shot 2022-12-02 at 14 03 53

Note that in this example, the `id` property is not added to the HashMap.

Maybe something like this could do the trick:
```
putIfNotNull("update" to existingToUpdatedProducts.mapNotNull { (existing, updated) ->
    val jsonBody = productModelToProductJsonBody(
        productModel = existing, updatedProductModel = updated
    )
    if (jsonBody.isNotEmpty()) {
        jsonBody.plus("id" to updated.remoteProductId)
    }else{
        null
    }
}
```

wzieba:
Thanks, @atorresveiga , you're entirely right. I've rushed with this change.

In 36da5cc I've added unit tests and then adjusted behavior in c575c8d. It should work fine now.

atorresveiga:
Great work @wzieba! Nice touch of adding unit tests to the request generation. I tested the code, and it is working great. I've one last nitpick advice, though. Maybe we can use `existingToUpdatedProducts.mapNotNull` instead of `existingToUpdatedProducts.map`. This way, only values that are different than `null` will be added to the list of products to update. 
Using `map` could end with a request/response like the one below, sending a `null` value that is not required and, because of that, receiving an error response between the results:
Screen Shot 2022-12-05 at 09 51 43

wzieba:
Thank you again @atorresveiga , fix applied in 070238ffedee775ef8e8f1c50d8f45ae4cb651b3

atorresveiga:
Great work, @wzieba! 
LGTM! :shipit:

wzieba:
You have to approve the PR @atorresveiga to make all check green 😊

atorresveiga (approved):
LGTM!

PR #2582, opened 2022-11-28
Jetpack focus add support for remote configuration for phases

### LINKED ISSUE #17331: Jetpack focus: Add support for remote configuration for Phases
Description

Retrieve remote values from the new remote-config endpoint

- Add support for getting the remote config values
- Implement the logic to fetch remote field configs

Ref: pe7hp4-4k-p2

AjeshRPai:
Closes https://github.com/wordpress-mobile/WordPress-Android/issues/17331

This PR adds the endpoint and the logic to retrieve the remote config values. 

This PR has the following changes
1. Adds remote config endpoint in wp-com-v2-endpoints.txt 
2. Adds class `RemoteConfigRestClient.kt` - contains the logic to fetch the remote config from endpoint and return Error or Remote config depending on API response
3. Adds `RemoteConfigRestClientTest.kt` - contains the unit test for `RemoteConfigRestClient`
4. Adds `RemoteConfigStore` - Gets the result from the Rest client and wraps it an RemoteConfigResult class for the Data layer 
5. Adds `RemoteConfigStoreTest.kt` - Unit tests for the logic in `RemoteConfigStore`

There is not much to test yet - just make sure the CI passes. Also make sure the endpoint and response structure is correct as per the project spec. Ref: pe7hp4-4k-p2


zwarm (approved):
@AjeshRPai - Everything looks good and the tests all pass. 👍

PR #2583, opened 2022-11-29
Add support for connecting to self-hosted sites without XMLRPC

hichamboushaba:
The PR adds a new way for fetching sites directly using the [Cookie authentication](https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/), this will allow two things:
1. Ability to call the Rest API using Cookie authentication for sites without XMLPPC (for example: plugin installation to allow Jetpack Connection).
2. Ability to create application passwords in the future.

The changes of the PR are:
1. Add a new `Origin` value to the `SiteModel`.
2. Create a new rest client `SiteWPAPIRestClient` that creates a `SiteModel` instance using the `/wp/v2/settings` response, and uses the Cookie authentication.
3. Integrates the new rest client to `SiteStore` and add a new action to fetch a site using Cookie Authentication.
4. Update the example app to test the new flow.
5. Update the WPAPI requests to `emit` authentication errors when they are encountered.

#### Test
1. Open the example app.
2. Click on "Sign in and fetch sites"
3. Enter your wp-admin credentials and the site URL, and make sure the Checkbox is kept unchecked.
4. Click on OK.
5. Confirm that the site name is correctly printed in the console.
6. Click on Plugins.
7. Test fetching/installing some plugins.
8. Confirm it works as expected.

Bonus: you can also use a plugin (for example this [one](https://wordpress.org/plugins/disable-xml-rpc-api/)) to disable XMLPRC to confirm that these changes work without it.


hichamboushaba:
On Hold for now.

PR #2585, opened 2022-12-01
Woo: Remove hardcoded eCommerce plan purchase

0nko:
FluxC now provides a way to fetch plan data, which can be then used to make a purchase, so there is no reason to use the hardcoded values to add a monthly eCommerce plan to a cart. This allows for more flexibility in the future when we add options to buy different plans/periods.

**To test:**
1. Start the example app
2. Go to Woo -> Store creation
3. Select a site
4. Tap on the Add plan to cart button
5. Verify a response with a cart content is returned in the console


JorgeMucientes (approved):

JorgeMucientes:
Looks good :shipit:

PR #2586, opened 2022-12-01
Add support for WordPress Application Passwords 

hichamboushaba:
Part of #2588 

This PR adds support for Application Passwords in FluxC, it does so by providing a class `ApplicationPasswordNetwork` that handles the network requests:
1. It uses the `ApplicationPasswordManager` to get or create a new password if needed.
2. The class `ApplicationPasswordManager` handles the password creation for both self-hosted sites using site credentials, or for Jetpack sites using the Jetpack tunnel.
3. The creation flow follows the internal documentation: pe5sF9-U3-p2

In the last [commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2586/commits/a8c58dba2287ab6ebb4349cfcae30d4a135e9f25), I updated the visibility of everything related to the implementation of the feature to `internal`, this way it won't leak to clients, the exception is `ApplicationPasswordNetwork` as it's needed in the WooCommerce module.

For testing, please refer to the PR #2587


hafizrahman:
Looking...

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordRestClient.kt
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        return if (response is WPAPIResponse.Error && response.error.volleyError?.networkResponse?.statusCode == 401) {
+            AppLog.w(
+                AppLog.T.MAIN,
+                "Authentication failure using application password, maybe revoked?" +
+                    " Delete the saved one then retry"
+            )
+            applicationPasswordManager.deleteLocalApplicationPassword(site)
+            executeGsonRequest(site, method, path, clazz, params, body)
```
hafizrahman: Can this create an infinite loop in some edge case?

hichamboushaba: Good point, this is a point I've been thinking off, and I'm still indecisive about it.
The logic shouldn't result in a infinite loop, as we check only for `401` error there, but maybe we can also add a check to avoid doing multiple retries, I'll add it later today.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordRestClient.kt
```diff
@@ -0,0 +1,118 @@
...
+
+private const val AUTHORIZATION_HEADER = "Authorization"
+
+@Singleton
+class ApplicationPasswordRestClient @Inject constructor(
+    @Named("no-cookies") private val requestQueue: RequestQueue,
+    private val userAgent: UserAgent,
+    private val applicationPasswordManager: ApplicationPasswordManager
+) {
+    suspend fun <T> executeGsonRequest(
```
hichamboushaba: I followed a different pattern for this REST Client, instead of passing a Request, the client exposes `suspend` functions, I think this is easier to understand than the other flow, and also allows hiding the implementation details properly, for example if we want to use something other than `Volley` for this client, it can be easily achievable, as we won't have anything about the implementation leaking to the feature `RestClients`, unlike before where the `Request` class was specifically a `Volley` request.

Let me know what you think about?
And let me know what you think about the naming? is calling it `RestClient` while still use it inside other `RestClient`s is OK and clear?

hafizrahman: > I followed a different pattern for this REST Client, instead of passing a Request,

I understand the approach and I like it. My feeling is that it might not be immediately intuitive for people coming in who already have existing knowledge of how the REST API previously works (and that's pretty much all of us). Maybe add your explanation into the code as well?

> And let me know what you think about the naming? is calling it RestClient while still use it inside other RestClients is OK and clear?

Using CouponRestClient for example, I think the naming structure now is similar but implies two different meaning:
- `CouponRestClient` : A REST client related to coupon areas
- `ApplicationPasswordRestClient`: Not a REST client related to application password areas (the class for that is instead ` WPApiApplicationPasswordGenerator`). Instead, this is a REST client for passing any kind of areas' functionalities using application password method.

Maybe we can rename it better to imply the passing part? `ApplicationPasswordBasedRestClient` or something?

hichamboushaba: > I understand the approach and I like it. My feeling is that it might not be immediately intuitive for people coming in who already have existing knowledge of how the REST API previously works (and that's pretty much all of us). Maybe add your explanation into the code as well?

For this, I feel like the code should document itself, it's not intuitive yet, we need to figure out why.
I'm biased as the one who wrote it, but it feels similar to how other libraries would work, it exposes just the building parts to execute basic requests, and doesn't expose anything about the implementation details, so that clients can build upon it.

> Maybe we can rename it better to imply the passing part? ApplicationPasswordBasedRestClient or something?

WDYT about `ApplicationPasswordNetwork`? this way it's clear that it's not a rest client, but a network implementation that Rest Clients need to use.

cc @JorgeMucientes @0nko, if you can give your opinion too about the two above points 🙏, I don't want to bother a lot of people with a ping about this, so let's limit it to the Activate team, and we'll trust our decision will be good enough 🙂.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordRestClient.kt
hichamboushaba: This issue should be solved now with fbce938c9009c3660b89e3fd529707d8aefcdffc, let me know if you have any thoughts.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackApplicationPasswordGenerator.kt
```diff
@@ -0,0 +1,119 @@
...
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackError
+import org.wordpress.android.fluxc.network.rest.wpcom.jetpacktunnel.JetpackTunnelGsonRequestBuilder.JetpackResponse.JetpackSuccess
+import org.wordpress.android.util.AppLog
+import org.wordpress.android.util.AppLog.T
+import javax.inject.Inject
+import javax.inject.Named
+import javax.inject.Singleton
+
+@Singleton
+class JetpackApplicationPasswordGenerator @Inject constructor(
```
0nko: This is just a nitpick, but when I saw the word `Generator` in other parts of the codebase the first thing that came to my mind was that it was some cryptographic key generation function, when really it's just a REST client, which also handles fetching and deletion.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordRestClient.kt
0nko: I like the `ApplicationPasswordNetwork` better. I left a comment about the use of `Generator` name before I read this thread. Maybe those could then be renamed?

0nko (commented):
Great work, @hichamboushaba! Please see my comments in the code.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/JetpackApplicationPasswordGenerator.kt
hichamboushaba: I agree, the name `Generator` is a leftover from an early attempt (at this [commit](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2586/commits/5f1284a9e0be592eba02de8db5b6af7a82eb5cbf)).
I'll rename them now.

Thanks @0nko.

JorgeMucientes:
Awesome job @hichamboushaba 🙌🏼  Code is easy to follow, and I can't think of anything concerning about the current implementation. If any, I wish Kotlin had package private visibility to ensure only `ApplicationPasswordNetwork` is exposed to the "outside FluxC world".  

I like the naming changes applied, good call in the `ApplicationPasswordNetwork` to avoid confusion with other rest clients.

JorgeMucientes (approved):

hichamboushaba:
Thanks for the review @JorgeMucientes .

> If any, I wish Kotlin had package private visibility to ensure only ApplicationPasswordNetwork is exposed to the "outside FluxC world".

I agree, the `internal` visibility is just not enough for some cases, but the language designers disagree, so 🤷.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordNetwork.kt
```diff
@@ -0,0 +1,118 @@
...
+        params: Map<String, String> = emptyMap(),
+        body: Map<String, Any> = emptyMap()
+    ): WPAPIResponse<T> {
+        val credentialsResult = applicationPasswordManager.getApplicationCredentials(site)
+        val credentials = when (credentialsResult) {
+            is ApplicationPasswordCreationResult.Existing -> credentialsResult.credentials
+            is ApplicationPasswordCreationResult.Created -> credentialsResult.credentials
+            is ApplicationPasswordCreationResult.Failure ->
+                return WPAPIResponse.Error(credentialsResult.error)
+            ApplicationPasswordCreationResult.NotSupported -> TODO()
```
hichamboushaba, line 43: The PR is already big enough, I'll work on this on an upcoming PR.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsStore.kt
```diff
@@ -0,0 +1,86 @@
...
+            createPrefs()
+        } catch (e: Exception) {
+            // In case we can't decrypt the file after a backup, let's delete it
+            AppLog.d(
+                AppLog.T.MAIN,
+                "Can't decrypt encrypted preferences, delete it and create new one"
+            )
+            context.deleteSharedPreferences(filename)
+            createPrefs()
+        }
```
hichamboushaba, line 78: To test this case, you can follow the following steps:
1. Use the CouponStore PR #2587 
2. Update the code in the catch block to throw it.
3. Install the app, then follow the step of the CouponStore PR (until the step 5).
4. Open Android Studio, then click on "Device File Explorer"
5. Navigate to `/data/data/org.wordpress.android.fluxc.example/shared_prefs`
6. Open the file `flux-example-android-encrypted-prefs`, then copy it to your laptop.
7. Go back to the mobile phone, then clear the app's data.
8. Re-open the app.
9. Open Android Studio, then use "Device File Explorer" to put the file `flux-example-android-encrypted-prefs` into the path `/data/data/org.wordpress.android.fluxc.example/shared_prefs`
10. Repeat steps of the CouponStore PR.
11. Notice that the app crashes.
12. Put back the `catch` block.
13. Repeat the steps, and confirm that it works as expected.

hichamboushaba:
This is ready for a final review @hafizrahman, but let's not merge it until we confirm it doesn't break anything in the client apps.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordRestClient.kt
hafizrahman: I like `ApplicationPasswordNetwork` too.

hafizrahman (approved):

hichamboushaba:
I tested the PR in the clients, and confirmed there is no breaking changes in it, I'll merge it now.

PR #2589, opened 2022-12-05
Application passwords: Add listener for detecting when the feature is unavailable

hichamboushaba:
This PR handles two changes:
1. Exposes the error `code` from the WPApi responses (check below)
2. Adds support for detecting when the Application Passwords are not available, it's done through an optional listener `ApplicationPasswordsUnavailableListener` that clients can add to the Dagger graph.

For part 1, I had to update the `WPAPIResponse` and the networking requests to expose a new subtype `WPAPINetworkError`, that's what the 5 first commits of the PR do, reviewing them separately will make the understanding easier.

#### Testing
For testing, please check the WCAndroid PR: https://github.com/woocommerce/woocommerce-android/pull/7989


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/OnWPAPIErrorListener.kt
```diff
@@ -0,0 +1,16 @@
...
+fun interface OnWPAPIErrorListener {
+    fun onErrorResponse(error: WPAPINetworkError)
+}
+
+class WPAPIErrorListenerWrapper(
+    private val listener: OnWPAPIErrorListener
+) : BaseRequest.BaseErrorListener {
+    override fun onErrorResponse(error: BaseNetworkError) {
+        listener.onErrorResponse(error as WPAPINetworkError)
+    }
```
hafizrahman, line 15: Just an aside that this wrapper pattern is cool and is something I looked for when doing REST API POC but wasn't able to quite figure out how to structure it.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordNetwork.kt
```diff
@@ -39,7 +42,7 @@ class ApplicationPasswordNetwork @Inject constructor(
             is ApplicationPasswordCreationResult.Existing -> credentialsResult.credentials
             is ApplicationPasswordCreationResult.Created -> credentialsResult.credentials
             is ApplicationPasswordCreationResult.Failure ->
-                return WPAPIResponse.Error(credentialsResult.error)
+                return WPAPIResponse.Error(credentialsResult.error.toWPAPINetworkError())
```
hafizrahman, line 53: Out of curiosity, why add a converter here, instead of converting `ApplicationPasswordCreationResult.error` to be a type of `WPAPINetworkError`?

hichamboushaba: The error can come from either the WPApi (when using site credentials) or from WPCom (when using the Jetpack Tunnel), so we can't have the same one.

hafizrahman: Now I have some thoughts that's not related to this PR, more like it should be for the previous PR:

Initially I'm thinking why there's a part in this class where we might get a non-WP API error, that then gets "force" converted into WP API error even though it isn't. It feels a bit unintuitive that a supposedly WP API-specific class `ApplicationPasswordNetwork` has a non-WP API call involved. Someone using/reading `ApplicationPasswordNetwork` in the future might not intuitively get that it can be used for WPCom sites too.

So I wonder if we should document this behavior in the code? About why we have to process the credentials each time and that it can support WPcom sites? But then again this is just from my perspective where in a couple of months I probably would've forgotten why this is set this way. So perhaps this is just a problem specific to me, and writing things down here is enough for future me to remember 😅

example/src/main/java/org/wordpress/android/fluxc/example/ExampleApp.kt
```diff
@@ -24,6 +25,7 @@ open class ExampleApp : Application(), HasAndroidInjector {
         component.inject(this)
         val wellSqlConfig = WellSqlConfig(applicationContext, WellSqlConfig.ADDON_WOOCOMMERCE)
         WellSql.init(wellSqlConfig)
+        applicationPasswordsUnavailableLogger.init()
```
hafizrahman: Just to clarify: this is an example on how to implement the `ApplicationPasswordsUnavailableListener`, but it can't be tested yet in the Example app?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordNetwork.kt
hichamboushaba: If you open the `ApplicationPasswordsManager`, does it clarify the code and what it does or not?

I'm asking because there are two points here:
1. For someone who just needs to consume the `Network` class, they doesn't have to go into details about what it does, so they don't need to know about the error conversion itself, all they need to know is just that this `Network` class allows communicating with the site using Application Passwords.
2. For someone who needs to make any changes to the implementation details, they'll get to see `ApplicationPasswordsManager`, which IMO (and I may be biased as I said before) should be clear on what it does, and why it returns a `BaseNetworkError` and not a `WPAPINetworkError` here.

> It feels a bit unintuitive that a supposedly WP API-specific class ApplicationPasswordNetwork has a non-WP API call involved

One clarification about this, it's not a "non-WPAPI call", it's still a WPAPI call, but sent using the Jetpack Tunnel, so the error is still the same, just wrapped differently, and that's why we need to unwrap it and put it inside the `WPAPINetworkError`.

example/src/main/java/org/wordpress/android/fluxc/example/ExampleApp.kt
hichamboushaba: Yes, it can be tested once we merge these changes to the coupons PR, but since we are in a Hack Week, and there is also an example in the WCAndroid, I didn't think it was necessary to do the merge this week.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordNetwork.kt
hichamboushaba: And btw, thanks for challenging me on this point, if there are still any unclear points, we should address them.

hafizrahman: > If you open the `ApplicationPasswordsManager`, does it clarify the code and what it does or not?

Right, this is what I meant before (but clearly failed to explain properly).

The code is clear in _what_ it does, but to me possibly unclear _why_ it does things that way. In my mind possible questions from reading the code might be "why do we generate passwords differently?" or "why do we support wpcom site too while they already have Jetpack connected?"

I already understood the answers to those questions because we've had conversations in various P2 / Slack / Zoom about them. I'm just wondering if we should add the same "why" reasoning directly in the codebase for the people who do not participate in the same conversation, but might come across the code in the future.

Or generally speaking, my question is more about "do we need to include business decisions into code comments/documentation?"

> One clarification about this, it's not a "non-WPAPI call", it's still a WPAPI call, but sent using the Jetpack Tunnel, so the error is still the same, just wrapped differently, and that's why we need to unwrap it and put it inside the WPAPINetworkError.

This is awesome, and now I understand that part better :)

hichamboushaba: > my question is more about "do we need to include business decisions into code comments/documentation?"

Oh, I see, sure, as Paolo mentioned yesterday, we will be adding documentation of the feature later when we are done with the implementation, so the answer to your question, is yes, we need it and we'll add it.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsUnavailableListener.kt
```diff
@@ -0,0 +1,8 @@
+package org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords
+
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.rest.wpapi.WPAPINetworkError
+
+interface ApplicationPasswordsUnavailableListener {
+    fun featureIsUnavailable(siteModel: SiteModel, networkError: WPAPINetworkError)
```
hafizrahman: This function name's pattern "X is Y" makes it feel like a function that should return a boolean. 

Maybe something like `handleFeatureUnavailable` is clearer?

hichamboushaba: Normally the name convention if it returned a boolean would be "isXY", the convention starting with `is` is like asking a question and expecting an answer.
But still, I wasn't also satisfied with the namings here, WDYT about doing the following:

```kotlin
interface ApplicationPasswordsListener {
    fun featureUnavailable(siteModel: SiteModel, networkError: WPAPINetworkError)
}
```

This will make the `Listener` related to the feature as a whole and not linked to just the event, and might even be extended with other events if necessary in the future (I'm not sure we'll need them, but having the possibility would be better than forcing a single event).

I didn't add the `handle` prefix because I still think the listener role is not about the handling, but notifying the client app 🤔 

Let me know what you think @hafizrahman 🙏

hafizrahman: > I didn't add the handle prefix because I still think the listener role is not about the handling, but notifying the client app 🤔

Oh right. Looking at the example in the Woo PR makes me understand this better. I think I'm still struggling with understanding what this function does based on just the name. I looked at other interfaces in FluxC and saw a pattern of using `onSomething`. Do you think `onFeatureUnavailable` works here? 

I like the idea of renaming the interface, making it more extensible as you mentioned.

hichamboushaba: Done in [48d3548](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2589/commits/48d35489a2de7b2a0e248de85cd2491d6044ab94)

I also added a new event `onNewPasswordCreated` in [e536bd6](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2589/commits/e536bd654f16cf54ea98172827ae995b45ea7702), this could be helpful for client apps for something like logging or Tracks, let me know what you think.

If you don't mind, let's defer updating the WCAndroid PR after your review, this way I won't have to work on updating the FluxC hash twice.

hafizrahman (approved):
Thanks for explaning patiently to my questions :) That's all I have and things look good to me now.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsUnavailableListener.kt
hafizrahman: Great idea with `onNewPasswordCreated`!

PR #2590, opened 2022-12-10
Call to the API only when needed

atorresveiga:
### Why
We optimized the data we sent to the API when bulk-updating products. But when the optimization is optimal (no data needs to be sent), we should avoid sending that request and can, instead, show a success message.

https://user-images.githubusercontent.com/18119390/206819018-dd5d06cd-5990-40b6-a69b-cb62caf69872.mp4

### Description
This small PR checks if no change needs to be executed on the API before sending the request. If no change needs to be executed, it sends an early success response

### Testing instruction
You can use the [bulk update PR](https://github.com/woocommerce/woocommerce-android/pull/7971) to test the changes
1. Enter on selection mode
2. Tap on more -> Update status
3. Select the same status as the product selected
4. Check that no request is sent
5. Check that a success message is shown


ThomazFB (approved):
Sorry for the long time to review. I had to get some context before reviewing it 😬. LGTM!

atorresveiga:
Thanks for your review @ThomazFB 🙏🏻

PR #2594, opened 2022-12-13
Application passwords unit and connected tests

### LINKED ISSUE #2588: Application Password support
- [x] Application Password generation and networking. #2586 
- [x] Handling of sites that don't support application passwords. #2589 
- [x] Unit tests and connected tests. https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2594

hichamboushaba:
This PR adds unit and connected tests to the Application Passwords feature.
Check the comments for some explanations of some decisions.

#### Testing
Just confirm that the added tests are green.

### DIFF
--- a/.buildkite/unit-tests.sh
+++ b/.buildkite/unit-tests.sh
@@ -6,10 +6,11 @@ echo "--- :closed_lock_with_key: Installing Secrets"
 cp gradle.properties-example gradle.properties && cp -a example/properties-example/ example/properties/
 
 echo "--- :gradle: Run Tests"
-./gradlew example:testRelease plugins:woocommerce:testRelease || EXIT_CODE=$?
+./gradlew example:testRelease fluxc:testRelease plugins:woocommerce:testRelease || EXIT_CODE=$?
 
 echo -e "\n--- :buildkite: Upload Test Results to Test Analytics"
 find example/build/test-results/testReleaseUnitTest -name "TEST-*.xml" -exec upload_buildkite_test_analytics_junit {} $BUILDKITE_ANALYTICS_TOKEN_UNIT_TESTS \;
+find fluxc/build/test-results/testReleaseUnitTest -name "TEST-*.xml" -exec upload_buildkite_test_analytics_junit {} $BUILDKITE_ANALYTICS_TOKEN_UNIT_TESTS \;
 find plugins/woocommerce/build/test-results/testReleaseUnitTest -name "TEST-*.xml" -exec upload_buildkite_test_analytics_junit {} $BUILDKITE_ANALYTICS_TOKEN_UNIT_TESTS \;
 
 echo -e "\n--- Emitting original exit code"
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordManager.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordManager.kt
@@ -46,7 +46,7 @@ internal class ApplicationPasswordManager @Inject constructor(
             createApplicationPassword(site, usernamePayload.userName).also {
                 if (it is ApplicationPasswordCreationResult.Created) {
                     applicationPasswordsStore.saveCredentials(
-                        usernamePayload.userName,
+                        site.domainName,
                         it.credentials
                     )
                 }
@@ -170,5 +170,5 @@ internal class ApplicationPasswordManager @Inject constructor(
     }
 
     private val SiteModel.domainName
-        get() = UrlUtils.removeScheme(url)
+        get() = UrlUtils.removeScheme(url).trim('/')
 }
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordManager.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordManager.kt
@@ -23,7 +23,7 @@ internal class ApplicationPasswordManager @Inject constructor(
     private val appLogWrapper: AppLogWrapper
 ) {
     @Suppress("ReturnCount")
-    internal suspend fun getApplicationCredentials(
+    suspend fun getApplicationCredentials(
         site: SiteModel
     ): ApplicationPasswordCreationResult {
         if (site.isWPCom) return ApplicationPasswordCreationResult.NotSupported(
@@ -122,7 +122,7 @@ internal class ApplicationPasswordManager @Inject constructor(
         }
     }
 
-    internal suspend fun deleteApplicationCredentials(
+    suspend fun deleteApplicationCredentials(
         site: SiteModel
     ): ApplicationPasswordDeletionResult {
         val payload = if (site.origin == SiteModel.ORIGIN_WPCOM_REST) {
@@ -165,7 +165,7 @@ internal class ApplicationPasswordManager @Inject constructor(
         }
     }
 
-    internal fun deleteLocalApplicationPassword(site: SiteModel) {
+    fun deleteLocalApplicationPassword(site: SiteModel) {
         applicationPasswordsStore.deleteCredentials(site.domainName)
     }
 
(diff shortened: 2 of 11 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordManager.kt
```diff
@@ -16,15 +16,14 @@ private const val NOT_FOUND = 404
 private const val APPLICATION_PASSWORDS_DISABLED_ERROR_CODE = "application_passwords_disabled"
 
 internal class ApplicationPasswordManager @Inject constructor(
-    context: Context,
+    private val applicationPasswordsStore: ApplicationPasswordsStore,
```
hichamboushaba, line 19: I moved to injecting this dependency to allow providing a custom behavior for tests.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordManager.kt
```diff
@@ -16,15 +16,14 @@ private const val NOT_FOUND = 404
...
     @ApplicationPasswordClientId private val applicationName: String,
     private val jetpackApplicationPasswordRestClient: JetpackApplicationPasswordRestClient,
-    private val wpApiApplicationPasswordRestClient: WPApiApplicationPasswordRestClient
+    private val wpApiApplicationPasswordRestClient: WPApiApplicationPasswordRestClient,
+    private val appLogWrapper: AppLogWrapper
 ) {
-    private val applicationPasswordsStore = ApplicationPasswordsStore(context, applicationName)
-
     @Suppress("ReturnCount")
-    internal suspend fun getApplicationCredentials(
```
hichamboushaba, line 27: Those `internal` modifiers didn't add any value, as the whole class is already `internal`. And they were causing issues when trying to consume the code from Java, I discovered when I was testing some changes that I didn't keep.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordNetwork.kt
```diff
@@ -103,6 +103,10 @@ class ApplicationPasswordNetwork @Inject constructor(
         }
     }
 
+    suspend fun deleteApplicationPassword(site: SiteModel) {
```
hichamboushaba, line 106: Since the `ApplicationPasswordNetwork` is the only class we expose, it needs a way to allow clients to request deleting passwords.

.buildkite/unit-tests.sh
```diff
@@ -6,10 +6,11 @@ echo "--- :closed_lock_with_key: Installing Secrets"
 cp gradle.properties-example gradle.properties && cp -a example/properties-example/ example/properties/
 
 echo "--- :gradle: Run Tests"
-./gradlew example:testRelease plugins:woocommerce:testRelease || EXIT_CODE=$?
+./gradlew example:testRelease fluxc:testRelease plugins:woocommerce:testRelease || EXIT_CODE=$?
```
hichamboushaba, line 9: @ParaskP7 I needed to add some unit tests to the `fluxc` module instead of the Example app one, so I updated the `build.gradle` and this script to support this. I needed this because for this feature, we are trying to hide everything about the implementation details from the clients, so the internal classes use the `internal` visibility modifier.

Is this change OK? If we need validation from someone else from the apps infrastructure team let me know.

ParaskP7: 👋 @hichamboushaba and thanks for the heads-up! 🥇 

I quickly reviewed your changes on the `fluxc` related `build.gradle` file and it all looks good to me, including the `includeAndroidResources` configuration you added, which is most probably due to Robolectric, please correct me if I am wrong. 👍 

As far as the test dependencies that you added, from the looks of it, it seems that you added all those because you do need them all, so I am okay with that too. 👍 

As such, I don't see anything alarming here and actually many thanks for configuring this module for testability! 💯 

PS: I was never in favor of having all the `fluxc` related unit tests on the `example` module anyway! 😄

hichamboushaba: Thanks so much for the quick review @ParaskP7 🙏 

> including the includeAndroidResources configuration you added, which is most probably due to Robolectric, please correct me if I am wrong. 👍

Yes, this was needed for ~~the connected tests~~ Robolectric because I needed it to work with Volley requests.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordManager.kt
```diff
@@ -47,7 +46,7 @@ internal class ApplicationPasswordManager @Inject constructor(
             createApplicationPassword(site, usernamePayload.userName).also {
                 if (it is ApplicationPasswordCreationResult.Created) {
                     applicationPasswordsStore.saveCredentials(
-                        usernamePayload.userName,
+                        site.domainName,
```
hichamboushaba, line 49: A bug that was caught by the tests 😅 🎉

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordNetwork.kt
```diff
@@ -103,6 +103,10 @@ class ApplicationPasswordNetwork @Inject constructor(
...
+        applicationPasswordManager.deleteApplicationCredentials(site)
```
hafizrahman, line 107: Do we not need to pass the return value from `deleteApplicationCredentials()`?

hichamboushaba: We agreed on making the call fail silently during the logout, but I agree it should still return something.
For not, this function is just temporary, I added it because the `ApplicationPasswordsManager` is internal, but I don't like it that much, I don't think the `Network` class should be the one responsible for this, it would be better to be part of `AccountStore` or `SiteStore`.

I will try to move it there, I already tried and was blocked by the fact that injecting `ApplicationPasswordsManager` in any one of those classes will require all of the FluxC clients to break until they provide an `@ApplicationPasswordsClientId` String, which I don't want to do, as the feature should be kept optional until it's needed.

Let's keep this way in this PR, I'll see what I can do in another PR, WDYT?

hichamboushaba: @hafizrahman I handled this change in this PR https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2600, I'll update the description this afternoon, anyway, it's not urgent, feel free to leave it until next week.

hafizrahman: Sounds like a good plan to me!

hafizrahman: > I don't think the Network class should be the one responsible for this, it would be better to be part of AccountStore or SiteStore.

It did feel out of place for me but I was struggling to articulate it properly. 

> will require all of the FluxC clients to break until they provide an @ApplicationPasswordsClientId String, which I don't want to do, as the feature should be kept optional until it's needed.

Interesting issue, I saw #2600 and so far couldn't think of a more simple solution for this. Almost makes me want to keep this in this class just to keep things simple. Calling password deletion manually doesn't feel like a common use case to me for a FluxC client, so perhaps it's OK to keep it tucked here?

hafizrahman (approved):
👍🏼 

Thanks for adding these tests!

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordNetwork.kt
hichamboushaba: > Calling password deletion manually doesn't feel like a common use case to me for a FluxC client

We will be calling it on logout, so it's needed.

> perhaps it's OK to keep it tucked here?

Can you elaborate? you don't like the solution done in #2600? I tried to solve the original issue with it, the feature is still optional, and even without providing `@ApplicationPasswordsClientId`, the clients will keep working, things would break only when they start using the feature itself at runtime.

PR #2587, opened 2022-12-01
REST API: Migrate CouponStore to use application passwords

hichamboushaba:
This PR migrates the CouponStore to use application passwords using the work done in the PRs #2586 and #2589, and #2592.

⚠️ Please don't merge before #2592.

##### WordPress.com login:
1. Open the example app, then sign in using your WordPress.com account.
2. Click on Woo, then Coupons.
3. Select a site.
4. Click on Fetch Coupons (or any other operation)
5. Confirm that it works successfully.
6. Open the users page in `wp-admin` and confirm that the app created an application password (the page is available only for self-hosted sites).
7. Revoke the password.
8. Retest the feature, and confirm it stills works as expected, the app will create a second password.

##### Site credentials login:
1. Apply the patch from below (we don't check for Woo installation for site credentials login yet)
2. Open the app, then sign in using your wp-admin credentials (enter site URL in the address form).
3. Repeat steps 2 to 8 from the above test.

```patch
Index: example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt b/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(revision 679271e71db3fb58b5e7fbc58545bd14886f4a40)
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(date 1653478977241)
@@ -7,6 +7,8 @@
 import dagger.android.support.DaggerFragment
 import kotlinx.android.synthetic.main.fragment_woo_products.*
 import org.wordpress.android.fluxc.example.R
+import org.wordpress.android.fluxc.example.SiteSelectorDialog
+import org.wordpress.android.fluxc.example.ui.common.showSiteSelectorDialog
 import org.wordpress.android.fluxc.example.ui.common.showStoreSelectorDialog
 import org.wordpress.android.fluxc.example.utils.toggleSiteDependentButtons
 import org.wordpress.android.fluxc.model.SiteModel
@@ -23,7 +25,7 @@
                 Button(context).apply {
                     text = "Select Site"
                     setOnClickListener {
[...]

PR #2595, opened 2022-12-14
[Woo] Support two networking implementations for REST API project

hichamboushaba:
⚠️ Please don't merge until #2594 is merged.

To test, repeat the test steps of #2587.


config/checkstyle.xml
```diff
@@ -95,7 +95,7 @@
 
     <!-- Specific to FluxC - Kotlin -->
     <module name="RegexpMultiline">
-        <property name="format" value="(?&lt;!@Singleton)\n.*class(.+\n)*.*:[\n\s]*Base[\w]*Client\("/>
+        <property name="format" value="(?&lt;!@Singleton)\n.*^class(.+\n)*.*:[\n\s]*Base[\w]*Client\("/>
```
hichamboushaba, line 98: Check the commit message for why I made this change.

hafizrahman (commented):
Nice work! My general sense (just from a workload perspective) with this implementation is that it won't require a lot more time compared if we're going with single networking implementation.

Do we also want to handle cases where at some point users might turn on/off Jetpack or turn on/off Application Passwords, when they're already using the app?

hichamboushaba:
Thanks for the review @hafizrahman 

> Do we also want to handle cases where at some point users might turn on/off Jetpack or turn on/off Application Passwords, when they're already using the app?

Since the decision on what path to take depends only on the instance of the `SiteModel` we pass, we can support any scenario we want in the app by passing the correct one:
1. If the site gets connected to Jetpack, we will just need to fetch sites using the traditional way `/me/sites`, then update the `SelectedSite` using the new instance, which should work automatically.
2. If there are any plans to support the other way around, we will do the same thing after requesting the user's wp-admin credentials.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooNetwork.kt
```diff
@@ -0,0 +1,83 @@
...
+    suspend fun <T : Any> executeGetGsonRequest(
+        site: SiteModel,
+        path: String,
+        clazz: Class<T>,
+        params: Map<String, String> = emptyMap()
+    ): WPAPIResponse<T> {
+        return when (site.origin) {
+            SiteModel.ORIGIN_WPCOM_REST -> jetpackTunnelWPAPINetwork.executeGetGsonRequest(site, path, clazz, params)
+                .toWPAPIResponse()
+            else -> applicationPasswordNetwork.executeGetGsonRequest(site, path, clazz, params)
```
hafizrahman: probably np: should this handle only `SiteModel.ORIGIN_XMLRPC`, and add error handling if it's `SiteModel.ORIGIN_UNKNOWN`?

As far as I can check, the value is always eventually set to either `ORIGIN_WPCOM_REST` or `ORIGIN_XMLRPC` in WooCommerce app's FluxC usage, so this should be OK. Is it worth being extra safe here?

hichamboushaba: Good point, done in d209e60

hafizrahman (approved):

PR #2596, opened 2022-12-14
Fix connected tests of Jetpack sites

hichamboushaba:
While working on #2594, I noticed that some unrelated connected tests were failing, after checking I found out that it's because of a similar issue that we had last year p1625713520033400/1625704778.033300-slack-C0QB79UV8

To avoid having the issues in the future, we decided to make a change on the code side to exclude the site protocol when comparing the URLs.

Note: it seems the test `testInstallAndDeleteSitePlugin` is still failing, but from what I see, this was failing for at least a month, I don't have time to spend investigating it, so I'll leave it to someone else 🙂.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ActivityLogTestJetpack.kt
@@ -39,6 +39,7 @@ import org.wordpress.android.fluxc.store.SiteStore.SiteErrorType
 import org.wordpress.android.fluxc.tools.FormattableContent
 import org.wordpress.android.util.AppLog
 import org.wordpress.android.util.AppLog.T
+import org.wordpress.android.util.UrlUtils
 import java.util.Date
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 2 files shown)


ParaskP7 (approved):
👋 @hichamboushaba !

I reviewed and test this PR as per the description, everything LGTM and thank you so much for addressing this! 💯 🥇 ❤️ 

> While working on https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2594, I noticed that some unrelated connected tests were failing, after checking I found out that it's because of a similar issue that we had last year p1625713520033400/1625704778.033300-slack-C0QB79UV8

👍 

> To avoid having the issues in the future, we decided to make a change on the code side to exclude the site protocol when comparing the URLs.

This was a good idea, thank you for doing that, I am almost certain that the scheme plays little role in those tests, so if they now become more robust by excluding it, then this is a win! ✅ 

> Note: it seems the test testInstallAndDeleteSitePlugin is still failing, but from what I see, this was failing for at least a month, I don't have time to spend investigating it, so I'll leave it to someone else 🙂.

I am shamelessly pinging @pachlava here in case he can help with investigating and fixing this `testInstallAndDeleteSitePlugin`. 😊 

-----

PS: It seems that if we fix this test we could then go ahead and re-enabled the `Connected Tests` check on `FluxC` PRs so that to start depending on them again afterwards. 🤞

pachlava:
Thanks for the ping @ParaskP7 👋 🙂 

I'll see what I can find.

hichamboushaba:
@ParaskP7 I'm merging this PR now.
@pachlava thanks for taking a look at the other test.

PR #2598, opened 2022-12-15
Align naming between the different components of the Application Passwords feature

hichamboushaba:
This PR just aligns the naming of the different components used in the Application Passwords feature:
1. Use "ApplicationPassword**s**" for the main components.
2. Use "ApplicationPassword" for the individual results and responses.

Previously we were using the "**s**" just for some classes, and it was missing from others, and I decided to stick to it to align with the namings in the WordPress repo.

⚠️ Please don't merge until the WCAnroid [PR](https://github.com/woocommerce/woocommerce-android/pull/8036) is approved.


hafizrahman (approved):
I have read way too many "application password" lately that I didn't notice the missing "s" in places. Good eye for noticing!

PR #2599, opened 2022-12-15
Support reply to a product review

wzieba:
## Description

This PR adds support for WPCOM endpoint to send a reply to a comment (review). See relevant iOS PR: https://github.com/woocommerce/woocommerce-ios/pull/7783

WCAndroid PR: https://github.com/woocommerce/woocommerce-android/pull/8045


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -965,6 +968,26 @@ class ProductRestClient @Inject constructor(
             ).handleResult()
         }
 
+    suspend fun replyToReview(
+        site: SiteModel,
+        remoteReviewId: Long,
+        replayContent: String?
+    ): Response<Unit> {
+        val url = WPCOMREST.sites.site(site.siteId).comments.comment(remoteReviewId).replies.new_.urlV1_1
```
wzieba: I didn't use `CommentsRestClient#createNewReply` as it uses the database entity model (`CommentEntity`) of WordPress comment as an output.

Source: https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/comment/CommentsRestClient.kt#L157-L184

ThomazFB (approved):
Looks good! This is a nice HACK week idea 😄 I'm leaving the PR approved since there are some conflicts that need to be fixed before merging!

wzieba:
Thanks @ThomazFB ! I'll mark it back as a draft though as there are changes I have to make here to support new REST endpoint. Internal details: p1671616312367749-slack-C03L1NF1EA3

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -987,6 +989,26 @@ class ProductRestClient @Inject constructor(
             ).handleResult()
         }
 
+    suspend fun replyToReview(
```
wzieba, line 992: @hichamboushaba - does it look good?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -987,6 +989,26 @@ class ProductRestClient @Inject constructor(
...
+        site: SiteModel,
+        productId: RemoteId,
+        reviewId: RemoteId,
+        replyContent: String?
+    ): WooPayload<Unit> {
+        val request = mutableMapOf(
+            "post" to productId.value,
```
wzieba, line 999: I had to add `post` parameter. Otherwise, the API was returning `Sorry, you are not allowed to create this comment without a post.` error message

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
hichamboushaba: Yes, it looks good, just two np remarks (not blockers 🙂):

1. Rename `request` to `body` to clarify they are request body and not the request itself and to stay consistent with the other usages.
2. Use named arguments for the function invocation.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
hichamboushaba: This is actually useful, as the comment will be linked to the product, and if the product gets deleted, the comments will be deleted too.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
hichamboushaba: Thanks so much btw 🙏

wzieba: Thank you! Good suggestions, I've applied them in eefb650503ade35443053053c831d81acd05320b

hichamboushaba (approved):
The code looks good, thanks for taking care of the migration to WooNetwork @wzieba 🙏.
I didn't test the feature, but I believe it would be tested in the WCAndroid PR.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
wzieba: True, but now we now have a functional dependency:
- `parent` points to some `Review` which points to a `Product`
- `post` points to a `Product`

I think just pointing to `parent` Review should work well - `parent` Review is removed when its parent `Product` (post) is removed, and then all replies of it as well.

Either way it's API contract, so no choice here 😅

wzieba:
> I didn't review the feature, but I believe it would be reviewed in the WCAndroid PR.

Yes, we'll review the change feature-wise on WCAndroid PR. 

Thank you @hichamboushaba !

PR #2600, opened 2022-12-16
Application Passwords: expose a function to delete passwords

hichamboushaba:
@hafizrahman let's continue the [discussion ](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2594#discussion_r1050465079) we started in the other PR here.

In this PR, I made two main changes:
1. I removed the `deleteApplicationPassword` from `ApplicationPasswordsNetwork` and moved it to `SiteStore`. For this part, I was torn between `SiteStore` and `AccountStore`, but I decided to go with `SiteStore` for two reasons: it's already in Kotlin 😄, and `AccountStore` is more about the WPCom account than this.
2. I updated the DI configuration to make `@ApplicationPasswordsClientId` as `Optional`, this allows moving from a compile error to a runtime error if it's missing, and with this we won't impact the other clients if they don't want to use the feature.

Let me know what you think about these two changes.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
@@ -9,19 +9,34 @@ import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComGson
 import org.wordpress.android.fluxc.utils.AppLogWrapper
 import org.wordpress.android.util.AppLog
 import org.wordpress.android.util.UrlUtils
+import java.util.Optional
 import javax.inject.Inject
 
 private const val CONFLICT = 409
 private const val NOT_FOUND = 404
 private const val APPLICATION_PASSWORDS_DISABLED_ERROR_CODE = "application_passwords_disabled"
 
+/**
+ * Note: the [ApplicationPasswordsClientId] is provided as [Optional] because we want to keep the feature optional and
+ * to not force the client apps to provide it. With this change, we will keep Dagger happy, and we move from a compile
+ * error to a runtime error if it's missing.
+ */
 internal class ApplicationPasswordsManager @Inject constructor(
     private val applicationPasswordsStore: ApplicationPasswordsStore,
-    @ApplicationPasswordsClientId private val applicationName: String,
+    @ApplicationPasswordsClientId private val applicationNameOptional: Optional<String>,
     private val jetpackApplicationPasswordsRestClient: JetpackApplicationPasswordsRestClient,
     private val wpApiApplicationPasswordsRestClient: WPApiApplicationPasswordsRestClient,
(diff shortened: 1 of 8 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
```diff
@@ -9,19 +9,34 @@ import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComGson
...
 
+/**
+ * Note: the [ApplicationPasswordsClientId] is provided as [Optional] because we want to keep the feature optional and
+ * to not force the client apps to provide it. With this change, we will keep Dagger happy, and we move from a compile
+ * error to a runtime error if it's missing.
+ */
 internal class ApplicationPasswordsManager @Inject constructor(
     private val applicationPasswordsStore: ApplicationPasswordsStore,
-    @ApplicationPasswordsClientId private val applicationName: String,
+    @ApplicationPasswordsClientId private val applicationNameOptional: Optional<String>,
```
hafizrahman: Question: what is the benefit of having it added here, instead of having this class re-using the `applicationNameOptional` value from `applicationPasswordsStore`?

hichamboushaba: Good question, it wasn't done for any particular reason.
Actually, before the commit 5b54a20b85487863a22d103615aa47834ac19e76, the roles were inverted, the `applicationName` was injected only in the `ApplicationPasswordsManager` class, and then supplied to the `ApplicationPasswordsStore`, and I liked this approach better, but I had to change it to make the `ApplicationPasswordsManager` testable (by allowing the injection of a mocked implementation of the `Store`).
Now with the `Optional` usage, we have a lot of redundancy.

So, if I'm understanding your suggestion correctly, the idea is to make `applicationName` public in `ApplicationPasswordsStore`, then use it instead of injecting the value twice? If I'm right in my understanding, then I like it, I'll add it.

hichamboushaba: Let me know if [45a42b5](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2600/commits/45a42b5dd5b603b08702a32d11010490a235e76a) aligns with your suggestion.

hafizrahman: I was initially unsure because I thought I might be missing some use case for it.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2600/commits/45a42b5dd5b603b08702a32d11010490a235e76a indeed looks like what I had in mind! Awesome!

hafizrahman (approved):
👍🏼  Let's ship it!

PR #2601, opened 2022-12-19
[Woo] Migrate Reviews to use WooNetwork.

hafizrahman:
For https://github.com/woocommerce/woocommerce-android/issues/8015

Test instructions:

### 1. For site credentials login

First, apply patch:

```patch
Index: example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt b/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(revision ab9379d307673d92659cf08607388d7b549488d7)
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(date 1671447606619)
@@ -7,6 +7,8 @@
 import dagger.android.support.DaggerFragment
 import kotlinx.android.synthetic.main.fragment_woo_products.*
 import org.wordpress.android.fluxc.example.R
+import org.wordpress.android.fluxc.example.SiteSelectorDialog
+import org.wordpress.android.fluxc.example.ui.common.showSiteSelectorDialog
 import org.wordpress.android.fluxc.example.ui.common.showStoreSelectorDialog
 import org.wordpress.android.fluxc.example.utils.toggleSiteDependentButtons
 import org.wordpress.android.fluxc.model.SiteModel
@@ -23,7 +25,7 @@
                 Button(context).apply {
                     text = "Select Site"
                     setOnClickListener {
-                        showStoreSelectorDialog(selectedPos, object : StoreSelectorDialog.Listener {
+                        showSiteSelectorDialog(selectedPos, object : SiteSelectorDialog.Listener {
                             override fun onSiteSelected(site: SiteModel, pos: Int) {
                                 onSiteSelected(site)
                                 selectedSite = site
```

1. Start Example app,
2. Sign in using wp-admin credentials and xmlrpc url,
3. Go to Woo > Select Site > pick the site,
4. Go to Products > Select Site > pick the site,
5. Test these buttons and check that changes apply in wp-admin:
   1. Fetch product reviews by product
   2. Fetch product reviews by site
   3. Fetch product review by ID
   4. Update product review status

[...]

PR #2602, opened 2022-12-19
Woo REST API: Migrate Products endpoints to use WooNetwork

hichamboushaba:
This PR migrates all of the remaining Jetpack requests in ProductRestClient to use `WooNetwork`.

### Notes for review
- You may want to review commit by commit to review small chunks of the changes at a time.
- Most of the changes are purely migration, but for three commits (e9dcab9babc5dad67c1ccccc49e446c819f01d96 and 31cabd4f99568099cba230ba03d298460afb4405 and 328e1d771b9dd000e0882c289bb38df6369d04c8) I did some small refactorings, I removed the extensive usage of extensions as it made the code harder to follow and read and wasn't consistent with what we do in the other places.

### Testing
IMO the code review should be the main part here, as the behavior was tested in previous PRs.
But if you want to test some endpoints, follow the below steps:

#### Site Credentials login
1. Apply this patch
```
IDEA additional info:
<+>UTF-8
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(revision ab9379d307673d92659cf08607388d7b549488d7)
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(date 1671447606619)
@@ -7,6 +7,8 @@
@@ -23,7 +25,7 @@
                 Button(context).apply {
                     text = "Select Site"
                     setOnClickListener {
                                 onSiteSelected(site)
                                 selectedSite = site
```
2. Start Example app.
3. Sign in using wp-admin credentials and xmlrpc url.
4. Go to Woo > Select Site > pick the site
5. Go to Products > Select Site > pick the site
6. Test some product related scenarios

#### WordPress.com login
1. Start the example app, then sign in using your WordPress.com account.
2. Repeat 4 to 6 from previous test.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -4,7 +4,6 @@ import android.content.Context
 import com.android.volley.RequestQueue
 import com.google.gson.JsonArray
 import com.google.gson.JsonParser
-import com.google.gson.reflect.TypeToken
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.action.WCProductAction
 import org.wordpress.android.fluxc.generated.WCProductActionBuilder
@@ -1200,7 +1199,6 @@ class ProductRestClient @Inject constructor(
             }
 
             val url = WOOCOMMERCE.products.categories.pathV3
-            val responseType = object : TypeToken<List<ProductCategoryApiResponse>>() {}.type
             val params = mutableMapOf(
                 "per_page" to pageSize.toString(),
                 "offset" to offset.toString(),
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -579,16 +579,18 @@ class ProductRestClient @Inject constructor(
             params["exclude"] = excludedCategoryIds.map { it }.joinToString()
         }
 
-        return WOOCOMMERCE.products.categories.pathV3
-            .requestCategoryTo(site, params)
-            .handleResultFrom(site)
-    }
+        val url = WOOCOMMERCE.products.categories.pathV3
 
-    @JvmName("handleResultFromProductCategoryApiResponse")
-    private fun JetpackResponse<Array<ProductCategoryApiResponse>>.handleResultFrom(site: SiteModel) =
-        when (this) {
-            is JetpackSuccess -> {
-                data
+        val response = wooNetwork.executeGetGsonRequest(
+            site = site,
+            path = url,
+            params = params,
+            clazz = Array<ProductCategoryApiResponse>::class.java
+        )
+
+        return when (response) {
+            is WPAPIResponse.Success -> {
+                response.data
                     ?.map {
                         it.asProductCategoryModel()
                             .apply { localSiteId = site.id }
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/utils/CoroutineEngineUtils.kt
```diff
@@ -0,0 +1,56 @@
...
+import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.runBlocking
+import org.mockito.Mockito.lenient
+import org.mockito.kotlin.any
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+import org.wordpress.android.fluxc.tools.CoroutineEngine
+
+fun initCoroutineEngine() = runBlocking {
```
hichamboushaba, line 14: This was copied from the unit tests in the example app.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/WPAPIResponseExt.kt
```diff
@@ -0,0 +1,11 @@
+package org.wordpress.android.fluxc.utils
+
+import org.wordpress.android.fluxc.network.rest.wpapi.WPAPIResponse
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.toWooError
+
+fun <T> WPAPIResponse<T>.handleResult() =
```
hafizrahman, line 7: nice!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -104,30 +102,37 @@ class ProductRestClient @Inject constructor(
      * @param [remoteShippingClassId] Unique server id of the shipping class to fetch
      */
     fun fetchSingleProductShippingClass(site: SiteModel, remoteShippingClassId: Long) {
-        val url = WOOCOMMERCE.products.shipping_classes.id(remoteShippingClassId).pathV3
-        val responseType = object : TypeToken<ProductShippingClassApiResponse>() {}.type
-        val params = emptyMap<String, String>()
-        val request = JetpackTunnelGsonRequest.buildGetRequest(url, site.siteId, params, responseType,
-                { response: ProductShippingClassApiResponse? ->
-                    response?.let {
+        coroutineEngine.launch(AppLog.T.API, this, "fetchSingleProductShippingClass") {
```
hafizrahman, line 105: ❓ What is the benefit of using coroutine here while it wasn't used before?

hafizrahman (approved):
This looks good to me. Just left a question.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
hichamboushaba: It's not just for benefit, but we are forced to switch to a coroutine.
Our `Network` class expose just suspendable functions, while this function (and multiple other ones) are still using the `Dispatcher`, so we have two choices to make them work with the new architecture:
1. Use the `CoroutineEngine` to move the work to a new Coroutine, as I did here.
2. Add duplicate functions to our `Network` that use callbacks same as the old `Request`s.

I think the first approach is better in this case, it doesn't require much work, and the ultimate goal is to make those functions suspendable in the future, WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/utils/WPAPIResponseExt.kt
hichamboushaba: The truth is that this is just a copy of a similar function for `JetpackAPIResponse` 😄.

I think we can even do better by providing a mapper function, I didn't spend too much time on this, but I might do it today, and suggest something to be used in future migrations.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
hafizrahman: Oh I was more curious because I didn't have to do this [in the Reviews PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2601). Turns out the ones I converted were already `suspend fun`.

👍🏼  Thanks for the explanation.

PR #2603, opened 2022-12-19
Woo REST API: Finish migration of ProductRestClient

hichamboushaba:
⚠️ Please don't merge until #2602 is merged.

This finishes the migration of ProductRestClient to the new architecture by introducing a new `Network` class: `WPComNetwork`, this allows getting rid of the base class and the unused arguments.

For review, it's better to review cce651717a5dda51e12694314424989c879ef269 separately, as it simply reformats the file to follow the new Kotlin coding style. 

### Test
This change applies only to WPCom login, we will check later if there is an WP or WC based API to handle the password change.

1. Open the example app then sign in using WordPress.com account.
2. Click on Woo, then select a site.
3. Click on Products then select a site.
4. Click on Update Product.
5. Enter a product ID.
6. Make sure visibility is set to `publish`.
7. Scroll to the bottom then change the password.
8. Confirm the change and fetch work as expected.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComNetwork.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/WPComNetwork.kt
@@ -1,0 +1,54 @@
+package org.wordpress.android.fluxc.network.rest.wpcom
+
+import android.content.Context
+import com.android.volley.RequestQueue
+import org.wordpress.android.fluxc.Dispatcher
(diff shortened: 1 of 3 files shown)


hafizrahman (approved):
LGTM, assuming we're planning to complete `WPComNetwork` later as needed for delete/put actions.

hichamboushaba:
Thanks @hafizrahman for the review.

> assuming we're planning to complete WPComNetwork later as needed for delete/put actions.

The WPCom request builder doesn't have them, so I assume they were never needed, but when they are needed, they can be added easily.

PR #2604, opened 2022-12-20
Fix Mocked Tests

hichamboushaba:
After merging #2601, we broke the instrumented tests, they don't build anymore.

This PR does two things:
1. Fixes the build issue by providing the necessary dependencies to the Mocked modules.
2. Fixes the mocked Products tests.


hafizrahman (approved):
`MockedStack_WCProductsTest` all green now. Thanks for the quick fix!

PR #2605, opened 2022-12-20
Enhance the mapper function from WPAPIResponse to WooPayload

hichamboushaba:
During the migration of `ProductRestClient`, I noticed the extension `JetpackResponse#handleResult`, and I created a similar one for `WPAPIResponse`, but I was concerned about two issues with the extension:
1. It considers success responses with `null` data as success results.
2. It doesn't allow a way to map the response to another type.

So in this PR, I attempted to solve the two issues, and I renamed it from `handleResult` to `toWooPayload`.

I'll try to explain my reasoning for why point 1 should be changed:
- This logic will have no impact on the existing Jetpack sites: given the [logic](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/b40e9c214fc20537f960d7fb8fa611cd6dd34812/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/GsonRequest.java#L87-L101) we have for parsing the JSON responses, there is no way we'll receive a `null` object from the Jetpack tunnel, because it will always return a `data` property (I can share the code internally if needed), this means that we will always get an object, what could be `null` if the site's response is empty is the object's properties.
- For application passwords login: if we get a `null` object because we receive an empty response, it's better to report it as an error, and not as a success, this would allow better detection for errors, and would lead to more robust handling.
- When needed and with the commit 66bb579, we can still accept `null` as a success by explicitly using a nullable type as the argument of `WooPayload` (check unit tests for an example).

### Testing
I don't think there is much to test, it's all just code refactoring, but feel free to repeat the steps from #2602.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooError.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooError.kt
@@ -32,7 +32,8 @@ enum class WooErrorType {
     INVALID_RESPONSE,
     AUTHORIZATION_REQUIRED,
     INVALID_PARAM,
-    PLUGIN_NOT_ACTIVE
+    PLUGIN_NOT_ACTIVE,
+    EMPTY_RESPONSE
 }
 
 fun WPComGsonNetworkError.toWooError(): WooError {
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -74,7 +74,7 @@ import org.wordpress.android.fluxc.store.WCProductStore.RemoteUpdateVariationPay
 import org.wordpress.android.fluxc.store.WCProductStore.RemoteUpdatedProductPasswordPayload
 import org.wordpress.android.fluxc.store.WCProductStore.RemoteVariationPayload
 import org.wordpress.android.fluxc.tools.CoroutineEngine
-import org.wordpress.android.fluxc.utils.handleResult
+import org.wordpress.android.fluxc.utils.toWooPayload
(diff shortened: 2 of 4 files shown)


hichamboushaba:
@JorgeMucientes I requested your review too since Hafiz is AFK today 🙏, this will be used in the future `RestClient` migrations, and I prefer to still base my PRs on top of `trunk` than other intermediate branches.

Thanks in advance 🙏.

JorgeMucientes:
Nice refactor @hichamboushaba 🥇 

 I agree with your reasoning on the `nullable` values for successful responses. I just have a question on this. If we have an API request where we expect an empty body and I just need to know if it was 200 success or whatever code, I'm I force to pass a nullable type? Can I use `Unit` or will `toWooPayload` return an error for that?

JorgeMucientes (approved):

hichamboushaba:

> I just have a question on this. If we have an API request where we expect an empty body and I just need to know if it was 200 success or whatever code, I'm I force to pass a nullable type? Can I use Unit or will toWooPayload return an error for that?

Assuming the site will return an empty JSON (so `{}` or just an empty string for direct REST API, which would translate to `{"data":{}}` and `{"data":""}` respectively for the Jetpack tunnel), passing `Unit::class.java` should return a non-null value with success, so `toWooPayload` would still translate to success.

We can confirm this with the following test:
```kotlin
class JsonTest {
    @Test
    fun `test empty json object`() {
        val result = parseJson("{}", Unit::class.java)

        assertNotNull(result)
    }

    @Test
    fun `test empty json`() {
        val result = parseJson("", Unit::class.java)

        assertNotNull(result)
    }

    private fun <T> parseJson(json: String, clazz: Class<T>): T {
        val gson = Gson()

        return gson.fromJson(json, clazz)
    }
}
```
The function `parseJson` matches the same logic as what `GsonRequest` does for us.

Note: this is true also if the site passes a valid JSON that we don't care about, passing `Unit::class.java` would still result in a non-null value.

I'll leave the PR unmerged to give you a chance to review my response here, if you agree with this, please feel free to merge it.

JorgeMucientes:
Thanks for clarifying 🙏🏼 

Let's 🚢 this!

PR #2608, opened 2022-12-21
Woo REST API: migrate RefundRestClient

### LINKED ISSUE #8069: [REST API] Migrate RefundRestClient

hichamboushaba:
Closes: https://github.com/woocommerce/woocommerce-android/issues/8069

This PR migrates `RefundRestClient` to the new architecture using `WooNetwork`.

The last commit e522154 is not related to the migration itself, but it's just an improvement to the Refunds test fragment to allow selecting a specific site.

### Test
As with all the migration PRs, the focus should be on the code review by confirming we didn't miss any argument or used the wrong HTTP method.

But a small smoke test to the feature wouldn't hurt:
##### WordPress.com login
1. Open the example app.
2. Sign in using WordPress.com.
3. Click on Woo, then pick a site.
4. Click on Refunds then pick a site.
5. Test some refund scenarios and confirm they work as expected.

##### Site Credentials login (bonus)
1. Apply the following patch
```patch
IDEA additional info:
<+>UTF-8
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(revision e522154c8bc2f74f9d2039917d9de31eaf2b3240)
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(date 1671625263544)
@@ -7,6 +7,8 @@
@@ -23,7 +25,7 @@
                 Button(context).apply {
                     text = "Select Site"
                     setOnClickListener {
                                 onSiteSelected(site)
                                 selectedSite = site
```
2. Open the app then sign in using site credentials (enter site address in the login dialog).
3. Repeat steps 3 to 5 from the above test case.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt
@@ -91,21 +91,12 @@ class RefundRestClient @Inject constructor(
     ): WooPayload<RefundResponse> {
         val url = WOOCOMMERCE.orders.id(orderId).refunds.refund(refundId).pathV3
 
-        val response = jetpackTunnelGsonRequestBuilder.syncGetRequest(
-                this,
-                site,
-                url,
-                emptyMap(),
-                RefundResponse::class.java
+        val response = wooNetwork.executeGetGsonRequest(
+            site = site,
+            path = url,
+            clazz = RefundResponse::class.java
         )
-        return when (response) {
-            is JetpackSuccess -> {
-                WooPayload(response.data)
-            }
-            is JetpackError -> {
-                WooPayload(response.error.toWooError())
-            }
-        }
+        return response.toWooPayload()
     }
 
     suspend fun fetchAllRefunds(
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt
```diff
@@ -1,37 +1,19 @@
...
-class RefundRestClient @Inject constructor(
-    dispatcher: Dispatcher,
-    private val jetpackTunnelGsonRequestBuilder: JetpackTunnelGsonRequestBuilder,
-    appContext: Context?,
-    @Named("regular") requestQueue: RequestQueue,
-    accessToken: AccessToken,
-    userAgent: UserAgent,
-    private val wooNetwork: WooNetwork
-) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
+class RefundRestClient @Inject constructor(private val wooNetwork: WooNetwork) {
```
hafizrahman, line 16: Very clean refactor ❤️

hafizrahman, line 16: Any reason we're removing the @Singleton annotation?

hafizrahman (commented):
Looks good overall, sorry for the delay on this due to my AFK. I left a question only.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/refunds/RefundRestClient.kt
hichamboushaba: The classes are stateless now and don't hold any information, there is no reason for them to be kept in memory if not needed.
Still, technically this won't make any change in the behavior, since the `Stores` are Singleton, so they will still behave the same as before, just without the constraint.

hafizrahman (approved):

PR #2611, opened 2022-12-22
Woo REST API: Migrate WooCommerceStore REST clients

### LINKED ISSUE #8070: [REST API] Migrate WooSystemRestClient

### LINKED ISSUE #8076: [REST API] Migrate WCUserRestClient

### LINKED ISSUE #8077: [REST API] Migrate WooCommerceRestClient

hichamboushaba:
Closes https://github.com/woocommerce/woocommerce-android/issues/8077
Closes https://github.com/woocommerce/woocommerce-android/issues/8076
Closes https://github.com/woocommerce/woocommerce-android/issues/8070

This PR migrates three rest clients:
1. `WooCommerceRestClient`
1. `WooSystemRestClient`
1. `WCUserRestClient` 

For the first rest client, as I wanted to keep the logic for customizing the timeout, I updated all the `Network` classes to do so by adding two new parameters: `requestTimeout` and `retries` to the GET functions (we can add them to the other functions if needed).
I didn't use the `RetryPolicy` as I don't want our `Network` classes to expose anything about `Volley`.

@JorgeMucientes I saw that we decided to keep the 15s timeout after the last experiment pe5sF9-vT-p2#comment-1227, is this specific to the `fetchSupportedWooApiVersion` function or all of the app's endpoints?
Anyway, we don't need to make any changes as part of this PR, since now with the `WooNetwork` class centralizing the logic for the WCAndroid app, we can easily specify a different timeout without impacting the other clients.

### Test
As with all the migration PRs, the focus should be on the code review by confirming we didn't miss any argument or used the wrong HTTP method.

For a small smoke test:
##### WordPress.com login

1. Open the example app.
2. Sign in using WordPress.com.
3. Click on Woo, then pick a site.
4. Click on "Log Woo Api Versions" and confirm it works.
5. Click on "Fetch WC Site Settings" and confirm it works.
6. Click on "Fetch WC Product Settings" and confirm it works.
7. Click on "GET USER_ROLE" and confirm it works.

###### Site Credentials login

1. Apply the following patch
```patch
IDEA additional info:
<+>UTF-8
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(revision e522154c8bc2f74f9d2039917d9de31eaf2b3240)
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/StoreSelectingFragment.kt	(date 1671625263544)
@@ -7,6 +7,8 @@
@@ -23,7 +25,7 @@
                 Button(context).apply {
                     text = "Select Site"
                     setOnClickListener {
                                 onSiteSelected(site)
                                 selectedSite = site
```
2. Open the app then sign in using site credentials (enter site address in the login
3. Repeat steps 3 to 7 from the above test case.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
@@ -35,6 +35,7 @@ import static org.wordpress.android.fluxc.network.xmlrpc.XMLRPCRequest.XmlRpcErr
 
 public abstract class BaseRequest<T> extends Request<T> {
     public static final int DEFAULT_REQUEST_TIMEOUT = 30000;
+    public static final int DEFAULT_MAX_RETRIES = DefaultRetryPolicy.DEFAULT_MAX_RETRIES;
     public static final int UPLOAD_REQUEST_READ_TIMEOUT = 60000;
 
     // Only used when enabledCaching is called - caching is off by default for all requests
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooCommerceRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooCommerceRestClient.kt
@@ -77,17 +77,12 @@ class WooCommerceRestClient @Inject constructor(
 
     suspend fun fetchSiteSettingsProducts(site: SiteModel): WooPayload<List<SiteSettingsResponse>> {
         val url = WOOCOMMERCE.settings.products.pathV3
-        val response = jetpackTunnelGsonRequestBuilder.syncGetRequest(
-            this,
-            site,
-            url,
-            emptyMap(),
-            Array<SiteSettingsResponse>::class.java
+        val response = wooNetwork.executeGetGsonRequest(
+            site = site,
+            path = url,
+            clazz = Array<SiteSettingsResponse>::class.java
         )
-        return when (response) {
-            is JetpackSuccess -> WooPayload(response.data?.toList())
-            is JetpackError -> WooPayload(response.error.toWooError())
-        }
+        return response.toWooPayload { it.toList() }
     }
 
     suspend fun enableCoupons(site: SiteModel): WooPayload<Boolean> {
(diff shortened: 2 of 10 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/BaseRequest.java
```diff
@@ -35,6 +35,7 @@
 
 public abstract class BaseRequest<T> extends Request<T> {
     public static final int DEFAULT_REQUEST_TIMEOUT = 30000;
+    public static final int DEFAULT_MAX_RETRIES = DefaultRetryPolicy.DEFAULT_MAX_RETRIES;
```
hichamboushaba, line 38: I just added this to avoid referencing `Volley`'s default value in the `Network` classes API.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooCommerceRestClient.kt
```diff
@@ -42,22 +28,18 @@ class WooCommerceRestClient @Inject constructor(
         overrideRetryPolicy: Boolean = false
     ): WooPayload<RootWPAPIRestResponse> {
         val url = "/"
-        val retryPolicy = when (overrideRetryPolicy) {
-            true -> jetpackTunnelGsonRequestBuilder.buildDefaultTimeoutRetryPolicy()
-            false -> null
+        val timeout = when (overrideRetryPolicy) {
+            true -> ROOT_ENDPOINT_TIMEOUT_MS
+            false -> BaseRequest.DEFAULT_REQUEST_TIMEOUT
```
JorgeMucientes, line 33: Thanks for making this cleaner and more explicit about which timeout policy is applied

JorgeMucientes:
Great job @hichamboushaba code looks good and works as expected!

> is this specific to the fetchSupportedWooApiVersion function or all of the app's endpoints?

Yes. Originally, I only changed the time-out for that request because that request was the one that seemed to make the user wait endlessly when picking a site from site picker. Refer to this spike I did in the past pe5sF9-t6-p2 where I mention `"One thing that got me investigating for a while and couldn’t find why it happens is that when WooCommerceStore: fetchSupportedWooApiVersion is triggered after clicking on the “Continue” button"`). Since I wanted to avoid taking any risks when changing the time-out policy for Jetpack tunnel without measuring first, I only updated the timeout for that single request. But after the A/B test showing there wasn't any real difference in success ratio between one timeout value and the other I think we are safe updating the default timeout value to 15s. In any case that is out of scope for this PR as you mentioned.

JorgeMucientes (approved):

PR #2612, opened 2022-12-22
Woo: Refactor logic of providing the JITM endpoint path

hichamboushaba:
As we are working on the REST API project, we are planning on migrating all of the `RestClients` in the app to a new architecture that uses the class `WooNetwork`. And while I was looking at the different `RestClients`, I was confused by how the JITM endpoint path was provided, as it references the `WPCom` base URL and the site ID, even though it's a WordPress REST API that uses the Jetpack namespace `/jetpack/v4`.

But after checking the code of `getJitmPath`, it clarified my confusion, as it simply returned a constant String `/jetpack/v4/jitm` without any reference to the other endpoint arguments.

This PR updates the logic to align with how we construct the endpoints in FluxC (the `txt` endpoint files), this would help avoid any confusion in the future when we want to migrate this class, the PR doesn't have any effect on the runtime:

| Before | After |
| --- | --- |
| Screen Shot 2022-12-22 at 13 03 13 | Screen Shot 2022-12-22 at 13 04 41 |

I tested and confirmed that the API still works as expected using local composite builds, but I don't think this requires updating FluxC in WCAndroid right now, the change will be picked up when we update FluxC to include other changes.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/jitm/JitmRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/jitm/JitmRestClientTest.kt
@@ -7,7 +7,7 @@ import org.junit.Before
 import org.junit.Test
 import org.mockito.kotlin.mock
 import org.mockito.kotlin.whenever
-import org.wordpress.android.fluxc.generated.endpoint.WPCOMREST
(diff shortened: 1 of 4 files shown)


AnirudhBhat (approved):
Thanks for fixing the path mapping @hichamboushaba. LGTM! :shipit:

PR #2614, opened 2022-12-22
Woo REST API: detect WooCommerce installation status

hichamboushaba:
⚠️ Please don't merge until #2611 is merged

With this PR, the `WooCommerceStore` will handle detecting the Woo installation status when fetching self-hosted sites, and it will persist the status in the column `hasWooCommerce`, with this.

With this change, we don't need the `patch` we've been using to test the example app with self-hosted sites, and we'll use this in the WCAndroid in the future to show an error if WooCommerce is not installed.

### Testing
1. Open the example app.
2. Sign in using site credentials.
3. Click on Woo.
4. Click on Select Site.
5. Confirm your site is shown.
6. Click on Log Sites.
7. Confirm the sites are fetched.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
@@ -10,6 +10,7 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.mockito.kotlin.any
 import org.mockito.kotlin.mock
+import org.mockito.kotlin.never
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import org.robolectric.RobolectricTestRunner
@@ -39,6 +40,7 @@ import org.wordpress.android.fluxc.network.rest.wpcom.wc.system.toDomainModel
 import org.wordpress.android.fluxc.persistence.WCSettingsSqlUtils.WCSettingsBuilder
 import org.wordpress.android.fluxc.persistence.WellSqlConfig
 import org.wordpress.android.fluxc.site.SiteUtils
+import org.wordpress.android.fluxc.store.AccountStore
 import org.wordpress.android.fluxc.store.SiteStore
 import org.wordpress.android.fluxc.store.SiteStore.OnSiteChanged
 import org.wordpress.android.fluxc.store.WooCommerceStore
@@ -59,6 +61,7 @@ class WooCommerceStoreTest {
     private val restClient = mock<WooSystemRestClient>()
     private val siteStore = mock<SiteStore>()
     private val wcrestClient = mock<WooCommerceRestClient>()
+    private val accountStore = mock<AccountStore>()
     private val settingsMapper = WCSettingsMapper()
 
     private val wooCommerceStore = WooCommerceStore(
@@ -69,7 +72,8 @@ class WooCommerceStoreTest {
         systemRestClient = restClient,
         wcCoreRestClient = wcrestClient,
         siteSqlUtils = TestSiteSqlUtils.siteSqlUtils,
-        settingsMapper = settingsMapper
+        settingsMapper = settingsMapper,
+        accountStore = accountStore
     )
     private val error = WooError(INVALID_RESPONSE, NETWORK_ERROR, "Invalid site ID")
     private val site = SiteModel().apply {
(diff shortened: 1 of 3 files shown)


example/src/test/java/org/wordpress/android/fluxc/wc/WooCommerceStoreTest.kt
```diff
@@ -270,10 +274,34 @@ class WooCommerceStoreTest {
...
+            whenever(siteStore.fetchSites(any())).thenReturn(OnSiteChanged(1, updatedSites = listOf(site)))
+
+            wooCommerceStore.fetchWooCommerceSites()
+
+            verify(siteStore).fetchSites(any())
+        }
+    }
+
+    @Test
+    fun `when the user is not signed in using WPCom, then don't fetch sites using WPCom API`() {
```
hafizrahman, line 290: 👍🏼

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -83,11 +91,12 @@ open class WooCommerceStore @Inject constructor(
...
-        // Fetch WooCommerce availability for jetpack cp sites
-        siteStore.sites.filter { it.isJetpackCPConnected }
+        // Fetch WooCommerce availability for non-Jetpack sites
+        siteStore.sites
+            .filterNot { it.origin == SiteModel.ORIGIN_WPCOM_REST && it.isJetpackConnected }
             .forEach { site ->
-                val isUpdated = fetchUpdatedSiteMetaData(site)
-                if (isUpdated.model == true && !fetchResult.updatedSites.contains(site)) {
+                val isUpdatedResult = fetchAndUpdateNonJetpackSite(site)
+                if (isUpdatedResult.model == true && !fetchResult.updatedSites.contains(site)) {
```
hafizrahman: np: I'd suggest `isResultUpdated` as variable name instead. Matches the pattern in `fetchAndUpdateNonJetpackSite()` where it has `val isMetadataUpdated` and `val isWooStatusUpdated`.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -100,38 +109,31 @@ open class WooCommerceStore @Inject constructor(
...
-                            )
-                    )
+        }
+
+        if (!site.isJetpackConnected || site.origin != SiteModel.ORIGIN_WPCOM_REST) {
+            fetchAndUpdateNonJetpackSite(site).also {
+                if (it.isError) return WooResult(it.error)
+
+                if (it.model == true) {
+                    emitChange(OnSiteChanged(1, listOf(site)))
```
hafizrahman: ❓ My understanding is a bit fuzzy here. 

I thought`OnSiteChanged`'s `updatedSites` parameter is meant to be filled with the latest site info from the fetching action. But here we are using the original `site` information that's passed into `fetchWooCommerceSite`. Can there be a case where the latest data after fetching is now different from the data in `site`?

hafizrahman (commented):
Overall looks good! I left a question about something that I haven't quite understood yet.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
hichamboushaba: This still returns the updated site, `SiteModel` is a mutable class, and we update its properties in the `fetchAndUpdateNonJetpackSite`.

But to make things clearer, I updated the logic to retrieve the object from the DB before returning it in f9cb65fc2cc2f00dcd4bfa8569b113d94a865146.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
hichamboushaba: I agree, done in ae7ca29dc64af1de143483056f35e2ffee67ff29

hichamboushaba:
Thanks @hafizrahman for the review, this is ready for another round.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -105,7 +105,7 @@ open class WooCommerceStore @Inject constructor(
 
         emitChange(OnSiteChanged(rowsAffected, fetchResult.updatedSites))
 
-        return WooResult(getWooCommerceSites())
+        return withContext(Dispatchers.IO) { WooResult(getWooCommerceSites()) }
```
hichamboushaba, line 108: I added this just to avoid hitting the DB on the main thread.

hafizrahman (approved):
Nice work, good to go :)

PR #2616, opened 2022-12-23
Application Passwords: Fix NPE errors

hichamboushaba:
This PR simply avoids using the double bang operator for reading the `data` property on success responses, as we found out that sometimes, the server would send an explicit `"data": null` (internal discussion p1671742442925879-slack-CGPNUU63E)

There is nothing to test, just code review.


malinajirka (approved):
LGTM, thanks!

PR #2619, opened 2022-12-27
Woo REST API: remove dependency on WPCom siteId in DB tables

hichamboushaba:
⚠️ Please don't merge yet.

Since we are going to support self-hosted sites, having a valid `siteId` in the `SiteModel` table is not garanteed anymore, since it comes from the WPCom API (internal discussion: pe5sF9-12J-p2).
In this PR, I update all the Woo tables that use the `siteId` for referencing the site, and replace it with the local site id.

Some additional improvements in the PR:
- In f51f013a0ff25879b64e77de0514ded4845b1312 I removed the table indices, as they were redundant (check the code [comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2619#discussion_r1057647398) for more context).
- In 690221335d74912de7745ae52b6760388a0846c3 I added indices to the foreign keys, Room was throwing a warning about it: `{column} column references a foreign key but it is not part of an index. This may trigger full table scans whenever parent table is modified so you are highly advised to create an index that covers this column`

The PR is big because of the DB schema file, and a bit of advice for an easier review: review by commit.

### Testing
Please refer to the WCAndroid [PR](https://github.com/woocommerce/woocommerce-android/pull/8084) for testing steps.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
```diff
@@ -627,3 +627,29 @@ internal val MIGRATION_20_21 = object : Migration(20, 21) {
         }
     }
 }
+
+@Suppress("MaxLineLength")
```
hichamboushaba, line 631: I like how each statement is being put in a line, it makes the reading easier IMO, let me know if yo don't think it's a good idea.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
```diff
@@ -627,3 +627,29 @@ internal val MIGRATION_20_21 = object : Migration(20, 21) {
...
+internal val MIGRATION_21_22 = Migration(21, 22) { database ->
```
hichamboushaba, line 632: I tried to use an Automigration with `RenameColumn` then override `postMigration`, and while it worked initially, it broke after adding some tables, as the order of the execution doesn't respect the order of declaration, and then foreign keys checks don't work.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/CouponEntity.kt
```diff
@@ -2,20 +2,20 @@ package org.wordpress.android.fluxc.persistence.entity
...
 import androidx.room.TypeConverters
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
+import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
 import org.wordpress.android.fluxc.persistence.converters.DiscountTypeConverter
 import java.math.BigDecimal
 
 @Entity(
     tableName = "Coupons",
-    primaryKeys = ["id", "siteId"],
-    indices = [Index("id", "siteId")]
```
hichamboushaba, line 14: SQLite [creates](https://www.sqlite.org/lang_createtable.html#unique_constraints) automatically an index for the primary key:
> UNIQUE and PRIMARY KEY constraints are implemented by creating a unique index in the database

So adding an additional index here was redundant (if not even being less performant).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/InboxNotesDao.kt
```diff
@@ -48,8 +49,8 @@ abstract class InboxNotesDao {
     }
 
     @Transaction
-    open suspend fun updateNote(siteId: Long, noteWithActions: InboxNoteWithActions) {
-        deleteInboxNote(siteId, noteWithActions.inboxNote.remoteId)
+    open suspend fun updateNote(localSiteId: LocalId, noteWithActions: InboxNoteWithActions) {
+        deleteInboxNote(noteWithActions.inboxNote.remoteId, localSiteId)
```
hichamboushaba, line 53: The arguments were mixed up for this function, so this change fixes a bug that we didn't know about.

0nko: Nice!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/migrations/Migrations.kt
0nko: Normally I split the lines to avoid the lint error, but I actually like it on a single line too, so let's keep it.. 😝

0nko (approved):
Excellent work, @hichamboushaba! I like how you opted to use `LocalId` amd `RemoteId` instead of `Int` and `Long`. Something I thought of while reading the code, maybe  we could make these wrapper classes [inline to improve the performance](https://kotlinlang.org/docs/inline-classes.html#members), WDYT? It's just a suggestion, not a blocker, though.

I've tested the associated WCAndroid PR and everything looks good.

hichamboushaba:
> I like how you opted to use LocalId amd RemoteId instead of Int and Long

All the credits goes to @wzieba for introducing them.

> maybe we could make these wrapper classes [inline to improve the performance](https://kotlinlang.org/docs/inline-classes.html#members), WDYT?

Yes, that's something that crossed my mind, WDYT @wzieba ?.
Still, I think it would probably be better to do it in a separate PR to make the purpose clearer.

wzieba:
> maybe we could make these wrapper classes [inline to improve the performance](https://kotlinlang.org/docs/inline-classes.html#members), WDYT?

Sounds good to me!

PR #2621, opened 2022-12-29
Woo REST API: Migrate InPersonPaymentsRestClient and JitmRestClient

### LINKED ISSUE #8093: [REST API] Migrate InPersonPaymentsRestClient

### LINKED ISSUE #8094: [REST API] Migrate JitmRestClient
JITMs work only for Jetpack sites, but we should still be able to fetch them if the user decides to use site credentials for accessing the site.

hichamboushaba:
Closes https://github.com/woocommerce/woocommerce-android/issues/8093
Closes https://github.com/woocommerce/woocommerce-android/issues/8094

This PR migrates InPersonPaymentsRestClient and JitmRestClient to use WooNetwork.
Even though these features depend on having a valid Jetpack connection in the site, there is still a possibility that we will allow people to connect using their site credentials, so with this PR, we'll make the API available to use for the two types of authentications, which would allow things to work regardless of the product decision later.

#### Testing
Normally no testing is required, the code review is what's necessary to confirm that the migration didn't miss anything.
There are also mocked and connected tests to confirm there is no regression:
1. `ReleaseStack_InPersonPaymentsWCPayTest`
2. `ReleaseStack_InPersonPaymentsStripeExtensionTest`
3. `MockedStack_InPersonPaymentsTest`
4. `MockedStack_JitmTest`

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_JitmTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_JitmTest.kt
@@ -16,6 +16,11 @@ class MockedStack_JitmTest : MockedStack_Base() {
 
     @Inject internal lateinit var interceptor: ResponseMockingInterceptor
 
+    private val testSite = SiteModel().apply {
+        origin = SiteModel.ORIGIN_WPCOM_REST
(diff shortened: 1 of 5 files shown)


kidinov (approved):
LGTM! Thanks!

PR #2626, opened 2023-01-05
[Blogging Prompts] Return list of prompts when calling fetchPrompts

### LINKED ISSUE #2624: [Blogging Prompts] fetchPrompts always returns null list on Success
The `fetchPrompts` method [signature](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/3cfaee1a4662cddb68a5a23773361412669c2f52/fluxc/src/main/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStore.kt#LL28-L32C58) leads users of that method to believe it will return the prompts that were fetched from the network and persisted in the local database, but it actually always returns a successful `BloggingPromptsResult` with `model = null`.

This seems to be intended behavior, based on the unit tests and original PR of this Store (#2383) but since the return type is `BloggingPromptsResult<List<BloggingPromptModel>>`, it feels like it should return a relevant List of `BloggingPromptModel`s containing the items that were fetched and inserted in the local database.

The implementation for this should be pretty straightforward, by making line 88 return the Result with `model = response.toBloggingPrompts()`, which is what is being inserted in the `DAO`.

https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/3cfaee1a4662cddb68a5a23773361412669c2f52/fluxc/src/main/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStore.kt#L83-L91

thomashorta:
Fixes #2624 
WordPress-Android Dependent PR: https://github.com/wordpress-mobile/WordPress-Android/pull/17722

## Issue
This was found while developing the Prompts List screen in WordPress-Android (https://github.com/wordpress-mobile/WordPress-Android/issues/17125).

As explained in issue #2624, it seems like this method returns a list of the prompts that were fetched and inserted in the local database, based on the `fetchPrompts` return type, but it was always returning an empty successful result instead.

## Changes
This PR fixes the issue above by making sure the list of inserted prompts is also returned in the `model` field of the `BloggingPromptsResult`

## Test Instructions

Use PR https://github.com/wordpress-mobile/WordPress-Android/pull/17722 to test this change.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStoreTest.kt
@@ -191,7 +191,7 @@ class BloggingPromptsStoreTest {
     }
 
     @Test
-    fun `when fetch prompts triggered, then all prompt model are inserted into db`() = test {
+    fun `when fetch prompts triggered, then all prompt models are inserted into db`() = test {
         val payload = BloggingPromptsPayload(PROMPTS_RESPONSE)
         whenever(
             restClient.fetchPrompts(
@@ -207,7 +207,7 @@ class BloggingPromptsStoreTest {
     }
 
     @Test
-    fun `given cards response, when fetch cards gets triggered, then all prompt model are returned in the result`() =
+    fun `given cards response, when fetch cards gets triggered, then all prompt models are returned in the result`() =
         test {
             val payload = BloggingPromptsPayload(PROMPTS_RESPONSE)
             whenever(
--- a/example/src/test/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStoreTest.kt
@@ -207,7 +207,7 @@ class BloggingPromptsStoreTest {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


example/src/test/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStoreTest.kt
```diff
@@ -207,7 +207,7 @@ class BloggingPromptsStoreTest {
     }
 
     @Test
-    fun `given cards response, when fetch cards gets triggered, then empty cards model is returned`() =
+    fun `given cards response, when fetch cards gets triggered, then all prompt model are returned in the result`() =
```
RenanLukas: Nitpick:
```suggestion
    fun `given cards response, when fetch cards gets triggered, then all prompt models are returned in the result`() =
```

RenanLukas (approved):
Thanks for the PR, @thomashorta. It works as expected.
I've left a super small comment.

LGTM :shipit:

example/src/test/java/org/wordpress/android/fluxc/store/bloggingprompts/BloggingPromptsStoreTest.kt
thomashorta: Yeah, I will also update the test above this one, which has the same mistake.

RenanLukas (approved):
LGTM :shipit:

PR #2254, opened 2022-01-31
Run fetchSites on background thread

kidinov:
Closes: #5729

### Description

Runs `fetchSites ` on the IO thread

### Testing instructions

* Open site selection screen via settings in debug
* Notice that the red frame is not visible anymore

#### Before:

https://user-images.githubusercontent.com/4923871/151559027-456f9a2d-ca14-4bcf-93aa-1645897c36ca.mp4

#### After:

https://user-images.githubusercontent.com/4923871/151559046-bd572f9c-14b5-439f-9fa2-2e19732f0389.mp4

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2628, opened 2023-01-10
Woo REST API: fix Woo detection for non-eligible users

hichamboushaba:
When fetching a site using REST API or Jetpack CP, we use the endpoint `/wc/v3/settings` to infer if Woo is installed or not, and we were inferring a valid installation when we get a success, while 404 results in Woo missing, and other errors are reported as fetch errors.

When working on the User role check for the REST API project, I faced an issue as the API returns `403`. This PR adjusts the logic to handle the `403` status code too, and if we get it we can infer that Woo is installed, since this means that API  call reached the Woo API controller and resulted in the permission error.

For the second commit of the PR, check the comments below.

### Testing
Please check the WCAndroid [PR](https://github.com/woocommerce/woocommerce-android/pull/8156)

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
@@ -1341,12 +1341,15 @@ open class SiteStore @Inject constructor(
     }
 
     suspend fun fetchSite(site: SiteModel): OnSiteChanged {
-        val updatedSite = if (site.isUsingWpComRestApi) {
-            siteRestClient.fetchSite(site)
-        } else {
-            siteXMLRPCClient.fetchSite(site)
+        return coroutineEngine.withDefaultContext(T.API, this, "Fetch site") {
+            val updatedSite = if (site.isUsingWpComRestApi) {
+                siteRestClient.fetchSite(site)
+            } else {
+                siteXMLRPCClient.fetchSite(site)
+            }
+
+            updateSite(updatedSite)
         }
-        return updateSite(updatedSite)
     }
 
     suspend fun fetchSites(payload: FetchSitesPayload): OnSiteChanged {
@@ -1357,7 +1360,9 @@ open class SiteStore @Inject constructor(
     }
 
     suspend fun fetchSitesXmlRpc(payload: RefreshSitesXMLRPCPayload): OnSiteChanged {
-        return updateSites(siteXMLRPCClient.fetchSites(payload.url, payload.username, payload.password))
+        return coroutineEngine.withDefaultContext(T.API, this, "Fetch sites") {
+            updateSites(siteXMLRPCClient.fetchSites(payload.url, payload.username, payload.password))
+        }
     }
 
     @Suppress("ForbiddenComment", "SwallowedException")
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -1341,12 +1341,15 @@ open class SiteStore @Inject constructor(
...
-        } else {
-            siteXMLRPCClient.fetchSite(site)
+        return coroutineEngine.withDefaultContext(T.API, this, "Fetch site") {
+            val updatedSite = if (site.isUsingWpComRestApi) {
+                siteRestClient.fetchSite(site)
+            } else {
+                siteXMLRPCClient.fetchSite(site)
+            }
+
+            updateSite(updatedSite)
```
hichamboushaba, line 1351: These functions in the `SiteStore` are suspendable, and yet, the DB logic wasn't offloaded correctly to the background thread, with this change, we follow the convention that the suspendable functions in the `Stores` use the default context.

This exact issue was fixed for `fetchSites` in #2254, and this PR updates the two other `fetch` functions.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
```diff
@@ -59,10 +63,14 @@ class WooSystemRestClient @Inject constructor(private val wooNetwork: WooNetwork
...
             is WPAPIResponse.Error -> {
-                if (response.error.type == NOT_FOUND) {
-                    WooPayload(false)
-                } else {
-                    WooPayload(response.error.toWooError())
+                when (response.error.volleyError?.networkResponse?.statusCode) {
+                    NOT_FOUND -> WooPayload(false)
+                    FORBIDDEN -> {
+                        // If we get a 403 status code, we can infer that Woo is installed,
+                        // the user just doesn't have permission to access the Woo API
```
hafizrahman, line 70: Just to discuss: 

when a user doesn't get a permission to access API, is this exclusively because of their user role? Or can it be because of security plugins blocking certain users from using API, for example?

If there can be multiple reasons, do they need to be handled in this PR (maybe be clarified in comments)? I'm mostly asking because this PR seems to be tied [to a task specifically about checking user role](https://github.com/woocommerce/woocommerce-android/pull/8156).

hafizrahman (approved):
This looks good to me, just added a comment for discussion.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
hichamboushaba: > when a user doesn't get a permission to access API, is this exclusively because of their user role? Or can it be because of security plugins blocking certain users from using API, for example?

We don't care about the cause here, our goal of this code is just to know whether Woo is installed or not, what happens later depends on other parts of the app: for example, we handle the role check, but if the cause is different, then the behavior won't be defined, same as what happens when accessing the site using WordPress.com or Jeptack CP.

Another point to keep in mind: this is needed just because we fetch the site using XMLRPC for now, if we get a chance to work on improving this part, we will handle the Woo detection differently, as we will be able to infer whether its present or not from fetching the root `/wp-json/` endpoint, which will be used also to fetch site details (check my discussion with Huong [here](https://github.com/woocommerce/woocommerce-ios/pull/8525#discussion_r1060061941)).

hafizrahman: Thank you for the explanation!

PR #2629, opened 2023-01-10
feat: bulk create variations

wzieba:
### Description

This PR adds `WCProductStore#createVariations`, which allows for bulk creation of variations required by https://github.com/woocommerce/woocommerce-android/issues/8157

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
@@ -1,0 +1,112 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.product
+
+import com.google.gson.JsonArray
+import com.google.gson.JsonParser
+import org.wordpress.android.fluxc.model.WCProductVariationModel
+
+object ProductVariationMapper {
+    /**
+     * Build json body of product items to be updated to the backend.
+     *
+     * This method checks if there is a cached version of the product stored locally.
+     * If not, it generates a new product model for the same product ID, with default fields
+     * and verifies that the [updatedVariationModel] has fields that are different from the default
+     * fields of [variationModel]. This is to ensure that we do not update product fields that do not contain any
+     * changes.
+     */
+    @Suppress("ForbiddenComment", "LongMethod", "ComplexMethod", "SwallowedException", "TooGenericExceptionCaught")
+    fun variantModelToProductJsonBody(
+        variationModel: WCProductVariationModel?,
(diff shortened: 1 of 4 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1575,6 +1578,59 @@ class WCProductStore @Inject constructor(
...
     }
 
+    suspend fun createVariations(
+        site: SiteModel,
+        productId: RemoteId,
+        variations: List<WCProductVariationModel>,
+    ): WooResult<BatchProductVariationsApiResponse> {
+        return coroutineEngine.withDefaultContext(API, this, "createVariations") {
+            val responses = variations
+                .chunked(VARIATIONS_CREATION_LIMIT)
```
wzieba, line 1588: API limits bulk creation to up to 100 variations per request. This part chunks a list of variations into sub-lists with size up to 100.

If any request is successful - we'll return success (we don't have a transactional system in our API so I think it's the best we can do).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
```diff
@@ -0,0 +1,112 @@
...
+            body["shipping_class"] = updatedVariationModel.shippingClass
+        }
+        // TODO: Once removal is supported, we can remove the extra isNotBlank() condition
+        if (storedVariationModel.image != updatedVariationModel.image && updatedVariationModel.image.isNotBlank()) {
+            body["image"] = updatedVariationModel.getImageModel()?.toJson() ?: ""
+        }
+        if (storedVariationModel.menuOrder != updatedVariationModel.menuOrder) {
+            body["menu_order"] = updatedVariationModel.menuOrder
+        }
+        if (storedVariationModel.attributes != updatedVariationModel.attributes) {
```
ThomazFB, line 101: I wish Kotlin had a way to easily compare fields linked by name to avoid having to manually implement this 😅 But it's good you left all this concentrated in a single mapper!

ThomazFB (approved):
Looks good! I'll leave this just approved if you want to control the best moment to introduce this to the Woo codebase!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductVariationMapper.kt
wzieba: Yes - it looks like something that could be automated. Maybe next time we could write our own solution (when there will be more time).

PR #2633, opened 2023-01-13
[REST API] Take into account UUID when deleting application password

### LINKED ISSUE #2632: [Application Passwords] Deleting a password will clear all site's passwords
Currently, we pass the `Application Password`'s name in the delete request, but the [API](https://github.com/WordPress/WordPress/blob/4191d80f708b99386f30d39aed129e8bff3fe752/wp-includes/rest-api/endpoints/class-wp-rest-application-passwords-controller.php#L78) doesn't support this flow, as it expects a UUID instead of the name. So it considers the body empty, and proceeds to clearing all of the passwords.

hafizrahman:
Fixes #2632

Test instructions in WC Android https://github.com/woocommerce/woocommerce-android/pull/8176

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
@@ -147,44 +147,44 @@ internal class ApplicationPasswordsManager @Inject constructor(
                     "UUID required for deletion is not found"
                 )
             )
-        } else {
-            val payload = if (site.origin == SiteModel.ORIGIN_WPCOM_REST) {
-                jetpackApplicationPasswordsRestClient.deleteApplicationPassword(
-                    site = site,
-                    uuid = uuid
-                )
-            } else {
-                wpApiApplicationPasswordsRestClient.deleteApplicationPassword(
-                    site = site,
-                    uuid = uuid
-                )
-            }
+        }
 
-            return when {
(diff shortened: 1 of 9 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
```diff
@@ -133,43 +137,55 @@ internal class ApplicationPasswordsManager @Inject constructor(
...
+                    "UUID required for deletion is not found"
+                )
             )
         } else {
-            wpApiApplicationPasswordsRestClient.deleteApplicationPassword(
-                site = site,
-                applicationName = applicationName
-            )
-        }
+            val payload = if (site.origin == SiteModel.ORIGIN_WPCOM_REST) {
```
hichamboushaba: np, since you have the `return` clause in the above `if` block, we can reduce nesting by removing the `else` clause and moving its block to the function's main block. (it would follow the early return pattern)

hichamboushaba (approved):
Nice work @hafizrahman 👏, :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
hafizrahman: Ah, good idea. I had wanted to simplify this area further. 👍🏼

PR #2634, opened 2023-01-13
Woo REST API: media upload using application passwords

### LINKED ISSUE #8017: [REST API] Add support for media upload.
As long as XMLRPC is still used for the initial login, media upload will continue to work using it.

But we might want to start working on using the REST API for the upload to be prepared for removing XMLRPC checks.

hichamboushaba:
Closes https://github.com/woocommerce/woocommerce-android/issues/8017

This PR adds a media rest client that supports application passwords, I'll explain some decisions in the code below.

tip: reviewing commit by commit might make understanding the changes easier.

### Testing
1. Open the example app.
2. Sign in using site credentials.
3. Open media screen.
4. Test some scenarios (make sure the app has the storage permission, Android removes it sometimes).
5. Confirm the scenarios work as expected.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_AppComponent.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_AppComponent.java
@@ -2,7 +2,6 @@ package org.wordpress.android.fluxc.release;
 
 import org.wordpress.android.fluxc.di.WCDatabaseModule;
 import org.wordpress.android.fluxc.example.di.AppConfigModule;
-import org.wordpress.android.fluxc.example.di.ApplicationPasswordsModule;
 import org.wordpress.android.fluxc.module.AppContextModule;
 import org.wordpress.android.fluxc.module.DatabaseModule;
 import org.wordpress.android.fluxc.module.MockedToolsModule;
@@ -25,7 +24,8 @@ import dagger.Component;
         MockedToolsModule.class,
         DatabaseModule.class,
         WCDatabaseModule.class,
-        ApplicationPasswordsModule.class
+        // TODO re-enable when we offer a way to fetch sites using the REST API
+        // ApplicationPasswordsModule.class
 })
 public interface ReleaseStack_AppComponent {
     void inject(ReleaseStack_AccountTest test);
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsConfiguration.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsConfiguration.kt
@@ -1,0 +1,21 @@
+package org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords
+
+import org.wordpress.android.fluxc.module.ApplicationPasswordsClientId
+import java.util.Optional
+import javax.inject.Inject
+
+data class ApplicationPasswordsConfiguration @Inject constructor(
+    @ApplicationPasswordsClientId private val applicationNameOptional: Optional<String>
+) {
+    val isEnabled: Boolean
+        get() = applicationNameOptional.isPresent
+
+    val applicationName: String
+        get() = applicationNameOptional.orElseThrow {
+            NoSuchElementException(
+                "Please make sure to inject a String instance with " +
+                    "the annotation @${ApplicationPasswordsClientId::class.simpleName} to the Dagger graph" +
+                    "to be able to use the Application Passwords feature"
+            )
+        }
+}
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsConfiguration.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsConfiguration.kt
@@ -4,7 +4,7 @@ import org.wordpress.android.fluxc.module.ApplicationPasswordsClientId
 import java.util.Optional
 import javax.inject.Inject
 
-data class ApplicationPasswordsConfiguration @Inject constructor(
+internal data class ApplicationPasswordsConfiguration @Inject constructor(
     @ApplicationPasswordsClientId private val applicationNameOptional: Optional<String>
 ) {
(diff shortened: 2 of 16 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_AppComponent.java
```diff
@@ -25,7 +24,8 @@
         MockedToolsModule.class,
         DatabaseModule.class,
         WCDatabaseModule.class,
-        ApplicationPasswordsModule.class
+        // TODO re-enable when we offer a way to fetch sites using the REST API
+        // ApplicationPasswordsModule.class
```
hichamboushaba, line 28: Currently, Application Passwords is used for XMLRPC sites, so we can't choose the API depending on the type of the site, which means that if we keep this, we'll cause all of the media tests to start using Application Passwords instead of XMLRPC.

When we get a chance to resume working on https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2583, we'll be able to differentiate between sites that expect XMLRPC and others that should use REST API, then we'll be able to re-enable the tests.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsConfiguration.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    val applicationName: String
+        get() = applicationNameOptional.orElseThrow {
+            NoSuchElementException(
+                "Please make sure to inject a String instance with " +
+                    "the annotation @${ApplicationPasswordsClientId::class.simpleName} to the Dagger graph" +
+                    "to be able to use the Application Passwords feature"
+            )
+        }
+}
```
hichamboushaba: Currently, a client app opts in for Application Passwords support by providing the `@ApplicationPasswordsClientId` to the dagger graph, to make this logic easier to understand and reuse, I added this class, it wraps the logic, and offers the application name to components that need it.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
```diff
@@ -48,51 +42,55 @@ import org.wordpress.android.util.AppLog
...
 import java.io.IOException
 import java.util.concurrent.ConcurrentHashMap
-import javax.inject.Inject
 import javax.inject.Named
-import javax.inject.Singleton
 
-@Singleton
-class WPV2MediaRestClient @Inject constructor(
-    dispatcher: Dispatcher,
+abstract class BaseWPV2MediaRestClient constructor(
```
hichamboushaba, line 47: Jetpack CP uses the `/wp/v2/media` API of WPCom, this API is exactly the same as the one from Core, so instead of creating a completely new rest client, I moved the common logic to this class, and created two subclasses that override the specific things for each API.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsConfiguration.kt
hafizrahman: Feels clear this way, nice improvement.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsConfiguration.kt
```diff
@@ -0,0 +1,21 @@
+package org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords
+
+import org.wordpress.android.fluxc.module.ApplicationPasswordsClientId
+import java.util.Optional
+import javax.inject.Inject
+
+data class ApplicationPasswordsConfiguration @Inject constructor(
+    @ApplicationPasswordsClientId private val applicationNameOptional: Optional<String>
```
hafizrahman: Just to clarify, merging this won't cause issue with how WC Android currently use this, yeah? Because the application name is already provided using the same annotation there?

hichamboushaba: Yes, there is no change to the API exposed to the clients, it's internal changes.

I think I will even mark `ApplicationPasswordsConfiguration` as internal, as it won't be needed outside of the `fluxc` module, I'll do it in a commit right now.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
```diff
@@ -123,10 +121,10 @@ class WPV2MediaRestClient @Inject constructor(
             }
 
             val request = Request.Builder()
-                    .url(url)
-                    .post(body = body)
-                    .header(WPComGsonRequest.REST_AUTHORIZATION_HEADER, "Bearer ${accessToken.get()}")
-                    .build()
+                .url(url)
+                .post(body = body)
+                .header(WPComGsonRequest.REST_AUTHORIZATION_HEADER, getAuthorizationHeader(site))
```
hafizrahman, line 126: In `ApplicationPasswordsMediaRestClient`'s implementation, this might return an empty string which I suppose will make the request fail.

Do we want to fail early here in that case? Or keep using the existing error handling below?

hafizrahman (approved):
This works well in my test and the code looks good as well. I left a couple of non-blocking questions (one of which you've already answered), feel free to merge whenever you're ready.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
hichamboushaba: Good point, we can do so, but I didn't want to complicate logic for an unlikely scenario: the media feature is deep in all of the clients, the user won't reach it until they have a valid application password, so the chances of having an empty string here should be very minimal.
Since the result is the same, just would take a slightly longer, I opted for code simplicity.

PR #2638, opened 2023-01-18
Woo REST API: handle UUID fetching when deleting a password

### LINKED ISSUE #2637: [Application Passwords] Handle duplicate application names
With the fix of https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/2632, the logic to handle duplicate application name doesn't work anymore, because we don't have the UUID to delete the password.

We need to add an intermediary API call to fetch the UUID before attempting the delete.

hichamboushaba:
Closes #2637 

⚠️ Please don't merge until the WCAndroid PR https://github.com/woocommerce/woocommerce-android/pull/8213 is merged.

After fixing #2632, the logic to handle duplicate application passwords stopped working, because when it happens, we will be missing the required UUID to delete the previous password.

With this PR, when we don't find a local UUID, we will attempt to fetch the password UUID from the site, then attempt the download.

#### Testing
1. Open the app, then sign in using site credentials.
2. Open the Woo tab, this will trigger a request using Application Passwords, which would generate a new password.
3. Go to the device settings, then clear the app's data.
4. Re-open the app.
5. Click on the Woo tab.
6. Confirm a new password was successfully created.
7. Optionally: you can also confirm using Flipper that a request to fetch the UUID was sent before the `DELETE` one.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordUUID.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordUUID.kt
@@ -1,0 +1,3 @@
+package org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords
+
+internal typealias ApplicationPasswordUUID = String
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ApplicationPasswordsLogger.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ApplicationPasswordsLogger.kt
@@ -42,9 +42,25 @@ class ApplicationPasswordsLogger @Inject constructor(
             })
     }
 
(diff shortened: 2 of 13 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordUUID.kt
```diff
@@ -0,0 +1,3 @@
+package org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords
+
+internal typealias ApplicationPasswordUUID = String
```
hichamboushaba, line 3: This typealias is just to improve readability.
I thought about using a value class, but since we don't do any validation nor operations, it won't bring any additional value.

0nko (approved):
The steps are working as described. I've merged the associated WCAndroid PR, so I'm going ahead and merging this one, as well. :shipit:

PR #2639, opened 2023-01-18
Woo REST API: some fixes

### LINKED ISSUE #8154: KeyStoreException: the master key android-keystore://_androidx_security_master_key_ exists but is unusable
Sentry Issue: [WOOCOMMERCE-ANDROID-5M2](https://sentry.io/organizations/a8c/issues/3847570654/?referrer=github_integration)

```
KeyStoreException: -62
    at com.woocommerce.android.util.ApplicationPasswordsTester$testApi$1.invokeSuspend(ApplicationPasswordsTester.kt:29)
...
(24 additional frame(s) were not displayed)

UnrecoverableKeyException: Failed to obtain information about key
    at com.woocommerce.android.util.ApplicationPasswordsTester$testApi$1.invokeSuspend(ApplicationPasswordsTester.kt:29)
...

KeyStoreException: the master key android-keystore://_androidx_security_master_key_ exists but is unusable
    at com.woocommerce.android.util.ApplicationPasswordsTester$testApi$1.invokeSuspend(ApplicationPasswordsTester.kt:29)
...
(17 additional frame(s) were not displayed)
```

### LINKED ISSUE #8173: NullPointerException
Sentry Issue: [WOOCOMMERCE-ANDROID-5PA](https://sentry.io/organizations/a8c/issues/3869687330/?referrer=github_integration)

```
NullPointerException: None
    at org.wordpress.android.fluxc.store.WooCommerceStore$fetchWooCommerceSite$updatedSite$1.invokeSuspend(WooCommerceStore.kt:140)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
    at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
    at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
...
(4 additional frame(s) were not displayed)
```

hichamboushaba:
Closes: https://github.com/woocommerce/woocommerce-android/issues/8154
Closes: https://github.com/woocommerce/woocommerce-android/issues/8173

This PR fixes two issues:

### [Issue 1](https://github.com/woocommerce/woocommerce-android/issues/8154)
It fixes an issue with the usage of the `EncryptedSharedPreferences`, as explained in the code comments, the library's documentation states that to avoid issues with decryption, the file should be excluded from cloud backup. But since we can't really handle this easily as it's part of a library, I opted for a destructive approach if we can't decrypt the file.
But after the test done with Jetpack users, 3 users faced an issue still, it's about the unusability of the encryption key, as I couldn't reproduce the issue, I just tried to follow the different recommendations from this bug report https://github.com/google/tink/issues/535:
1. Make the code thread-safe: the init part was already thread-safe because of using `lazy`, so I just made the other functions of reading and writing synchronized.
2. Remove the key when deleting preferences, here I made a small improvement over the suggested solution in the issue by checking the existence of the key before attempting to delete it.

### [Issue 2](https://github.com/woocommerce/woocommerce-android/issues/8173)
By checking some logs of some crashes there, I noticed this happens on logout, and this made sense: a race condition might result in deleting the site after the network fetch and before reading it from the DB.
I tried to fix this by gracefully failing if the site doesn't exist in the DB instead of using the double bang operator.

### Testing
Since both bugs are not reproducible, I don't there is anything to test, except maybe confirming a non-regression by testing login using site credentials in the example app and testing some scenarios in the Woo section.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
@@ -67,7 +67,6 @@ open class WooCommerceStore @Inject constructor(
         const val WOO_API_NAMESPACE_V3 = "wc/v3"
     }
 
-
     private val SiteModel.needsAdditionalCheckForWooInstallation
         get() = origin != SiteModel.ORIGIN_WPCOM_REST || isJetpackCPConnected
 
@@ -137,7 +136,15 @@ open class WooCommerceStore @Inject constructor(
             }
         } else false
 
-        val updatedSite = withContext(Dispatchers.IO) { siteStore.getSiteByLocalId(site.id)!! }
+        val updatedSite = withContext(Dispatchers.IO) {
+            siteStore.getSiteByLocalId(site.id)
+        } ?: return WooResult(
+            WooError(
+                type = GENERIC_ERROR,
+                message = "Site not found in the DB",
+                original = UNKNOWN
+            )
+        )
 
         if (isSiteUpdated) {
             emitChange(OnSiteChanged(1, listOf(updatedSite)))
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -147,19 +154,19 @@ open class WooCommerceStore @Inject constructor(
     }
 
     fun getWooCommerceSites(): MutableList<SiteModel> =
-            siteSqlUtils.getSitesWith(SiteModelTable.HAS_WOO_COMMERCE, true).asModel
+        siteSqlUtils.getSitesWith(SiteModelTable.HAS_WOO_COMMERCE, true).asModel
```
hichamboushaba, line 157: I accidentally reformated the file, so this change and the below ones are not related to the PR, the only relevant change for the PR in this file is the one above.

hafizrahman:
**Issue 1:**
Those are two long threads! Indeed deleting the file & key, and marking as `@synchronized` are the only mentioned workarounds that I could find there.

hafizrahman (approved):
Have checked the fix for both issue 1 and 2 and they look good to me. Hopefully these help with the majority of our users :shipit:

PR #2640, opened 2023-01-20
Better error mapping for create order endpoint

kidinov:
Maps `NO_CONNECTION`, `INVALID_SSL_CERTIFICATE` and `NETWORK_ERROR` to `NETWORK_ERROR`, which is better representation than `API_ERROR`

@JorgeMucientes if this mapping makes sense to you, maybe use this change in your PR right away? E.g. to show network-related error in both cases - Timeout and Network_error, wdyt?


JorgeMucientes (approved):

JorgeMucientes:
Thanks for applying this changes @kidinov. 

> @JorgeMucientes if this mapping makes sense to you, maybe use this change in your PR right away? E.g. to show network-related error in both cases - Timeout and Network_error, wdyt?

Yes let's merge this and I'll update fluxC version on my PR to update my error handling implementation 👍🏼

PR #2636, opened 2023-01-16
Remove CircleCI build status shield from `README.md`

mokagio:
We no longer use CircleCI.

PR #2641, opened 2023-01-24
[Test] Re-Enable Release Stack Encrypted Log Connected Test

ParaskP7:
This connected test was temporarily disabled as it was only failing in CircleCI. However, since CircleCI is not longer being used for this project, and instead, this project is now using Buildkite, this test is being re-enabled.

FYI: This work is based on my comment [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2636#pullrequestreview-1249917384).

PS: Those test were disabled as part of [this PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1662).

-----

### To test

1. There is nothing much to test here.
2. Verifying that all the CI checks are successful should be enough.

Having said the above, note the fact that for some time now the `Connected Tests` job is not required for `FluxC` PRs to be merged. This is because we have some tests still failing there that need to be fixed before we can re-enabled that check. Thus, we will need to check if that specific `ReleaseStack_EncryptedLogTest` test passes in Firebase manually.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_EncryptedLogTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_EncryptedLogTest.kt
@@ -5,7 +5,6 @@ import org.hamcrest.CoreMatchers.`is`
 import org.hamcrest.CoreMatchers.hasItem
 import org.junit.Assert.assertThat
 import org.junit.Assert.assertTrue
-import org.junit.Ignore
 import org.junit.Test
 import org.wordpress.android.fluxc.TestUtils
 import org.wordpress.android.fluxc.generated.EncryptedLogActionBuilder
@@ -27,7 +26,6 @@ private const val NUMBER_OF_LOGS_TO_UPLOAD = 2
 private const val TEST_UUID_PREFIX = "TEST-UUID-"
 private const val INVALID_UUID = "INVALID_UUID" // Underscore is not allowed
 
-@Ignore("Temporarily disabled: Tests are only failing in CircleCI and will be debugged next week")
 class ReleaseStack_EncryptedLogTest : ReleaseStack_Base() {
     @Inject lateinit var encryptedLogStore: EncryptedLogStore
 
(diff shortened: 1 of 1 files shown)


ParaskP7:
FYI: Actually, I am still seeing 1/2 test on this `ReleaseStack_EncryptedLogTest` suite failing, the `testQueueForUploadForInvalidUuid` (see test results [here](https://console.firebase.google.com/project/api-project-108380595987/testlab/histories/bh.f08c7ad6c1a268e8/matrices/6632754086758371012/executions/bs.c435089ec0a2124c/test-cases) and [here](https://console.firebase.google.com/project/api-project-108380595987/testlab/histories/bh.f08c7ad6c1a268e8/matrices/8927845962297825105/executions/bs.62c5328ea034fc69/test-cases)). Thus, I am not sure if that test suite is ready after all... 🤔

PS: The other [testQueueForUpload](https://console.firebase.google.com/project/api-project-108380595987/testlab/histories/bh.f08c7ad6c1a268e8/matrices/6632754086758371012/executions/bs.c435089ec0a2124c/testcases/325/test-cases) test seems to be working as expected. 🤷

malinajirka (approved):
Thanks for looking into this @ParaskP7! These things are easy to forget about and I truly appreciate you proactively fixing these types of 'rests'. 

Unfortunately, I'm not sure why the test is failing and my lack of knowledge+experience with encrypted logging doesn't help. As we discussed I'd suggest putting `@ignore` on the failing test so we improve the state a bit for now. Later, we can discuss it with someone who has experience with the logs as they might be able to figure out what's going on faster than I'd.

ParaskP7:
I agree @malinajirka and thanks so much for reviewing this change, unfortunately it wasn't that straightforward... 😊

As suggested, I'll proceed with ignoring this one failing test and merge the PR afterwards. 👍

ParaskP7:
FYI: I did the change, but now tests are failing because of an unrelated `testLoadImageFromHTTPAuthSite` test failure. 🤷 

PS: I'll proceed with merging this anyway as I don't think this is related. We also know that connected tests are not stable anyway, thus that shouldn't block us, especially since the failing test failing is not related to the `ReleaseStack_EncryptedLogTest` test suite.

PR #2643, opened 2023-01-27
Add support for connecting to self-hosted sites without XMLRPC

hichamboushaba:
Part of #2648 
Based on the work done in the previous PR https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2583

The PR adds a new way for fetching sites directly using the [Cookie authentication](https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/), the changes of the PR are:

- Add a new Origin value to the SiteModel.
- Create a new rest client SiteWPAPIRestClient that creates a SiteModel instance using the /wp-json response, and uses the Cookie authentication.
- Integrates the new rest client to SiteStore and add a new action to fetch a site using Cookie Authentication.
- Update the example app to test the new flow.
- Update the WPAPI requests to emit authentication errors when they are encountered.

### Test
1. Open the example app.
2. Click on "Sign in and fetch sites"
3. Enter your wp-admin credentials and the site URL, and make sure the Checkbox is kept unchecked.
4. Click on OK.
5. Confirm that the site name is correctly printed in the console.
6. Click on Plugins.
7. Test fetching/installing some plugins.
8. Confirm it works as expected.

Bonus: you can also use a plugin (for example this [one](https://wordpress.org/plugins/disable-xml-rpc-api/)) to disable XMLPRC to confirm that these changes work without it.


hichamboushaba:
@hafizrahman this should be ready to review now, I implemented what we discussed this morning in [42b01f5](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2643/commits/42b01f5f0f87fe39c1e601876423bc8f35ef82c4), and in [806b5d5](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2643/commits/806b5d59b6a7e00475de53e93ac7ca989ef2a925) I updated the comment about handling the `wp-login.php` status code to include the discussion I mentioned this morning.

hafizrahman (approved):
Works well in my test 👍🏼 

We also already pair programmed to go through the commits in this PR, and I'm comfortable with pre-approving this now. Thanks for walking me through the changes.

There was a Detekt error previously, but otherwise this should be good to go.

PR #2645, opened 2023-01-30
[Blogging Prompts] Add isPromptsCardOptedIn field to BloggingReminders entity

### LINKED ISSUE #2644: [Blogging Prompts] Add isPromptsCardOptedIn field to Reminders table
FluxC still doesn't fetch the Blogging Prompts / Reminders settings from the API (WIP here: #2532)). 

But we need to match the API fields in our local Room database and domain objects to be able to properly implement and persist (at least locally for now) the "Blogging Prompts" enabled settings, which lets the user enable/disable the Prompts feature in the Jetpack app (https://github.com/wordpress-mobile/WordPress-Android/issues/17803).

This means adding an `isPromptsCardOptedIn` field to those models so we can use it in the Jetpack app to display, or not, the feature to users, also giving them control to toggle it on/off.

Although this is a setting exclusive to the Prompts feature, the way the model and endpoint were built in the past actually merge the settings of both reminders and prompts into the `BloggingRemindersStore` and its models, so that's where the changes should happen.

thomashorta:
Fixes #2644 
WordPress-Android Dependent PR: https://github.com/wordpress-mobile/WordPress-Android/pull/17857

## Issue 
As explained in #2644, the `BloggingRemindersStore` still doesn't use the backend to fetch remotely saved data, but in order to have the same model fields we are adding the `isPromptsCardOptedIn` to the local `BloggingReminders` room Entity.

This is also being done so we can move forward with improvements to allow the user to enable or disable the Blogging Prompts feature, which will be controlled by the new `isPromptsCardOptedIn` field (only locally for now).

## Changes
This new `isPromptsCardOptedIn` field was added to the `BloggingReminders` Entity and defaults to true. To make sure the new DB is compatible with existing data in the user's device, an auto-migration was added from version 10 to version 11 of the Room DB.

Added a test for the new Auto-migration from version 10 to 11.

## Test Instructions
Use PR **WIP** to test this change.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/persistence/room/WPAndroidDatabaseMigrationTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/persistence/room/WPAndroidDatabaseMigrationTest.kt
@@ -1,7 +1,6 @@
 package org.wordpress.android.fluxc.persistence.room
 
 import androidx.room.testing.MigrationTestHelper
-import androidx.sqlite.db.framework.FrameworkSQLiteOpenHelperFactory
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.runner.AndroidJUnit4
 import org.assertj.core.api.Assertions.assertThat
@@ -22,8 +21,7 @@ class WPAndroidDatabaseMigrationTest {
     @JvmField
     val helper: MigrationTestHelper = MigrationTestHelper(
         InstrumentationRegistry.getInstrumentation(),
-        WPAndroidDatabase::class.java.canonicalName,
-        FrameworkSQLiteOpenHelperFactory()
+        WPAndroidDatabase::class.java,
     )
 
     @Test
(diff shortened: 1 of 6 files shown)


example/src/androidTest/java/org/wordpress/android/fluxc/persistence/room/WPAndroidDatabaseMigrationTest.kt
```diff
@@ -49,9 +48,9 @@ class WPAndroidDatabaseMigrationTest {
 
         assertThat(cursor.count).isEqualTo(1)
         cursor.moveToFirst()
-        assertThat(cursor.getInt(0)).isEqualTo(1000)
-        assertThat(cursor.getInt(2)).isEqualTo(1)
-        assertThat(cursor.getInt(4)).isEqualTo(0)
+        assertThat(cursor.getInt(cursor.getColumnIndex("localSiteId"))).isEqualTo(1000)
```
RenanLukas, line 51: 💯

RenanLukas (approved):
LGTM :shipit:

PR #2647, opened 2023-02-01
Longer timeout for the iap/orders call

kidinov:
Sometimes `iap/orders` may be slow to response. This PR adds a longer timeout for this endpoint

To test this change follow the IAP flow in the app


JorgeMucientes (approved):

JorgeMucientes:
Good catch on the timeout issues with `iap/orders` endpoint @kidinov 🥇. I did another test pointing my local branch to this fluxC branch and the flow worked well. It took really long, but everything worked as expected after all. 

:shipit:

Issue #2648, opened 2023-02-01
Woo REST API: Get rid of XMLRPC steps


hichamboushaba:
Currently, the WP REST API support in the WCAndroid app still depends on the XMLRPC discovery and authentication, but in reality, it's not needed.
This issue tracks the progress toward removing these steps:
- [x] Introduce a way to authenticate and fetch sites without XMLRPC. https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2643
- [x] Detect WooCommerce installation status. https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2643
- [x] Update `WooNetwork` to support the new site's type. https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2649
- [ ] Fetch WPCom's `siteId`, which can also allow detecting if Jetpack is installed and connected.

PR #2649, opened 2023-02-01
Application Passwords support without XMLRPC

hichamboushaba:
Part of #2648 

This PR builds on top of #2643, it adds the following changes:

1. Updates `WooNetwork` to work with the new SiteModel `origin`.
2. Update the logic for the WooCommerce installation check to be skipped for non `WPAPI_ORIGIN` sites.
3. Updates `MediaStore` to restrict Application Passwords rest client for the new origin.
4. Enables the Application Passwords connected tests.

#### Testing
1. Open the example app.
2. Sign in using site credentials (make sure to keep the XMLRPC option unchecked)
3. Click on Woo.
4. Select your site.
5. Do some testing, and confirm it works.


hafizrahman (approved):
LGTM! Also just for transparency, we've also paired together to go through these commits. All good.

PR #2650, opened 2023-02-02
Woo REST API: validate Application Passwords username before using it

hichamboushaba:
While working on the XMLRPC removal, I found a major bug with the way we store and retrieve the application passwords, but the fix is simple.
Let's start with the issue: during the logout, we delete the application password locally and remotely, but sometimes, we have some APIs in the queue, and when they get fired, they would trigger the creation of a new password, this is not really bad by itself, as the password is not harmful unless it's leaked, which won't happen. But the issue happens if we try to sign in later with a different username for the same site, the app will keep using the old password, as we don't check if it matches the correct username.

The fix is simple, when we find an existing password, we validate that it has the correct username before using it.

While the fix should work, and it's needed in all cases, we also need to start thinking about a way to cancel all pending network calls on logout.

Check the WCAndroid [PR](https://github.com/woocommerce/woocommerce-android/pull/8302) for testing steps.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
@@ -44,7 +44,8 @@ internal class ApplicationPasswordsManager @Inject constructor(
             )
         )
         val existingPassword = applicationPasswordsStore.getCredentials(site.domainName)
-        if (existingPassword != null) {
+        if (existingPassword != null &&
+            (site.username == existingPassword.userName || site.isUsingWpComRestApi)) {
             return ApplicationPasswordCreationResult.Existing(existingPassword)
         }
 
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
```diff
@@ -44,7 +44,8 @@ internal class ApplicationPasswordsManager @Inject constructor(
             )
         )
         val existingPassword = applicationPasswordsStore.getCredentials(site.domainName)
-        if (existingPassword != null) {
+        if (existingPassword != null &&
+            (site.username == existingPassword.userName || site.isUsingWpComRestApi)) {
```
hichamboushaba, line 48: I'm ignoring WPCom rest sites from the check here, and the reason is that the wp-admin username is not persisted for them in the DB, also the chances of the issue happening for them are minor if not non-existing even, because the API call to create the password depends on the WPCom token which gets cleared on logout.

0nko (approved):
:shipit:

PR #2651, opened 2023-02-03
Woo REST API: validate username on password deletion

hichamboushaba:
This is a continuation of #2650, in the previous PR I missed the scenario of password deletion, which might still cause issues in some rare cases: if a password exists locally for a different account, when attempting to delete the password, we will be using the wrong UUID, and the request will fail.

Check the WCAndroid PR for testing steps https://github.com/woocommerce/woocommerce-android/pull/8306

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsStore.kt
@@ -109,10 +109,6 @@ internal class ApplicationPasswordsStore @Inject constructor(
         }
     }
 
-    fun getUuid(site: SiteModel): ApplicationPasswordUUID? {
-        return encryptedPreferences.getString(site.uuidPrefKey, null)
-    }
-
     private val SiteModel.domainName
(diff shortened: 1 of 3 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsStore.kt
```diff
@@ -28,37 +30,38 @@ internal class ApplicationPasswordsStore @Inject constructor(
...
-            )
-        } else {
-            null
+    fun getCredentials(site: SiteModel): ApplicationPasswordCredentials? {
+        val username = encryptedPreferences.getString(site.usernamePrefKey, null)
+        val password = encryptedPreferences.getString(site.passwordPrefKey, null)
+        val uuid = encryptedPreferences.getString(site.uuidPrefKey, null)
+
+        return when {
+            !site.isUsingWpComRestApi && site.username != username -> null
```
hichamboushaba, line 39: Instead of doing the validation in multiple places in `ApplicationPasswordsManager`, I moved this logic to the `ApplicationPasswordsStore`.

0nko (approved):
Works as expected.

:shipit:

PR #2652, opened 2023-02-03
Woo: Domain price

0nko:
This is a subtask of woocommerce/woocommerce-android#8238. The PR adds the ability to fetch a price for a domain. This is useful for specific premium domains, which don't have the cost specified in a domain suggestion response.

I also added a parameter to specify a product type when fetching products (to be used with `domains` type).

This will be used in the domain change feature.

**To test:**

1. Start the example app and sign in
2. Tap on the Domains button
3. Tap on the Fetch domain price button
4. Enter a premium domain name, for example `blablabla.blog`
5. Verify that a very expensive price is returned :)
6. Try it again with a random word
7. Verify the UNKNOWN error type is returned


JorgeMucientes (approved):

JorgeMucientes:
Good works @0nko 🏅! Code looks good and works well. 

It kind of sucks we are going to have to trigger an API request for each premium domain in the domain suggestions list. Wasn't expecting that.

PR #2656, opened 2023-02-08
[Blogging Prompts] Set default prompts setting to site isPotentialBloggingSite

### LINKED ISSUE #2655: [Blogging Prompts] Set default prompts setting based on site being a blog
After some internal discussion, we decided to change some logic regarding the availability/visibility of the Blogging Prompts setting, and it will show to all sites (except self-hosted) so we also want to set the default of that setting (`BloggingRemindersModel.isPromptsCardEnabled`) to the `SiteModel.isPotentialBloggingSite` value.

With this change, the prompts feature will show up by default for blogging sites and will not for non-blogging sites (but non-blogging sites should be able to turn the feature on after WP issue https://github.com/wordpress-mobile/WordPress-Android/issues/17930).

thomashorta:
Fixes #2655 
WP-Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/17932

Update the `BloggingRemindersStore` to use the `isPotentialBloggingSite` field of `SiteModel` as default for `isPromptsCardEnabled` (a setting that defines if the prompts feature is enabled in the Jetpack app). 

This makes sure the feature defaults to OFF in sites that are not blogs, to keep the experience consistent in the new version of Jetpack.

This change can be tested in this Jetpack app PR: https://github.com/wordpress-mobile/WordPress-Android/pull/17932

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/store/BloggingRemindersStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/store/BloggingRemindersStoreTest.kt
@@ -13,6 +13,7 @@ import org.mockito.kotlin.whenever
 import org.wordpress.android.fluxc.model.BloggingRemindersMapper
 import org.wordpress.android.fluxc.model.BloggingRemindersModel
 import org.wordpress.android.fluxc.model.BloggingRemindersModel.Day.MONDAY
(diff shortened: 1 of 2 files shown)


RenanLukas (approved):
LGTM :shipit:

PR #2661, opened 2023-02-19
Feature/8260 fetch onboarding task

JorgeMucientes:
Part of: [#8260 ](https://github.com/woocommerce/woocommerce-android/issues/8260)

### Description
Fetch onboarding tasks from API. 

### Test 
1. Run FluxC Example app
2. Click on Woo option
3. Scroll down and click on Store Onboarding
4. Select a site 
5. Click Fetch onboarding task
6. See the fetched `TaskDto` data is displayed in the log output correctly. 

More in depth test will be done using WooCommerce app from [this PR ](https://github.com/woocommerce/woocommerce-android/pull/8359)

### DIFF
--- a/fluxc-processor/src/main/resources/wp-com-endpoints.txt
+++ b/fluxc-processor/src/main/resources/wp-com-endpoints.txt
@@ -147,5 +147,3 @@
 /products
 
 /plans
-
-/wc-admin/onboarding/tasks
--- a/fluxc-processor/src/main/resources/wp-com-endpoints.txt
+++ b/fluxc-processor/src/main/resources/wp-com-endpoints.txt
@@ -147,3 +147,5 @@
 /products
 
 /plans
+
+/wc-admin/onboarding/tasks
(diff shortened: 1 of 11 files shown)


fluxc-processor/src/main/resources/wp-com-endpoints.txt
```diff
@@ -146,4 +146,6 @@
 
 /products
 
-/plans
\ No newline at end of file
+/plans
+
+/wc-admin/onboarding/tasks
```
hichamboushaba: @JorgeMucientes I believe this endpoint should be in `wc-wp-api-endpoints` and not `wp-com-endpoints` 🙂.

Also, to align with how we declare the endpoints, we should declare only the path (`/onboarding/tasks`) in the `txt` file, and then the `namespace` has to be added to the `WCWPAPIEndpoint` class, same as the other namespaces `/wc/v3/...

JorgeMucientes: Hey @hichamboushaba thank you for the feedback TIL. I've addressed the suggestion in my last commit.

hichamboushaba: Thanks @JorgeMucientes

0nko (approved):
LGTM 👍

PR #2662, opened 2023-02-24
Add support for local product search

malinajirka:
This PR adds support for quering products by a searchQuery from the local database. It is used in [this PR in the woocommerce repo](https://github.com/woocommerce/woocommerce-android/pull/8391).

To Test:
[More on WCAndroid repo PR.](https://github.com/woocommerce/woocommerce-android/pull/8391)

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
@@ -176,7 +176,8 @@ object ProductSqlUtils {
         site: SiteModel,
         filterOptions: Map<ProductFilterOption, String>,
         sortType: ProductSorting = DEFAULT_PRODUCT_SORTING,
-        excludedProductIds: List<Long>? = null
+        excludedProductIds: List<Long>? = null,
+        searchQuery: String? = null,
     ): List<WCProductModel> {
         val queryBuilder = WellSql.select(WCProductModel::class.java)
                 .where().beginGroup()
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -199,7 +199,14 @@ object ProductSqlUtils {
             queryBuilder.contains(WCProductModelTable.CATEGORIES, categoryFilter)
         }
         if (searchQuery?.isNotEmpty() == true) {
-            queryBuilder.contains(WCProductModelTable.NAME, searchQuery)
+            queryBuilder
```
kidinov, line 202: I see there is a test of this class. Maybe to validate this code there as well?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -176,7 +176,8 @@ object ProductSqlUtils {
         site: SiteModel,
```
kidinov, line 176: np: maybe to put in the name that search query is also accepted by the method?

kidinov (approved):
LGTM! I left a couple of not blocking comments

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
malinajirka: I missed that, thanks! Added in 89f70ebbe923fe64fc50d28081b9f12d24372882

malinajirka:
Thanks so much for the review @kidinov ! I've implemented both suggestions and requested a re-review in case you'd have some feedback on the method name or the test method. Thanks 🙇

PR #2664, opened 2023-02-28
Add new endpoint for launching store

JorgeMucientes:
Part of [#8425](https://github.com/woocommerce/woocommerce-android/issues/8425)

### Description 
Adds new endpoint to launch stores that are unpublished

### Testing 

1. Create a new WP.com site with `eCommerce` plan 
2. Run example app
3. Navigate to Woo/Store Onboarding/ Select your new site 
4. Click Launch Site
5. See the success log 
6. Click again 
7. Check "Site has already been launched..." error

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
@@ -73,11 +73,11 @@ import org.wordpress.android.fluxc.model.RoleModel
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.SitesModel
 import org.wordpress.android.fluxc.network.BaseRequest.BaseNetworkError
-import org.wordpress.android.fluxc.network.rest.wpapi.site.SiteWPAPIRestClient
 import org.wordpress.android.fluxc.network.rest.wpapi.WPAPINetworkError
 import org.wordpress.android.fluxc.network.rest.wpapi.WPAPIResponse
 import org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords.ApplicationPasswordDeletionResult
 import org.wordpress.android.fluxc.network.rest.wpapi.applicationpasswords.ApplicationPasswordsManager
+import org.wordpress.android.fluxc.network.rest.wpapi.site.SiteWPAPIRestClient
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComGsonNetworkError
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Error
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequestBuilder.Response.Success
@@ -2046,6 +2046,20 @@ open class SiteStore @Inject constructor(
         return coroutineEngine.withDefaultContext(T.API, this, "Fetch domain price") {
             when (val response =
                 siteRestClient.fetchDomainPrice(domainName)) {
+                is Success -> {
+                    WPAPIResponse.Success(response.data)
+                }
+                is Error -> {
+                    WPAPIResponse.Error(WPAPINetworkError(response.error))
+                }
+            }
+        }
+    }
+
+    suspend fun launchSite(site: SiteModel): WPAPIResponse<Unit> {
+        return coroutineEngine.withDefaultContext(T.API, this, "Launch site") {
+            when (val response =
+                siteRestClient.launchSite(site)) {
                 is Success -> {
                     WPAPIResponse.Success(response.data)
                 }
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -2055,4 +2055,18 @@ open class SiteStore @Inject constructor(
             }
         }
     }
+
+    suspend fun launchSite(site: SiteModel): WPAPIResponse<Unit> {
```
hichamboushaba: Not a review yet, just a remark about the use of `WPAPIResponse` here.
The `WPAPIResponse` is a class that represents a network response that uses the Core WP REST API, while here we are using a WPCom API, and just [above](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ca3a62405614f9a67d2a9c25f5e9388c75af095d/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt#L279), we already expose the network result in the equivalent response class for WPCom API `WPComGsonRequestBuilder.Response`.

Historically, the Stores in FluxC would return a `Payload` class, in Woo we have the generic class `WooPayload` which makes this easier (thanks to Ondrej for adding it a long time ago), but in the pure WP stores, we tend to create different classes to represent success and error 🤷‍♂️.

JorgeMucientes: Thanks for providing this context @hichamboushaba. I'll refactor this into a custom `Payload` class. Tbh I find it a bit confusing some return types used stores, as well as where these Payload data classes should live.

hichamboushaba: > Tbh I find it a bit confusing some return types used stores, as well as where these Payload data classes should live.

I agree, but 🤷‍♂️ 😄

JorgeMucientes: Changes applied! Thanks for the suggestion!

0nko (commented):
I'm getting an error that says the site was already launched, even though it wasn't 🤔 

image

Response:
```
{"code":"already-launched","message":"This site has already been launched","data":null}
```

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
hichamboushaba: Thanks @JorgeMucientes, I think I caused a mistake here 🤦, truly sorry.
To be consistent with the rest of the events, the base class should be `OnChanged` instead of `Payload`, this will also make it easier to consume this in WCAndroid as a `Result`, as we have the `OnChangedException` to wrap those types.

Sorry again, I still get confused about the differences between `Payload` and `OnChanged` (designed for the original FluxC architecture that we don't use anymore).

JorgeMucientes: No worries @hichamboushaba, no big deal 👍🏼  I'll refactor it. 

So, just to get it clear this time 😅 . The way to go right now in FluxC stores to keep things consistent is to return an implementation of `OnChange` class right? And this on `OnChange` class is a legacy type that made sense back when fluxC used an event-based (using the bus) architecture. Is that correct? 
This also triggers another question. Do you know why `Payload` class was createdm what's it's purpose? Is essentially the same as `OnChange` class but with a clone method.

JorgeMucientes:
hmm, interesting @0nko. Was it a new site created from scratch? Was it a site you created a long time ago? I remember experiencing the same, with an old site I had subscribed to the `eCommerce` plan but that I had created couple months ago. So I'm wondering if only new sites are "launchable" using that endpoint. I'm going to do few more test though to see if I can reproduce that again and report back to somewhereWarm.

0nko:
@JorgeMucientes It's a fairly new site I created a few days ago when I was testing domain registration. I created it from the app.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
hichamboushaba: > The way to go right now in FluxC stores to keep things consistent is to return an implementation of OnChange class right? And this on OnChange class is a legacy type that made sense back when fluxC used an event-based (using the bus) architecture. Is that correct?

Yes, but whether it's still useful for the coroutines usage or not is something to discuss, I think it can still have some usefulness, but it needs some improvements to make it scale well: use a generic type and a better representation of errors.

> Do you know why Payload class was createdm what's it's purpose?

I wasn't around when the legacy architecture was still used, but from the looks of it, the `Payload` classes are intended as arguments for the `Action` classes, and not as representation of the events, which is the role of the `OnChanged` ones.

0nko (approved):
Since based on our conversation it's evident that there is a backend sync issue, the PR is good to go! Nice work! 🎖️

PR #2665, opened 2023-02-28
Excluded the constantly failing/crashing Connected Tests.

pachlava:
### Why

As discussed in pdcxQM-1St-p2, there's a number of connected tests that always fail, and this slows down the process of Diffs review a lot.

#### The following tests constantly fail:

- `org.wordpress.android.fluxc.ditests.OkHttpInjectionTest#assertNoRedirectsOkHttpClientBuilderHasExpectedSettings`
- `org.wordpress.android.fluxc.mocked.MockedStack_EditorThemeStoreTest#testEditorSettingsOrgUrl`
- `org.wordpress.android.fluxc.persistence.room.WPAndroidDatabaseMigrationTest#migrate10to11`
- `org.wordpress.android.fluxc.release.ReleaseStack_EncryptedLogTest#testQueueForUpload`
- `org.wordpress.android.fluxc.release.ReleaseStack_ReactNativeWPAPIRequestTest#testAuthenticatedCallToSelfSignedSslSite`
- `org.wordpress.android.fluxc.release.ReleaseStack_WPApiPluginTest#testFetchPlugins`
- `org.wordpress.android.fluxc.release.ReleaseStack_TaxonomyTestXMLRPC#testFetchTags`

#### For these three files, the execution leads to crash, leaving all remaining tests unexecuted

- `org.wordpress.android.fluxc.release.ReleaseStack_PostListTestWpCom`
- `org.wordpress.android.fluxc.release.ReleaseStack_PostListTestXMLRPC`
- `org.wordpress.android.fluxc.release.ReleaseStack_WCOrderListTest`
- 
**java.lang.IllegalStateException: No instrumentation registered! Must run under a registering instrumentation.**

### How

1. _I disabled only the tests that fail always._ I did not touch the ones that _might_ fail when they are executed as a part of the whole package, or the ones that fail on CI only. While this is not the best solution, it allows:
    - Have much more tests executed on CI (~540 VS initial ~22)
    - Run FluxC tests for Diffs much faster

2. Increased (60c43062400c65374ec47aa5383410b0b60cc43c) a timeout for tests execution from 30 min to 60 min, because the initial time is not enough for a full run.

### To test

- As can be seen from [related Connected Tests execution](https://console.firebase.google.com/u/1/project/api-project-108380595987/testlab/histories/bh.f08c7ad6c1a268e8/matrices/8934336767943555808/details?stepId=bs.476dcb244d25a340&testCaseId=539), while they still fail, now there are 539 executed tests, opposed to 22 previously.
- Optionally, run the tests locally. Most likely, they won't pass all at once, but they should pass at least by executing selected files.

### DIFF
--- a/.buildkite/connected-tests.sh
+++ b/.buildkite/connected-tests.sh
@@ -21,7 +21,7 @@ gcloud firebase test android run \
 	--type "instrumentation" \
 	--app "example/build/outputs/apk/debug/example-debug.apk" \
 	--test "example/build/outputs/apk/androidTest/debug/example-debug-androidTest.apk" \
-	--timeout "30m" \
+	--timeout "60m" \
 	--device "model=Pixel3,version=30,locale=en,orientation=portrait" \
 	--project "api-project-108380595987" \
 	--verbosity info \
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/ditests/OkHttpInjectionTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/ditests/OkHttpInjectionTest.kt
@@ -5,6 +5,7 @@ import okhttp3.OkHttpClient
 import okhttp3.internal.tls.OkHostnameVerifier
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Before
+import org.junit.Ignore
 import org.junit.Test
 import org.wordpress.android.fluxc.network.BaseRequest
 import javax.inject.Inject
@@ -42,6 +43,7 @@ class OkHttpInjectionTest {
     }
 
     @Test
+    @Ignore("Disabling as a part of effort to exclude flaky or failing tests. pdcxQM-1St-p2")
     fun assertNoRedirectsOkHttpClientBuilderHasExpectedSettings() {
         noRedirectsOkHttpClient.apply {
             assertThat(cookieJar).isEqualTo(customCookieJar)
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/ditests/OkHttpInjectionTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/ditests/OkHttpInjectionTest.kt
@@ -43,7 +43,8 @@ class OkHttpInjectionTest {
     }
 
     @Test
-    @Ignore("Disabling as a part of effort to exclude flaky or failing tests. pdcxQM-1St-p2")
+    @Ignore("Disabling as a part of effort to exclude flaky or failing tests." +
+        "See https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2665")
     fun assertNoRedirectsOkHttpClientBuilderHasExpectedSettings() {
         noRedirectsOkHttpClient.apply {
             assertThat(cookieJar).isEqualTo(customCookieJar)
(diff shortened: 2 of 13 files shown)


.buildkite/connected-tests.sh
```diff
@@ -21,7 +21,7 @@ gcloud firebase test android run \
 	--type "instrumentation" \
 	--app "example/build/outputs/apk/debug/example-debug.apk" \
 	--test "example/build/outputs/apk/androidTest/debug/example-debug-androidTest.apk" \
-	--timeout "30m" \
+	--timeout "60m" \
```
pachlava, line 24: The initial value is not enough for a full run.

example/src/androidTest/java/org/wordpress/android/fluxc/ditests/OkHttpInjectionTest.kt
```diff
@@ -42,6 +43,7 @@ class OkHttpInjectionTest {
     }
 
     @Test
+    @Ignore("Disabling as a part of effort to exclude flaky or failing tests. pdcxQM-1St-p2")
```
jostnes: nitpick: pdcxQM-1St-p2 when copied and pasted into a browser doesn't link to the page directly (we'll need to paste it somewhere on Github to work). maybe we can create an issue for this and have it link to that issue instead for context?

jostnes (commented):
thanks for making this update @pachlava! i added a nitpicky comment and have some questions:
1. are you planning to create another follow-up PR for this to try to get it to a "pass" state (i.e. ignore all flaky/failing tests)? (also not sure if possible as i'm not sure how "flaky" the tests are)
2. i'm not too familiar with the process of excluding steps and was wondering since the "Connected Tests" step results are ignored by Github, if that step is still running and all other steps have passed, will we be able to merge the PR without waiting for that step to complete? (since timeout is updated, that step is ~1hr)

example/src/androidTest/java/org/wordpress/android/fluxc/ditests/OkHttpInjectionTest.kt
pachlava: Good point @jostnes! I knew that the link won't work anywhere outside GH, but it didn't cross my mind to refer to a GH issue instead (🤦) - good idea 👍 ! This is what I did in 6df9a1d042f4e11ae056d2c448a8810073c760c3 - the comments now refer to the _current_ PR, where the link to P2 can be found at the very beginning.

pachlava:
Thank you for the review @jostnes ! 👋 

> 1. are you planning to create another follow-up PR for this to try to get it to a "pass" state (i.e. ignore all flaky/failing tests)? (also not sure if possible as i'm not sure how "flaky" the tests are)

Yes, this is planned as my next step. This is not mentioned in this PR, but it is mentioned in the P2 (quite deep however, it's hard to notice): pdcxQM-1St-p2#comment-1695 (last paragraph)

> 2. i'm not too familiar with the process of excluding steps and was wondering since the "Connected Tests" step results are ignored by Github, if that step is still running and all other steps have passed, will we be able to merge the PR without waiting for that step to complete? (since timeout is updated, that step is ~1hr)

That's a good question 👍 I didn't know this either, but after a commit to address your [other note](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2665#discussion_r1126025833), I had a chance to check, and see the `All checks have passed` note while the Connected Tests are still running. I still can't see the merge button enabled because there are no approving reviews, but I tend to think that it should be possible, because of `All checks have passed` status.

jostnes (approved):
Thanks for making the update and replies! Considering that this isn't final (there's a follow-up PR) and it doesn't block other PRs from merging sooner, it's a 👍 from me! 

:shipit:

pachlava:
Thanks for the re-review, @jostnes ! 

> Considering that this isn't final (there's a follow-up PR)

I think this is something I did not mention here in detail: the purpose of this PR is different from the follow-up one. They are not related. With this PR, I only aimed for excluding the tests that fail _locally_ (to make the Diffs review faster). Excluding the tests that fail on CI has a different purpose, which is to have a CI job with a set of connected tests that pass. This is why I did not want to mix those two things, so that it's easier to track down the changes in future, if needed 🙂

PR #2666, opened 2023-02-28
Get subscription details using order

### LINKED ISSUE #8376: [WooCommerce Subscriptions] Get subscriptions details using the order id
We need to request the subscription information using the order id (the goal will be to display the subscription information in the order detail screen).
This task is about the networking work to request the data.
Subscriptions are stored as child orders of the main order, so a subscription will have the order's id as the `parent_id` field.
We can get all Subscriptions from an order by querying all orders with the `parent_id` as the `order_id`. 
`query: {parent:[order_id]}` using `path=/wc/v3/subscriptions/`

```
curl --location --request GET 'https://public-api.wordpress.com/rest/v1.1/jetpack-blogs/214253715/rest-api/?json=true&locale=en_US&path=/wc/v3/subscriptions/%26_method=get&query=%7B%22parent%22%3A%20%5B355%5D%2C%22_fields%22%3A%22id%2Cparent_id%22%7D%0A%0A%0A' \
```
p1677103732340039-slack-CNA89KY6M

atorresveiga:
Close: https://github.com/woocommerce/woocommerce-android/issues/8376
### Why
We need to request the subscription information using the order id (the goal will be to display the subscription information in the order detail screen).
### Description
This PR adds the logic necessary to fetch the Order's subscription info using the order id. It does that by filtering the subscriptions and setting the parent query field equal to the order-id
### Testing instructions
I think that the best way to test this code is to follow the testing instructions on the App counterpart https://github.com/woocommerce/woocommerce-android/pull/8440


wzieba:
hi @atorresveiga , just want to confirm - is this ready for review? I'm asking because of the draft state.

atorresveiga:
> is this ready for review?

Not yet @wzieba 😁

atorresveiga:
This is ready for review now @wzieba 😁

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/subscription/SubscriptionRestClient.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import org.wordpress.android.fluxc.generated.endpoint.WOOCOMMERCE
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.Response
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooNetwork
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
+import org.wordpress.android.fluxc.utils.toWooPayload
+import javax.inject.Inject
+
+class SubscriptionRestClient @Inject constructor(private val wooNetwork: WooNetwork) {
```
wzieba: If I'm not mistaken, we don't need this in FluxC anymore, right?

wzieba (approved):

PR #2667, opened 2023-03-01
Magic Link: allow clients to define custom values for the flow and source

hichamboushaba:
This PR is needed for https://github.com/woocommerce/woocommerce-android/issues/8427, check the discussion pe5sF9-1e3-p2 for why.

Although it was possible to add more values to the already defined enums `AuthEmailPayloadFlow`, I opted to convert it to an interface instead because I think that the other solution doesn't scale well, and would be confusing to have values that are not reusable with other clients, let me know what you think about this.

Since the existing enum types implement the new interfaces, and we kept the constructors that use them, the change shouldn't break anything in the Login library nor the WPAndroid app.

Check the WCAndroid PR https://github.com/woocommerce/woocommerce-android/pull/8447 for testing instructions.


hafizrahman (approved):
👍🏼 The reasoning makes sense to me.

PR #2677, opened 2023-03-15
Improve error handling of the Cookie+Nonce authentication

hichamboushaba:
This PR improves the error handling for the Cookie+Nonce authentication, the goal is to allow differentiating between the different causes of the issue:

1. If we receive a 200 from the wp-login.php URL, then we try to extract the error message from the received page, and then we return it with a status "NOT_AUTHENTICATED", and if the extraction fails, we consider the error as "INVALID_RESPONSE".
2. If we receive 404 from the wp-login.php, we report it as CUSTOM_LOGIN_URL
3. If we reveive 404 from wp-admin URL, we report it as CUSTOM_ADMIN_URL (this can't be reproduced, check p1678898799819319-slack-C046HDZL87J for more details.
4. If we receive other errors, we report them as GENERIC.

##### Testing
Check the WCAndroid PR https://github.com/woocommerce/woocommerce-android/pull/8562

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
@@ -1,5 +1,6 @@
 package org.wordpress.android.fluxc.network.rest.wpapi
 
+import androidx.core.text.HtmlCompat
 import com.android.volley.NoConnectionError
 import com.android.volley.RequestQueue
 import org.wordpress.android.fluxc.Dispatcher
@@ -12,10 +13,14 @@ import org.wordpress.android.fluxc.network.rest.wpapi.Nonce.FailedRequest
 import org.wordpress.android.fluxc.network.rest.wpapi.Nonce.Unknown
 import org.wordpress.android.fluxc.network.rest.wpapi.WPAPIResponse.Error
 import org.wordpress.android.fluxc.network.rest.wpapi.WPAPIResponse.Success
+import org.wordpress.android.fluxc.store.SiteStore
 import org.wordpress.android.fluxc.utils.CurrentTimeProvider
+import org.wordpress.android.fluxc.utils.extensions.slashJoin
 import javax.inject.Inject
 import javax.inject.Named
 import javax.inject.Singleton
+
+private const val NOT_FOUND_STATUS_CODE = 404
 
 @Singleton
 class NonceRestClient @Inject constructor(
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
```diff
@@ -104,18 +129,32 @@ class NonceRestClient @Inject constructor(
...
             }
         }
     }
 
-    private fun slashJoin(begin: String, end: String): String {
-        val noSlashBegin = begin.replace("/$".toRegex(), "")
-        val noSlashEnd = end.replace("^/".toRegex(), "")
-        return "$noSlashBegin/$noSlashEnd"
+    private fun extractErrorMessage(htmlResponse: String): String? {
+        val regex = Regex("<div[^>]*id=\"login_error\"[^>]*>([\\s\\S]+?)</div>")
```
hafizrahman, line 150: Not a blocker: this function sounds like it's a generic way to find error message from a HTML response, but in practice it's specifically targeting a div with id="login_error". Should we maybe explain here why we specifically target it? In future core might change and this might not be the correct id anymore, and it will help troubleshoot things.

fluxc/src/main/java/org/wordpress/android/fluxc/utils/SiteErrorUtils.java
```diff
@@ -19,7 +23,17 @@ public static SiteError genericToSiteError(BaseNetworkError error) {
...
         }
 
-        SiteError siteError = new SiteError(errorType, error.message);
+        if (error instanceof WPAPINetworkError) {
+            String errorCode = ((WPAPINetworkError) error).getErrorCode();
+            wpapiError = new WPAPIError(
+                    WPAPIErrorType.Companion.fromString(errorCode),
+                    error.hasVolleyError() && error.volleyError.networkResponse != null
+                            ? error.volleyError.networkResponse.statusCode
+                            : 200 // volleyError will be missing when the request succeeds but the response is invalid
```
hafizrahman, line 32: 👍🏼

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
```diff
@@ -920,6 +921,23 @@ open class SiteStore @Inject constructor(
         UNABLE_TO_READ_SITE
     }
 
+    data class WPAPIError(
```
hafizrahman, line 924: ❓ not a blocker but wondering about the naming.

This class and the subsequent `WPAPIErrorType` do not seem to match the issues to me. Custom login URL and custom admin URL do not feel like an API issue, but more of a native login problem (i.e: error that happens when trying to login to wp-admin via app). Or maybe `NonceRequestError`?

Other classes in this file seems more contextual, like `SuggestDomainErrorType` or `AccessCookieErrorType`. Should we rename it to be more clear as well?

hafizrahman (approved):
Nice work, this works nicely in the tests I do in the Woo PR. Left a couple of non-blocking questions but it's OK to go otherwise.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
hichamboushaba: I'll skip it, the content of the function is simple enough to be considered self-documenting, and adding more comments to this file will just make it less maintainable (it already has many comments).

> In future core might change and this might not be the correct id anymore, and it will help troubleshoot things.

This is unlikely, as it would break all of the existing themes, if you check the history of the `wp-login.php` in Github, you'll see that the id has been then since it was first introduced (2013), it didn't change.
But even if it happens, as I said above, the function is quite simple, and it's gonna be easy to find it.

fluxc/src/main/java/org/wordpress/android/fluxc/store/SiteStore.kt
hichamboushaba: Currently in FluxC, Cookies+Nonce authentication is used only for WPApi, and this SiteError class is common for WPCom, XMLRPC and WPApi, so I wanted this additional field to be clear that it's used only for WPApi.

Anyway, all of this might be temporary, when we introduce the WebView login (regardless of the solution we will follow), we'll need to work on a change to differentiate the login step from the site fetching explicitly, so we'll move those files somewhere else, and hopefully we'll have a clearer architecture.

PR #2669, opened 2023-03-06
IPP feedback survey bug fix

AnirudhBhat:
### Please make sure to review the corresponding [WooCommerce PR](https://github.com/woocommerce/woocommerce-android/pull/8482) as soon as this PR gets merged. Or better yet, review the WooCommerce PR and get it ready before merging this one.

This PR contains fluxc changes to fix the [IPP survey feedback display logic bug](https://github.com/woocommerce/woocommerce-android/issues/8445). 

### Changes in this PR
1. Store "receipt_url" metadata into the metadata column in the Order database
2. Created a new database migration. The reason for this migration is explained as a comment in the `Migrations.kt` file

PR #2680, opened 2023-03-16
Push the WCAndroidDatabase schema file for version 23

hichamboushaba:
In the PR https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2669, the WCAndroidDatabase version was increased, but we missed adding the schema file of the new version.
As this file is needed to support [auto-migrations](https://developer.android.com/training/data-storage/room/migrating-db-versions#automated), I'm adding this small PR to add the file.


AnirudhBhat (approved):
Oops. Sorry for missing this. Thank you for adding this file. LGTM! :shipit:

PR #2681, opened 2023-03-16
[WP Individual JP Plugin] Save and retrieve filtered jetpack CP connected sites

### LINKED ISSUE #2678: [WordPress Individual JP Plugin] Save and retrieve filtered jetpack CP connected sites
We want to be able to retrieve the sites that were filtered out when fetching the user sites. Currently the code [allows the clients to filter](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/44a7e0f95fae507c7cc64f5d8a93f6227f9d9e8f/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt#L141) Jetpack CP (connected package / plugin ???) connected sites, as they have [some issues](https://github.com/wordpress-mobile/WordPress-Android/issues/15540#issuecomment-993752880).

WordPress is one of the clients that filter out those sites, but now we want to at least know if and what sites were filtered so we can show an informative modal dialog to the users saying they will get better support for those sites in the Jetpack app.

To do so, we want to store the filtered sites (instead of just ignoring them) preferably in a DB table, then add methods in the `SiteStore` to retrieve those sites. The saved sites don't need to be saved as as full `SiteModel` as we just need some basic information like:
- name
- url
- ids
- activeJetpackConnectionPlugins

thomashorta:
Fixes #2678

Create a new DAO in Room for storing user sites that have Jetpack Connection Package connection as well as add methods to `SiteStore` to retrieve those sites, which will be needed for https://github.com/wordpress-mobile/WordPress-Android/issues/18114.

Mechanisms were added in place to make sure those sites are properly reset upon fetching the sites and logging out from the app.

Some unit tests were also added for those new public methods.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/SitesModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/SitesModel.java
@@ -37,7 +37,7 @@ public class SitesModel extends Payload<BaseNetworkError> {
     }
 
     public void setSites(List<SiteModel> sites) {
-        this.mSites = sites;
+        mSites = sites;
     }
 
     public void setJetpackCPSites(List<SiteModel> jetpackCPSites) {
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/SitesModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/SitesModel.java
@@ -11,19 +11,36 @@ import java.util.List;
 public class SitesModel extends Payload<BaseNetworkError> {
     private List<SiteModel> mSites;
 
+    private List<SiteModel> mJetpackCPSites;
+
     public SitesModel() {
         mSites = new ArrayList<>();
+        mJetpackCPSites = new ArrayList<>();
     }
 
     public SitesModel(@NonNull List<SiteModel> sites) {
         mSites = sites;
+        mJetpackCPSites = new ArrayList<>();
+    }
+
+    public SitesModel(@NonNull List<SiteModel> sites, @NonNull List<SiteModel> jetpackCPSites) {
+        mSites = sites;
+        mJetpackCPSites = jetpackCPSites;
     }
 
     public List<SiteModel> getSites() {
         return mSites;
     }
 
+    public List<SiteModel> getJetpackCPSites() {
+        return mJetpackCPSites;
+    }
+
     public void setSites(List<SiteModel> sites) {
         this.mSites = sites;
+    }
+
+    public void setJetpackCPSites(List<SiteModel> jetpackCPSites) {
+        mJetpackCPSites = jetpackCPSites;
     }
 }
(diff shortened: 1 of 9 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/SitesModel.java
```diff
@@ -11,19 +11,36 @@
...
+    public List<SiteModel> getJetpackCPSites() {
+        return mJetpackCPSites;
+    }
+
     public void setSites(List<SiteModel> sites) {
         this.mSites = sites;
     }
+
+    public void setJetpackCPSites(List<SiteModel> jetpackCPSites) {
+        mJetpackCPSites = jetpackCPSites;
```
RenanLukas, line 44: Nitpick: would it make sense to use `this.mJetpackCPSites`? I know it doesn't make a difference since we're using Hungarian notation. I'm just bringing it up because `setSites` is using `this.mSites`.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -135,13 +135,15 @@ class SiteRestClient @Inject constructor(
         return when (response) {
             is Success -> {
                 val siteArray = mutableListOf<SiteModel>()
+                val jetpackCPSiteArray = mutableListOf<SiteModel>()
                 for (siteResponse in response.data.sites) {
                     val siteModel = siteResponseToSiteModel(siteResponse)
                     // see https://github.com/wordpress-mobile/WordPress-Android/issues/15540#issuecomment-993752880
+                    if (siteModel.isJetpackCPConnected) jetpackCPSiteArray.add(siteModel)
```
RenanLukas, line 141: Would it make sense to move this line so it's above the comment in line [141](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2681/files#diff-ac504807ac3755802bc839947ec85b4b045a906264eaf35ab97885016704df8aR141) so that the comment stays together with line [143](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2681/files#diff-ac504807ac3755802bc839947ec85b4b045a906264eaf35ab97885016704df8aR143)?

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/JetpackCPConnectedSitesDao.kt
```diff
@@ -0,0 +1,85 @@
...
+    )
+    data class JetpackCPConnectedSiteEntity(
+        val remoteSiteId: Long? = null,
+        val localSiteId: Int,
+        val url: String,
+        val name: String,
+        val description: String,
+        val activeJetpackConnectionPlugins: String,
+    ) {
+        fun toJetpackCPConnectedSite() = JetpackCPConnectedSiteModel(
```
RenanLukas: Nitpick: would it make sense to explicitly add the return type to this single expression function just because it's public?

RenanLukas (approved):
Thanks for the changes, @thomashorta 

Code LGTM :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/model/SitesModel.java
thomashorta: TBH I think it makes more sense to remove `this.` from `setSites` as it's the only place in the class using `this` reference to access a class field.

PR #2683, opened 2023-03-17
[Woo REST API] Prepare for Application Passwords' Web authorization

hichamboushaba:
This PR is needed to support the Web Authorization flow for the application passwords login, this flow differs from the other flow where the app would use the REST API to create the application password, instead the flow would be like this:
1. Fetches the REST root endpoint (`/wp-json/`), this will allow both checking if application passwords are enabled, and would give us the authorization URL.
2. Open the authorization URL.
3. The user will be asked to sign-in, then approve our connection.
4. When it's done, a redirection URL will be triggered, and it'll contain the created password.
5. Save the password to the `ApplicationPasswordsStore`, then proceed to finishing the signing.

This means that we'll have these two main changes to the rest of the flow:
1. We can't re-generate the password if it gets revoked.
2. For deleting the password, we need to use it itself for the authentication, as we don't have the site credentials.

To achieve the above, this PR has these changes:

1. Update the Application Passwords implementation to allow working with web authorization, and this includes the following changes:
    - Ability to save passwords directly by the clients 780f42d
    - Handle missing user credentials when attempting to regenerate the password 7501ab0
    - Handle deleting passwords using the Application Password itself 4682aa3
7. Update the WPApiAuthenticator class (called now `CookieNonceAuthenticator` for clarity) to work on network level, this is just an improvement to have a better code architecture ad815bd and d739927
8. Update the WPApi authentication to separate the login from the fetch, we need this since we need to fetch the site without authentication now 84fc30d
9. Handle fetching the Application Passwords authorization URL f2f60b9 and 548e889
10. Fix and Improve tests b449c4d and a2e0730

Reviewing each part separately will make the review easier.

The first part of WCAndroid changes to align with this PR is in https://github.com/woocommerce/woocommerce-android/issues/8598, then will add another PR to implement the Web Authorization.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
@@ -174,21 +174,32 @@ internal class ApplicationPasswordsManager @Inject constructor(
     suspend fun deleteApplicationCredentials(
         site: SiteModel
     ): ApplicationPasswordDeletionResult {
-        val uuid = applicationPasswordsStore.getCredentials(site)?.uuid
-            ?: fetchApplicationPasswordUUID(site).let {
+        val credentials = applicationPasswordsStore.getCredentials(site)
+
+        val payload = if (credentials == null) {
+            // If we don't have any saved credentials, let's fetch the UUID then delete the password using
+            // either the WP.com token or the self-hosted credentials.
+            val uuid = fetchApplicationPasswordUUID(site).let {
                 if (it.isError) return ApplicationPasswordDeletionResult.Failure(it.error)
                 it.uuid
             }
 
-        val payload = if (site.origin == SiteModel.ORIGIN_WPCOM_REST) {
-            jetpackApplicationPasswordsRestClient.deleteApplicationPassword(
-                site = site,
-                uuid = uuid
-            )
+            if (site.origin == SiteModel.ORIGIN_WPCOM_REST) {
+                jetpackApplicationPasswordsRestClient.deleteApplicationPassword(
+                    site = site,
+                    uuid = uuid
+                )
+            } else {
+                wpApiApplicationPasswordsRestClient.deleteApplicationPassword(
+                    site = site,
+                    uuid = uuid
+                )
+            }
         } else {
+            // If we have an Application Password, we can use it itself for the delete request.
             wpApiApplicationPasswordsRestClient.deleteApplicationPassword(
                 site = site,
-                uuid = uuid
+                credentials = credentials
             )
         }
 
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
@@ -36,7 +36,7 @@ internal class ApplicationPasswordsManager @Inject constructor(
      */
     private val SiteModel.supportsApplicationPasswordsGeneration
         get() = origin == SiteModel.ORIGIN_WPCOM_REST ||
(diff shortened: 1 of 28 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
```diff
@@ -79,15 +82,33 @@ internal class ApplicationPasswordsManager @Inject constructor(
...
+        } else {
+            return ApplicationPasswordCreationResult.Failure(
+                WPAPINetworkError(
+                    BaseNetworkError(
+                        GenericErrorType.UNKNOWN,
+                        "Site password is missing. " +
+                            "The application password was probably authorized using the Web flow"
+                    )
+                )
+            )
```
hafizrahman: ❓  What are your thoughts on moving this statement up and utilizing early returns?

Currently, it is somewhat difficult to determine the condition for this failure. By moving the if statement up and explicitly detailing the conditions, it should be more apparent to future readers.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
```diff
@@ -79,13 +83,36 @@ internal class ApplicationPasswordsManager @Inject constructor(
...
+                        "Site password is missing. " +
+                            "The application password was probably authorized using the Web flow",
+                        VolleyError(
+                            NetworkResponse(
+                                UNAUTHORIZED, null, true, System.currentTimeMillis(), emptyList()
+                            )
+                        )
+                    )
+                )
+            )
```
hafizrahman: ❓ What are your thoughts on moving this statement up and utilizing early returns?

Currently, it is somewhat difficult to determine the condition for this failure. By moving the if statement up and explicitly detailing the conditions, it should be more apparent to future readers.

hafizrahman:
❓As this is a significant PR, I think it would be helpful to list the architectural differences between the current and future methods of handling application passwords in the PR description. This list could include differences in how passwords are generated, regenerated, deleted, and so on, particularly related to handling results from webview-based authentication. 

While I appreciate the list of changes currently provided, I can't intuitively follow the purpose behind each one without further context. For instance, it is not immediately clear why "Handle fetching the Application Passwords authorization URL" is needed without reviewing the code. I believe that including a list of architectural differences would clarify the rationale behind these changes, and make future research on this topic easier when have to debug/troubleshoot.

I understand this is still a draft PR though, and that the current description might still be in the works and to be updated soon.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/applicationpasswords/ApplicationPasswordsManager.kt
hafizrahman: idea: maybe quickly check for existing credentials and return early if missing?

hichamboushaba: We can change the order of the ifs if you think it would make the code easier to follow, but it's not as simple as checking for existing credentials, we still have the support for Jetpack sites even if we don't use it in our app, and for them, the credentials will also be missing.

I'll work on a small refactoring, hopefully, it'll make the flow easier to follow, I'll let you know when I do it.

hichamboushaba: Let me know if [4d430b3](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2683/commits/4d430b3217d45362b8957c8b9d6a2156a0abe4bf) makes things easier to follow, it uses an early return but with a property extension that has a comment explaining the different paths.

hafizrahman (approved):
Just to document that we've paired to review this together, and I'm good with merging these changes now 👍🏼

PR #2684, opened 2023-03-18
Save product variation metadata

atorresveiga:
Closes https://github.com/woocommerce/woocommerce-android/issues/8389

### Why 
We need to save the product variants metadata locally to be able to get the subscription details information from the subscription variants.

### Description 
This PR adds the metadata column to the WCProductVariationModel and uses the field to save the metadata of product variants locally

### Testing
Use the [woo PR](https://github.com/woocommerce/woocommerce-android/pull/8601) and check that the metadata value of product variants is saved locally

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
@@ -36,7 +36,7 @@ open class WellSqlConfig : DefaultWellConfig {
     annotation class AddOn
 
     override fun getDbVersion(): Int {
-        return 180
+        return 181
     }
 
     override fun getDbName(): String {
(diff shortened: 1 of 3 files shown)


wzieba:
Please consider filtering the Product's metadata, as we do for Orders. See: `org.wordpress.android.fluxc.network.rest.wpcom.wc.order`StripOrderMetaData`, this prevents issues like https://wp.me/p91TBi-7qQ

atorresveiga:
> Please consider filtering the Product's metadata, as we do for Orders. See: `org.wordpress.android.fluxc.network.rest.wpcom.wc.order`StripOrderMetaData`, this prevents issues like p91TBi-7qQ-p2

Do you mean filtering only to save subscription information? 
Taking into account that subscription info starts with `_`, the same filter as `isDisplayableAttribute` would make no sense here.

At first, I didn't see a significant problem taking into account that it is the metadata for the ProductVariation (in my mind, it should have fewer values than Orders), and we don't even have a similar filter for Products. 
That said, I agree that only saving the data we will use will save the user's resources, and the filter is easy to implement. Thanks for your recommendation @wzieba

ThomazFB (approved):
Tested and could verify the column created and correctly mapped. Looks good!

ThomazFB:
Oh, I just noticed that now the discussion about the metadata filter. I'm sorry for merging the PR right away 😰

atorresveiga:
No problem @ThomazFB . Thanks for the review 🙏 .
I'll add the metadata strip in a different PR 😄

PR #2686, opened 2023-03-22
Urgent: Fix WP API root endpoint deserialization

hichamboushaba:
In the PR #2683, I added support for deserializing the `application-passwords` data in the `authentication` field, but in doing so, I accidentally removed the parent `JsonObjectOrEmptyArray` for the Authentication class, and this is causing a deserialization issue when the response is empty, which would block login for sites with empty `authentication` response.


JorgeMucientes (approved):

JorgeMucientes:
Change looks good @hichamboushaba. Just double-checking if there is anything specific I should test before merging. Is there a way to test application passwords login using FluxC `example` app? 

EDIT: Just saw that I can test the changes[ from this PR ](https://github.com/woocommerce/woocommerce-android/pull/8619). But how do I replicate an empty authentication response?

hichamboushaba:
@JorgeMucientes sorry for not putting enough context, to simulate it, just install a plugin that disables Application Passwords, normally it's the only option available by default, then try to login to the site (preferably in both WordPress.com by having Jetpack connected and site credentials)

PR #2687, opened 2023-03-22
Save only used metadata

atorresveiga:
Closes: https://github.com/woocommerce/woocommerce-android/issues/8621

### Description
This PR strips the Product Variation Metadata before saving it on the database. This way, the product variation item saved gets smaller, and we can prevent memory issues.

### Testing instruction
Install the app from the [woo PR](https://github.com/woocommerce/woocommerce-android/pull/8622) and check that only the supported keys are saved in the database. You ca use the database inspector or FluxC

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt
@@ -1,0 +1,32 @@
+package org.wordpress.android.fluxc.model
+
+import com.google.gson.Gson
+import com.google.gson.JsonArray
+import com.google.gson.JsonObject
+import javax.inject.Inject
+
+class StripProductVariationMetaData @Inject internal constructor(private val gson: Gson) {
+    operator fun invoke(metadata: String?): String? {
+        if (metadata == null) return null
+
+        val filteredMetadata = gson.fromJson(metadata, JsonArray::class.java)
+            .mapNotNull { it as? JsonObject }
+            .filter { jsonObject ->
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt
```diff
@@ -0,0 +1,32 @@
...
+                key in SUPPORTED_KEYS
+            }
+            .filter { jsonObject ->
+                val value = jsonObject[WCMetaData.VALUE]?.asString ?: ""
+                value.isBlank().not()
+            }
+            .toList()
+
+        return if (filteredMetadata.isEmpty()) null else gson.toJson(filteredMetadata)
+    }
```
ThomazFB: np: since this function has two filter rules, I recommend we verify them at once and operate them as a sequence to reduce the number of logical checks we do for each element, maybe something like this:

```suggestion
    operator fun invoke(metadata: String?): String? {
        if (metadata == null) return null

        return gson.fromJson(metadata, JsonArray::class.java)
            .mapNotNull { it as? JsonObject }
            .asSequence()
            .filter { jsonObject ->
                jsonObject[WCMetaData.KEY]?.asString.orEmpty() in SUPPORTED_KEYS &&
                    jsonObject[WCMetaData.VALUE]?.asString.orEmpty().isNotBlank()
            }.toList()
            .takeIf { it.isNotEmpty() }
            ?.let { gson.toJson(it) }
    }
```

ThomazFB (approved):
Looks good @atorresveiga! Thanks for adding tests to the `StripProductVariationMetaData` too! It made way easier for me to review the code!

atorresveiga:
Thanks for your recommendation, @ThomazFB! 🙏

PR #2692, opened 2023-03-27
Renewal subscriptions support

atorresveiga:
Closes: https://github.com/woocommerce/woocommerce-android/issues/8658

### Description
This PR adds the changes needed to get the renewal orders subscriptions information. It adds a new endpoint to fetch the subscription information using the subscription id, and it also adds the `SUPPORTED_KEYS` set to save metadata keys needed by the app (in this case, the `_subscription_renewal` metadata).

### Testing
It is better to test this PR along with [the woo PR](https://github.com/woocommerce/woocommerce-android/pull/8659).
Check that the "_subscription_renewal" field is saved in the local database for renewal subscriptions.

PR #2693, opened 2023-03-28
Display only displayable metadata fields

atorresveiga:
### Description
With the changes introduced in [2692](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2692) to display subscription information for renewal orders, we started storing non-displayable metadata in order metadata fields. The goal of this PR is to add the necessary functions to be able to retrieve only the displayable metadata for orders when we want to. For example, when displaying the custom fields screen.

### Testing
It is easy to test this PR along with [the woo PR](https://github.com/woocommerce/woocommerce-android/pull/8661)

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderMetaData.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderMetaData.kt
@@ -48,6 +48,7 @@ class StripOrderMetaData @Inject internal constructor(private val gson: Gson) {
             localSiteId = localSiteId,
             orderId = orderId,
             key = key.orEmpty(),
-            value = value.toString().replace(htmlRegex, "")
+            value = value.toString().replace(htmlRegex, ""),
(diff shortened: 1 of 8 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderMetaData.kt
```diff
@@ -48,6 +48,7 @@ class StripOrderMetaData @Inject internal constructor(private val gson: Gson) {
             localSiteId = localSiteId,
             orderId = orderId,
             key = key.orEmpty(),
-            value = value.toString().replace(htmlRegex, "")
+            value = value.toString().replace(htmlRegex, ""),
+            isDisplayable = isDisplayableAttribute
```
atorresveiga, line 52: The idea of this PR is to have only one source-of-true for whether a metadata field is displayable or not, the `isDisplayableAttribute` extension function

ThomazFB (approved):
Code reviewed and changes tested alongside https://github.com/woocommerce/woocommerce-android/pull/8661. Looks good!

PR #2703, opened 2023-04-05
Fix crash when order error message during fetching is null

hafizrahman:
This is to the FluxC part for issue https://github.com/woocommerce/woocommerce-android/issues/8729

Not sure how to test this since I'm not sure how to replicate an empty fetch error message, but probably OK to go without?

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCOrdersTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCOrdersTest.kt
@@ -541,6 +541,21 @@ class MockedStack_WCOrdersTest : MockedStack_Base() {
         assertEquals(payload.error.type, OrderErrorType.INVALID_RESPONSE)
     }
 
+    // Test for any order-related call that returns error with no message to ensure
+    // the app doesn't crash, by converting the null value to empty string.
+    @Test
+    fun testAnyOrderCallErrorWithNoMessage() = runBlocking {
+        val errorJson = JsonObject().apply {
+            addProperty("error", "some_error_message")
+        }
+        val remoteOrderId = 88L
+        interceptor.respondWithError(errorJson)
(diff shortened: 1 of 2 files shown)


kidinov (commented):
@hafizrahman thank you for working on that! 

> Not sure how to test this since I'm not sure how to replicate an empty fetch error message, but is it probably OK to go without?

Look into `MockedStack_WCOrdersTest` - you can test it fairly easily, and I think this import to do so, to avoid this kind of issues in future refactorings

hafizrahman:
@kidinov Thank you for your suggestion. Can you help me out with the commit I just pushed?

I feel that the test setup make sense. But, when I debug `fetchSingleOrder()` before `wpAPINetworkErrorToOrderError()` is called, `response.error.message` is already an empty string. I traced it out and found that in `BaseNetworkError(@NonNull VolleyError volleyError)`, a null message is always converted to empty message (`this.message = "")`.

So in practice I still can't replicate this null message issue, and this test I'm adding also returns true without the additional `.orEmpty()` fix on the first commit. Do you have any suggestions how to improve the test?

kidinov:
@hafizrahman 

I didn't go deep, but it has something to do with the error code (200 or not 200). Apparently, we parse those differently.

I fixed the test, so now it reproduces the crash and merges the PR so it will lend to the next release

be79030191f0508a633599f69bdb41e13d780fdc

kidinov:
@hafizrahman Yeah, I spent some time trying to figure out - no luck. It has to be something with condition on `WPComGsonRequest:162` 

Probably in some cases 

```
        if (error.hasVolleyError() && error.volleyError.networkResponse != null
                && error.volleyError.networkResponse.statusCode >= 400) {
```
That's false and then a message doesn't fall back to the empty string.

In any case, it's Friday, and let's merge what we have and see if this actually fixes the crash. I removed the test - it may be misleading

kidinov (approved):

PR #2707, opened 2023-04-11
[Woo REST API] Allow Jetpack connection and user fetching using application passwords as alternative to cookie+nonce authentication.

hafizrahman:
Part of https://github.com/woocommerce/woocommerce-android/issues/8749 and https://github.com/woocommerce/woocommerce-android/pull/8802

In Woo Mobile, after adding REST API login flow, we're exploring the possibility of completely using webview to generate application passwords for logging in, without any native login component. In this situation, we won't have any site credentials anymore, so we can't do API calls that require cookie+nonce authentication.

Currently, for the Jetpack activation feature in REST API login flow, there are some API calls involved that still use the cookie+nonce authentication:

1. Call for `/jetpack/connection/url`
2. Call for `/jetpack/connection/data`

This PR adds the possibility in the `JetpackStore` and `JetpackWPAPIRestClient` to do the above calls using application passwords.

> **Note**
> Please note that not using cookie+nonce is only needed for REST API login flow. Fetching that uses cookie+nonce is still possible, to make sure [Jetpack installation in WPCom login flow](https://github.com/woocommerce/woocommerce-android/pull/7889) is still supported.

Lastly, registering a Jetpack site is not required during Jetpack activation in REST API login flow, so we make sure it's not done in the REST API case.

For testing instruction, please test using https://github.com/woocommerce/woocommerce-android/pull/8785

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/jetpack/JetpackWPAPIRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/jetpack/JetpackWPAPIRestClient.kt
@@ -117,16 +117,7 @@ class JetpackWPAPIRestClient @Inject constructor(
             )
             return when (response) {
                 is Success<JetpackConnectionDataResponse> -> JetpackWPAPIPayload(
-                        response.data?.let {
-                            JetpackUser(
-                                    isConnected = it.currentUser.isConnected ?: false,
-                                    isMaster = it.currentUser.isMaster ?: false,
-                                    username = it.currentUser.username.orEmpty(),
-                                    wpcomEmail = it.currentUser.wpcomUser?.email.orEmpty(),
-                                    wpcomId = it.currentUser.wpcomUser?.id ?: 0L,
-                                    wpcomUsername = it.currentUser.wpcomUser?.login.orEmpty()
-                            )
-                        }
+                        response.data?.toJetpackUser()
                 )
 
                 is Error -> JetpackWPAPIPayload(response.error)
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/jetpack/JetpackWPAPIRestClient.kt
```diff
@@ -27,14 +28,29 @@ class JetpackWPAPIRestClient @Inject constructor(
...
     @Named("custom-ssl") requestQueue: RequestQueue,
     @Named("no-redirects") private val noRedirectsRequestQueue: RequestQueue,
     userAgent: UserAgent
 ) : BaseWPAPIRestClient(dispatcher, requestQueue, userAgent) {
     suspend fun fetchJetpackConnectionUrl(
-        site: SiteModel
+        site: SiteModel,
+        useApplicationPasswords: Boolean = false
     ): JetpackWPAPIPayload<String> {
+        if (useApplicationPasswords) {
```
hafizrahman, line 41: Note that Jetpack activation that uses cookie+nonce is still possible, to make sure [Jetpack installation in WPCom login flow](https://github.com/woocommerce/woocommerce-android/pull/7889) is still supported. This additional code is just adding application-passwords fetch on top of the existing call.

kidinov (approved):
Approving because it's not blocking. 

But I would advise:
* [Avoid passing booleans to a function](https://alexkondov.com/should-you-pass-boolean-to-functions/#:~:text=Booleans%20make%20it%20hard%20to,based%20on%20the%20call%20site.&text=Single%20argument%20functions%20and%20passing,named%20values%20improves%20the%20readability.). It's better to split them into 2
* Add tests

You may address that latter if you think it makes sense

hafizrahman:
@kidinov Thanks! Learned something new about avoiding passing booleans, it makes a lot of sense! I'll merge this first and work on your suggested improvements hopefully next week.

Issue #2710, opened 2023-04-11
[Blogging Prompts] Update endpoint to V3


thomashorta:
Refs pctCYC-KA-p2

Update the endpoint to V3 and make appropriate changes based on the reference above, since some parameters have different names and the API is slightly different.

PR #2711, opened 2023-04-11
Add support for WP.com V3 endpoints to code generator

thomashorta:
Part of #2710 

The text below was based on the [V2 endpoint PR](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/640).

This adds a new endpoint text file, `wp-com-v3-endpoints.txt`, which is used to generate an endpoint class for WordPress.com V3 endpoints.

These are generally used just like any other WP.com (V2) REST endpoints (as part of a `WPComGsonRequest`, and within a network client that extends `BaseWPComRestClient`). The only difference is to use the generated `WPCOMV3` class instead of `WPCOMV2` to build the endpoint URL.

Note: The V3 endpoints don't seem to currently have a versioning structure, so all FluxC offers is the one kind of URL (https://public-api.wordpress.com/wpcom/v2/...). We can add new URL methods to `WPComV3Endpoint.java` in the future if needed (if `/wpcom/v3.1/` and such are used).

To test, rebuild the project on this branch and run the unit tests (`./gradlew testDebug`).

---

This also adds the `blogging-prompts` v3 endpoint initially though it's still not being used in this PR.


ovitrif (approved):
LGTM :shipit:

PR #2712, opened 2023-04-12
[JPAndroid] Dashboard cards: Add activity to endpoint handling

zwarm:
Parent Issue: https://github.com/wordpress-mobile/WordPress-Android/issues/18207
Parent PR: In Progress

This PR updates the dashboard card endpoint to handle Activity Log data
Note: I had hoped to reuse `ActivityLogRestClient.ActivitiesResponse`, but the response from the dashboard cards includes an error, I thought it was best to create a copy and add the new response fields. I'm open to suggestions.
**Update**: After chatting with @RenanLukas, it was decided to move `error: String? = null ` to the existing `ActivitiesResponse` class and remove the copy from `CardsRestClient`. 

**Test**
- Do a code review
- Make sure the unit tests in `CardsRestClientTest.kt` and `CardsStoreTest.kt` pass
- Test the changes using the Parent PR

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/dashboard/CardsRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/dashboard/CardsRestClient.kt
@@ -6,7 +6,9 @@ import com.google.gson.annotations.SerializedName
 import org.wordpress.android.fluxc.Dispatcher
 import org.wordpress.android.fluxc.generated.endpoint.WPCOMV2
 import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.model.activity.ActivityLogModel
 import org.wordpress.android.fluxc.model.dashboard.CardModel
+import org.wordpress.android.fluxc.model.dashboard.CardModel.ActivityCardModel
 import org.wordpress.android.fluxc.model.dashboard.CardModel.PagesCardModel
 import org.wordpress.android.fluxc.model.dashboard.CardModel.PagesCardModel.PageCardModel
 import org.wordpress.android.fluxc.model.dashboard.CardModel.PostsCardModel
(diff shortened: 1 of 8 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/dashboard/CardsRestClient.kt
```diff
@@ -183,3 +190,59 @@ fun WPComGsonNetworkError.toCardsError(): CardsError {
     }
     return CardsError(type, message)
 }
+
+fun ActivitiesResponse.toActivityCardModel(): ActivityCardModel {
+    val error = error?.let { toActivityCardError(it) }
+
+    val activities = current?.orderedItems?.mapNotNull {
+        when {
+            it.activity_id == null -> {
```
RenanLukas: 💡 I was trying to think of a way to simplify these checks considering we don't have fallthrough for `when`. Do you think it would make sense to have the `== null` checks merged into a single branch?

e.g.
```
it.activity_id == null || it.summary == null || it.content?.text == null || it.published == null -> {
    null
}
```

RenanLukas (approved):
Thanks for the changes, @zwarm  
I've left a small optional suggestion but it's approved in case you don't think it makes sense.
LGTM :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/dashboard/CardsRestClient.kt
zwarm: @RenanLukas - You are correct here, no fallthrough here because I don't think it needs one. Since we are either returning `null` if it matches one of the cases or it returns a value from the `else` block when none of the previous cases match.

I can go either way on the readability, so I will take your suggestion and merge into a single line. Thanks for the suggestion. Thanks

PR #2716, opened 2023-04-14
Fix crash when Product call response error message is null

hafizrahman:
This is to fix issue https://github.com/woocommerce/woocommerce-android/issues/8823

Similar to a previous fix on https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2703 , I still wasn't able to replicate how a null error message is possible, so unfortunately this fix comes with no unit test.


kidinov (approved):
LGTM

Is this already clear if the fix for the orders worked?

PR #2719, opened 2023-04-26
Allow updating specific fields when updating media details

SiobhyB:
# Related PRs

* `Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/18327

# Description

At the moment, when a request is made to edit a media item, its `title`, `description`, `caption`, `alt`, and `parent_id` fields are all updated ([reference](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/5805ef65e9cebb0153c38ebf98a8f6fab646648c/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/media/MediaRestClient.java#L615-L641)).

With this PR and the related [FluxC PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/20606), the logic is updated to allow us to specify the fields that we want to update. The reason for this change is to avoid updating more fields than necessary when saving a post and updating the unattached media items (i.e. only update the field postID).

Further background and reasoning behind making this change can be found in https://github.com/wordpress-mobile/gutenberg-mobile/issues/5697, which describes a bug in the VideoPress block that's caused by the title being set for unattached items. The iOS equivalent to this change can be found at https://github.com/wordpress-mobile/WordPress-iOS/pull/20606.

# Testing

Please refer to [the Android PR](https://github.com/wordpress-mobile/WordPress-Android/pull/18327) for the central place, with the most up-to-date testing instructions.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
@@ -93,7 +93,7 @@ public class MediaModel extends Payload<BaseNetworkError> implements Identifiabl
     private boolean mDeleted;
 
     // Default fields to update upon request
-    private String[] fieldsToUpdate = {"parent_id", "title", "description", "caption", "alt"};
+    private String[] mFieldsToUpdate = {"parent_id", "title", "description", "caption", "alt"};
 
     @Override
     public boolean equals(Object other) {
@@ -331,11 +331,11 @@ public class MediaModel extends Payload<BaseNetworkError> implements Identifiabl
     }
 
     public String[] getFieldsToUpdate() {
-        return fieldsToUpdate;
+        return mFieldsToUpdate;
     }
 
     public void setFieldsToUpdate(String[] fieldsToUpdate) {
-        this.fieldsToUpdate = fieldsToUpdate;
+        this.mFieldsToUpdate = fieldsToUpdate;
     }
 
     //
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
@@ -365,7 +365,7 @@ public class MediaModel extends Payload<BaseNetworkError> implements Identifiabl
     //
     // Legacy methods
     //
-    
+
     public boolean isVideo() {
         return MediaUtils.isVideoMimeType(getMimeType());
     }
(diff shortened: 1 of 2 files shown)


jhnstn (approved):
The reasoning for the changes sound good.
I tested via  https://github.com/wordpress-mobile/WordPress-Android/pull/18327#pullrequestreview-1409391277

fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
```diff
@@ -92,6 +92,9 @@ public static MediaUploadState fromString(String stringState) {
     // Set to true on a successful response to delete via WP.com REST API, not stored locally
     private boolean mDeleted;
 
+    // Default fields to update upon request
+    private String[] mFieldsToUpdate = {"parent_id", "title", "description", "caption", "alt"};
```
antonis: This is not a change request but a suggestion. Wdyt of using an enum of strings for the default fields to update. 
E.g.
```
public enum DefaultFields {
    PARENT_ID("parent_id"),
    TITLE("title"),
    DESCRIPTION("description"),
    CAPTION("caption"),
    ALT("alt");
}
```
This way we could reference those directly in the `mFieldsToUpdate`, the [switch statement ](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2719/files#diff-c462b67fadc7580d065193713bb955927793d9a2131a0bf92808f068edcd4215R628) or [from WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/pull/18327/files#diff-5651461484d5ed6882ac4835e8256677dd133b228711d73aba1573471743b04cR18).

antonis (approved):
Great work @SiobhyB 👍 
The change LGTM and work as expected. I've added [a suggestion](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2719/files#r1183516951) which I don't consider blocking for the PR 🎉

fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java
SiobhyB: Thank you, Antonis! I've gone ahead to implement that suggestion in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2719/commits/8192ed142cae22f853f9c7e3c217fc8d0d6b857f and https://github.com/wordpress-mobile/WordPress-Android/pull/18327/commits/aabf39b4c13cd0052727f843b8f001569d7f5c2a 🙇‍♀️

PR #2720, opened 2023-04-27
[Room] Fix the DB schema for version 14

thomashorta:
The currently committed version 14.json has the field `version` as 13, which is odd and breaks builds when trying to update the Room version (in other PR).


RenanLukas (approved):
LGTM :shipit:

PR #2721, opened 2023-04-28
[Prompts] Use Blogging Prompts v3 endpoint

### LINKED ISSUE #2710: [Blogging Prompts] Update endpoint to V3

thomashorta:
Fixes: #2710 
WP-Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/18340

Start pointing to the v3 API for the Blogging Prompts endpoint. Some adjustments were needed to match the new response type, and how to request it (since it returns a root-level list in the JSON, instead of an object).

Other changes include:
- Update room DB version
- Update room Entity for prompts
- Update Prompts data models
- Create a new method for doing requests with Type Tokens instead of classes

Keep in mind this **replaces** the v2 endpoint, so we are not able to hit the old endpoint after this change. Currently, there are some https://github.com/Automattic/jetpack/pull/30255 on the way for the v3 to properly support the return of a single prompt per day (for future support of multiple prompts per day).

This can be tested[^1] in the WordPress Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/18340

[^1]: Notice that since the backend changes are still not deployed, testing it involves pulling backend code in a sandbox environment and pointing/proxying the app to that sandbox, which is only possible for Automatticians.


RenanLukas (approved):
LGTM :shipit:

RenanLukas (approved):
Thanks, @thomashorta . I've checked the latest changes and they LGTM :shipit:

PR #2722, opened 2023-05-03
Fix crash on sold items exceeding integer range

wzieba:
## Description

This PR addresses https://github.com/woocommerce/woocommerce-android/issues/8825, by introducing a custom deserializer which will assign `null` value if API value exceeds integer limits or is negative. 

### Testing

I don't see a reasonable way to test this fix - I think unit tests are sufficient. When this PR is merged, I'll prepare a PR to update FluxC tag in the WooCommerce repository.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCRevenueStatsModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCRevenueStatsModel.kt
@@ -2,6 +2,7 @@ package org.wordpress.android.fluxc.model
 
 import com.google.gson.Gson
 import com.google.gson.JsonParser
+import com.google.gson.annotations.JsonAdapter
 import com.google.gson.annotations.SerializedName
 import com.google.gson.reflect.TypeToken
 import com.yarolegovich.wellsql.core.Identifiable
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCRevenueStatsModel.kt
```diff
@@ -63,6 +65,7 @@ data class WCRevenueStatsModel(@PrimaryKey @Column private var id: Int = 0) : Id
         val netRevenue: Double? = null
         @SerializedName("avg_order_value")
         val avgOrderValue: Double? = null
+        @JsonAdapter(NonNegativeIntJsonDeserializer::class)
         @SerializedName("num_items_sold")
         val itemsSold: Int? = null
```
atorresveiga, line 70: WDYT if instead of limiting `itemsSold` to fit within the Int threshold, we use Long type on stats and have more bandwidth for bigger numbers?
From the [documentation](https://www.sqlite.org/datatype3.html), SQLite INTEGER type supports Long values, and the Long value Max in [Kotlin](https://kotlinlang.org/docs/numbers.html) is huge. This way, we will still present analytics to merchants with big numbers in order's sell

wzieba: The maximum value of `Int` is `2,147,483,647` - I believe it's more than enough to hold values of a number of sold items for any shop. I believe that values like `15032797007` app receives in `num_items_sold` are bugs, not deliberate values. WDYT?

wzieba: hi @atorresveiga, friendly reminder about this issue - maybe we can continue discussion and close this task this week? 🙂

atorresveiga (approved):
LGTM! :shipit:

PR #2723, opened 2023-05-03
[Prompts] Re-add Prompts v2 endpoint support temporarily

thomashorta:
Add a way to fetch prompts from the v2 endpoint in a way that's easy to revert in the future (separate function and duplication of needed response models). It's worth noting the v3 support was added in #2721 but needs a backend change (https://github.com/Automattic/jetpack/pull/30255) to be tested in production[^1], which is not yet deployed.

The Jetpack client app will be responsible for passing a value to `useV2Endpoint` when calling `fetchPrompts` in the Prompts Store.

This can be tested[^1] with the WordPress-Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/18361

[^1]: Notice that since the backend changes are still not deployed, testing it involves pulling backend code in a sandbox environment and pointing/proxying the app to that sandbox, which is only possible for Automatticians.


thomashorta:
We decided to not continue v2 support, not even for the transition. So I'm closing this PR.

PR #2725, opened 2023-05-08
Support core stock statuses

### LINKED ISSUE #8976: [Product Bundles]  Used core stock status value by default
To match wp-admin's behavior, we should use core's stocks status by default. This will allow us to match how wp-admin displays the backorder status.

### Test
1. Create a bundle product
2. Set the stock status to `backorder`
3. Open the product list screen
4. Check that the bundled product's status is displayed as `in stock` 

#### Expected
1. Create a bundled product
2. Set the stock status to `backorder`
3. Open the product list screen
4. Check that the bundled product's status is displayed as `backorder` matching wp-admin behavior

atorresveiga:
Close https://github.com/woocommerce/woocommerce-android/issues/8976

### Description
This PR goal is to match wp-admin behavior for bundled products when the product is on backorder status. Currently, we are using the  `bundle_stock_status` by default for bundled products, but this logic has a flaw. The bundled product field doesn't support the `backorder` stock status value, and this caused an inconsistency between what was displayed in the apps vs wp-admin. To solve this issue, we are now using the stock_status field for the cases when the bundle_stock_status is a core stock status.

### Testing instructions
My recommendation is to test this changes using this PR https://github.com/woocommerce/woocommerce-android/pull/8962 and composite builds
1. Create a Bundle product with backorder stock status
2. Open the product list screen
3. Check that the backorder stock status is displayed in the bundled product stock status description
4. Change one of the inner products in the bundle to out-of-stock
5. Open the product list screen
6. Check that the insufficient stock status is displayed in the bundled product stock status description

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/CoreProductStockStatus.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/CoreProductStockStatus.kt
@@ -10,6 +10,7 @@ enum class CoreProductStockStatus(val value: String) {
 
     companion object {
         private val valueMap = values().associateBy(CoreProductStockStatus::value)
+        val ALL_VALUES = valueMap.keys
 
         /**
          * Convert the base value into the associated CoreProductStockStatus object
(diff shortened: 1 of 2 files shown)


ThomazFB (approved):
Tested alongside https://github.com/woocommerce/woocommerce-android/pull/8962#pullrequestreview-1417503565. Looks good!

PR #2726, opened 2023-05-08
[EditorTheme] Detect block-based themes

RenanLukas:
**Parent issue**: https://github.com/wordpress-mobile/WordPress-Android/issues/18368
**Parent PR**: https://github.com/wordpress-mobile/WordPress-Android/pull/18385

Thanks @mkevins for working on [the POC](https://github.com/wordpress-mobile/WordPress-FluxC-Android/tree/try/support-block-based-themes) related to this issue. I've used much of your code 👍 

This PR introduces two new properties: `BlockEditorSettings#isBlockBasedTheme` and `BlockEditorSettings#hasBlockTemplates`. To detect if a Site's theme is block-based, the clients can call `EditorThemeStore#getIsBlockBasedTheme`.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/EditorThemeStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/EditorThemeStore.kt
@@ -63,6 +63,11 @@ class EditorThemeStore
         return editorThemeSqlUtils.getEditorThemeForSite(site)
     }
 
+    fun getIsBlockBasedTheme(site: SiteModel) =
+        getEditorThemeForSite(site)?.themeSupport?.let { themeSupport ->
+            themeSupport.isBlockBasedTheme || themeSupport.isBlockBasedTheme
+        }
+
     @Subscribe(threadMode = ThreadMode.ASYNC)
     override fun onAction(action: Action<*>) {
         val actionType = action.type as? EditorThemeAction ?: return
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/EditorThemeStore.kt
```diff
@@ -63,6 +64,14 @@ class EditorThemeStore
         return editorThemeSqlUtils.getEditorThemeForSite(site)
     }
 
+    fun getIsBlockBasedTheme(site: SiteModel): Boolean =
+        getEditorThemeForSite(site)?.themeSupport?.let { editorThemeSupport ->
+            getIsBlockBasedTheme(editorThemeSupport)
+        } ?: false
+
+    fun getIsBlockBasedTheme(editorThemeSupport: EditorThemeSupport): Boolean =
+        editorThemeSupport.isBlockBasedTheme || (editorThemeSupport.hasBlockTemplates ?: false)
```
thomashorta: 💡 Suggestion: if this only uses fields from `EditorThemeSupport` and doesn't really need anything from the `EditorThemeStore`, wouldn't it be better to move this logic to a computed field of the `EditorThemeSupport` model or an extension function?

I feel like having this in the store as an overload to a method that does IO is misleading. When I first read the code [here](https://github.com/wordpress-mobile/WordPress-Android/pull/18385/files#diff-f21797dc299c709abeec486de81a6a28239c0335aa587f10bfa517757c38c0d9R165-R169) I thought we needed to fetch 2 things to get the answer back, but that second one is just this check on local fields.

RenanLukas: Thanks for the suggestion, it definitely makes sense 👍

RenanLukas:
Hey, @thomashorta 👋 Thanks for taking the time to review my PR. I've implemented the suggestion you made in your initial review.

thomashorta (approved):
LGTM! :shipit:

PR #2730, opened 2023-05-12
Revert "[EditorTheme] Detect block-based themes"

RenanLukas:
Revert https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2726

@thomashorta found out that the changes on FluxC introduced a SQLite statement (`RENAME COLUMN`) that is only supported in more recent SQLite versions (3.25+) which is only present on Android [API 30 and newer](https://developer.android.com/reference/android/database/sqlite/package-summary), so it causes a crash when updating in older APIs.


thomashorta (approved):
Thanks for creating the revert PR. Let's merge it today so the next code freeze doesn't pick it.

PR #2735, opened 2023-05-16
Order creation - Search product by SKU (exact match) support

AnirudhBhat:
**Note:** Please make sure to review and get the corresponding [WooCommerce PR](https://github.com/woocommerce/woocommerce-android/pull/9053) ready before merging this one.

This PR adds support for searching products by exact match SKU. This functionality is needed for the "Scanned and Deliver" project where we add the ability to add products by scanning the barcode.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -85,6 +85,11 @@ class WCProductStore @Inject constructor(
         override fun toString() = name.toLowerCase(Locale.US)
     }
 
+    data class SkuSearchOptions(
+        val isSkuSearch: Boolean = false,
+        val isExactSkuSearch: Boolean = false,
+    )
+
     class FetchProductSkuAvailabilityPayload(
         var site: SiteModel,
         var sku: String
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -85,6 +85,11 @@ class WCProductStore @Inject constructor(
         override fun toString() = name.toLowerCase(Locale.US)
     }
 
+    data class SkuSearchOptions(
```
samiuelson: ##### 💡 Small suggestion

What do you think about refactoring `SkuSearchOptions` to an enum? 
```kotlin
enum class SkuSearchOptions { 
    Disabled, ExactSearch, PartialMatch
}
```
In my opinion, this would make the API a bit more clear and resistant to misuse like this:
```kotlin
skuSearchOptions = SkuSearchOptions(isSkuSearch=false, isExactSkuSearch=true)
```

samiuelson (commented):
Good work @AnirudhBhat 👍🏼  I've shared a small improvement suggestion. Let me know what you think.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
AnirudhBhat: Good suggestion. Implemented here: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2735/commits/f23763e8cf34e8e4389da79436988802e58a9f98

samiuelson (approved):
LGTM!

PR #2737, opened 2023-05-17
feat: add user ip country code as account property

wzieba:
### Description

This PR adds support for storing country code based on IP, which we get from /me/ endpoint from REST Api.


ThomazFB (approved):
Tested and reviewed through https://github.com/woocommerce/woocommerce-android/pull/9064

PR #2744, opened 2023-06-05
Support Geo API

wzieba:
See https://github.com/woocommerce/woocommerce-android/pull/9192/ for all details.


fluxc/src/main/java/org/wordpress/android/fluxc/model/AccountModel.java
```diff
@@ -39,8 +39,6 @@ public class AccountModel extends Payload<BaseNetworkError> implements Identifia
     @Column private boolean mTracksOptOut;
     @Column private boolean mUsernameCanBeChanged;
 
-    @Column private String mUserIpCountryCode;
```
wzieba, line 42: I've decided to, for now, not remove user ip country code field from the database. It'd require a risky migration of the whole `AccountModel` table (SQLite doesn't support removing a single table) and I don't think the benefits of removing one column are enough to justify this work and risk.

atorresveiga (approved):
LGTM! :shipit:

PR #2746, opened 2023-06-08
Add Jetpack AI API

### LINKED ISSUE #9213: Jetpack AI: FluxC API implementation
Implement Jetpack AI's API in FluxC to support AI work in WooCommerce Android.

hafizrahman:
Closes https://github.com/woocommerce/woocommerce-android/issues/9213

This PR adds a store and rest client dealing with the Jetpack AI API that allows for prompt completions.

`JetpackAIStore` offers two public functions:

1. `fetchJetpackAICompletionsForSite()`
2. `fetchJetpackAICompletionsForPost()`

Both are nearly similar, with the difference being the second function allows for marking a post as having AI generated content suggested to it. More detailed comments are also added to the relevant code.

The store intentionally does not use the Event Bus architecture anymore. A client can simply run the store functions in a coroutine and handle the result accordingly, like in the Example app sample.

I only added an Example test for `fetchJetpackAICompletionsForSite()`, and not for `fetchJetpackAICompletionsForPost()`. For the latter the API seems to work naively for now: it does not return error if the entered post id is not found or invalid, and instead returns completions still.

### To test:

1. Run Example app,
2. Sign in to a WordPress.com account,
3. Scroll down to find the "Jetpack AI" button, select it,
4. On the next screen, use "Select site" to pick a WordPress.com site,
5. Tap "Generate Haiku", and notice that the log area displays a haiku generated by AI.


0nko (approved):
Everything looks good. Nice work, @hafizrahman! 🏅

PR #2748, opened 2023-06-08
Adjust WCProductStore::fetchProducts to allow partial data sync

### LINKED ISSUE #2747: Adjust `WCProductStore::fetchProducts` to allow data update without wiping db table
Goal: modify the `WCProductStore::fetchProducts` function to allow data update without wiping db table - provide additional arg `forceRefresh: Boolean` indicating if the Products table should be wiped when fetching the first page of products.

This change is needed for: https://github.com/woocommerce/woocommerce-android/issues/9200

samiuelson:
Closes: #2747

This PR adds a new arg to the `WCProductStore::fetchProducts` - `forceRefresh: Boolean` allowing to do a partial data sync instead of full sync (with Product table wiping).

This change is needed for https://github.com/woocommerce/woocommerce-android/pull/9202

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -1451,7 +1451,9 @@ class WCProductStore @Inject constructor(
         }
     }
 
-    // Returns a boolean indicating whether more coupons can be fetched
+    /**
+     * @return Boolean indicating whether more products can be fetched.
+     */
     @Suppress("ComplexCondition")
     suspend fun fetchProducts(
         site: SiteModel,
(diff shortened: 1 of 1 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1460,7 +1462,8 @@ class WCProductStore @Inject constructor(
         sortType: ProductSorting = DEFAULT_PRODUCT_SORTING,
         includedProductIds: List<Long> = emptyList(),
         excludedProductIds: List<Long> = emptyList(),
-        filterOptions: Map<ProductFilterOption, String> = emptyMap()
+        filterOptions: Map<ProductFilterOption, String> = emptyMap(),
+        forceRefresh: Boolean = false
```
malinajirka: 💡  As we discussed, I'd suggest keeping the default to `true`. The main reason is that this method is a suspendable version of [#1](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/2b51e55b271b150e0c77607711f4549d23d25284/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt#L1091) and [#2](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/2b51e55b271b150e0c77607711f4549d23d25284/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt#L1678) and I think keeping them consistent for now might pay off. Wdyt?

malinajirka (commented):
Thanks @samiuelson, overall LGTM - I mentioned one idea, but feel free to keep it as is. I'll run thorough tests later.

malinajirka (approved):
Great job, works as expected!

PR #2751, opened 2023-06-12
WC analytics customers endpoint

kidinov:
Adds `wc_analytics/reports/customers` end point

Can be tested via tests or the PR - https://github.com/woocommerce/woocommerce-android/pull/9212


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/customer/CustomerRestClient.kt
```diff
@@ -107,4 +114,49 @@ class CustomerRestClient @Inject constructor(private val wooNetwork: WooNetwork)
 
         return response.toWooPayload()
     }
+
+    /**
+     * Makes a GET call to `wc-analytics/reports/customers` to fetch customers
+     *
+     */
+    suspend fun fetchCustomersFromAnalytics(
```
AnirudhBhat, line 122: NP: ❓ What do you think about writing unit tests for this class?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/customer/WCCustomerMapper.kt
```diff
@@ -50,36 +51,71 @@ class WCCustomerMapper @Inject constructor() {
...
+            lastName = model.lastName,
+            username = model.username,
+            billing = billing,
+            shipping = shipping
         )
     }
+
+    @Suppress("ComplexMethod")
+    fun mapToModel(site: SiteModel, dto: CustomerFromAnalyticsDTO): WCCustomerModel {
+        return WCCustomerModel().apply {
```
AnirudhBhat, line 89: NP: ❓ What do you think about writing tests for this mapper? I'm specifically thinking about testing the pieces that involve some kind of business logic like `isPayingCustomer`, `firstName`, and `lastName`.

AnirudhBhat (approved):
LGTM :shipit: 

Just left a couple of minor comments that are non-blocking.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/customer/CustomerRestClient.kt
kidinov: Added basic test here b379f268dad4976100218436adb051f6b7affb4d

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/customer/WCCustomerMapper.kt
kidinov: Good point

324aea8486512735ad2deccae352f8a10abf325a

PR #2753, opened 2023-06-14
Add new field can_blaze to site model

JorgeMucientes:
Part of this [task](https://github.com/woocommerce/woocommerce-android/issues/9209). 

Adds a new available API field to site model: `can_blaze`. The field is nested inside the site's data options when fetching sites, for example from `me/sites` endpoint. 

There's nothing to test on this PR. Changes can be tested in [this PR](https://github.com/woocommerce/woocommerce-android/pull/9228) where the new field is used.


hafizrahman (approved):
Looks good! 🚢

PR #2756, opened 2023-06-20
Strip product metadata

### LINKED ISSUE #9218: [Performance] Strip Product Metadata
We are requesting and saving the product's metadata entirely, even fields we are not using in the app. This metadata can be huge and increase the product size significantly. In recent research, I have found fields of more than 8MG that can even crash the app for being too big.

Image

This task is for stripping the metadata field as we do on Orders and only save the data we use in the app

atorresveiga:
Closes [9218](https://github.com/woocommerce/woocommerce-android/issues/9218)
### Why
Moving big chunks of data can impact the app's performance. This PR is part of the optimization we are working on as part of the Reliability project to improve the app's performance perception.

### Description
This PR adds the `StripProductMetaData` class to remove all unused fields from the product's metadata. Unused metadata fields can be huge, and removing them will result in better use of the device's resources, improving the app's performance and preventing OOM errors.

### Testing instructions
Test it with composite builds and, using App inspection or Flipper, check that only StripProductMetaData.SUPPORTED_KEYS are saved in the DB


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductMetaData.kt
```diff
@@ -0,0 +1,33 @@
...
+        return gson.fromJson(metadata, JsonArray::class.java)
+            .mapNotNull { it as? JsonObject }
+            .asSequence()
+            .filter { jsonObject ->
+                val isNullOrEmpty = jsonObject[WCMetaData.VALUE].isElementNullOrEmpty()
+                jsonObject[WCMetaData.KEY]?.asString.orEmpty() in SUPPORTED_KEYS && isNullOrEmpty.not()
+            }.toList()
+            .takeIf { it.isNotEmpty() }
+            ?.let { gson.toJson(it) } ?: EMPTY_JSON_ARRAY
+    }
```
ThomazFB, line 25: Just out of curiosity, I remember reviewing a code like this in the past. Did we implement a metadata strip before that got removed? I'm not finding anything like this in the trunk code base  😅

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -1486,6 +1497,7 @@ class ProductRestClient @Inject constructor(
                         val newModel = product.asProductModel().apply {
                             id = product.id?.toInt() ?: 0
                             localSiteId = site.id
+                            metadata = stripProductMetaData(metadata)
```
ThomazFB, line 1500: I I was wondering if maybe the `stripProductMetaData` use case could be called inside the `asProductModel()` transformation since all data strip happens right after it for every scenario. It could also be easier to maintain and enforce the metadata removal for future implementations.

But I also understand that since the stripping happens through a use case, the dependency injection will get weird and we would have to create the use case object manually, so if you think it's not worth the effort, it's OK too. I'm only concerned with making sure that we won't easily forget to strip that metadata in future changes.

WDYT?

ThomazFB (approved):
Tested focusing on verifying if the product add-ons feature is working just fine after the changes here, and looks good! I left some minor comments, but nothing is blocking, so I'm moving forward with the PR approval!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductMetaData.kt
atorresveiga: > I remember reviewing a code like this in the past.

Yes, but it was for product variations, not products. This is the strip use case class `StripProductVariationMetaData`

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
atorresveiga: >  I'm only concerned with making sure that we won't easily forget to strip that metadata in future changes.

Good point @ThomazFB. ☝️ 

> But I also understand that since the stripping happens through a use case, the dependency injection will get weird and we would have to create the use case object manually

That was also why I avoided adding the use case manually. I decided to go this way because I saw that we also had this line `localSiteId = site.id` on every call, adding the site id to the product model (so, we always pre-process the product before saving it to the database).
That said, I believe we can do better, so I'll add a task to refactor the `asProductModel()` and create a map class that can handle the injection of the use case. I will tackle this in a different PR.

ThomazFB: Nice! Thanks for that!

PR #2758, opened 2023-06-20
Resilient strip product variation

atorresveiga:
### Description 
This PR prevents the StripProductVariationMetaData from crashing when it receives unexpected data. 
The line `gson.fromJson(metadata, JsonArray::class.java)` will crash when the metadata value is empty `""`
Calling ` jsonObject[WCMetaData.VALUE]?.asString` will crash when the jsonObject value is an array with more than one element

### Testing instructions

1. Apply the patch and check that unit tests fail.
2. Revert the changes, run unit tests again, and check that everything pass.

<details>
  <summary>Patch with the old version</summary>

```
Index: plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt
IDEA additional info:
<+>UTF-8
diff --git a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt	(revision fdec197d8f5ffe04d2141f32071a4d0e6045a754)
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt	(date 1687280251643)
@@ -3,19 +3,18 @@
 import com.google.gson.Gson
 import com.google.gson.JsonArray
 import com.google.gson.JsonObject
-import org.wordpress.android.fluxc.utils.isElementNullOrEmpty
 import javax.inject.Inject
 
 class StripProductVariationMetaData @Inject internal constructor(private val gson: Gson) {
     operator fun invoke(metadata: String?): String? {
-        if (metadata.isNullOrEmpty()) return null
+        if (metadata == null) return null
 
         return gson.fromJson(metadata, JsonArray::class.java)
             .mapNotNull { it as? JsonObject }
             .asSequence()
             .filter { jsonObject ->
-                val isNullOrEmpty = jsonObject[WCMetaData.VALUE].isElementNullOrEmpty()
-                jsonObject[WCMetaData.KEY]?.asString.orEmpty() in SUPPORTED_KEYS && isNullOrEmpty.not()
+                jsonObject[WCMetaData.KEY]?.asString.orEmpty() in SUPPORTED_KEYS &&
+                    jsonObject[WCMetaData.VALUE]?.asString.orEmpty().isNotBlank()
             }.toList()
             .takeIf { it.isNotEmpty() }
             ?.let { gson.toJson(it) }

```

</details>

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductVariationMetaData.kt
@@ -3,18 +3,19 @@ package org.wordpress.android.fluxc.model
 import com.google.gson.Gson
 import com.google.gson.JsonArray
 import com.google.gson.JsonObject
+import org.wordpress.android.fluxc.utils.isElementNullOrEmpty
 import javax.inject.Inject
 
 class StripProductVariationMetaData @Inject internal constructor(private val gson: Gson) {
     operator fun invoke(metadata: String?): String? {
-        if (metadata == null) return null
+        if (metadata.isNullOrEmpty()) return null
 
         return gson.fromJson(metadata, JsonArray::class.java)
             .mapNotNull { it as? JsonObject }
             .asSequence()
             .filter { jsonObject ->
-                jsonObject[WCMetaData.KEY]?.asString.orEmpty() in SUPPORTED_KEYS &&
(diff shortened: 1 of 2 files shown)


ThomazFB (approved):
Good catch and nice work adding test coverage to the crash scenario! 😄

PR #2763, opened 2023-07-01
[JPAandroid] Add endpoints and DB layer for Blaze Campaigns

zwarm:
This PR adds support for:

- Calling the blaze search campaign endpoint
- DB layer for storing blaze campaign data and associated paging information
- Clicks and impressions - Defaulting null to "0" 
- `budgetCents` is stored as a Long in the DB, it's up to the UI layer to format as USD.
- URL is set as nullable, even though it is required for JP blaze. (Tumblr doesn't require an image, so setting it as nullable will help prevent possible crashes in the future
- A new constructor was added to `BaseWPComRestClient.java` to accept a `AcceptHeaderStrategy`. This strategy encapsulates Accept Headers that need to be applied to the request. Other strategies can be easily added and injected into child rest client classes.

This PR will use [JPAndroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/18686) as a testing companion. However, if needed, a separate JP Android PR will be created with a hash only.


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCampaignsRestClient.kt
```diff
@@ -0,0 +1,96 @@
...
+class BlazeCampaignsRestClient @Inject constructor(
+    private val wpComGsonRequestBuilder: WPComGsonRequestBuilder,
+    dispatcher: Dispatcher,
+    appContext: Context?,
+    @Named("regular") requestQueue: RequestQueue,
+    accessToken: AccessToken,
+    userAgent: UserAgent
+) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent, true) {
+    suspend fun fetchBlazeCampaigns(
+        site: SiteModel,
```
AjeshRPai, line 34: Disclaimer: I know that you are not finished with this PR, just sharing something I noticed, Feel free to ignore if you intend to address this later 

❓ question: do we need the `sitemodel` class or `siteId` will be enough

zwarm: I'm on the fence. In some cases we use the entire model and in other cases just the site. I'm going to leave it for the time being and we can revisit if it causes problems. Thanks

zwarm:
@AjeshRPai - This PR has been replaced with #2768 . This branch will be deleted.

PR #2764, opened 2023-07-03
feat: export application password

wzieba:
We need to export application password header so we can pass it to React Native instance to enable application password authentication there.


ThomazFB (approved):
After a very long review, looks good to me! 👍

Edit: It seems the function could be actually a extended computed property of SiteModel, but since it will be used in RN, I'm afraid of compatibility issues.

PR #2765, opened 2023-07-04
[Jetpack Social] Parse features in SiteModel plan object

RenanLukas:
Parent issue: https://github.com/wordpress-mobile/WordPress-Android/issues/18730
Parent PR: https://github.com/wordpress-mobile/WordPress-Android/pull/18735

This PR parses the `features` object in `SiteModel` and adds `hasSocialShares1000Active` property so we can use it to determine if we should show the social sharing limit to a site.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
@@ -252,6 +252,9 @@ public class SiteModel extends Payload<BaseNetworkError> implements Identifiable
     @Column
     private Boolean mCanBlaze;
 
+    // Jetpack Social
+    @Column private boolean mHasSocialShares1000Active;
+
     @Override
     public int getId() {
         return mId;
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
```diff
@@ -252,6 +252,9 @@ public class SiteModel extends Payload<BaseNetworkError> implements Identifiable
     @Column
     private Boolean mCanBlaze;
 
+    // Jetpack Social
+    @Column private boolean mHasSocialShares1000Active;
```
thomashorta: 💡 I think exposing this is too specific to the actual client needs (WP-Android). I would suggest exposing a `mPlanActiveFeatures` with a list of the active features for the plan, returned by the API, or empty if none was returned / no existing plan for some reason.

I believe it would need to be stored in the local DB as a comma-separated String but it would be nice if it could be exposed as a `List<String>` here in the `SiteModel`.

Wdyt?

thomashorta (commented):
Nice work! Thanks.

I added a suggestion that might change this PR significantly so I will not approve it now and wait for an answer/changes if that's OK.

RenanLukas:
Thanks for the suggestion, @thomashorta . Makes sense to have generic fields in `SiteModel` instead of something specific to one client.

It's updated 👍

fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java
```diff
@@ -251,6 +252,9 @@ public class SiteModel extends Payload<BaseNetworkError> implements Identifiable
     private String mApplicationPasswordsAuthorizeUrl;
     @Column
     private Boolean mCanBlaze;
+    // Comma-separated list of active features in the site's plan
+    @Column
+    private String mActiveFeatures;
```
thomashorta: 🔍  Nitpick: since this is related to the features of the site plan and not features of the actual site, does it make sense to have a specific name here, like `mPlanActiveFeatures`?

thomashorta (approved):
This is looking great! Thanks for the changes! 🙇 

I added a small nitpicky comment but feel free to ignore it if it doesn't make sense to you.

PR #2767, opened 2023-07-04
Top performers new endpoint

### LINKED ISSUE #9311: [Performance] Replace top performers requests with single analytics report request
## Suggested changes

On the web, the Products analytics report makes a single request to `wc-analytics/reports/products`. We could replace the current requests to populate Top Performers with this request. This should return all the data we need for `TopEarnerStats` in a single response, with less processing required.

Parameters for the query:

* `orderby=items_sold`
* `order=desc`
* `per_page=5`
* `page=1`
* `before={start date of time period}`
* `after={end date of time period}`
* `extended_info=true`

## Current behavior

Currently we populate the Top Performers list on My Store using multiple requests:

* A request to `/wc-analytics/leaderboards` or `/wc-analytics/leaderboards/products` to fetch the store's leaderboards.
* A request to `products` to get product details for any products not in storage.

The leaderboards endpoint is what the web uses for the "Top products - Items sold" section in the analytics overview page, but the endpoint isn't optimized for mobile and doesn't contain all the data we display in the app. It requires additional processing in the client to extract the data the leaderboard response (including extracting the product ID to fetch the product details) and convert the response into `TopEarnerStats`. See how [leaderboards are converted into top earner stats](https://github.com/woocommerce/woocommerce-ios/blob/5a4235c29a6aeb5ef3937ae18d1f876c061e9e57/Yosemite/Yosemite/Stores/StatsStoreV4.swift#L524) and how a [leaderboard row is converted into a top earner stats item](https://github.com/woocommerce/woocommerce-ios/blob/5a4235c29a6aeb5ef3937ae18d1f876c061e9e57/Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift#L65) for more details on that work.

Note: The `wc-analytics/reports/products` endpoint has been supported in Woo Core since version 6.5, the same version when support for `wc-analytics/reports/revenue/stats` was added (used for order stats on My Store).

atorresveiga:
Closes: https://github.com/woocommerce/woocommerce-android/issues/9311
### Why
On the web, the Products analytics report makes a single request to wc-analytics/reports/products. We could replace the current requests to populate Top Performers with this request. This should return all the data we need for TopEarnerStats in a single response, requiring less processing.

### Description
This PR brings support for the `wc-analytics/reports/products` endpoint so that we can process top-performer products in a single request.

### Testing
Check that all Unit tests are green.
It is better to test this PR with the [Woo PR](https://github.com/woocommerce/woocommerce-android/pull/9366).


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCLeaderboardsStore.kt
```diff
@@ -108,6 +113,56 @@ class WCLeaderboardsStore @Inject constructor(
...
+            startDate = startDate,
+            endDate = endDate,
+            quantity = quantity
+        )
+
+        return when {
+            response.isError -> WooResult(response.error)
+            response.result != null -> {
+                val period = DateUtils.getDatePeriod(startDate, endDate)
+                val currency = wooCommerceStore.getSiteSettings(site)?.currencyCode ?: ""
```
atorresveiga, line 148: The report request doesn't include the currency code, a field we save into the database. I decided to fetch this field from the site settings, but if the settings are not cached yet, use blank, considering that we are not using the currency to display top-performer products.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCLeaderboardsStore.kt
```diff
@@ -54,8 +58,9 @@ class WCLeaderboardsStore @Inject constructor(
     ): WooResult<List<TopPerformerProductEntity>> {
         val startDate = granularity.startDateTime(site)
         val endDate = granularity.endDateTime(site)
-        val interval = OrderStatsRestClient.OrderStatsApiUnit.fromStatsGranularity(granularity).toString()
-        return fetchTopPerformerProducts(
+        val interval = OrderStatsRestClient.OrderStatsApiUnit.fromStatsGranularity(granularity)
+            .toString()
+        return fetchTopPerformerProductsLegacy(
```
atorresveiga, line 63: We need this function to fetch analytics for sites using an older version

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/reports/ReportsProductApiResponse.kt
```diff
@@ -0,0 +1,35 @@
...
+    companion object {
+        private val SRC_REGEX = Regex("src=\"(.*?)\"")
+    }
+
+    fun getImageUrl() = imageHTML
+        ?.let {
+            SRC_REGEX.find(it)?.value
+                ?.replace("src=", "")
+                ?.replace("\"", "")
+        }
```
ThomazFB: very minor NP: This sound like a perfect scenario to adopt a computed property instead of a getter function, like:

```suggestion
        val imageUrl: String?
        get() = imageHTML
            ?.let { SRC_REGEX.find(it)?.value }
            ?.replace("src=", "")
            ?.replace("\"", "")
```

ThomazFB (approved):
Nice work updating the unit tests! Code looks good and tests are all green in my local environment and CI. Only left a very minor suggestion that you can ignore if you want to 😄

PR #2768, opened 2023-07-09
[JPAndroid] Add endpoints and DB layer for Blaze Campaigns

zwarm:
This PR adds support for:

- Calling the blaze search campaign endpoint
- DB layer for storing blaze campaign data and associated paging information
- Clicks and impressions - Defaulting null to "0"
- budgetCents is stored as a Long in the DB, it's up to the UI layer to format as USD.
- URL is set as nullable, even though it is required for JP blaze. (Tumblr doesn't require an image, so setting it as nullable will help prevent possible crashes in the future
- A new constructor was added to BaseWPComRestClient.java to accept a AcceptHeaderStrategy. This strategy encapsulates Accept Headers that need to be applied to the request. Other strategies can be easily added and injected into child rest client classes.

This PR will use https://github.com/wordpress-mobile/WordPress-Android/pull/18686 as a testing companion. However, if needed, a separate JP Android PR will be created with a hash only.

This PR replaces #2763 - there were too many database changes, that it was less risky to start fresh.


AjeshRPai (approved):
Hey @zwarm 

I have updated the PR with trunk and bumped the database version. I tested the database migration by installing the app from trunk and using this branch. Everything looks good to me. 

Approving and Merging 👍🏼

PR #2771, opened 2023-07-10
Remove unused method from SiteModel

RenanLukas:
Removes unused method from `SiteModel` added in the context of JP Social feature.


thomashorta (approved):
LGTM :shipit:

PR #2774, opened 2023-07-13
Integration with jetpack-social endpoint

RenanLukas:
Integration with `wpcom/v2/sites/{siteURL}/jetpack-social` endpoint and implementation of local persistence.


fluxc/src/main/java/org/wordpress/android/fluxc/model/jetpacksocial/JetpackSocialMapper.kt
```diff
@@ -0,0 +1,36 @@
...
+            JetpackSocialEntity(
+                siteLocalId = siteLocalId,
+                isShareLimitEnabled = isShareLimitEnabled ?: false,
+                toBePublicizedCount = toBePublicizedCount ?: -1,
+                shareLimit = shareLimit ?: -1,
+                publicizedCount = publicizedCount ?: -1,
+                sharedPostsCount = sharedPostsCount ?: -1,
+                sharesRemaining = sharesRemaining ?: -1,
+                isEnhancedPublishingEnabled = isEnhancedPublishingEnabled ?: false,
+                isSocialImageGeneratorEnabled = isSocialImageGeneratorEnabled ?: false,
```
thomashorta, line 19: ❓Do you know if there are any valid scenarios that can return `null` for those fields? If there are any, do these fallbacks represent that scenario properly?

I ask that because I feel it might be a bit odd requesting that data and receiving something like `shareLimit = -1` and we might probably need to account for that in the client code to avoid showing something like "-1 shares remaining" in the UI, wdyt?

thomashorta (approved):
Thanks for the PR, Renan. 
I only added a minor question, other than that everything is looking good to me! 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/model/jetpacksocial/JetpackSocialMapper.kt
RenanLukas: Hey @thomashorta , good question.
As discussed, we've decided to keep the fallbacks and map them to a meaningful domain object in the client, considering the context of the feature we're working on.

PR #2776, opened 2023-07-14
Increase Cursor Window Size to load larger posts

### LINKED ISSUE #18727: SQLiteBlobTooBigException: Row too big to fit into CursorWindow requiredPos=0, totalRows=1
Sentry Issue: [JETPACK-ANDROID-87M](https://a8c.sentry.io/issues/4268768055/?referrer=github_integration)

```
SQLiteBlobTooBigException: Row too big to fit into CursorWindow requiredPos=0, totalRows=1
    at android.database.sqlite.SQLiteConnection.nativeExecuteForCursorWindow(SQLiteConnection.java)
    at android.database.sqlite.SQLiteConnection.executeForCursorWindow(SQLiteConnection.java:942)
    at android.database.sqlite.SQLiteSession.executeForCursorWindow(SQLiteSession.java:838)
    at android.database.sqlite.SQLiteQuery.fillWindow(SQLiteQuery.java:62)
    at android.database.sqlite.SQLiteCursor.fillWindow(SQLiteCursor.java:153)
...
```

By July 3rd, 2023, this issue has impacted 4 users. Seems it was recently introduced as it's first seen in version `22.5`.

ravishanker:
Increase the cursor window size to 5MB for devices running API 28 and above. This should reduce the number of SQLiteBlobTooBigExceptions.

Fixes [#18727](https://github.com/wordpress-mobile/WordPress-Android/issues/18727)

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
@@ -1988,13 +1988,13 @@ open class WellSqlConfig : DefaultWellConfig {
     }
 
     /**
-     * For debug builds we want a cursor window size of 5MB so we can test for any problems caused by
-     * a larger size. Once we're confident this works we'll return 5MB in release builds to hopefully
-     * reduce the number of SQLiteBlobTooBigExceptions. Note that this is only called on API 28 and
-     * above since earlier versions don't allow adjusting the cursor window size.
+     * Increase the cursor window size to 5MB for devices running API 28 and above. This should
+     * reduce the number of SQLiteBlobTooBigExceptions.
+     * NOTE: this is only called on API 28 and above since earlier versions don't allow adjusting
+     * the cursor window size.
      */
     @Suppress("MagicNumber")
-    override fun getCursorWindowSize() = if (BuildConfig.DEBUG) (1024L * 1024L * 5L) else 0L
+    override fun getCursorWindowSize() = (1024L * 1024L * 5L)
 
     /**
      * Drop and create all tables


thomashorta (approved):
@ravishanker thanks for the PR.

I believe it makes sense to return this as 5Mb as I don't think we stumbled in issues regarding that for debug builds so far, but maybe it would be nice to ask for a quick look from someone in the WooCommerce team as well since they also use this library.

I am aware they actually [override this config](https://github.com/woocommerce/woocommerce-android/blob/46c3009d456e4f997fe7946e9ea84daee3edd2ac/WooCommerce/src/main/kotlin/com/woocommerce/android/WooWellSqlConfig.kt#L73) by passing 5Mb in their prod builds, so I believe this would be fine for them, but maybe good to ask anyway and give them visibility of this change, wdyt?

Otherwise, it looks good to me!

ravishanker:
👋 @jd-alexander, @malinajirka - Could I get your advice on this change since you've done this before 🙇 
- How's the result after increasing size to 5MB?  
- Is that enough or should it be increased to 20MB like it is already done on Develop branch?

malinajirka:
I think we might want to consider keeping this decision app-based. I believe the Woo app doesn't need 20MB cursor window so increasing it in FluxC feels off. I see the [WP and Jetpack apps are already using 20MB](https://github.com/wordpress-mobile/WordPress-Android/blob/bef5f609d43421864a1ce6934842feb56e914014/WordPress/src/debug/java/org/wordpress/android/WPWellSqlConfig.kt#L49) so I believe this change in FluxC won't make any difference, wdyt?

> How's the result after increasing size to 5MB?

We have seen a decrease in the number of crashes. However, it's important to keep in mind that this is a workaround not an actual fix of the core issue - we ideally wouldn't store post body in SQL, we'd store it for example in a file and store only the path in the SQL.

thomashorta:
> I think we might want to consider keeping this decision app-based. I believe the Woo app doesn't need 20MB cursor window so increasing it in FluxC feels off. I see the [WP and Jetpack apps are already using 20MB](https://github.com/wordpress-mobile/WordPress-Android/blob/bef5f609d43421864a1ce6934842feb56e914014/WordPress/src/debug/java/org/wordpress/android/WPWellSqlConfig.kt#L49) so I believe this change in FluxC won't make any difference, wdyt?

Just a reminder that the `20MB` config on WPAndroid only affects debug builds as mentioned [here](https://github.com/wordpress-mobile/WordPress-Android/issues/18727#issuecomment-1633309414), so it doesn't override the FluxC config for prod builds. But I agree that this should be a per-app config, with this 5 MB being just a base size for all clients (since the actual common date is handled by FluxC anyway, so it should be similar for all clients). 

If WPAndroid needs a bigger window size then we would need to create a new release Well SQL config class overriding that value, but I think it's ok to start with those 5MB from the base config and see how it goes.

PR #2745, opened 2023-06-06
Primitive boolean deserializer

atorresveiga:
Closes: https://github.com/woocommerce/woocommerce-android/issues/9153
Closes: https://github.com/woocommerce/woocommerce-android/issues/9154

### Description
Some extensions (YITH WooCommerce Gift Cards, for example) change the type of the purchasable field to be an integer resulting in a [JsonSyntaxException](https://a8c.sentry.io/issues/3923384523/events/fd2a9bacec4b41e3ba2501400d232210/?project=1459556&referrer=next-event) when trying to deserialize a boolean, but an integer value was received instead. This PR tries to solve this issue by introducing a more resilience deserializer that will try to recover from the error and deserialize the purchasable field from primitives values (Boolean, Int, String)

### Testing Instructions
Unit tests should cover all the different scenarios

PR #2778, opened 2023-07-19
Fix sold individually as number

atorresveiga:
Part of: https://github.com/woocommerce/woocommerce-android/issues/9154

### Description
Some extensions change the type of the `sold individually` field to be an integer resulting in a [JsonSyntaxException](https://a8c.sentry.io/issues/3923384523/events/fd2a9bacec4b41e3ba2501400d232210/?project=1459556&referrer=next-event) when trying to deserialize a boolean, but an integer value was received instead. This PR tries to solve this issue by using the more resilience deserializer `PrimitiveBooleanJsonDeserializer` (introduced [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2745)) that will try to recover from the error and deserialize the `sold individually` field from primitives values (Boolean, Int, String)

### Testing Instructions
Unit tests should cover all the different scenarios

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/product/ProductReliabilityTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/product/ProductReliabilityTest.kt
@@ -1,0 +1,25 @@
+package org.wordpress.android.fluxc.wc.product
+
+import com.google.gson.Gson
+import org.assertj.core.api.Assertions.assertThat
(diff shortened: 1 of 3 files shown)


ThomazFB (approved):
Good resilience improvement! I noticed that another field of the ProductApiResponse also uses this same JsonAdapter strategy. Looks like any boolean can be victim of this problem 😅 

Anyway, looks good!

PR #2779, opened 2023-07-20
Feature/jetpack ai update endpoint

hafizrahman:
This implements new endpoint as outlined in pe5sF9-1Il-p2

Part of: https://github.com/woocommerce/woocommerce-android/issues/9425

### Description
This PR updates the text completion to use the WPCom `text-completion` endpoint. As part of this, some changes are added (the endpoint now requires `feature` value, and no longer require site ID). The new endpoint also does not have the `skipCache` parameter, but it seems like it always fetches fresh and so should not be a problem.

The use of this endpoint requires a valid JWT token, which has to be fetched first. The token is valid for a few minutes, so it's saved in FluxC's shared preferences to be re-used. If during re-use it doesn't work, then FluxC will attempt to fetch once and use the fresh token.

### Design decisions
1. FluxC will handle the token fetching, persisting, and checking.
2. FluxC will only attempt to fetch token once, or pass error. This applies both for the case whether a token already exists or not.
3. FluxC does not format the resulting completion from the endpoint. Sometimes the resulting string can have elements like the `\n` endline character. The client is responsible for handling these if needed.
4. This PR does not delete functions related to the previously used endpoints. This is in case the store functions are also used in JP/WP apps. Some research will need to be done if they're actually used, and if not, they can be deleted safely in a separate PR.

### Testing
1. Run Example app, login to a WordPress.com account, then go to "Jetpack AI".
2. Select a WordPress.com site, then click "Generate Haiku". Make sure it works and returns value as shown in the console area.
3. Try again quickly and make sure it generates something different (to show that it's not cached).

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpackai/JetpackAIRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpackai/JetpackAIRestClient.kt
@@ -27,9 +27,9 @@ class JetpackAIRestClient @Inject constructor(
     userAgent: UserAgent
 ) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
     suspend fun fetchJetpackAIJWTToken(
-        siteId: Long
+        site: SiteModel
     ) : JetpackAIJWTTokenResponse {
-        val url = WPCOMV2.sites.site(siteId).jetpack_openai_query.jwt.url
+        val url = WPCOMV2.sites.site(site.siteId).jetpack_openai_query.jwt.url
         val response = wpComGsonRequestBuilder.syncPostRequest(
             restClient = this,
             url = url,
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpackai/JetpackAIRestClient.kt
```diff
@@ -25,6 +26,57 @@ class JetpackAIRestClient @Inject constructor(
...
+        }
+    }
+
+    suspend fun fetchJetpackAITextCompletion(
+        token: String,
+        prompt: String,
+        feature: String
+    ): JetpackAICompletionsResponse {
+        val url = WPCOMV2.text_completion.url
+        val body = mutableMapOf<String, Any>()
```
JorgeMucientes: Super minor np that can be ignored. In this case, we know all the parameters are `Strings` so we could strongly type the map by setting a `mutableMapOf<String, String>()`

fluxc/src/main/java/org/wordpress/android/fluxc/store/jetpackai/JetpackAIStore.kt
```diff
@@ -58,4 +68,69 @@ class JetpackAIStore @Inject constructor(
...
+                    jwtTokenResponse.token,
+                    prompt,
+                    feature
+                )
+            }
+        }
+    }
+
+     private suspend fun fetchJetpackAIJWTToken(site: SiteModel)
+     : JetpackAIJWTTokenResponse = coroutineEngine.withDefaultContext(
```
JorgeMucientes, line 129: Np formatting: 

```suggestion
    private suspend fun fetchJetpackAIJWTToken(site: SiteModel)
        : JetpackAIJWTTokenResponse = coroutineEngine.withDefaultContext(
```

JorgeMucientes (approved):

JorgeMucientes:
Nice work @hafizrahman! Works as expected. Just left some minor nitpicks, feel free to merge.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpackai/JetpackAIRestClient.kt
JorgeMucientes:
```suggestion
        val body = mutableMapOf<String, String>()
```

PR #2780, opened 2023-07-21
Add: "_fields" parameter for text completion API call.

hafizrahman:
The `text-completion` API call returns a bit more data than needed. With this change, we request only to get the "completion" field, to reduce the amount of data being transferred.

### To test:
1. Start example app, login to a WPCom account with a WPCom Woo site.
2. Go to Jetpack AI -> Select a site -> tap "Generate Haiku". Ensure that it still works and return value properly.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpackai/JetpackAIRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpackai/JetpackAIRestClient.kt
@@ -26,6 +26,10 @@ class JetpackAIRestClient @Inject constructor(
     accessToken: AccessToken,
     userAgent: UserAgent
 ) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
+    companion object {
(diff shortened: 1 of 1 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/jetpackai/JetpackAIRestClient.kt
```diff
@@ -26,6 +26,10 @@ class JetpackAIRestClient @Inject constructor(
     accessToken: AccessToken,
     userAgent: UserAgent
 ) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent) {
+    companion object {
+        private const val FIELDS_TO_REQUEST = "completion"
```
hafizrahman, line 30: using plural "FIELDS" here even though the current value is only 1 field, because:
1. the corresponding parameter name is plural "_fields",
2. technically it's possible to change this in the future to include multiple fields, e.g: "completion,anotherfield"

JorgeMucientes (approved):

JorgeMucientes:
Works as expected :shipit:

PR #2785, opened 2023-07-24
Analytics Performance: Add access to revenue stats from range ID queries

ThomazFB:
⚠️ This PR has breaking changes that are connected with https://github.com/woocommerce/woocommerce-android/pull/9453, do not merge it until everything is approved.

Summary
==========
Introduces an external way to define a Revenue Stats range through an ID, making sure that we can insert/update/select any Revenue stats data through a range ID when available.

How to Test
==========
1. I recommend testing the changes introduced here through https://github.com/woocommerce/woocommerce-android/pull/9453. Just make sure all unit tests are OK.

Update release notes:

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
@@ -41,7 +41,7 @@ open class WellSqlConfig : DefaultWellConfig {
     annotation class AddOn
 
     override fun getDbVersion(): Int {
-        return 190
+        return 191
     }
 
     override fun getDbName(): String {
@@ -1958,6 +1958,9 @@ open class WellSqlConfig : DefaultWellConfig {
                 }
                 189 -> migrate(version) {
                     db.execSQL("ALTER TABLE SiteModel ADD PLAN_ACTIVE_FEATURES TEXT")
+                }
+                190 -> migrate(version) {
+                    db.execSQL("ALTER TABLE WCRevenueStatsModel ADD RANGE_ID INTEGER")
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


atorresveiga (approved):
Great work, Thomaz! The code looks good and is working as expected. Tested alongside the Woo PR.

LGTM! :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
```diff
@@ -1963,6 +1963,9 @@ open class WellSqlConfig : DefaultWellConfig {
                     db.execSQL("ALTER TABLE PostModel ADD AUTO_SHARE_MESSAGE TEXT")
                     db.execSQL("ALTER TABLE PostModel ADD AUTO_SHARE_ID INTEGER")
                 }
+                191 -> migrate(version) {
+                    db.execSQL("ALTER TABLE WCRevenueStatsModel ADD RANGE_ID INTEGER")
```
atorresveiga, line 1967: Nice!

atorresveiga (changes requested):
I almost miss something 🤦 

You must replace 191 ->` migrate(version)` to `migrateAddOn(ADDON_WOOCOMMERCE, version)` otherwise the WP app will crash because it doesn't have the `WCRevenueStatsModel` table

ThomazFB:
Wow, sorry that I forgot about this @atorresveiga, thank you so much for finding it in time to avoid a huge problem 😬 It's fixed now!

atorresveiga (approved):
Awesome work Thomaz! Using the new revenueRangeId is a great addition that will make our life easier when working with the stats.

LGTM! :shipit:

PR #2786, opened 2023-07-24
Check for orders enhancement

### LINKED ISSUE #9462: [Performance] Check if the site has orders using the local database
In the My Store screen, we constantly check whether the store has orders for each stats tab. We request a random order using a network request to check if the store has orders. This task is about optimizing this behavior by including a local database check when checking for the existence of orders for a particular store.

atorresveiga:
Closes: https://github.com/woocommerce/woocommerce-android/issues/9462

### Why
As part of the Reliability project, we are reducing the number of requests we make to the API.

### Description
This PR adds the new `hasOrders` function to the orders store. This new function first checks the existence of orders on the local database before sending the request to the API. If there is at least one order in the local database, we can assume that the store has orders, and we don't need to request this data from the API.

### Testing
It is better to test this PR using the [Woo PR](https://github.com/woocommerce/woocommerce-android/pull/9464)

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/order/WCOrderStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/order/WCOrderStoreTest.kt
@@ -15,9 +15,11 @@ import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.mockito.kotlin.any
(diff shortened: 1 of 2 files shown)


ThomazFB (approved):
Tested alongside https://github.com/woocommerce/woocommerce-android/pull/9464#pullrequestreview-1551055553. Looks good! :shipit:

PR #2777, opened 2023-07-18
Parse auto-share message and ID in PostModel metadata

RenanLukas:
Parses auto-share message and ID in `PostModel` metadata.

WordPress-Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/18790

PR #2788, opened 2023-07-25
Parse skip publicize connections metadata in PostModel

RenanLukas:
Parses publicize connections metadata in PostModel.

WordPress-Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/18851


fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -544,6 +547,15 @@ public String getAutoShareMessage() {
         return StringUtils.notNullStr(mAutoShareMessage);
     }
 
+    public void setPublicizeSkipConnections(final String publicizeSkipConnections) {
+        mPublicizeSkipConnections = publicizeSkipConnections;
+    }
+
+    @Override
+    @NonNull public String getPublicizeSkipConnections() {
+        return StringUtils.notNullStr(mPublicizeSkipConnections);
```
develric: [Yeah](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2777#discussion_r1270841011) 😄 👍  !

develric (commented):
I see the companion WPAndroid PR is a WIP, but meanwhile I checked the code and looks good 👍 . I can go over a test round when the companion PR is up, just ping me @RenanLukas  🙇

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -592,7 +630,9 @@ && getAutoSaveRevisionId() == otherPost.getAutoSaveRevisionId()
                 && StringUtils.equals(getAutoSaveModified(), otherPost.getAutoSaveModified())
                 && StringUtils.equals(getAutoSavePreviewUrl(), otherPost.getAutoSavePreviewUrl())
                 && StringUtils.equals(getAutoShareMessage(), otherPost.getAutoShareMessage())
-                && getAutoShareId() == otherPost.getAutoShareId();
+                && getPublicizeSkipConnectionsJson().equals(otherPost.getPublicizeSkipConnectionsJson())
+                && getAutoShareId() == otherPost.getAutoShareId()
+                && StringUtils.equals(getPublicizeSkipConnectionsJson(), otherPost.getPublicizeSkipConnectionsJson());
```
develric: There seems to be a duplicated check on getPublicizeSkipConnectionsJson here, wdyt? (and FWIW `StringUtils.equals(` seems more in line with the rest of the code here 🙇 )

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -544,6 +550,38 @@ public String getAutoShareMessage() {
...
+        try {
+            if (mPublicizeSkipConnectionsJson != null) {
+                final Type publicizeSkipConnectionsType = new TypeToken<List<PublicizeSkipConnection>>() {}.getType();
+                final List<PublicizeSkipConnection> list =
+                        new Gson().fromJson(mPublicizeSkipConnectionsJson, publicizeSkipConnectionsType);
+                if (list.size() > 0) {
+                    return list;
+                } else {
+                    return Collections.emptyList();
+                }
```
develric: I think fromJson in principle might return null 🤔 ...also if the list has a size or is empty maybe we can just return it, let me know wdyt 🙇

develric (commented):
Thanks for the changes @RenanLukas 🙇 . Did a pass and left a couple of comments, let me know 👍 ! I'm running out of time today, but I'll continue tomorrow (also with the companion WPAndroid PR 🙇 )

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
RenanLukas: Thanks!

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
RenanLukas: Good catch, thanks!

RenanLukas:
Thanks for the review, @develric . I've addressed your comments.

fluxc/src/main/java/org/wordpress/android/fluxc/model/PostModel.java
```diff
@@ -544,6 +550,38 @@ public String getAutoShareMessage() {
...
+                        new Gson().fromJson(mPublicizeSkipConnectionsJson, publicizeSkipConnectionsType);
+                if (list != null) {
+                    return list;
+                } else {
+                    return Collections.emptyList();
+                }
+            }
+            return Collections.emptyList();
+        } catch (final Exception exception) {
+            return Collections.emptyList();
```
develric, line 583: Sorry for not spotting this before 🙇 , I think we should AppLog.e here (it seems the common way we use in other similar places of the library afaiu). Also, maybe for a separate PR (let's create a GH issue in case), but t would be good to propagate this error somewhere (like to tracks maybe 🤔 ) to spot if it happens often. I see the consumer of this method is both the FluxC lib and the EditPostPublishSettingsViewModel in the WPAndroid app. We might wrap the returned value from this method in a Success/Failure sealed class and log-only in fluxc while add a track event in WPAndroid. Wdyt? 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
```diff
@@ -708,6 +717,26 @@ private Map<String, Object> postModelToParams(PostModel post) {
...
+                for (final PublicizeSkipConnection connection : publicizeSkipConnections) {
+                    if (connection.getKey() != null && !connection.getKey().isEmpty()) {
+                        publicizeSkipConnectionsMetadata.put("key", connection.getKey());
+                    }
+                    if (connection.getId() > 0) {
+                        publicizeSkipConnectionsMetadata.put("id", connection.getId());
+                    }
+                    if (connection.getValue() != null && !connection.getValue().isEmpty()) {
+                        publicizeSkipConnectionsMetadata.put("value", connection.getValue());
+                    }
```
develric: Since this part was already there in iOS since some time, we should probably double check there, but does it make sense to add a key potentially without the value (if that's null or empty) or a value without it's key? 🤔...wdyt?

develric (commented):
Thanks for the changes @RenanLukas. I did another pass and tested with the companion app. I left couple of comments here if you can look at it 🙇 , and I'm reporting an issue with updating the metadata I'm facing in the WPAndroid PR 🙇

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
RenanLukas: I agree it doesn't make sense to add a `key` without the `value` (or if something really bad happens, a `value` without a `key`). Good catch 👍

RenanLukas:
Thanks for the second round of review, @develric . I've addressed the comments again 🎉

develric (approved):
Thanks for the changes @RenanLukas 🙇 . This change [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2788/commits/5288d75c7bf547f922204ef2d294be8585059709) made updating active sharing connections from the app work now 👍 ....I'm approving this PR 🎉

PR #2789, opened 2023-07-26
Issue/wc 9372 local SKU product search

malinajirka:
This PR adds support for local SKU product search to the `getProducts` method. This will be useful for improving the Order Creation flow in the Woo app - https://github.com/woocommerce/woocommerce-android/issues/9372.

Props to @rossanafmenezes - we solved this issue together during a couple pair programming sessions.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -245,14 +247,28 @@ object ProductSqlUtils {
...
+                        .contains(WCProductModelTable.NAME, searchQuery)
+                        .or()
+                        .contains(WCProductModelTable.DESCRIPTION, searchQuery)
+                        .or()
+                        .contains(WCProductModelTable.SHORT_DESCRIPTION, searchQuery)
+                        .endGroup()
+                }
+                SkuSearchOptions.ExactSearch -> {
+                    queryBuilder.beginGroup()
+                        .equals(WCProductModelTable.SKU, searchQuery.lowercase())
```
samiuelson: Why do we need to lowercase the query here?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
```diff
@@ -245,14 +247,28 @@ object ProductSqlUtils {
...
+                        .endGroup()
+                }
+                SkuSearchOptions.PartialMatch -> {
+                    queryBuilder.beginGroup()
+                        .contains(WCProductModelTable.SKU, searchQuery)
```
samiuelson, line 269: Is the `contains()` method case-sensitive?

samiuelson (commented):
Thanks for your work. Looks good overall! 

I left some questions about parts that were unclear to me.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
malinajirka: Nope, `contains` in SQLite is case insensitive, AFAIK. (eg https://database.guide/how-to-make-sqlites-like-operator-case-sensitive/)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/ProductSqlUtils.kt
malinajirka: Great question - I wanted to return to this but I forgot. 

Using `lowerCase` here doesn't give us any benefit at all, since we are converting only the searchQuery but not the data in the database - I removed it in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2789/commits/ef80d315afa38f3a553f2a4384cd4c0db0cb4f6f.

Unfortunately, it seems WellSQL doesn't support COLLATE NOCASE, so I can't think of an elegant solution. For SkuSearchOptions.ExactSearch, we'll be doing a case sensitive search - which I guess might be expected. This method with ExactSearch is not being used at the moment in the Woo app.

malinajirka:
Thanks so much for the review @samiuelson, it's ready for another round.

samiuelson (approved):
Looks good! :shipit:

PR #2792, opened 2023-08-03
Fix NPE NonceRestClient crash

mkevins:
This fixes a NullPointerException because the `username` field is nullable in the model table for `SiteModel`, and it is accessed as a platform type from Kotlin code. The fix uses the `@Nullable` annotation, and follows up by making the dependent `Nonce` return type optional.

Note: other fields in this (as well as other) models are also left unannotated, even when they are nullable. These are not addressed here to limit the scope of this fix to this single crash (which is currently our [most impactful NPE](https://github.com/wordpress-mobile/WordPress-Android/issues/18623)).

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClientTest.kt
@@ -67,7 +67,7 @@ class NonceRestClientTest {
 
         val actual = subject.requestNonce(site)
 
-        TestCase.assertEquals(Nonce.Available(expectedNonce, site.username), actual)
+        TestCase.assertEquals(Nonce.Available(expectedNonce, site.username!!), actual)
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


mkevins:
**Update**

I originally began by marking the `SiteModel`'s username field as `@Nullable`, since it is nullable in the db (i.e. there is no `NOT NULL` constraint on the column). However, this surfaces many type mismatches between the not null and nullable usages of the username getter.

I also began exploring a different approach: to make this field `@NonNull`, and add a constraint to the database column (via a migration). As a preliminary step, I examined the current database contents (via App Inspection), and noticed that immediately following a successful login, the `USERNAME` field is actually `NULL`, which I did not expect. Imo, this is worth investigating further, since `username` should be available as a field by the time the sites are loaded. Note: this was observed even in the case when logging in with a username and password.

It seems, in some cases, we instantiate `SiteModel()`, and the username can be `null`, even when being annotated as `@NonNull`. This is because, afaik, this is only checked _after_ constructor invocation is finished (i.e. setting the field). But, if it is never set, it can remain null even when annotated otherwise. It may be useful to next explore the extent of this pattern usage, since ideally (and as we move more toward Kotlin), addressing this with a well-typed data class (or even sealed classes) would be a better way to go.

PR #2794, opened 2023-08-04
Improve sites loading time

kidinov:
Currently, we fetch nonjetpack sites sequentially. This PR makes it parallel

To test, use composite build with WOO app and open Site Picker. Notice that the requests for non-JP sites now run in parallel. Smoke test site selection and login flow

More context - peaMlT-97-p2

image


wzieba (approved):
Looks good!

PR #2796, opened 2023-08-07
[JP Social/New Post] Add the for=mobile query param on post creation

thomashorta:
This new param identifies that the post is being created from the mobile apps in order to make sure it properly updates the metadata for JP social sharing and only shares the post to social networks selected by the user (sent as metadata in the post creation request).

Testing can be done using the version from PR: https://github.com/wordpress-mobile/WordPress-Android/pull/18920

Ref: p1691410760076329/1691399612.659809-slack-C05362VSXFU


RenanLukas (approved):
Thanks for the PR, @thomashorta . I've tested these changes using https://github.com/wordpress-mobile/WordPress-Android/pull/18920 and it's working as expected :shipit:

PR #2805, opened 2023-08-14
Send new app installation date to WcTracker

JorgeMucientes:
Part of: https://github.com/woocommerce/woocommerce-android/issues/9584

Adds required FluxC changes to send the app installation date in ISO 8601 `yyyy-MM-dd'T'HH:mm:ss'Z'`

The new parameter is optional so the API changes are non-breaking. To test the changes, follow instructions from this PR: https://github.com/woocommerce/woocommerce-android/pull/9594

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/tracker/TrackerRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/tracker/TrackerRestClient.kt
@@ -4,23 +4,28 @@ import org.wordpress.android.fluxc.generated.endpoint.WOOCOMMERCE
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooNetwork
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
+import org.wordpress.android.fluxc.utils.putIfNotEmpty
 import org.wordpress.android.fluxc.utils.toWooPayload
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 internal class TrackerRestClient @Inject constructor(private val wooNetwork: WooNetwork) {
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/tracker/TrackerStore.kt
```diff
@@ -7,7 +7,11 @@ import javax.inject.Singleton
 
 @Singleton
 class TrackerStore @Inject internal constructor(private val restClient: TrackerRestClient) {
-    suspend fun sendTelemetry(appVersion: String, site: SiteModel): WooPayload<Unit> {
-        return restClient.sendTelemetry(appVersion, site)
+    suspend fun sendTelemetry(
+        appVersion: String,
+        site: SiteModel,
+        installationDateIso8601: String? = null
```
wzieba: I wouldn't introduce a default argument here - there's no case when we want to invoke `sendTelemetry` method without `installationDateIso8601` argument

```suggestion
        installationDateIso8601: String?
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/tracker/TrackerRestClient.kt
```diff
@@ -4,23 +4,28 @@ import org.wordpress.android.fluxc.generated.endpoint.WOOCOMMERCE
...
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 internal class TrackerRestClient @Inject constructor(private val wooNetwork: WooNetwork) {
-    suspend fun sendTelemetry(appVersion: String, site: SiteModel): WooPayload<Unit> {
+    suspend fun sendTelemetry(
+        appVersion: String,
+        site: SiteModel,
+        installationDate: String?
```
wzieba, line 17: ~Should `installationDate` be nullable? If I'm not mistaken, it cannot be.~ Sorry, just realised it can be 👍

wzieba (approved):

PR #2807, opened 2023-08-14
[Hazmat Shipping] Add hazmat category parameter support

### LINKED ISSUE #9595: [Hazmat Shipping] Adjust Shipping package data models to support the `hazmat` parameter
To properly create a Shipping Label declaring hazardous materials, we must adapt our internal package models to properly support the `hazmat` parameter with all the possible values expected.

Request body example:
``` json
        "origin": {
		...
	},
	"destination": {
		...
	},
	"packages": [ //Package and Customs data altogether
		{
                     ...
                     "hazmat": "CLASS_9_UNMARKED_LITHIUM"
		},		
                {
                     ... // No hazmat-checked package
		}
	],
        "email_receipt": true
```

ThomazFB:
:warning: Depends on https://github.com/woocommerce/woocommerce-android/pull/9601

Summary
==========
Partially fix issue [#9575](https://github.com/woocommerce/woocommerce-android/issues/9595) by introducing the Hazmat parameter support to the Repository and Store layers of the Shipping Label classes.

How to Test
==========
1. Ensure the code is good and CI and unit tests are green. No testable behavior was introduced here.

Update release notes:


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/shippinglabels/WCShippingLabelModel.kt
```diff
@@ -140,7 +140,8 @@ class WCShippingLabelModel(@PrimaryKey @Column private var id: Int = 0) : Identi
         val length: Float,
         val width: Float,
         val weight: Float,
-        @SerializedName("is_letter") val isLetter: Boolean = false
+        @SerializedName("is_letter") val isLetter: Boolean = false,
+        val hazmat: HazmatCategory? = null
```
wzieba, line 144: ❓ What would happen if JSON will have a value outside `HazmatCategory` set? We'll have `null` here or crash?

wzieba (approved):
Looks good, leaving one question

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/shippinglabels/WCShippingLabelModel.kt
ThomazFB: Considering Gson main directive to handle serialization of nullifying non-matching fields, I would expect this to be null, which is why I've set this as an optional nullable field, so the model creation doesn't crash when Gson tries to insert a null value there. Usually, the Gson library [expects you to define a TypeAdapter if you want to force a default value behavior](https://github.com/google/gson/issues/608), if it doesn't exist, it will resort to null. WDYT? 🤔

PR #2810, opened 2023-08-15
Add subscription metadata to supported keys

atorresveiga:
### Why
As part of the Reliability project, we introduced a new metadata strip function to save only the allowed product's metadata keys in the database and optimize the app's memory usage. Unfortunately, this optimization broke the simple subscription read-only functionality by not saving the subscription metadata needed to display the subscription information.

### Description
This PR adds the subscription metadata keys to the list of the supported private keys. This change prevents the strip function from removing the subscription metadata keys and restores the subscription's read-only functionality.

### Testing
It is better to test this change with the Woo PR  https://github.com/woocommerce/woocommerce-android/pull/9600

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductMetaData.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductMetaData.kt
@@ -3,6 +3,7 @@ package org.wordpress.android.fluxc.model
 import com.google.gson.Gson
 import com.google.gson.JsonArray
 import com.google.gson.JsonObject
+import org.wordpress.android.fluxc.model.WCProductModel.SubscriptionMetadataKeys
 import org.wordpress.android.fluxc.model.WCProductModel.AddOnsMetadataKeys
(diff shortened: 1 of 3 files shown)


ThomazFB (approved):
Fix is good, great work catching it, @atorresveiga! I just wonder if it would be valuable for us to create some unit test scenarios that can enforce that expected metadata are not getting stripped out again, so we can guarantee that this regression can't return and give us confidence of doing those changes again in the future. WDYT?

I understand that the supported private keys addition solves that, but I'm unsure of future regressions around any other possible metadata that we can end up stripping by accident.

atorresveiga:
> I just wonder if it would be valuable for us to create some unit test scenarios that can enforce that expected metadata are not getting stripped out again, so we can guarantee that this regression can't return and give us confidence of doing those changes again in the future. WDYT?

Makes sense, @ThomazFB. I added 3 unit tests to ensure supported keys contain all necessary keys and prevent any regression here https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2810/commits/eb020a44b2d36a027ba3d300456a0f9ad3024c8a

PR #2815, opened 2023-08-21
[Woo] Add a store to allow interaction with WC Admin Options API

hichamboushaba:
This PR adds a Store to allow interacting with the Woo endpoint `wc-admin/options`, this is needed for the task https://github.com/woocommerce/woocommerce-android/pull/9629.

### Testing instructions
1. Open the example app.
2. Make sure you are signed in.
3. Click on Woo, then "WooCommerce Admin"
4. Select a site.
5. Click on Update options.
6. Enter some sample data, for example: `test1:1,test2:2`
7. Confirm the call succeeds
8. Click on Fetch options
9. Enter some keys, for example: `test1,test2` (to match what we updated in 6).
10. Confirm the correct values are fetched (the output should be `Admin Options: test1:1, test2:2`

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/ui/wooadmin/WooAdminFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/ui/wooadmin/WooAdminFragment.kt
@@ -1,0 +1,78 @@
+package org.wordpress.android.fluxc.example.ui.wooadmin
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.lifecycle.lifecycleScope
+import com.google.gson.Gson
+import kotlinx.android.synthetic.main.fragment_woo_admin.*
+import kotlinx.coroutines.launch
+import org.wordpress.android.fluxc.example.R
+import org.wordpress.android.fluxc.example.prependToLog
+import org.wordpress.android.fluxc.example.ui.StoreSelectingFragment
+import org.wordpress.android.fluxc.example.utils.showSingleLineDialog
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.admin.WooAdminStore
+import javax.inject.Inject
+
+class WooAdminFragment : StoreSelectingFragment() {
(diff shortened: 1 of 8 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/admin/WooAdminStore.kt
```diff
@@ -0,0 +1,22 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.admin
+
+import org.wordpress.android.fluxc.model.SiteModel
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooResult
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class WooAdminStore @Inject constructor(
```
hichamboushaba, line 9: Instead of making this work about the store profiler alone, I opted to make this class generic for everything related to the `wc-admin` endpoints, this would allow using it for other things if we need too, and wouldn't give the false impression that this endpoint is only for store profiler questions.

example/src/main/java/org/wordpress/android/fluxc/example/ui/wooadmin/WooAdminFragment.kt
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+
+        update_options.setOnClickListener {
+            lifecycleScope.launch {
+                selectedSite?.let { site ->
+                    val keyValues = showSingleLineDialog(
+                        activity = requireActivity(),
+                        message = "Please enter the options to update in the format key:value separated by a comma",
+                        isNumeric = false
```
hichamboushaba, line 62: A quick and dirty way of supplying key-values, I didn't want to spend much time on this since it's just for testing purposes.

JorgeMucientes (approved):

JorgeMucientes:
Nicely done @hichamboushaba. Code looks good and works as expected. 

:shipit:

PR #2816, opened 2023-08-22
Include bundle rules

atorresveiga:
Part of https://github.com/woocommerce/woocommerce-android/issues/9541
### Why

### Description
This small PR includes the bundle fields needed for configuring a bundle product.

### Testing
It is better to test this PR alongside the Woo PR

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCBundledProduct.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCBundledProduct.kt
@@ -11,6 +11,6 @@ class WCBundledProduct(
     @SerializedName("stock_status") val stockStatus: String,
     @SerializedName("quantity_min") val quantityMin: Long,
     @SerializedName("quantity_max") val quantityMax: Long,
-    @SerializedName("quantity_default") val quantityDefault: Long = 0,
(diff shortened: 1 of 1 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCBundledProduct.kt
```diff
@@ -2,10 +2,15 @@ package org.wordpress.android.fluxc.model
...
     @SerializedName("bundled_item_id") val id: Long,
     @SerializedName("product_id") val bundledProductId: Long,
     @SerializedName("menu_order") val menuOrder: Int,
     @SerializedName("title") val title: String,
-    @SerializedName("stock_status") val stockStatus: String
+    @SerializedName("stock_status") val stockStatus: String,
+    @SerializedName("quantity_min") val quantityMin: Long,
+    @SerializedName("quantity_max") val quantityMax: Long,
+    @SerializedName("quantity_default") val quantityDefault: Long = 0,
+    @SerializedName("optional") val isOptional: Boolean = false,
```
wzieba: As this is a DTO, I'd advise against introducing default values here. I think those models should map exactly the response from the Core and not add additional logic. It's not a blocking comment, though.

wzieba (approved):

PR #2820, opened 2023-08-24
Woo: Rest client and store functionalities for saving site title.

hafizrahman:
This adds API call to save site title, as well as onboarding store functionality to do the same.

Some points:

1. the API call is put in `WooSystemRestClient` to follow an existing function that also hits the same endpoint, `fetchSiteSettings()`
2. The store function is in `OnboardingStore` because currently saving site is used in the context of site onboarding.

There is nothing to test here but instead it should be tested with the accompanying WooCommerce android PR (coming soon)

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
@@ -16,6 +16,7 @@ class WooSystemRestClient @Inject constructor(private val wooNetwork: WooNetwork
     companion object {
         private const val NOT_FOUND = 404
         private const val FORBIDDEN = 403
+        private const val SAVE_SITE_TITLE_RESPONSE_FIELD = "title"
     }
 
     suspend fun fetchInstalledPlugins(site: SiteModel): WooPayload<WCSystemPluginResponse> {
@@ -87,6 +88,26 @@ class WooSystemRestClient @Inject constructor(private val wooNetwork: WooNetwork
 
         return response.toWooPayload()
     }
+
+    suspend fun saveSiteTitle(site: SiteModel, siteTitle: String): WooPayload<SaveSiteTitleResponse> {
+        val url = WPAPI.settings.urlV2
+
+        val response = wooNetwork.executePostGsonRequest(
+            site = site,
+            path = url,
+            clazz = SaveSiteTitleResponse::class.java,
+            body = mapOf(
+                "title" to siteTitle,
+                "_fields" to SAVE_SITE_TITLE_RESPONSE_FIELD
+            )
+        )
+
+        return response.toWooPayload()
+    }
+
+    data class SaveSiteTitleResponse(
+        @SerializedName("title") val title: String? = null
+    )
 
     data class SSRResponse(
         val environment: JsonElement? = null,
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OnboardingStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OnboardingStore.kt
@@ -35,6 +35,7 @@ class OnboardingStore @Inject constructor(
                         .firstOrNull { it.id == ONBOARDING_TASKS_KEY }
                         ?.tasks ?: emptyList()
                 )
+
                 else -> WooResult(WooError(GENERIC_ERROR, UNKNOWN))
             }
         }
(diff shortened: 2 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OnboardingStore.kt
```diff
@@ -34,6 +38,25 @@ class OnboardingStore @Inject constructor(
...
         }
+
+    suspend fun saveSiteTitle(site: SiteModel, siteTitle: String): WooResult<String> =
+        coroutineEngine.withDefaultContext(API, this, "saveSiteTitleOnboarding") {
+            val response = systemRestClient.saveSiteTitle(site, siteTitle)
+
+            when {
+                response.isError -> WooResult(response.error)
+                response.result?.title != null -> {
+                    val fetchResult = wooCommerceStore.fetchWooCommerceSite(site)
```
hafizrahman, line 50: The intent here is to update the stored site information after site title is changed. But is doing it this way a bit overkill, and there's another way to do it in a more simple way?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OnboardingStore.kt
```diff
@@ -31,6 +35,31 @@ class OnboardingStore @Inject constructor(
...
+            }
+        }
+
+    suspend fun saveSiteTitle(site: SiteModel, siteTitle: String): WooResult<String> =
+        coroutineEngine.withDefaultContext(API, this, "saveSiteTitleOnboarding") {
+            val response = systemRestClient.saveSiteTitle(site, siteTitle)
+
+            when {
+                response.isError -> WooResult(response.error)
+                response.result?.title != null -> {
```
hafizrahman, line 49: Not checking for empty string because in WordPress, setting empty string as Site Title is acceptable.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
```diff
@@ -88,6 +89,26 @@ class WooSystemRestClient @Inject constructor(private val wooNetwork: WooNetwork
...
+    suspend fun saveSiteTitle(site: SiteModel, siteTitle: String): WooPayload<SaveSiteTitleResponse> {
+        val url = WPAPI.settings.urlV2
+
+        val response = wooNetwork.executePostGsonRequest(
+            site = site,
+            path = url,
+            clazz = SaveSiteTitleResponse::class.java,
+            body = mapOf(
+                "title" to siteTitle,
+                "_fields" to SAVE_SITE_TITLE_RESPONSE_FIELD
```
JorgeMucientes, line 101: Out of curiosity, what is this field for? It's kind of unclear to me from reading the constant name: `SAVE_SITE_TITLE_RESPONSE_FIELD`

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OnboardingStore.kt
JorgeMucientes: hmm, unsure I follow what do you mean?

JorgeMucientes (approved):

JorgeMucientes:
Nicely done @hafizrahman! :shipit:

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
hafizrahman: By default, the response will contain a lot of unneeded things (all site settings properties like site title, site tagline, etc etc).

Using `_fields` tells WP to only return the `title` property in the response since that's the only thing we care about here. Saves response size.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/OnboardingStore.kt
hafizrahman: So if site title is successfully changed, then the change doesn't automatically gets reflected in the site's entry in the app's DB. 

Here I'm using `wooCommerceStore.fetchWooCommerceSite(site)` to update the entry. But I'm also wondering if it's too much and if you have suggestions how to do it better (just update the site title directly in the DB, for example?)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/system/WooSystemRestClient.kt
hafizrahman: `_fields` is a global parameter that can be used for WP API requests to make the response smaller (https://developer.wordpress.org/rest-api/using-the-rest-api/global-parameters/#_fields).

hafizrahman:
I'm merging for now and happy to add improvements to a separate PR!

PR #2848, opened 2023-09-06
HACK Week: Increase max attribute terms parameter

### LINKED ISSUE #9583: [Variable Product] - Existing options for an attribute are not displayed in add options if more than 10 options available
**Describe the bug**
Attributes are not displayed fully in add attribute option screen and only show up to 10 options, i.e. if there are more than 10 options, only the last 10 are displayed. 

This was first reported in 6605997-zen and I was able to reproduce the same issue with the steps below.

**To Reproduce**
1. Log in to a store that has more than 10 options for an attribute type in a variable product (e.g. 25 different colors)
2. Create New Variable Product > Add Variations > Select existing attribute (e.g. Color)
3. Scroll down on the attribute screen to select an existing option
4. See that not all options are displayed for that attribute

**Screenshots**

https://github.com/woocommerce/woocommerce-ios/assets/17252150/7ffebfba-bab3-4085-961d-a70184d547d1

**Expected behavior**
All options should be displayed and selectable from the add options screen

- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).

**Mobile Environment**
Please include:
- Device: Pixel 6a
- OS version: Android 12
- WooCommerce version: v14.8-rc-1 (current beta)

**WordPress Environment**
<details>
```
Copy and paste the system status report from **WooCommerce > System Status** in WordPress admin.
```
</details>

ThomazFB:
Why
==========
This is a temporary fix for issue https://github.com/woocommerce/woocommerce-android/issues/9583. 

The current attributes term fetching doesn't use any pagination strategy and the `per_page` default value is 10. Because of this limitation, some users are reporting problems to see all terms of an attribute. 

How
==========
This PR solves this problem with brute force by making the `per_page` maxed to 100, but this won't be the final solution.

In the upcoming PRs under work right now, a proper pagination system is being introduced to make the attributes term loading more intelligent and capable of loading any amount of terms.

How to Test
==========
1. Make sure unit tests and CI are green.

Update release notes:

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/attributes/ProductAttributeRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/attributes/ProductAttributeRestClient.kt
@@ -42,9 +42,13 @@ class ProductAttributeRestClient @Inject constructor(private val wooNetwork: Woo
 
     suspend fun fetchAllAttributeTerms(
         site: SiteModel,
-        attributeID: Long
+        attributeID: Long,
+        pageSize: Int = 100
     ) = WOOCOMMERCE.products.attributes.attribute(attributeID).terms.pathV3
-            .request<Array<AttributeTermApiResponse>>(site)
+            .request<Array<AttributeTermApiResponse>>(
+                site = site,
+                params = mapOf("per_page" to pageSize.toString())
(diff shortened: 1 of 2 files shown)


atorresveiga (approved):
Sorry for the late review Thomaz. 🙏 
Everything is working as expected!

LGTM :shipit:

PR #2855, opened 2023-09-15
Enables fetching product reviews by specific list of review ids

JorgeMucientes:
Enables retrieving product reviews from database by product review id. Additionally it adds a small refactor to make behavior more explicit about clearing the product reviews database after fetching reviews from the API. 

Nothin to test in this PR. The new function will be tested in real scenario in this PR: https://github.com/woocommerce/woocommerce-android/pull/9753

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -84,6 +84,7 @@ class WCProductStore @Inject constructor(
 
         override fun toString() = name.toLowerCase(Locale.US)
     }
+
     enum class SkuSearchOptions {
         Disabled, ExactSearch, PartialMatch
     }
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1200,8 +1235,8 @@ class WCProductStore @Inject constructor(
             } else {
                 // Clear existing product reviews if this is a fresh fetch (loadMore = false).
                 // This is the simplest way to keep our local reviews in sync with remote reviews
-                // in case of deletions.
-                if (!response.loadedMore) {
+                // in case of deletions or status updates.
+                if (deletePreviouslyCachedReviews) {
                     ProductSqlUtils.deleteAllProductReviewsForSite(response.site)
```
JorgeMucientes, line 1239: The way this worked was a bit "hidden" in the code. I added this new parameter that clearly states it's purpose and allows for better control on when to reset the existing cached reviews. For example, when applying unread filter we do not need to wipe out the existing cached reviews.

hichamboushaba: I agree, but just a small np, I would remove the default value from this, or at least make it `false`: deleting the cached reviews by default doesn't seem very safe, as normally by default we shouldn't remove them, and we should do it only for some specific conditions.

JorgeMucientes: Entirely agree. Default values are risky on it's own. I initially added it to avoid "breaking" changes. But it was pretty easy to fix that. So removed the default value 👍🏼

hichamboushaba (approved):
Nice work @JorgeMucientes :shipit:

PR #2857, opened 2023-09-16
Preparations for Jetpack AI assistant eligibility checks

hichamboushaba:
This PR contains two changes to make it possible for checking the site's eligibility for AI assistant usage:

##### Make sure the site's plan features are fetched correctly 5f96f8f

In the PR https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1936, the URL used for fetching user's sites changed from `/v1.1/me/sites` to `/v1.2/me/sites`, but this had an impact that wasn't noticed: the response of the v1.2 doesn't include the plan's features list.

This PR attempts to fix this by following this logic:
1. If `fetchSites` includes any filtering, it will make two requests: `/v1.2/me/sites` and `/v1.1/me/sites/features`, then combine their responses.
2. If `fetchSites` was called without filters, it will simply call `/v1.1/me/sites`

The above will make sure we always include the plan's features, while optimizing the time of the request when we don't need any filtering.

##### Allow checking if a product is a sample product generated by WooExpress bb10a96
This is a simple change that makes sure the metadata for sample products is saved and exposed to client apps following the logic pecCkj-Zo-p2#comment-506

---
Testing will be done in the WCAndroid PR https://github.com/woocommerce/woocommerce-android/pull/9798

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
@@ -132,16 +132,36 @@ class SiteRestClient @Inject constructor(
         val site: SiteModel? = null
     ) : Payload<SiteError>()
 
+    @Suppress("ComplexMethod")
     suspend fun fetchSites(filters: List<SiteFilter?>, filterJetpackConnectedPackageSite: Boolean): SitesModel {
+        val useV2Endpoint = filters.isNotEmpty()
         val params = getFetchSitesParams(filters)
-        val url = WPCOMREST.me.sites.urlV1_2
+        val url = WPCOMREST.me.sites.let { if (useV2Endpoint) it.urlV1_2 else it.urlV1_1 }
         val response = wpComGsonRequestBuilder.syncGetRequest(this, url, params, SitesResponse::class.java)
+
+        val siteFeatures = if (useV2Endpoint) {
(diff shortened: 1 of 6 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
```diff
@@ -132,23 +132,44 @@ class SiteRestClient @Inject constructor(
         val site: SiteModel? = null
     ) : Payload<SiteError>()
 
+    @Suppress("ComplexMethod")
     suspend fun fetchSites(filters: List<SiteFilter?>, filterJetpackConnectedPackageSite: Boolean): SitesModel {
+        val useV2Endpoint = filters.isNotEmpty()
```
hafizrahman, line 145: np: I can't help but think that this part might need an explanatory comment. Someone can read further by tracking back to this PR but I wish it's understandable at a glance right away why function call without filter reverts back to using v1.

Not a big deal in the end, though.

hafizrahman (approved):
looks good to me, just left a small comment.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt
hichamboushaba: Good point, done in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2857/commits/0c191a93520e2c3ee4194c91213b0f1751c89780

PR #2861, opened 2023-10-02
Add support for order 'needs_payment' and 'needs_processing' properties

### LINKED ISSUE #6992: [Orders] Add support for `is_editable`, `needs_payment`, and `needs_processing`
Reference: p1658236518810329-slack-C025A8VV728

Stores using WC6.6+ can use the new Order-related [endpoints](https://github.com/woocommerce/woocommerce/pull/32888) `is_editable`, `needs_payment`, `needs_processing`. 

### Why?

At the moment, certain checks like order statuses, payment eligibility, or refund eligibility, may depend on a [nil-check](https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/model/Order.kt#L52) for the Order's `datePaid` property where we assume that if an order has a non-nil `datePaid`, it has been effectively paid, this however causes different problems, for example:

- If the merchant switches order status manually
- Plugins that modify `datePaid`
- Custom order statuses
- Orders where all products are Virtual and Downloadable

More context: p2-p91TBi-8to

### Logic to be implemented:

This can be taken directly from the Woocommerce core implementation:
- `is_editable`: If Order status is `pending`, `on-hold`, or `auto-draft`, the Order is editable: https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/includes/class-wc-order.php#L1520-L1523
- `needs_payment`: If the Order status is `pending`, or `failed`, and the Order total is > 0, the Order needs payment: https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/includes/class-wc-order.php#L1395-L1402
- `needs_processing`: All Orders need processing, except these where the Order only contains Virtual and Downloadable products: https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/includes/class-wc-order.php#L1537-L1567 . Note that "needsProcessing" is not the same as "Order Status processing", the first one is an Order state, while the second one is an Order status.
- We also need fallbacks for each property, in case the merchant is not on WC6.6+. Here's the fallback implementation example on iOS: https://github.com/woocommerce/woocommerce-ios/blob/bf903fb6b3ece4535f17076644f0dc922c61a9e9/Networking/Networking/Model/Order.swift#L190-L195

toupper:
Closes: https://github.com/woocommerce/woocommerce-android/issues/6992

With this PR we add support for `needs_payment` and `needs_processing` in the order entity. As Woo clients older than 6.6 do not provide these data in the order JSON structure, we make these new properties nullable in the database, so we don't add any incorrect default value.

Together with these changes, we add a migration to drop the table and create it anew, forcing that way to update the orders information with these new values.

Once this PR is merged I will create a PR in the android WooCommerce repo to point to these changes in FluxC.

### DIFF
--- a/plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapperTest.kt
+++ b/plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapperTest.kt
@@ -67,7 +67,7 @@ internal class OrderDtoMapperTest {
     }
 
     @Test
-    fun `when needs_payment is not is json the order dto needs_payment property is null`() {
+    fun `when needs_payment is not in json the order dto needs_payment property is null`() {
         val json = JsonObject()
         val orderDto = Gson().fromJson(json, OrderDto::class.java)
         val (orderEntity, _) = sut.toDatabaseEntity(orderDto, localSiteId)
@@ -87,7 +87,7 @@ internal class OrderDtoMapperTest {
     }
 
     @Test
-    fun `when needs_processing is not is json the order dto needs_processing property is null`() {
+    fun `when needs_processing is not in json the order dto needs_processing property is null`() {
         val json = JsonObject()
         val orderDto = Gson().fromJson(json, OrderDto::class.java)
         val (orderEntity, _) = sut.toDatabaseEntity(orderDto, localSiteId)
--- a/plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapperTest.kt
+++ b/plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapperTest.kt
@@ -65,4 +65,44 @@ internal class OrderDtoMapperTest {
         }
         return Gson().fromJson(json, OrderDto::class.java)
     }
+
+    @Test
+    fun `when needs_payment is not is json the order dto needs_payment property is null`() {
+        val json = JsonObject()
+        val orderDto = Gson().fromJson(json, OrderDto::class.java)
+        val (orderEntity, _) = sut.toDatabaseEntity(orderDto, localSiteId)
+
+        assertThat(orderEntity.needsPayment).isNull()
+    }
+
+    @Test
+    fun `when needs_payment is in json the order dto needs_payment property is parsed`() {
+        val json = JsonObject().apply {
+            addProperty("needs_payment", "true")
+        }
+        val orderDto = Gson().fromJson(json, OrderDto::class.java)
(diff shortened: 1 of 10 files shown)


plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapperTest.kt
```diff
@@ -65,4 +65,44 @@ internal class OrderDtoMapperTest {
         }
         return Gson().fromJson(json, OrderDto::class.java)
     }
+
+    @Test
+    fun `when needs_payment is not is json the order dto needs_payment property is null`() {
```
rossanafmenezes: NP: just a typo, I believe. Shouldn't it be ..`is not in json.`.. ?

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapperTest.kt
```diff
@@ -65,4 +65,44 @@ internal class OrderDtoMapperTest {
...
+            addProperty("needs_payment", "true")
+        }
+        val orderDto = Gson().fromJson(json, OrderDto::class.java)
+        val (orderEntity, _) = sut.toDatabaseEntity(orderDto, localSiteId)
+
+        assertThat(orderEntity.needsPayment).isTrue()
+    }
+
+    @Test
+    fun `when needs_processing is not is json the order dto needs_processing property is null`() {
```
rossanafmenezes: Same question as above 😬

rossanafmenezes (approved):
LGTM. 

Found two typos in a test name, but not blocking.

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapperTest.kt
toupper: Good point, fixed in 348f2bd52a3d725bf0e521791c0e75566add5fe1

plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/OrderDtoMapperTest.kt
toupper: Yeah, fixed in 348f2bd52a3d725bf0e521791c0e75566add5fe1

PR #2862, opened 2023-10-04
Woo: ability to bulk add categories

hichamboushaba:
Needed for https://github.com/woocommerce/woocommerce-android/issues/9878

This PR adds support for bulk creating categories, using the `/batch` endpoint from the Woo API.
As agreed, there is no advanced handling for individual errors, and the consumers are expected to fetch updated resources before sending a request to create new ones.
I still added a statement to log a warning if some categories couldn't be created.

#### Testing
1. Open the example app and sign in.
2. Click on Woo, then pick a site.
3. Click on Products then pick a site (We need to fix this 😅)
4. Click on Update product.
5. Enter a product id.
6. Click on Select Categories.
7. Click on Add categories.
8. Enter some names separated by comma.
9. Confirm the creation succeeds, and the categories are shown in the list.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductCategoryApiResponse.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductCategoryApiResponse.kt
@@ -1,15 +1,16 @@
 package org.wordpress.android.fluxc.network.rest.wpcom.wc.product
 
+import com.google.gson.JsonObject
+import com.google.gson.annotations.SerializedName
 import org.wordpress.android.fluxc.model.WCProductCategoryModel
 import org.wordpress.android.fluxc.network.Response
 
-@Suppress("PropertyName")
 class ProductCategoryApiResponse : Response {
     val id: Long = 0L
-    var localSiteId = 0
     var name: String? = null
     var slug: String? = null
     var parent: Long? = null
+    val error: JsonObject? = null
 
     fun asProductCategoryModel(): WCProductCategoryModel {
         val response = this
@@ -21,3 +22,8 @@ class ProductCategoryApiResponse : Response {
         }
     }
 }
+
+data class ProductCategoryBatchApiResponse(
+    @SerializedName("create")
+    val createdCategories: List<ProductCategoryApiResponse> = emptyList()
+)
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -19,6 +19,7 @@ import org.wordpress.android.fluxc.model.WCProductReviewModel
 import org.wordpress.android.fluxc.model.WCProductShippingClassModel
 import org.wordpress.android.fluxc.model.WCProductTagModel
 import org.wordpress.android.fluxc.model.WCProductVariationModel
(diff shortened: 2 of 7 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductCategoryApiResponse.kt
```diff
@@ -1,15 +1,16 @@
...
 
+import com.google.gson.JsonObject
+import com.google.gson.annotations.SerializedName
 import org.wordpress.android.fluxc.model.WCProductCategoryModel
 import org.wordpress.android.fluxc.network.Response
 
-@Suppress("PropertyName")
 class ProductCategoryApiResponse : Response {
     val id: Long = 0L
-    var localSiteId = 0
```
hichamboushaba, line 9: This wasn't used

hichamboushaba:
@JorgeMucientes I tried to move all of the category related functions to separate Store/RestClient classes, but since it still uses EventBus for some functions, it wasn't straightforward, as I needed to refactor also the Actions, so I decided to skip it for now, after finishing the project, if we have time left, I'll go back to make the functions suspendable in the new classes.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -1434,6 +1438,55 @@ class ProductRestClient @Inject constructor(
...
+        site: SiteModel,
+        categories: List<WCProductCategoryModel>
+    ): WooPayload<List<WCProductCategoryModel>> {
+        val path = WOOCOMMERCE.products.categories.batch.pathV3
+
+        val body = mutableMapOf(
+            "create" to categories.map { category ->
+                mapOf(
+                    "name" to category.name,
+                    "parent" to category.parent.toString()
```
JorgeMucientes, line 1451: Minor np, why `toString` ? `parent` is an int value corresponding to the category parent id right?

JorgeMucientes:
Great job @hichamboushaba! Code looks good and works as expected. 

> @JorgeMucientes I tried to move all of the category related functions to separate Store/RestClient classes, but since it still uses EventBus for some functions, it wasn't straightforward, as I needed to refactor also the Actions, so I decided to skip it for now, after finishing the project, if we have time left, I'll go back to make the functions suspendable in the new classes.

Tbh what I initially had in mind when we talked about this, wasn't to migrate all the Categories/Tags code to its own Store and Rest client, because as you well noticed that is a bigger refactor. What I had in mind was more of a "starting point" refactor were we'd only add the new functions the the new store and rest client and then if there was time start moving the rest of the existing code. Anyway, not a big a deal, at least the functions are suspendable and the call chain is easy to follow through. So, I wouldn't spend much time migrating existing code. But that's just my opinion 🙂

JorgeMucientes (approved):

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
hichamboushaba: Good point, thanks for pointing it up, I checked the APIs doc, and it expects a number here, so we don't need the `toString` call.
Fixed in 84bea3a

hichamboushaba:
> What I had in mind was more of a "starting point" refactor were we'd only add the new functions the the new store and rest client and then if there was time start moving the rest of the existing code.

I didn't think about it this way, and personally I think it's better to keep them together until we have time to move them all to the new classes, so let's keep it this way for now 🙂.

PR #2865, opened 2023-10-06
Jetpack AI: improve caching logic of the JWT token

hichamboushaba:
This PR implements three changes to improve/fix the JWT token caching logic:
1. Remove the storage caching, and uses instead just an in-memory caching.
2. Validate the JWT token's expiry date before using, this will avoid delays caused by useless requests due to using an expired token.
3. Validate the `blog_id` property of the JWT token before using it, this fixes an issue where we might use a token for the wrong store.

### Testing
1. Open the example app and sign in using WordPress.com
2. Click on Jetpack AI
3. Select a site.
4. Open Logcat or use Flipper to monitor requests.
5. Click on Generate Haiku
6. Confirm there is a call to generate a new token `JetpackAIStore: fetch Jetpack AI JWT token`
7. Repeat 5, and confirm the token was reused.
8. Select a new site, and repeat 5.
9. Confirm a new token was generated.
10. Wait for 2 minutes.
11. Repeat 5.
12. Confirm a new token was generated.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/JWTTokenTests.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/JWTTokenTests.kt
@@ -1,0 +1,47 @@
+package org.wordpress.android.fluxc
+
+import android.util.Base64
+import org.junit.Assert
+import org.junit.Test
+import org.wordpress.android.fluxc.model.JWTToken
+
+class JWTTokenTests {
+    @Test
+    fun given_a_valid_token__when_takeIfValid_is_called__then_return_it() {
+        val token = generateToken(expired = false)
+
+        val result = token.takeIfValid()
+
+        Assert.assertNotNull(result)
+    }
+
+    @Test
+    fun given_an_expired_token__when_takeIfValid_is_called__then_return_null() {
+        val token = generateToken(expired = true)
+
+        val result = token.takeIfValid()
+
+        Assert.assertNull(result)
+    }
+
+    private fun generateToken(expired: Boolean): JWTToken {
+        val expirationTime = System.currentTimeMillis() / 1000 + if (expired) -100 else 100
+
+        // Sample token from https://jwt.io/ modifier with an expiration time
+        val header = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9"
+        val payload = Base64.encode(
+            """
+                {
+                    "sub": "1234567890",
+                    "name": "John Doe",
+                    "iat": 1516239022,
+                    "exp": $expirationTime,
+                    "expires": $expirationTime
+                }
+            """.trimIndent().toByteArray(), Base64.DEFAULT
+        ).decodeToString()
+        val signature = "SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"
+
+        return JWTToken("$header.$payload.$signature")
+    }
+}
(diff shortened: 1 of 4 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/model/JWTToken.kt
```diff
@@ -0,0 +1,35 @@
+package org.wordpress.android.fluxc.model
+
+import android.util.Base64
+import org.json.JSONObject
+
+@JvmInline
+value class JWTToken(
```
hichamboushaba, line 7: I added this class just to make testing the behavior easier, but if you feel this is unnecessary or have a better suggestion, please share it.

fluxc/src/main/java/org/wordpress/android/fluxc/store/jetpackai/JetpackAIStore.kt
```diff
@@ -78,59 +76,46 @@ class JetpackAIStore @Inject constructor(
...
 
-     private suspend fun fetchJetpackAIJWTToken(site: SiteModel)
-     : JetpackAIJWTTokenResponse = coroutineEngine.withDefaultContext(
-        tag = AppLog.T.API,
-        caller = this,
-        loggedMessage = "fetch Jetpack AI JWT token"
-    ) {
-        jetpackAIRestClient.fetchJetpackAIJWTToken(site)
-    }
+    private fun JWTToken.validateBlogId(blogId: Long): JWTToken? =
```
hichamboushaba, line 119: I made this an extension instead of a member function because the `blog_id` is not a standard JWT property, unlike the expiration time, where `exp` is a standard property.

example/src/androidTest/java/org/wordpress/android/fluxc/JWTTokenTests.kt
```diff
@@ -0,0 +1,47 @@
+package org.wordpress.android.fluxc
+
+import android.util.Base64
+import org.junit.Assert
+import org.junit.Test
+import org.wordpress.android.fluxc.model.JWTToken
+
+class JWTTokenTests {
+    @Test
+    fun given_a_valid_token__when_validateExpiryDate_is_called__then_return_it() {
```
JorgeMucientes, line 10: Minor np. Is there a reason why these unit tests are inside the `example/androidTest` directory? Shouldn't they be inside `example/test` ->  `org.wordpress.android.fluxc.model` package ?

Also, I'm curious why not follow this syntax for test naming. I saw other FluxC unit test using it. 

`fun 'given a valid token, when validateExpiryDate is called, then return it' () {  `

JorgeMucientes (approved):

JorgeMucientes:
Nice improvement, thanks for working on it. Work as expected @hichamboushaba. Left a couple questions but nothing blocking. Ready for 🚢

example/src/androidTest/java/org/wordpress/android/fluxc/JWTTokenTests.kt
hichamboushaba: Because this use the Android class `Base64`, so they need to be instrumented tests and not unit tests.

> Also, I'm curious why not follow this syntax for test naming.

For the above reason, since the tests will run in an Android runtime, using the spaces with backticks causes the tests to fail (https://stackoverflow.com/a/51565546)

JorgeMucientes: Thanks for clarifiying! Ship it!

PR #2867, opened 2023-10-11
Adds fetch all-domains api call

antonis:
Part of https://github.com/wordpress-mobile/WordPress-Android/issues/19165

This PR adds the ability to fetch all domains for the logged in user.

Note: For simplicity and given that the domains management feature will use [the existing fetchSiteDomains call](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ba92dd07fbcc2987c91a8d8a37fd5952881d3286/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.kt#L840) I decided to add the new call in the `SiteStore` though it is not directly related to the current site.

**To test:**

1. Start the example app and sign in (note: [setup oauth first](https://github.com/wordpress-mobile/WordPress-FluxC-Android#oauth2-authentication))
2. Tap on the DOMAINS button
3. Tap on the FETCH ALL DOMAINS button
4. Verify that your domains are listed
5. Verify that the domains count is correct

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClientTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClientTest.kt
@@ -549,6 +549,46 @@ class SiteRestClientTest {
         }
     }
 
+    @Test
+    fun `given a network error, when all domains are requested, then return api error`() = test {
+        val error = WPComGsonNetworkError(BaseNetworkError(GenericErrorType.NETWORK_ERROR))
+        initAllDomainsResponse(error = error)
+
+        val response = restClient.fetchAllDomains(noWpCom = true)
+        assert(response is Response.Error)
+        with((response as Response.Error).error) {
+            assertThat(type).isEqualTo(GenericErrorType.NETWORK_ERROR)
+            assertThat(message).isNull()
+        }
+    }
+
+    @Test
+    fun `given timeout, when all domains are requested, then return timeout error`() = test {
+        val error = WPComGsonNetworkError(BaseNetworkError(GenericErrorType.TIMEOUT))
+        initAllDomainsResponse(error = error)
+
+        val response = restClient.fetchAllDomains(noWpCom = true)
+        assert(response is Response.Error)
+        with((response as Response.Error).error) {
+            assertThat(type).isEqualTo(GenericErrorType.TIMEOUT)
+            assertThat(message).isNull()
+        }
+    }
+
+    @Test
+    fun `given not authenticated, when all domains are requested, then retun auth required error`() = test {
+        val tokenErrorMessage = "An active access token must be used to query information about the current user."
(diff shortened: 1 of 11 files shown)


example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClientTest.kt
```diff
@@ -526,6 +529,26 @@ class SiteRestClientTest {
...
+        initAllDomainsResponse(data = response)
+
+        val responseModel = restClient.fetchAllDomains(noWpCom = true)
+        assert(responseModel is Success)
+        with((responseModel as Success).data) {
+            assertThat(domains).hasSize(2)
+            assertThat(domains[0].domain).isEqualTo("some.test.domain")
+            assertThat(domains[0].wpcomDomain).isFalse()
+            assertThat(domains[1].domain).isEqualTo("some.test.domain 2")
+            assertThat(domains[1].wpcomDomain).isTrue()
```
justtwago: _Optional nitpick_: Android Studio recommends using property access syntax here instead of calling getters.
```suggestion
            assertThat(domains[0].wpcomDomain).isFalse
            assertThat(domains[1].domain).isEqualTo("some.test.domain 2")
            assertThat(domains[1].wpcomDomain).isTrue
```

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClientTest.kt
```diff
@@ -561,6 +584,13 @@ class SiteRestClientTest {
         return initGetResponse(SitesFeaturesRestResponse::class.java, data ?: mock(), error)
     }
 
+    private suspend fun initAllDomainsResponse(
+        data: AllDomainsResponse? = null,
+        error: WPComGsonNetworkError? = null
```
justtwago, line 629: _Optional nitpick_: the property is never used. If we don't want to test the error cases for the "fetch all domains" flow, we can remove it and pass `null` directly, or we can have a test for a bad case too. What do you think?

justtwago (approved):
Tested the changes, and it works as described. 🚀 



Here is the fetched list of my domains.

The code looks good! I just left some minor comments that don't block the PR.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClientTest.kt
antonis: Good catch @justtwago 🙇 
I've committed the suggestion.

example/src/test/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClientTest.kt
antonis: Thank you for catching this @justtwago 🙇 
I've added some tests for the error cases with 5bed1ee94a53a20c09e8242d31c8e1988abe8818

justtwago (approved):
Thank you a lot for the changes! 🚀

PR #2868, opened 2023-10-11
[Nullability Annotations to Java Classes] Add Missing Nullability Annotations to `Comment` Network Client Classes (`safe`)

ParaskP7:
Parent #2798

This PR adds any missing nullability annotation to [CommentRestClient.java](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ee1af4fea6995ca7a97e579a67942bf3ddfa1ec9/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/comment/CommentRestClient.java#L4) and [CommentXMLRPCClient.java](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ee1af4fea6995ca7a97e579a67942bf3ddfa1ec9/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/comment/CommentXMLRPCClient.java#L4), all its related response classes and anything in between (ie. `CommentStore`). See response classes below:
- [CommentWPComRestResponse.java](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ee1af4fea6995ca7a97e579a67942bf3ddfa1ec9/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/comment/CommentWPComRestResponse.java#L4)

FYI.1: This change is `safe`, meaning that there shouldn't be any (major) compile-time changes associated with this change. Having said that, testing is highly recommended since the changes in this PR can affect one or more clients (ie. [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android), [WCAndroid](https://github.com/woocommerce/woocommerce-android), etc)

FYI.2: An analysis on `CommentStore` and its usages with our main clients suggests that this store is only being utilizing by [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android). AFAIA [WCAndroid](https://github.com/woocommerce/woocommerce-android) is not using that store and as such can be safely ignored from testing.

PS: Note that [CommentStore.java](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ee1af4fea6995ca7a97e579a67942bf3ddfa1ec9/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java#L4) and their respective network client classes, both `REST` and `XMLRPC`, seem to be deprecated. The [CommentsStore.kt](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ee1af4fea6995ca7a97e579a67942bf3ddfa1ec9/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentsStore.kt#L4) and their respective network client classes, both `REST` and `XMLRPC`, seem to be the ones that are actually being actively used atm. Also note that the [CommentWPComRestResponse.java](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ee1af4fea6995ca7a97e579a67942bf3ddfa1ec9/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/comment/CommentWPComRestResponse.java#L4) response is shared between those two stores. As such, it seems that all comment related API calls from JP/WPAndroid are currently flowing via the new [CommentsStore.kt](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ee1af4fea6995ca7a97e579a67942bf3ddfa1ec9/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentsStore.kt#L4), while only FluxC is using the deprecated [CommentStore.java](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/ee1af4fea6995ca7a97e579a67942bf3ddfa1ec9/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java#L4). Be mindful of all that when testing this PR. 🤷 

-----

@wzieba let me know how I did on this PR, that is, in terms of updating these network client classes and adding nullability annotations on them and their main `CommentWPComRestResponse` response class, in order to utilize the nullability annotations added on its fields and thus make its use null proof. If you also feel that this is a good way to proceed with such client PRs, I'll then use this pattern across the board and follow a similar approach on every such PR.

-----

Nullability Annotation List:

1. [Add missing nn-a to likes utils pvd field on cmt rest client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/9fef4111ea753d93ebf67662916e0d083e6795fd)
2. [Add missing n-a to fetch comments on comment clients](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/1c258619e392a03a31027ac072782d32858922ee)
3. [Add missing n-a to push comment on comment clients](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/de7f769a454ba426efd41693107bffe6eecc80b8)
4. [Add missing n-a to fetch comment on comment clients](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/9f6c87b5d493ad8813372a5d0b7034d11a0d3774)
5. [Add missing n-a to fetch comment likes on comment rest client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/14729c21d3d172a6d469e4a7f89c6e5fc9caa756)
6. [Add missing n-a to delete comment on comment clients](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/12fe3cfc80ec316ca2211924e0e498ae88722c1d)
7. [Add missing n-a to create new reply/comment on comment clients](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/53e14a0bde415d92fede132f261b3161f843e846)
8. [Add missing n-a to like comment on comment rest client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/d044e7519fd416ed476a422cd935f49db4333bc5)
9. [Add missing n-a to comments response to comment list](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/c9ccc7295366e3719075726327c38d37abbfa03d)
10. [Add missing nl-a to requested status field on payload](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/726ea21602596f2e9b118ace46d7ef4a01aee818)
11. [Add missing n-a to comment response to comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/057436b1bfbf2748c6f3f8151452a0400354ed42)
12. [Add missing n-a to get cmt status from xmlrpc status string](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/484463a62cebdfcfd230e5a7bbaa2a8408b549f8)
13. [Add missing n-a to comments response to comment list](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/c6d0db782e6bb39f362f197c15f9f6d93b07905a)
14. [Add missing n-a to comment response to comment](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/8b2e4ace98fd41ac91cc93e1370c961028a57fd9)
15. [Add missing n-a to comment wp com rest response fields](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/2e855cd8fca3e827ef0d3b23a4018254e40af44e)

Nullability Checks List:

1. [Remove unnecessary payload likes if condition on comment store](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/b73fbb0e48b212a11025043cfcdbb6daa21402f2)

Warnings Resolutions List:

1. [Create missing switch branch on comment xmlrpc client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/35d09356d308ead0619a10e0c3b519d4c298f51e)
2. [Remove complexity from get cmt status from xmlrpc status str](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/246005f0538f73a68105fef7ec32c5ade2fc625b)
3. [Replace explicit type argument with <> on payload](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/a736e7481727ea919217b736410aaffd967900e3)
4. [Replace if with switch on comment xmlrpc client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/41d27ca6bcaba71dd9fbb41bd19a8750a1369b15)
5. [Replace explicit type argument with <> on payload](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/816d1e3d83bce3ad6b41767feace4ede30eb66dd)
6. [Make inner classes static on comment wp com rest response](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/85fa7bfcd30dfe407c056dfb5c4629f470ca11b2)

Warnings Suppression List:

1. [Suppress long parameter list warning on comments clients](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/fd0532864d540dea0318eeeeb06af5a8672898d3)
2. [Suppress too many functions warning on comments xmlrpc client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/d0df40f8166a7f0f124b22cb3b3e4b54c0d75362)
3. [Suppress unused warnings on comment store](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/0bc6ba314d9c040cd4a752911ac5ed646ec3640d)
4. [Suppress unused return value warning on comment store](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/fb46f9433b4e7352c9a694777a6c1ad6036a1795)
5. [Suppress raw types warning on comment store](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/f34249e0c5495e86159e04c99f02aade1331d5a0)

Refactor List:

1. [Reformat comment rest client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/13840f1a8419e6961d0854e025d7d05cde206b57)
2. [Reformat comment xmlrpc client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/ae9066ca1cb0b3f66e154bea20ce4fd0acff90e8)
3. [Replace anonymous classes with lambda on comment rest client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/a4c64eff1406fa6386bbcce94112d8f21d277274)
4. [Replace anonymous classes with lambda on comment xmlrpc client](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/c6458944eeff0e395863858a175f9fe82de9051e)
5. [Remove test only comment store method](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/6bdbfa424bc8e92935067d8a8ef5f4732601ce0f)

Test List:

1. [Suppress new class naming convention for connected tests](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/89c0d962ac16f60a79d4ee712f46875a5b060664)
2. [Remove interrupted exception from test method's throws list](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/0ba2a4d3787beddf858f9caba6837fea83e2526d)
3. [Simplify assert null assertion for connected tests](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/04d7934a3b2a1f0aa6adb859382fb100b37a666b)
4. [Resolve robolectric related application deprecation warnings](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/47d40ba5b027b8e4478fcdc5cc26fe2ddc7a6947)
5. [Guard usages of get comment by site and remote id method](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/21ca96ece9ad12c067cd2dfcd7a289db07d0b237)
6. [Simplify assert null assertion for unit tests](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/048fe853d40c2b0114efdaa76b9d6017b7e95933)
7. [Remove unnecessary comment response local variable from unit test](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/015f66438c755f918af85bd976df6a7238ef72f3)
8. [Fix typo with comments response on unit test](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/2bab5c02435f82f025eac5ec30a397ad41ad9582)
9. [Suppress unchecked cast on unit test](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/88d0a1e4bad9734b4bedd3c83c9560bf77bd555e)
10. [Fix typo with updated comment on unit test](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2868/commits/dd40ab2ccf66ffe75fed7fdf7150f3e5da6d0580)

-----

## Associated Clients

It is highly recommending that this change is being tested on the below associated clients:

- [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android)

-----

## To Test (`REST`):

- Smoke test the `FluxC Example` app via the `COMMENTS` screen. Verify everything is working as expected.
- In addition to the above, using [local-builds.gradle](https://github.com/wordpress-mobile/WordPress-Android/blob/44a6d0611175f132ff6920b4d6a3e16d565259b6/local-builds.gradle-example#L18) on [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android), smoke test any `CommentStore` related functionality on both, the WordPress and Jetpack apps, and see if everything is working as expected. For a couple of examples, you can expand and follow the inner and more explicit test steps within:

## To Test (`XMLRPC`):

- Create a new self-hosted WordPress site for XMLRPC testing purposes ([jurassic.ninja](https://jurassic.ninja/)).
- Smoke test the `FluxC Example` app via the `COMMENTS` screen. Verify everything is working as expected.
- In addition to the above, using [local-builds.gradle](https://github.com/wordpress-mobile/WordPress-Android/blob/44a6d0611175f132ff6920b4d6a3e16d565259b6/local-builds.gradle-example#L18) on [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android), smoke test any `CommentStore` related functionality on both, the WordPress and Jetpack apps, and see if everything is working as expected. For a couple of examples, you can expand and follow the inner and more explicit test steps within:

<details>
    <summary>1. Comment Details Screen from Comments Screen [CommentDetailFragment.java]</summary>

ℹ️ This test applies to both, the `Jetpack` and `WordPress` app.

- Go to `Comments` screen and tap on any comment.
- Verify that the `Comment Details` screen is shown and functioning as expected.
- For example, try replying, approving, marking as spam, liking, trashing, editing or copying a comment's link address. Additionally, make sure to verify that you can still swipe left-right to navigate between comments.

</details>

<details>
    <summary>2. Comment Details Screen from Notification Tab [CommentDetailFragment.java]</summary>

ℹ️ This test applies to the `Jetpack` app only.

- Go to `Notifications` tab, then its `COMMENTS` sub-tab, and tap on any comment.
- Verify that the `Comment Details` screen is shown and functioning as expected.
- For example, try replying to a comment. Additionally, make sure to verify that you can still swipe left-right to navigate between comments.

</details>

<details>
    <summary>3. Edit Comment Screen from Comments Screen [UnifiedCommentsEditActivity.java]</summary>

ℹ️ This test applies to both, the `Jetpack` and `WordPress` app.

- Go to `Comments` screen and tap on any comment.
- From the `More` on the right side, at the bottom of the comment, select `Edit`.
- Verify that the `Edit Comment` screen is shown and functioning as expected.
- For example, try editing/adding the comment's `Name`, `Web address`, `Email address` and its `Comment` itself.

</details>

<details>
    <summary>4. Edit Comment Screen from Notification Tab [UnifiedCommentsEditActivity.java]</summary>

ℹ️ This test applies to the `Jetpack` app only.

- Go to `Notifications` tab, then its `COMMENTS` sub-tab, and tap on any comment.
- From the `More` on the right side, at the bottom of the comment, select `Edit`.
- Verify that the `Edit Comment` screen is shown and functioning as expected.
- For example, try editing/adding the comment's `Name`, `Web address`, `Email address` and its `Comment` itself.

</details>

<details>
    <summary>5. Reader Comments Screen [ReaderCommentListActivity.java]</summary>

ℹ️ This test applies to the `Jetpack` app only.

- Go to `Reader` tab, then its `DISCOVER` sub-tab, and tap on any post.
- Tap on the `Comments` button at the bottom of the post.
- Verify that the `Reader Comments` screen is shown and functioning as expected.
- For example, try (un)following, enabling/disabling in-app notifications, replying to post, and then sharing, replying or liking a specific comment.

</details>

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/common/CommentsMapperTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/common/CommentsMapperTest.kt
@@ -43,7 +43,10 @@ class CommentsMapperTest {
 
     @Test
     fun `xmlrpc dto list is converted to entity list`() {
-        val commentList = getDefaultCommentList(false).map { it.copy(id = 0, authorProfileImageUrl = null) }
+        val commentList = getDefaultCommentList(false).map { it.copy(
+            id = 0,
+            authorProfileImageUrl = null
+        ) }
         val site = SiteModel().apply {
             id = commentList.first().localSiteId
             selfHostedSiteId = commentList.first().remoteSiteId
@@ -120,27 +123,27 @@ class CommentsMapperTest {
         val entity = this
         return CommentWPComRestResponse().apply {
             ID = entity.remoteCommentId
-            URL = entity.url
+            URL = entity.url ?: ""
             author = Author().apply {
                 ID = entity.authorId
-                URL = entity.authorUrl
-                avatar_URL = entity.authorProfileImageUrl
-                email = entity.authorEmail
-                name = entity.authorName
+                URL = entity.authorUrl ?: ""
+                avatar_URL = entity.authorProfileImageUrl ?: ""
+                email = entity.authorEmail ?: ""
+                name = entity.authorName ?: ""
             }
-            content = entity.content
-            date = entity.datePublished
+            content = entity.content ?: ""
+            date = entity.datePublished ?: ""
             i_like = entity.iLike
             parent = CommentParent().apply {
                 ID = entity.parentId
             }
             post = Post().apply {
                 type = "post"
-                title = entity.postTitle
+                title = entity.postTitle ?: ""
                 link = "https://public-api.wordpress.com/rest/v1.1/sites/185464053/posts/85"
                 ID = entity.remotePostId
             }
-            status = entity.status
+            status = entity.status ?: ""
         }
     }
 
@@ -188,22 +191,22 @@ class CommentsMapperTest {
 
     private fun getDefaultComment(allowNulls: Boolean): CommentEntity {
         return CommentEntity(
-                id = 0,
-                remoteCommentId = 10,
-                remotePostId = 100,
-                authorId = 44,
+                id = 0L,
+                remoteCommentId = 10L,
+                remotePostId = 100L,
+                authorId = 44L,
                 localSiteId = 10_000,
-                remoteSiteId = 100_000,
-                authorUrl = if (allowNulls) null else "https://test-debug-site.wordpress.com",
-                authorName = if (allowNulls) null else "authorname",
-                authorEmail = if (allowNulls) null else "email@wordpress.com",
-                authorProfileImageUrl = if (allowNulls) null else "https://gravatar.com/avatar/111222333",
-                postTitle = if (allowNulls) null else "again",
+                remoteSiteId = 100_000L,
+                authorUrl = if (allowNulls) "" else "https://test-debug-site.wordpress.com",
+                authorName = if (allowNulls) "" else "authorname",
+                authorEmail = if (allowNulls) "" else "email@wordpress.com",
+                authorProfileImageUrl = if (allowNulls) "" else "https://gravatar.com/avatar/111222333",
+                postTitle = if (allowNulls) "" else "again",
                 status = APPROVED.toString(),
-                datePublished = if (allowNulls) null else "2021-05-12T15:10:40+02:00",
+                datePublished = if (allowNulls) "" else "2021-05-12T15:10:40+02:00",
                 publishedTimestamp = 1_000_000,
-                content = if (allowNulls) null else "content example",
-                url = if (allowNulls) null else "https://test-debug-site.wordpress.com/2021/02/25/again/#comment-137",
+                content = if (allowNulls) "" else "content example",
+                url = if (allowNulls) "" else "https://test-debug-site.wordpress.com/2021/02/25/again/#comment-137",
                 hasParent = true,
                 parentId = 1_000L,
                 iLike = false
@@ -213,10 +216,10 @@ class CommentsMapperTest {
     private fun getDefaultCommentList(allowNulls: Boolean): CommentEntityList {
         val comment = getDefaultComment(allowNulls)
         return listOf(
-                comment.copy(id = 1, remoteCommentId = 10, datePublished = "2021-07-24T00:51:43+02:00"),
-                comment.copy(id = 2, remoteCommentId = 20, datePublished = "2021-07-24T00:51:43+02:00"),
-                comment.copy(id = 3, remoteCommentId = 30, datePublished = "2021-07-24T00:51:43+02:00"),
-                comment.copy(id = 4, remoteCommentId = 40, datePublished = "2021-07-24T00:51:43+02:00")
+                comment.copy(id = 1L, remoteCommentId = 10L, datePublished = "2021-07-24T00:51:43+02:00"),
+                comment.copy(id = 2L, remoteCommentId = 20L, datePublished = "2021-07-24T00:51:43+02:00"),
+                comment.copy(id = 3L, remoteCommentId = 30L, datePublished = "2021-07-24T00:51:43+02:00"),
+                comment.copy(id = 4L, remoteCommentId = 40L, datePublished = "2021-07-24T00:51:43+02:00")
         )
     }
 }
--- a/example/src/test/java/org/wordpress/android/fluxc/common/CommentsMapperTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/common/CommentsMapperTest.kt
@@ -189,7 +189,7 @@ class CommentsMapperTest {
         )
     }
 
-    private fun getDefaultComment(allowNulls: Boolean): CommentEntity {
+    private fun getDefaultComment(withEmpty: Boolean): CommentEntity {
         return CommentEntity(
                 id = 0L,
                 remoteCommentId = 10L,
@@ -197,24 +197,25 @@ class CommentsMapperTest {
                 authorId = 44L,
                 localSiteId = 10_000,
                 remoteSiteId = 100_000L,
-                authorUrl = if (allowNulls) "" else "https://test-debug-site.wordpress.com",
-                authorName = if (allowNulls) "" else "authorname",
-                authorEmail = if (allowNulls) "" else "email@wordpress.com",
-                authorProfileImageUrl = if (allowNulls) "" else "https://gravatar.com/avatar/111222333",
-                postTitle = if (allowNulls) "" else "again",
+                authorUrl = if (withEmpty) "" else "https://test-debug-site.wordpress.com",
+                authorName = if (withEmpty) "" else "authorname",
+                authorEmail = if (withEmpty) "" else "email@wordpress.com",
+                authorProfileImageUrl = if (withEmpty) "" else "https://gravatar.com/avatar/111222333",
+                postTitle = if (withEmpty) "" else "again",
                 status = APPROVED.toString(),
-                datePublished = if (allowNulls) "" else "2021-05-12T15:10:40+02:00",
+                datePublished = if (withEmpty) "" else "2021-05-12T15:10:40+02:00",
                 publishedTimestamp = 1_000_000,
-                content = if (allowNulls) "" else "content example",
-                url = if (allowNulls) "" else "https://test-debug-site.wordpress.com/2021/02/25/again/#comment-137",
+                content = if (withEmpty) "" else "content example",
+                url = if (withEmpty) "" else "https://test-debug-site.wordpress.com/2021/02/25/again/#comment-137",
                 hasParent = true,
                 parentId = 1_000L,
                 iLike = false
         )
     }
 
-    private fun getDefaultCommentList(allowNulls: Boolean): CommentEntityList {
-        val comment = getDefaultComment(allowNulls)
+    @Suppress("SameParameterValue")
+    private fun getDefaultCommentList(withEmpty: Boolean): CommentEntityList {
+        val comment = getDefaultComment(withEmpty)
         return listOf(
                 comment.copy(id = 1L, remoteCommentId = 10L, datePublished = "2021-07-24T00:51:43+02:00"),
                 comment.copy(id = 2L, remoteCommentId = 20L, datePublished = "2021-07-24T00:51:43+02:00"),
--- a/example/src/test/java/org/wordpress/android/fluxc/common/CommentsMapperTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/common/CommentsMapperTest.kt
@@ -11,6 +11,8 @@ import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.comments.CommentsMapper
 import org.wordpress.android.fluxc.network.rest.wpcom.comment.CommentParent
 import org.wordpress.android.fluxc.network.rest.wpcom.comment.CommentWPComRestResponse
+import org.wordpress.android.fluxc.network.rest.wpcom.comment.CommentWPComRestResponse.Author
+import org.wordpress.android.fluxc.network.rest.wpcom.comment.CommentWPComRestResponse.Post
 import org.wordpress.android.fluxc.persistence.comments.CommentEntityList
 import org.wordpress.android.fluxc.persistence.comments.CommentsDao.CommentEntity
 import org.wordpress.android.fluxc.utils.DateTimeUtilsWrapper
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
@@ -520,11 +520,19 @@ public class CommentStore extends Store {
         emitChange(event);
     }
 
-    private void fetchComment(RemoteCommentPayload payload) {
+    private void fetchComment(@NonNull RemoteCommentPayload payload) {
         if (payload.site.isUsingWpComRestApi()) {
-            mCommentRestClient.fetchComment(payload.site, payload.remoteCommentId, payload.comment);
+            mCommentRestClient.fetchComment(payload.site, getPrioritizedRemoteCommentId(payload), payload.comment);
         } else {
-            mCommentXMLRPCClient.fetchComment(payload.site, payload.remoteCommentId, payload.comment);
+            mCommentXMLRPCClient.fetchComment(payload.site, getPrioritizedRemoteCommentId(payload), payload.comment);
+        }
+    }
+
+    private long getPrioritizedRemoteCommentId(@NonNull RemoteCommentPayload payload) {
+        if (payload.comment != null) {
+            return payload.comment.getRemoteCommentId();
+        } else {
+            return payload.remoteCommentId;
         }
     }
 
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
@@ -258,8 +258,9 @@ public class CommentStore extends Store {
         return CommentSqlUtils.getCommentsCountForSite(site, statuses);
     }
 
+    @Nullable
     @SuppressWarnings("UnusedReturnValue")
-    public CommentModel getCommentBySiteAndRemoteId(SiteModel site, long remoteCommentId) {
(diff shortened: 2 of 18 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentsStore.kt
```diff
@@ -677,20 +678,25 @@ class CommentsStore @Inject constructor(
...
             // Create a new reply to a specific Comment
             createNewReply(
                     payload.site,
                     commentsMapper.commentLegacyModelToEntity(payload.comment),
                     commentsMapper.commentLegacyModelToEntity(payload.reply)
             )
+        } else {
+            throw IllegalStateException(
+                "Either post or reply must be not null and both can't be not null at the same time!"
+            )
```
wzieba, line 699: This will likely crash client apps in production if this scenario happens - clients don't have possibility to handle this exception. Can we change it to a log and failed `OnCommentChange` event?

fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
```diff
@@ -520,11 +508,19 @@ private void handlePushCommentResponse(RemoteCommentResponsePayload payload) {
...
+    private void fetchComment(@NonNull RemoteCommentPayload payload) {
         if (payload.site.isUsingWpComRestApi()) {
-            mCommentRestClient.fetchComment(payload.site, payload.remoteCommentId, payload.comment);
+            mCommentRestClient.fetchComment(payload.site, getPrioritizedRemoteCommentId(payload), payload.comment);
+        } else {
+            mCommentXMLRPCClient.fetchComment(payload.site, getPrioritizedRemoteCommentId(payload), payload.comment);
+        }
+    }
+
+    private long getPrioritizedRemoteCommentId(@NonNull RemoteCommentPayload payload) {
```
wzieba, line 519: What's the motivation behind this method? Are we sure it's valid to use `remoteCommentId` if `comment` is null? It seems to change a logic of many methods.

wzieba (commented):
I haven't finished review yet, leaving some comments to start discussion though.

fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentsStore.kt
ParaskP7: 👋 @wzieba and thanks for this comment, this is indeed an interesting one, let me explain:

1. First, you can take a look at this commit's description: 53e14a0bde415d92fede132f261b3161f843e846
2. From there, you will notice that I did that because the two `RemoteCreateCommentPayload` constructors are also defined as such and thus the callers should also respect their contract in order to avoid any NPE or an illegal app state. I also chose to keep the `payload.reply == null` condition and add the extra `payload.post == null` condition because the `payload.reply == null` condition was already present.
3. So, all in all, we should never reach this `else` case and thus if we do it is end-up being an illegal state of the app. Instead of trying to hide this state and somehow mute it I think it is better to throw an exception, that is, if and whenever someone changes how the `RemoteCreateCommentPayload` is constructed, because if they don't, as it is now, it will never happen anyway.

Does that makes sense? 🙏

fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentStore.java
ParaskP7: First all all, thanks for this comment @wzieba ! 💯 

As discussed, this was done to avoid duplication on having the same logic with the `REST` and `XMLRPC` clients themselves, and also risking it be on one of the clients and not the other.

FYI:

1. First such commit, which creates this extra method: 9f6c87b5d493ad8813372a5d0b7034d11a0d3774
2. Second such commit, which reused this extra method: 12fe3cfc80ec316ca2211924e0e498ae88722c1d
3. There are move such commits following the same pattern above.

ParaskP7:
👋 @wzieba and thanks for starting the review and any discussion on the changes, looking forward for us to finish the review! 🙇 ❤️

fluxc/src/main/java/org/wordpress/android/fluxc/store/CommentsStore.kt
wzieba: > the two RemoteCreateCommentPayload constructors are also defined as such and thus the callers should also respect their contract in order to avoid any NPE or an illegal app state

Thanks. When writing this comment, I was under the impression that `RemoteCreateCommentPayload` is a mapped response from the API. But then I looked at `WordPress-Android` codebase and understood, that it's a request created via constructor. Let's keep the exception, then!

example/src/test/java/org/wordpress/android/fluxc/common/CommentsMapperTest.kt
```diff
@@ -186,22 +191,22 @@ class CommentsMapperTest {
 
     private fun getDefaultComment(allowNulls: Boolean): CommentEntity {
```
wzieba: `allowNulls` doesn't seem to be valid anymore: how about `allowEmpty` or `withEmpty` because "allow" doesn't seem to fit here as well.

example/src/test/java/org/wordpress/android/fluxc/common/CommentsMapperTest.kt
```diff
@@ -186,22 +191,22 @@ class CommentsMapperTest {
...
+                authorId = 44L,
                 localSiteId = 10_000,
-                remoteSiteId = 100_000,
-                authorUrl = if (allowNulls) null else "https://test-debug-site.wordpress.com",
-                authorName = if (allowNulls) null else "authorname",
-                authorEmail = if (allowNulls) null else "email@wordpress.com",
-                authorProfileImageUrl = if (allowNulls) null else "https://gravatar.com/avatar/111222333",
-                postTitle = if (allowNulls) null else "again",
+                remoteSiteId = 100_000L,
+                authorUrl = if (allowNulls) "" else "https://test-debug-site.wordpress.com",
```
wzieba: What's the motivation behind assigning empty string to these fields instead of `null`? `null` is a valid value for many fields in `CommentEntity`.

ParaskP7: Good question @wzieba ! 👍

Briefly, this is done because otherwise this test suite will fail due to the fact that a some fields on the `CommentWPComRestResponse` are now defined as `@NonNull`. Let me explain:

1. You can try and change `authorUrl = if (withEmpty) "" else "https://test-debug-site.wordpress.com",` to `authorUrl = if (withEmpty) null else "https://test-debug-site.wordpress.com",` and see one test failing.
2. Now, in order to fix that you would need to update `URL = entity.authorUrl ?: ""` to `URL = entity.authorUrl ?: null`.
3. However, you will quickly notice that this is not compiling due to the fact that `Author.URL` is defined as `@NonNull` on `CommentWPComRestResponse`.

Does that make sense? 🙏

example/src/test/java/org/wordpress/android/fluxc/common/CommentsMapperTest.kt
ParaskP7: Thanks for this comment @wzieba ! 👍 

I agree, I am also not liking the currently `allowNulls`. As per your suggestion I have now updated this and another such instance of it with `withEmpty`. 💯 

Done here: 49bded048198442aebcc75ef125955ef8fb83466

wzieba (approved):
Thanks for answering the questions @ParaskP7 , looks good to me! I've tested the change on the FluxC example and Jetpack apps and haven't spotted any crashes/problems.

ParaskP7:
Awesome @wzieba , thanks for reviewing, testing and raising all those questions, you rock! 🙇 ❤️ 🚀

PR #2869, opened 2023-10-12
Enhances the fetch all-domains api call to resolve domain status

antonis:
Part of https://github.com/wordpress-mobile/WordPress-Android/issues/19165

**Based on:** https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2867

This PR adds the `resolve_status` parameter in the all-domains api call as introduced in the updated spec (internal ref: pc8HXX-1gR-p2#comment-1182).
It also converts string date to object in the all-domains response model for easier handling.

Note: The `locale` parameter [is already passed by the client implementation](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/BaseWPComRestClient.java#L30).

**To test:**

1. Start the example app and sign in (note: [setup oauth first](https://github.com/wordpress-mobile/WordPress-FluxC-Android#oauth2-authentication))
2. Tap on the DOMAINS button
3. Tap on the FETCH ALL DOMAINS button
4. Verify that your domains are listed and the `status` is printed for each domain
5. Verify that the domains count is correct

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/AllDomainsResponse.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/AllDomainsResponse.kt
@@ -1,7 +1,12 @@
 package org.wordpress.android.fluxc.network.rest.wpcom.site
 
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
+import com.google.gson.JsonElement
+import com.google.gson.JsonParseException
 import com.google.gson.annotations.JsonAdapter
 import com.google.gson.annotations.SerializedName
+import java.lang.reflect.Type
 
 data class AllDomainsResponse(val domains: List<AllDomainsDomain>)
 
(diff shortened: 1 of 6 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/AllDomainsResponse.kt
```diff
@@ -24,9 +31,11 @@ data class AllDomainsDomain(
     @JsonAdapter(BooleanTypeAdapter::class)
     val hasRegistration: Boolean = false,
     @SerializedName("registration_date")
-    val registrationDate: String? = null,
+    @JsonAdapter(AllDomainsDateAdapter::class)
+    val registrationDate: Date? = null,
     @SerializedName("expiry")
-    val expiry: String? = null,
+    @JsonAdapter(AllDomainsDateAdapter::class)
+    val expiry: Date? = null,
```
mkevins, line 38: One minor question (non-blocking): I had used `LocalDate` in the UI.. do you think it makes sense to use that here as well? Or stick with `Date`?

mkevins (approved):
Thank you for these changes Antonis!

I had one question, but it's non-blocking.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/AllDomainsResponse.kt
antonis: Good suggestion @mkevins 👍 
I had a quick look at this and I think it is not straightforward due to the [minSdkVersion(24)](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/build.gradle#L63) in the FluxC project and LocalDate being available [on api level 26](https://developer.android.com/reference/java/time/LocalDate). Since the conversion seems to be easy [on the WPAndroid side](https://github.com/wordpress-mobile/WordPress-Android/blob/57a414e6b95a5216cdbb32909208daa1a384b921/WordPress/src/main/java/org/wordpress/android/ui/jetpackoverlay/JetpackFeatureRemovalBrandingUtil.kt#L98) would it be ok to keep this as is?
My understanding is that though WPAndroid has the same minSdkVersion [the Desugar library](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/WordPress/build.gradle#L520) [makes the api available](https://developer.android.com/studio/write/java8-support-table).

antonis:
> Thank you for these changes Antonis!

Thank you for reviewing so quickly @mkevins 🙇 

> I had one question, but it's non-blocking.

I'll proceed with merging to unblock the dependent PRs but I'll be happy to iterate on the the suggestion if needed https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2869#discussion_r1357860874

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/AllDomainsResponse.kt
mkevins: This makes sense, thanks for the explanation. 👍

PR #2875, opened 2023-10-18
Update UpdateOrderRequest model

atorresveiga:
Part of https://github.com/woocommerce/woocommerce-android/issues/9541

### Why

### Description
This PR updates the lineItems field inside the UpdateOrderRequest model to use a more flexible model (a map). LineItem -> Map. This change will help us create more flexible LineItems with optional fields without modifying the LineItems model.

### Testing


ThomazFB (approved):
Looks good!

PR #2876, opened 2023-10-19
Configuration key

atorresveiga:
### Why

### Description
This PR updates the allowed line item metadata to cache the bundle item ID. It also adds the getConfigurationKey function to retrieve the lineItem configuration key (bundle item ID in the case of bundle products)

### Testing
It is better to test these changes in the Woo PR https://github.com/woocommerce/woocommerce-android/pull/9984

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
@@ -13,6 +13,7 @@ import org.wordpress.android.fluxc.model.LocalOrRemoteId.RemoteId
 import org.wordpress.android.fluxc.model.WCProductModel.AddOnsMetadataKeys.ADDONS_METADATA_KEY
 import org.wordpress.android.fluxc.model.WCProductVariationModel.ProductVariantOption
 import org.wordpress.android.fluxc.model.addons.RemoteAddonDto
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.product.CoreProductType
 import org.wordpress.android.fluxc.network.utils.getBoolean
 import org.wordpress.android.fluxc.network.utils.getLong
 import org.wordpress.android.fluxc.network.utils.getString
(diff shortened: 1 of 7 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -124,7 +125,17 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
...
 
-    @Suppress("SwallowedException", "TooGenericExceptionCaught") private val WCMetaData.addons
+    val isConfigurable: Boolean
+        get() = when (type) {
+            CoreProductType.BUNDLE.value -> {
+                Gson().fromJson(bundledItems, Array<WCBundledProduct>::class.java)
+                    ?.any { it.isConfigurable() } ?: false
+            }
+            else -> false
+        }
```
ThomazFB, line 135: NP: I wonder if it would be better to set the `isConfigurable` value during the `WCProductModel` creation instead of using a computed property. I say this because creating an entire `Gson` object for every call could be a costly operation. 

I see this is a problem with other computed properties around the WCProductModel, so I understand that fixing this would probably be outside of the scope of this PR.

Also, I'm not sure how often this check will be called, so please ignore it if you believe this will be called only once for every Product.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -124,7 +125,17 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
...
+
+    @Suppress("SwallowedException", "TooGenericExceptionCaught")
+    private val WCMetaData.addons
```
ThomazFB, line 138: One possible alternative for this suppression would be replace the `try / catch` block with a `runCatching {  }.getOrNull()` call. But since this is a code outside of the scope of your changes, feel free to ignore 😛

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/order/LineItem.kt
```diff
@@ -38,4 +38,17 @@ data class LineItem(
...
+        return when {
+            bundledBy.isNullOrBlank().not() -> {
+                (metaData?.first { meta ->
+                    meta.key == WCMetaData.BundleMetadataKeys.BUNDLED_ITEM_ID
+                }?.value.toString())
+                    .toLongOrNull()
+            }
+
+            else -> null
+        }
```
ThomazFB: NP: Using the `takeIf` can simplify your code since it already implies that if the condition is not met, a `null` will be returned, also, this looks like a computed property scenario, as the function is described as a getter:

```suggestion
    val configurationKey
        get() = bundledBy.takeIf { it.isNullOrBlank().not() }
            ?.let { metaData }
            ?.find { it.key == WCMetaData.BundleMetadataKeys.BUNDLED_ITEM_ID }
            ?.value.toString().toLongOrNull()
```

ThomazFB (approved):
Nice work adding the tests and with the overall implementation. I left some comments but they're all optional and non-blocking. Feel free to adopting them or not 😄

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
atorresveiga: Great point, Thomaz! Because this change will involve creating a new version of the DB, I will leave it to a later PR and will merge this code to the feature branch

PR #2877, opened 2023-10-20
[ WC Manual Taxes] Add customerId to OrderEntity

rossanafmenezes:
This PR is part of the effort to fix the behaviour when we try to delete a customer from an order.

Right now we can’t remove customer details from the order in case there are one or more products added.

Context:
p1694438541127429-slack-C025A8VV728
https://github.com/woocommerce/woocommerce-android/pull/9518#issuecomment-1663878435

Further investigation and discussion:

p1697561882534789-slack-CGPNUU63E

This PR does not affect WC at this point.


samiuelson (approved):
LGTM!

PR #2880, opened 2023-10-25
Configurable field

atorresveiga:
### Why

### Description
This PR moves the configuration validation to the DB, so after getting the cached products, we can check if a product is configurable or not.

### Testing


ThomazFB (approved):
Looks good! 👍

PR #2882, opened 2023-10-27
Float deserializer

atorresveiga:
### Why
When getting float values from the API, if the value we are trying to fetch is `null`, the API will return an EMPTY string. This will throw an exception when trying to convert this value (empty string) to Float.

### Description
This PR introduces the `JsonElementToFloatSerializerDeserializer` a more reliable serializer that will try to convert JSON elements to Float or return null for cases it doesn't know how to handle.

### Testing
Unit test should cover all major scenarios


ThomazFB (approved):
Looks good! Nice solution and great unit test coverage!

PR #2883, opened 2023-10-27
Bundle quantity rules support

atorresveiga:
### Why

### Description
This PR brings support to the bundle quantity rules in the App. It fetches the bundle min and max quantity rules and saves them as part of the product metadata.

### Description

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponse.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponse.kt
@@ -47,28 +47,28 @@ data class ProductApiResponse(
     val tax_class: String? = null,
     val manage_stock: String? = null,
     @JsonAdapter(NonNegativeDoubleJsonDeserializer::class)
-    val stock_quantity:Double? = 0.0,
+    val stock_quantity: Double? = 0.0,
     val stock_status: String? = null,
     val date_on_sale_from: String? = null,
     val date_on_sale_to: String? = null,
     val date_on_sale_from_gmt: String? = null,
     val date_on_sale_to_gmt: String? = null,
     val backorders: String? = null,
-    val backorders_allowed:Boolean = false,
-    val backordered:Boolean = false,
+    val backorders_allowed: Boolean = false,
+    val backordered: Boolean = false,
     @JsonAdapter(PrimitiveBooleanJsonDeserializer::class)
-    val sold_individually:Boolean = false,
+    val sold_individually: Boolean = false,
     val weight: String? = null,
     val dimensions: JsonElement? = null,
     val shipping_required: Boolean = false,
-    val shipping_taxable:Boolean = false,
+    val shipping_taxable: Boolean = false,
     val shipping_class: String? = null,
-    val shipping_class_id:Int = 0,
-    val reviews_allowed:Boolean = true,
+    val shipping_class_id: Int = 0,
+    val reviews_allowed: Boolean = true,
     val average_rating: String? = null,
-    val rating_count:Int = 0,
-    val parent_id:Long = 0L,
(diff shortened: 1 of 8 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponse.kt
```diff
@@ -169,7 +174,26 @@ data class ProductApiResponse(
...
+                response.bundle_max_size?.let { value ->
+                    WCMetaData.addAsMetadata(response.metadata, BUNDLE_MAX_SIZE, value)
+                }
+                response.bundle_min_size?.let { value ->
+                    WCMetaData.addAsMetadata(response.metadata, BUNDLE_MIN_SIZE, value)
+                }
+                response.metadata.toString()
+            } else {
+                response.metadata?.toString() ?: ""
+            }
```
ThomazFB: np: It seems we're adding multiple checks over the bundle rules. I also see one at the top of the `asProductModel` call. Maybe we could benefit from a code extraction to a function in this response class to make it easier to concentrate on all the bundle behaviors.

ThomazFB (approved):
Tested alongside the changes of https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2882. Looks good! I only left a comment over code organization as the `asProductModel` is getting bigger and more complex.

atorresveiga:
> Maybe we could benefit from a code extraction to a function in this response class to make it easier to concentrate on all the bundle behaviors.

Fixed here https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2883/commits/7c508d2c7b1df58f61fc35add413275c2aade818 @ThomazFB. Please let me know WDYT?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponse.kt
```diff
@@ -169,15 +170,40 @@ data class ProductApiResponse(
...
             response.dimensions?.asJsonObject?.let { json ->
                 length = json.getString("length") ?: ""
                 width = json.getString("width") ?: ""
                 height = json.getString("height") ?: ""
             }
         }
+        return applyBundledProductChanges(model)
+    }
+
+    private fun applyBundledProductChanges(model: WCProductModel): WCProductModel {
```
hichamboushaba, line 186: @atorresveiga I'm working on some [refactorings](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3070) to allow extracting the product `metadata` to a separate table, and while doing so I stumbled upon this logic that you added to add the properties regarding `bundle` to the `meta_data` Json instead of exposing them as regular properties in the `WCProductModel`.

Can I ask what was the reason for choosing this path? This was causing the `WCMetaData` object to be returned with null `id`s, Gson allowed this as it doesn't support Kotlin's nullability, but it's still problematic. 

With the upcoming PR, I'm intending to remove the `meta_data` column completely from the `WCProductModel`, and I will need to make changes to this logic too, is it OK if I make these keys as regular properties of the model instead of the current logic?

atorresveiga: Hi Hicham, thanks for the ping 

> Can I ask what was the reason for choosing this path? 

I chose this path with the goal in mind of keeping the product model small.

 > is it OK if I make these keys as regular properties of the model instead of the current logic?

Yes, feel free to make those regular properties of the product model 👍

hichamboushaba: Great, thanks for the clarifications @atorresveiga, I'll make those changes.

PR #2885, opened 2023-10-31
Payments overview all endpoint

kidinov:
The PR adds `payments/deposits/overview-all` end point

https://github.com/woocommerce/woocommerce-android/issues/10064


samiuelson (approved):
LGTM

PR #2888, opened 2023-11-01
Adds new field target_urn to BlazeCampaign response model

JorgeMucientes:
Adds a new field `target_urn` to Blaze Campaign response object and DB entity. This new field contains the `productId` associated to a Blaze campaign. That value will enable us hide the Blaze CTA in product detail screen, when the product is already promoted with Blaze. 

Changes to be tested in this PR: https://github.com/woocommerce/woocommerce-android/pull/10079

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/blaze/BlazeCampaignsModel.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/blaze/BlazeCampaignsModel.kt
@@ -19,5 +19,5 @@ data class BlazeCampaignModel(
     val budgetCents: Long,
     val impressions: Long,
     val clicks: Long,
-    val targetUrn: String,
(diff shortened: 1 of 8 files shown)


0nko (approved):
LGTM! :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/model/blaze/BlazeCampaignsModel.kt
```diff
@@ -18,5 +18,6 @@ data class BlazeCampaignModel(
     val uiStatus: String,
     val budgetCents: Long,
     val impressions: Long,
-    val clicks: Long
+    val clicks: Long,
+    val targetUrn: String,
```
0nko: Should this be nullable, as well?

0nko (approved):
Nice work!

PR #2889, opened 2023-11-01
Bundle variable support

atorresveiga:
### Why

### Description
This PR brings support to the bundle variable rules in the App. It fetches the default_variation_attributes and allowed_variations fields from the metadata and update the need configuration function to take variable products into account.

### Description


ThomazFB (approved):
Hey @atorresveiga! 👋 Is the Woo counterpart of this PR already available? 

I didn't find it in my PR requests, so I reviewed only the code introduced here. Since it's a small change, I didn't detect any problem. I believe I can test the working scenario directly through the Woo part.

PR #2895, opened 2023-11-03
Set a default value of false for is_single_user_site

zwarm:
This PR is an attempt to figure out why accessing `site.isSingleUserSite` returns a null. This may not be the reason why, but the only thing I can think of is that old data in the DB is left as null or the `Boolean` wrapper in SiteModel can have three values: true, false, or null.

Also gson will set boolean as false if the field is not in the response or if the field in the response is null. 🤷 

wdyt?


zwarm:
Closing this PR; as it's a nice to have update and not a necessity. If future time permits, I will return to this. FYO @AjeshRPai

PR #2902, opened 2023-11-14
Update Product rest client to sync metada values on product updates

JorgeMucientes:
Part of https://github.com/woocommerce/woocommerce-android/issues/10124. Update `ProductRestClient` to send any `metada` fields as a json array whenever any of these values change. 
Changes can be tested following the instructions from this PR https://github.com/woocommerce/woocommerce-android/pull/10171

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
@@ -1909,6 +1909,15 @@ class ProductRestClient @Inject constructor(
                 }
             }
         }
+        if (storedWCProductModel.metadata != updatedProductModel.metadata) {
+            JsonParser().apply {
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductRestClient.kt
```diff
@@ -1909,6 +1909,15 @@ class ProductRestClient @Inject constructor(
                 }
             }
         }
+        if (storedWCProductModel.metadata != updatedProductModel.metadata) {
+            JsonParser().apply {
+                body["meta_data"] = try {
+                    parse(updatedProductModel.metadata).asJsonArray
+                } catch (ex: Exception) {
+                    JsonArray()
```
hichamboushaba, line 1918: I think it would be great to log an error here using AppLog, this way, if a user faces an issue then reaches out to support, we'll have something to help investigate.

hichamboushaba (approved):
Nice work, it works well @JorgeMucientes, I left one suggestion, please check it out.

PR #2903, opened 2023-11-17
Add logic to persist the subscription's payment sync date metadata

hichamboushaba:
This PR adds the key `_subscription_payment_sync_date` to the list of keys we persist and expose when parsing product and variation details.

This is needed for https://github.com/woocommerce/woocommerce-android/pull/10208

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductVariationModel.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductVariationModel.kt
@@ -150,27 +150,6 @@ data class WCProductVariationModel(@PrimaryKey @Column private var id: Int = 0) 
         return gson.fromJson(attributes, responseType) as? List<ProductVariantOption> ?: emptyList()
     }
 
-    object SubscriptionMetadataKeys {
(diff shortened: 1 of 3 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductVariationModel.kt
```diff
@@ -150,27 +150,6 @@ data class WCProductVariationModel(@PrimaryKey @Column private var id: Int = 0)
         return gson.fromJson(attributes, responseType) as? List<ProductVariantOption> ?: emptyList()
     }
 
-    object SubscriptionMetadataKeys {
```
hichamboushaba, line 153: Having two objects with the same data was confusing first, and caused me a waste of time while investigating why they key I added wasn't working in WCAndroid, so I opted to remove this one, and keep using just the `WCProductModel.SubscriptionMetadataKeys` for variations as well, let me know if you don't think this was the right call.

JorgeMucientes: Makes sense thanks for the change!

JorgeMucientes (approved):

PR #2907, opened 2023-11-24
Improvements to the WPCom themes fetching endpoint

hichamboushaba:
⚠️ this has a breaking change for WPAndroid, please don't merge, I'll synchronize merging when this and the accompanying WPAndroid PR https://github.com/wordpress-mobile/WordPress-Android/pull/19668 is approved.

------ 
This PR adds some improvements that WCAndroid needs to implement theme selection support for WooExpress sites (https://github.com/woocommerce/woocommerce-android/issues/10224):

1. Allows passing a `filter` query to the fetching endpoint.
2. Allows customizing the results limit of the fetching endpoint.
3. Adds a function that allows querying themes from the DB using a list of IDs (will be useful for us given the hardcoded list of themes we will support)
4. Fix theme installation on atomic sites: it seems theme installation was broken in the Jetpack app for quite a while, because we try to add a suffix `-wpcom` to the theme's ID, while we should do it only for self-hosted Jetpack sites, Calypso does a similar thing [here](https://github.com/Automattic/wp-calypso/blob/5d2ed9937b86f0b840f9392fe1577d3d026ebcc6/client/state/themes/actions/suffix-theme-id-for-install.js#L17-L19), the commit 89d2c38 fixes this.

#### Testing
1. Open the example app, and sign in.
2. Click on Themes.
3. Click on Fetch WP.Com themes.
7. Test different scenarios, and confirm they work as expected.
8. Pick an atomic site.
9. Type the ID of a theme.
10. Click on Install WP.COM theme on Jetpack Site
11. Confirm it works.

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ThemeTestJetpack.java
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/release/ReleaseStack_ThemeTestJetpack.java
@@ -23,6 +23,7 @@ import org.wordpress.android.fluxc.store.SiteStore.FetchSitesPayload;
 import org.wordpress.android.fluxc.store.SiteStore.OnSiteChanged;
 import org.wordpress.android.fluxc.store.SiteStore.OnSiteRemoved;
 import org.wordpress.android.fluxc.store.ThemeStore;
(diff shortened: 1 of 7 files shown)


0nko (approved):
LGMT 👍

PR #2886, opened 2023-10-31
[Nullability Annotations to Java Classes] Add Missing Nullability Annotations to `Media` Model Classes (`breaking`)

ParaskP7:
Parent #2799
Closes #2832

This PR adds any missing nullability annotation to [MediaModel.java](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/7fd7840a9bd900fde8e32ad8bd5632ad8f05dff3/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java#L4) model and all related store & sql-util classes.

FYI: This change is `breaking`, meaning that any client that depended on the [MediaModel.java](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/7fd7840a9bd900fde8e32ad8bd5632ad8f05dff3/fluxc/src/main/java/org/wordpress/android/fluxc/model/MediaModel.java#L4) model should update to the new APIs (`non-default constructors`) as the existing API (`default constructor`) are now deprecated. As such, this change is inherently `risky`, meaning that there are compile-time changes associated with this change and thus needs testing to verify correctness, both on the library's and client's side.

PS: Any other changes on any other class are just some propagating missing nullability annotation changes, which stem from the main set of classes being updated, this PR's focus, plus, some other extra minor analysis, refactor and test related changes.

-----

PS.1: @AjeshRPai I added you as the main reviewer, not so randomly (it being a continuation of #2878), since I just wanted someone from the Jetpack/WordPress mobile team to be aware of and sign-off on that change. Feel free to merge this PR directly yourself if you deem so.

PS.2: @0nko I am also adding you as an extra reviewer, not so randomly (it being a continuation of #2878), since I want someone from the WooCommerce mobile team to be aware of this change as well. However, there is no need for you to review it as well, thus duplicating the review work. Having one team member from either of the mobile teams reviewing and testing this change is enough for this change.

-----

Nullability Annotations List (`Model`):

1. [Add missing final keyword to field name on media fields enum](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2886/commits/07c5b4a66c357f1575a81ee383dad1572b94ece0)
2. [Add missing nl-a to equals parameter on media model](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2886/commits/71cd024084657b7b22875f6c85eb38376d82a450)
[...]

PR #2908, opened 2023-11-27
Avoid NPE when source_url is missing from `/wp/v2/media` response

hichamboushaba:
The changes introduced in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2886 are causing a crash in WCAndroid (https://github.com/woocommerce/woocommerce-android/issues/10264): we changed the `url` property of `MediaModel` to be non-null, but it seems that the API still returns null values (or the property is missing from the response).

This PR aligns the behavior in this case with what we [do](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/63d05c6137b822ad5767bfe34ecfb8108a9a2b83/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/media/MediaXMLRPCClient.java#L545) for XMLPRC, we default to an empty String when it's null, and it's up to the clients to handle the error in this case.


hichamboushaba:
@0nko @ParaskP7 a friendly ping about this PR, it's a small change that won't take much of your time 🙏

0nko:
I assumed @ParaskP7 might want to review this since he worked on the nullability efforts but I'll go ahead with the review. Sorry for the wait @hichamboushaba.

0nko (approved):
Thanks for fixing this! :shipit:

PR #2910, opened 2023-11-28
Feature/10265 fetch theme demo pages

JorgeMucientes:
**Closes:** [#10265](https://github.com/woocommerce/woocommerce-android/issues/10265)

This PR adds an endpoint to fetch all pages given a theme demo URL. 

 ## Testing
The endpoint doesn't require any authentication.  So just run the sample appl, tap on `Themes` -> `Fetch demo pages for theme`. Check a list of pages for the hardcoded theme URL `https://zainodemo.wpcomstaging.com` is returned.


hafizrahman (approved):
Nice work, works as expected @JorgeMucientes 

Looking [at the documentation](https://developer.wordpress.org/rest-api/reference/pages/#list-pages), the `per_page` argument defaults to only 10 pages. I'm not sure if theme demos generally have less or more than that, but it seems safe to assume that more complex themes might have quite a bit of demo pages.

I'm not seeing that handled here, do you think we should allow for more pages to be requested?

Approving in advance but would love to hear your thoughts :)

JorgeMucientes:
That is actually a excellent point @hafizrahman! Thanks for sharing. I completely missed the `per_page` param existence. I've added it to the request setting the default value to 30. I think 30 should be enough for demo site.

PR #2915, opened 2023-11-29
Fix: Add missing customer id migration

ThomazFB:
Summary
==========
Fix a missing room migration to the `OrderEntity` class regarding the `customerId` field. Causing subsequent database migration to crash due to the absence of field declaration.

Since this contains a new schema file, the diff looks huge, but the PR is quite small actually.

How to Test
==========
1. Integrate this PR with the Woo app and run it on top of a `trunk` installation to properly trigger the migration. Make sure the app runs as expected with no crashes during the startup.

Update release notes:


rossanafmenezes (approved):
LGTM

Thanks for fixing it.

PR #2925, opened 2023-12-13
Fix: site.username NullPointerException

zwarm:
This PR prevents a crash in `NonceRestClient` when `site.username` is null. 

- The`Nonce` interface was updated to allow for nulls in username
- In `NonceRestClient`, a null check gate was added to `requestNonce` and when found, will return an `Unknown` Nonce instance.

Testing instructions can be found in the WP companion PR (Coming soon)

**Merge Instructions**
- On success testing of WP PR
- Remove the Not Ready For Merge label
- Merge as normal


AjeshRPai (approved):
Hey @zwarm 
I have tested this PR with the Companion WP android PR and everything is working as expected. Nice work. Approving and merging this PR.

PR #2931, opened 2023-12-22
Add exception catch block to Gson BundledProduct parsing operation

ThomazFB:
Summary
==========
Fix issue https://github.com/woocommerce/woocommerce-android/issues/10449 by adding an exception recovery, where if it's not possible to acquire the Bundled Products info, the `isConfigurable` is set to `false` immediately. 

How to Test
==========
1. Add this FluxC version to the Woo app and verify that all Bundled Product features in the Order Creation flow still work as expected.

Update release notes:

### DIFF
--- a/plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/model/WCProductModelTest.kt
+++ b/plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/model/WCProductModelTest.kt
@@ -5,6 +5,7 @@ import com.google.gson.JsonObject
 import org.assertj.core.api.Assertions
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/test/java/org/wordpress/android/fluxc/model/WCProductModelTest.kt
```diff
@@ -39,4 +40,15 @@ class WCProductModelTest {
 
         Assertions.assertThat(result).isNotEmpty
     }
+
+    @Test
+    fun `isConfigurable should return false when bundledItems is malformed JSON`() {
```
atorresveiga, line 46: Nice!

atorresveiga (approved):
Tested alongside the lates version of the Woo app.

Everything is working as expected.

LGTM! :shipit:

PR #2932, opened 2023-12-26
Add date type parameter

### LINKED ISSUE #10455: Dashboard: fetch revenue analytics that are consistent with products reports
Corresponding iOS issue: https://github.com/woocommerce/woocommerce-ios/issues/11496

More details and example screenshots in pe5pgL-49w-p2
Core changes in pe5pgL-49w-p2#update

**Describe the bug**

As illustrated in pe5pgL-49w-p2, the revenue analytics are currently based on `date_paid` by default (this can be changed in the core analytics settings), while products analytics are always based on `date_created` because they get data from two different database tables. When the date field base is different, the analytics can show non-zero stats for the revenue but zero or mismatching products in the top performers section in the dashboard and analytics hub.

The plan is to add a parameter to the revenue stats endpoint so that the analytics are always based on the order creation date for revenue and products.

**To Reproduce**

Prerequisite: in the store has analytics settings `wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Fsettings`, `Date type` option has `Date paid`. The store has at least one order with at least one product, created on a previous day, but is pending and not completed.

- Launch the app, log in to the store if needed
- Go to the orders tab, tap on the order in the prerequisite (created on a previous day and pending), then fulfill it like `Collect Payment` > `Cash` > `Mark as paid`
- Go to the my store tab, pull down to refresh (if the new order isn't shown, try refreshing the data after like 30-60 seconds) --> in the dashboard, notice the fulfilled order is reflected in the revenue stats, but not in the top performing products section. Same is in the analytics tab

**Screenshots**

dashboard - revenue | dashboard - products  | analytics hub
-- | -- | --
 |  | 

**Expected behavior**

The revenue stats should be consistent with the product stats.

- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).

**Mobile Environment**
Please include:
- Device: any
- iOS version: any
- WooCommerce iOS version: any, the latest being 16.6

atorresveiga:
Fixes https://github.com/woocommerce/woocommerce-android/issues/10455

## Why

As described in pe5pgL-49w-p2, because the app displays the revenue and product stats on the same analytics screen, when the revenue stats are fetched with the WC date type option from the analytics settings (`date_paid` by default) that's different from the date column (`date_created`) used in product stats, the data could look inconsistent for orders completed on a different date from the creation date. As a relatively simpler solution, a new parameter `date_type` was added to the revenue stats endpoint in core https://github.com/woocommerce/woocommerce/pull/42938 that we can set to `date_created` so that the revenue stats can always be based on the same date column as product stats.

## How

In `OrderStatsRestClient.fetchRevenueStats`, a new parameter `dateType="date_created"` was added. This new parameter is backward compatible, so stores without the parameter support just won't have the date type column overridden on the API. The core PR isn't merged yet but the main spec is pretty settled.

## Testing instructions

Hand-testing is optional, as the parameter doesn't have any effect in existing stores right now.
You can test the changes here using this Woo PR https://github.com/woocommerce/woocommerce-android/pull/10494

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/stats/WCStatsStoreTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/stats/WCStatsStoreTest.kt
@@ -1249,7 +1249,7 @@ class WCStatsStoreTest {
     @Test
     fun testFetchCurrentDayRevenueStatsDate() = runBlocking {
         val plus12SiteDate = SiteModel().apply { timezone = "12" }.let {
-            whenever(mockOrderStatsRestClient.fetchRevenueStats(any(), any(), any(), any(), any(), any(), any())
+            whenever(mockOrderStatsRestClient.fetchRevenueStats(any(), any(), any(), any(), any(), any(), any(), any())
             ).thenReturn(FetchRevenueStatsResponsePayload(it, DAYS, WCRevenueStatsModel()))
             val startDate = DateUtils.getStartDateForSite(it, DateUtils.formatDate("yyyy-MM-dd'T'00:00:00", Date()))
             val payload = FetchRevenueStatsPayload(it, StatsGranularity.DAYS, startDate)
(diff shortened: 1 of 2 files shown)


hafizrahman (approved):
LGTM 👍🏼 Thanks for making the comments more complete as well.

PR #2933, opened 2023-12-27
Blaze/mock target endpoints

hichamboushaba:
This PR adds the following changes:

1. Domain models for the Blaze targeting endpoints.
2. Caching logic for the supported types.
3. Mocked rest client to return some fake data for the tests.
4. Support for localization of the endpoints according to this peeHDf-2fw-p2#comment-1621

There is nothing to test for now, just code review and having green CI is enough.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/model/blaze/BlazeTargetingLocation.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/model/blaze/BlazeTargetingLocation.kt
@@ -1,0 +1,8 @@
+package org.wordpress.android.fluxc.model.blaze
+
+data class BlazeTargetingLocation(
+    val id: Long,
+    val name: String,
+    val type: String,
+    val parent: BlazeTargetingLocation?
+)
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/FakeBlazeTargetingRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/FakeBlazeTargetingRestClient.kt
@@ -12,6 +12,7 @@ import javax.inject.Inject
 private const val FAKE_DELAY = 500L
 
 class FakeBlazeTargetingRestClient @Inject constructor() {
+    @Suppress("MagicNumber")
     suspend fun fetchBlazeLocations(
         query: String
     ): BlazeTargetingPayload<List<BlazeTargetingLocation>> {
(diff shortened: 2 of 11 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/FakeBlazeTargetingRestClient.kt
```diff
@@ -0,0 +1,296 @@
...
+import org.wordpress.android.fluxc.model.blaze.BlazeTargetingLocation
+import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComGsonNetworkError
+import org.wordpress.android.fluxc.persistence.blaze.BlazeTargetingDeviceEntity
+import org.wordpress.android.fluxc.persistence.blaze.BlazeTargetingLanguageEntity
+import org.wordpress.android.fluxc.persistence.blaze.BlazeTargetingTopicEntity
+import javax.inject.Inject
+
+private const val FAKE_DELAY = 500L
+
+class FakeBlazeTargetingRestClient @Inject constructor() {
```
hichamboushaba, line 14: I put the mocks in this separate class, later we can decide whether to keep a separate class for those endpoints, or to move them to the existing class.

fluxc/src/main/java/org/wordpress/android/fluxc/model/blaze/BlazeTargetingLocation.kt
```diff
@@ -0,0 +1,8 @@
+package org.wordpress.android.fluxc.model.blaze
+
+data class BlazeTargetingLocation(
+    val id: Long,
+    val name: String,
+    val type: String,
+    val parent: BlazeTargetingLocation?
+)
```
hichamboushaba, line 8: I skipped some fields that I don't think we'll need for the mobile implementation, and since we are not caching results, adding them would be pretty easy without any DB migration.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/FakeBlazeTargetingRestClient.kt
```diff
@@ -0,0 +1,296 @@
...
+    @Suppress("MagicNumber", "UNUSED_PARAMETER")
+    suspend fun fetchBlazeLocations(
+        query: String,
+        locale: String
+    ): BlazeTargetingPayload<List<BlazeTargetingLocation>> {
+        if (query.length < 3) {
```
hafizrahman, line 20: given that this is a requirement from the server that a query needs to be at least 3 char, can we put a comment here to explain, or perhaps add a unit test?

hichamboushaba: Good idea, but for now this is just a temporary mocking class, we can add it to the real implementation.

hafizrahman (approved):
This looks good to me and matches the requirements including about the localization.

PR #2938, opened 2024-01-05
Fix locale issue for v3 endpoints keeping previous fallback to locale

RenanLukas:
The same as https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2937 but keeping the previous fallback to `locale` instead of `_locale`

Context p1704484851389609/1704423861.366279-slack-C0180B5PRJ4

Thanks @zwarm for thinking of this implementation


zwarm (approved):
@RenanLukas - All is working as expected. I pulled the branch down and tested using the WP companion PR.

zwarm:
> Thanks @zwarm for thinking of this implementation

It will suffice for now, but we should come up with something more concise at a later date. 😉

PR #2941, opened 2024-01-09
Implement unit tests for BaseWPComRestClient locale query parameter decision

RenanLukas:
Add unit tests to the logic in `BaseWPComRestClient` that decides whether to use `locale` or `_locale` query parameter based on the request URL.


zwarm (approved):
@RenanLukas - The unit test works as expected. Thanks for adding it 👍

PR #2944, opened 2024-01-11
Woo: Removed fields that are no longer returned by the API from "deposits/overview-all"

kidinov:
To address changes in the API the PR reviews the fields that are no longer returned.

```
GET deposits/overview-all
Response deposit.next_scheduled should no longer be returned.
Response balance.instant[].transaction_ids should no longer be returned.
Response balance.pending[].deposits_count should no longer be returned.
```

https://github.com/Automattic/woocommerce-payments-server/issues/4296

### DIFF
--- a/plugins/woocommerce/schemas/org.wordpress.android.fluxc.persistence.WCAndroidDatabase/33.json
+++ b/plugins/woocommerce/schemas/org.wordpress.android.fluxc.persistence.WCAndroidDatabase/33.json
@@ -1,0 +1,1615 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 33,
+    "identityHash": "43fd0900518251fbd271b5ffe7a0d89f",
+    "entities": [
+      {
+        "tableName": "AddonEntity",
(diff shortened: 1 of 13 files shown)


plugins/woocommerce/schemas/org.wordpress.android.fluxc.persistence.WCAndroidDatabase/33.json
```diff
@@ -0,0 +1,1615 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 33,
```
backwardstruck, line 4: Is there some documentation about how to do this DB version update in case I need to do the same thing in the future?

backwardstruck (commented):
As far as I can tell, this is all correct. Just had one question for my reference.

plugins/woocommerce/schemas/org.wordpress.android.fluxc.persistence.WCAndroidDatabase/33.json
kidinov: Yep, those are just schemas of the specific version of the databases. You can take a look there -  

https://developer.android.com/training/data-storage/room/migrating-db-versions#export-schemas

kidinov: @backwardstruck if that looks good - please approve the PR so it can be merged

backwardstruck (approved):
Looks good!

PR #2949, opened 2024-01-28
[Credential Manager] Add raw challenge JSON data

### LINKED ISSUE #10557: Implement Passkey fetching through the Credential Manager API
# Summary
We discovered that a new version 1.2.0 of the Credential Manager was released after the Passkey project development, containing lots of async functions counterparts within the API with it. E.g.:

> The [getCredential](https://href.li/?https://developer.android.com/reference/androidx/credentials/CredentialManager#getCredential(android.content.Context,androidx.credentials.PrepareGetCredentialResponse.PendingGetCredentialHandle)) function, which is a suspend function used to retrieve a Passkey data from a Challenge, is now followed by a [getCredentialAsync](https://href.li/?https://developer.android.com/reference/androidx/credentials/CredentialManager#getCredentialAsync(android.content.Context,androidx.credentials.PrepareGetCredentialResponse.PendingGetCredentialHandle,android.os.CancellationSignal,java.util.concurrent.Executor,androidx.credentials.CredentialManagerCallback)) function, which is not suspend and relies on a callback instead.

This seems to be a very intentional move from Google to support Java usage with the Credential Manager, and it might be a very good alternative to avoid a drastic change inside the WordPress Login library. We can verify in the docs statements like This API uses callbacks instead of Kotlin coroutines.

Our goal here is to ensure we're adopting the Credential Manager through this callback structure since the WordPress Login library still doesn't have proper support for the Kotlin Coroutines usage.

```[tasklist]
### Tasks
- [x] Introduce basic Credential Manager operation with callback structure
- [x] Connect Credential Manager handler usage with Passkey auth flow
- [ ] Ensure the Credential Manager is only called when API 34 or higher is available
```

ThomazFB:
⚠️⚠️⚠️ Merge only when https://github.com/wordpress-mobile/WordPress-Login-Flow-Android/pull/135 is approved.

# Summary
Small adjustment to allow the Webauthn payload to contain the entire JSON data instead of the parsed values. The Credential Manager API requires the raw JSON, while the FIDO2 API requires only some fields, so this PR adjusts the payload to contain only the JSON file, as we're moving away from FIDO2 to support only the Credential Manager.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/auth/Authenticator.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/auth/Authenticator.java
@@ -24,7 +24,6 @@ import org.wordpress.android.fluxc.generated.endpoint.WPCOMREST;
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest;
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComErrorListener;
 import org.wordpress.android.fluxc.network.rest.wpcom.WPComGsonRequest.WPComGsonNetworkError;
-import org.wordpress.android.fluxc.network.rest.wpcom.auth.webauthn.WebauthnChallengeInfo;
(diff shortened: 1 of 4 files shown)


zwarm (approved):
LGTM @ThomazFB.  All works as expected

PR #2957, opened 2024-02-16
Add DB timestamp in PostModel

pantstamp:
This PR adds a new field to the PostModel table in the database that represents the timestamp of the moment that this post was inserted or updated in our local database. 

This field will be exposed by the PostImmutableModel.dbTimestamp property. 

We need this so we can determine if a post object in our database needs to be refetched so we can avoid conflicts on post editing.


zwarm (approved):
@pantstamp - LGTM. I tested using the associated WP PR. All is working as expected. I am approving, but will leave the merging and hash updating for you. Thanks.

PR #2959, opened 2024-02-21
Refresh ListStore component on database updates

malinajirka:
**PR Not Ready For Merge** - It is ready for review though. We'll want to merge it when we merge the Tablet M2 feature branch into trunk in WCAndroid.

~This is still a draft since I haven't tested the changes yet - I'm seeking an early feedback and plan to test the changes after CI publishes this version.~

This PR updates OrderUpdateStore to ensure ListStore component is informed about changes to OrderModel and can refresh the list to display the up-to-date data.

cc @oguzkocer Would appreciate if you could check if this makes sense to you. One of my first PRs after a couple years and tbh I don't remember the ListStore component in such a detail to properly evaluate if this is something we should do. But I feel like, this is a valid use-case => we inform the component about a change we make to orders in the database. [I tried to mirror the approach from WCOrderStore](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/a0ca8b62889fadca9961b44f4637d85d8a99e8f5/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt#L864). Wdyt?

**Update 2/22/2024**

This PR is ready for review now. It consolidates updates of the ListStore component for OrderList in WooAndroid. Whenever an item is modified in the database an update is emitted and the list reloaded. This ensures the UI is always up to date with the state of the database and improves the UX by making the updates smoother and timely.

I recommend reviewing this by commits - this should simplify the review as you can better follow my train of thoughts.

To Test:
- Unit test on CI
- Other tests will be done in https://github.com/woocommerce/woocommerce-android/pull/10888


malinajirka:
I'm exploring an option to instead create a decorator around OrderDao object and emit the event there => this will increase chances the updates are automatically propagated during future updates.

oguzkocer:
@malinajirka I think it's perfectly fine - or even encouraged - to use the `newListDataInvalidatedAction` & `newListRequiresRefreshAction` actions whenever the relevant data is updated. The difference between the two is that, `newListDataInvalidatedAction` will trigger the UI to reload the existing data - for example when the contents of a post is updated - whereas `newListRequiresRefreshAction` will trigger the list to be re-fetched from the server, such as when a post is deleted and shouldn't be shown anymore in the list.

In that context, I think the more appropriate action to use in `fetchSingleOrder` & `updateOrderStatus` would be to use `newListDataInvalidatedAction` because the list itself is unchanged, but the contents need to be reloaded. Although with `updateOrderStatus`, `newListRequiresRefreshAction` might still be the correct choice if filters are involved. That's because the orders in any given list might need to be updated as well - either an existing one would be removed or a new one would be added.

Hope that helps!

malinajirka:
Thanks @oguzkocer! 

> The difference between the two is that, newListDataInvalidatedAction will trigger the UI to reload the existing data - for example when the contents of a post is updated - whereas newListRequiresRefreshAction will trigger the list to be re-fetched from the server, such as when a post is deleted and shouldn't be shown anymore in the list.

I spotted this when I realized, I got into an infinite cycle of fetches from remote 😅. Glad you confirmed that my current understanding is correct!

> Although with updateOrderStatus, newListRequiresRefreshAction might still be the correct choice if filters are involved.

I didn't think of this scenario!

Thank you for your input, it's very helpful and much appreciated! 🙇

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/dao/OrdersDao.kt
```diff
@@ -10,6 +10,9 @@ import org.wordpress.android.fluxc.model.LocalOrRemoteId.LocalId
 import org.wordpress.android.fluxc.model.OrderEntity
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.CoreOrderStatus
 
+/***
+ * AVOID USING THIS CLASS DIRECTLY - Use OrdersDaoDecorator to ensure the ListStore component keeps receiving events.
+ */
```
samiuelson:
```suggestion
/**
 * ⚠️AVOID USING THIS CLASS DIRECTLY -
 * Use [OrdersDaoDecorator] to ensure the [org.wordpress.android.fluxc.store.ListStore] component keeps receiving events.
 */
```
💡 kdoc's nitpick: by putting a class name in square brackets we can make it cmd-clickable. If the class is in a different package we need to provide a full path: 
Screenshot 2024-02-22 at 10 08 44

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/InsertOrder.kt
```diff
@@ -1,21 +1,26 @@
...
 
 class InsertOrder @Inject internal constructor(
-    private val ordersDao: OrdersDao,
+    private val ordersDaoDecorator: OrdersDaoDecorator,
     private val ordersMetaDataDao: OrderMetaDataDao,
     private val transactionExecutor: TransactionExecutor
 ) {
-    suspend operator fun invoke(vararg ordersPack: Pair<OrderEntity, List<OrderMetaDataEntity>>) {
+    suspend operator fun invoke(
+        localSiteId: LocalOrRemoteId.LocalId,
```
samiuelson, line 21:
```suggestion
```
❓ it seems to be unused now

samiuelson (approved):
Thanks for working on this @malinajirka, that's a great improvement. I left a few minor comments. Feel free to ignore.

PR #2961, opened 2024-02-27
Handles Default User-Agent Failure

antonis:
Fixes Partially https://github.com/wordpress-mobile/WordPress-Android/issues/20147#issuecomment-1961238187

**`WordPress-Android` PR:** https://github.com/wordpress-mobile/WordPress-Android/pull/20296

## Description
This PR converts the `UserAgent` class to Kotlin and handles the exception thrown when the default user agent cannot be retrieved preventing a runtime crash.
(internal ref: pcdRpT-5Us-p2)

Note: A similar implementation is employed in [the WordPress app initialiser](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/WordPress/src/main/java/org/wordpress/android/AppInitializer.kt#L1105)

## To Test
To test the app behaviour in the exception case you can use this [fallback.patch](https://github.com/wordpress-mobile/WordPress-FluxC-Android/files/14418192/fallback.patch)


fluxc/src/main/java/org/wordpress/android/fluxc/network/UserAgent.kt
```diff
@@ -0,0 +1,34 @@
+package org.wordpress.android.fluxc.network
+
+import android.content.Context
+import android.webkit.WebSettings
+import org.wordpress.android.util.PackageUtils
+
+@SuppressWarnings("SwallowedException", "TooGenericExceptionCaught", "MemberNameEqualsClassName")
+class UserAgent(appContext: Context?, appName: String) {
+    val userAgent: String
```
antonis, line 9: Leaving the member name equal with the class name to avoid changes in other files and possible breakages

fluxc/src/test/java/org/wordpress/android/fluxc/network/UserAgentTest.kt
```diff
@@ -0,0 +1,45 @@
...
+private const val APP_VERSION = "1.0"
+
+@RunWith(RobolectricTestRunner::class)
+class UserAgentTest {
+    private val context = RuntimeEnvironment.application.applicationContext
+
+    @Test
+    fun testUserAgent() = withMockedPackageUtils {
+        mockStatic(WebSettings::class.java).use { settings ->
+            settings.`when`<Any> { WebSettings.getDefaultUserAgent(context) }.thenReturn(USER_AGENT)
```
jarvislin: I'd suggest using `whenever` instead of `when` because 'when' is a reserved keyword in Kotlin.

```suggestion
            whenever(WebSettings.getDefaultUserAgent(context)).thenReturn(USER_AGENT)
```

fluxc/src/test/java/org/wordpress/android/fluxc/network/UserAgentTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.wordpress.android.util.PackageUtils
+import kotlin.test.assertEquals
+
+private const val APP_NAME = "App Name"
+private const val USER_AGENT = "Default User Agent"
+private const val APP_VERSION = "1.0"
+
+@RunWith(RobolectricTestRunner::class)
+class UserAgentTest {
+    private val context = RuntimeEnvironment.application.applicationContext
```
jarvislin: `RuntimeEnvironment.application` is deprecated so we can use `RuntimeEnvironment .getApplication()` instead.

```suggestion
    private val context = RuntimeEnvironment.getApplication().applicationContext
```

jarvislin (approved):
The logic looks good, I only left comments related to Mockito and Robolectric but they are not blockers.

I have applied the patch, but I'm not sure the steps to reproduce the crash

fluxc/src/test/java/org/wordpress/android/fluxc/network/UserAgentTest.kt
antonis: Thank you for the suggestion @jarvislin 🙇 
I hadn't realised that I could use `whenever` in the `mockStatic` context. I changed this with b760c39e369040ee0cf400555b21edab8ba0ea9b

antonis:
Thank you for reviewing and your suggestions @jarvislin 🙇 

> I have applied the patch, but I'm not sure the steps to reproduce the crash

I wasn't able to reproduce the crash either but tested that the app would perform as expected when the fallback exception case is used similar to how it works in [the WordPress app initialiser](https://github.com/wordpress-mobile/WordPress-Android/blob/trunk/WordPress/src/main/java/org/wordpress/android/AppInitializer.kt#L1105)

PR #2963, opened 2024-02-29
Migrate Blaze campaigns endpoint

JorgeMucientes:
⚠️ **Please do not merge**. I'll handle the merging of this PR once both dependant PRs from WordPress and WooCommerce repos are ready to be merged. 

### Description

Migrates existing endpoint for fetching Blaze campaigns to new endpoint. 

I initially approached this with the aim of keeping the old and the new rest client implementations coexisting until Jetpack was able to migrate, and reuse the existing models at Storage level. However, the changes in the API models were too big, and keeping both implementations would essentially require to duplicate all the work from rest client, storage and database. 
Also, the new API avoids using `budgetCents`:

> We wanted to switch from cents to make it more universal (not all currencies have cents). So in v1.1 I planned to use natural decimal values (JSON number) for money amounts everywhere (campaign create, forecast).

Another relevant thing to mention is the new API has a different way for paginating results:

Screenshot 2024-03-01 at 11 01 42

- `skip` to indicate the number of items to skip when loading the next batch
- `limit` would be the equivalent to `pageSize`

These changes are breaking changes that impacts both WordPress/Jetpack Android apps as well as WooCommerce.
Following are the PRs from both repositories with the required changes to fetch the campaigns using the new API. Please test and review the implementation from those: 

- WooCommerce: https://github.com/woocommerce/woocommerce-android/pull/10980
- Jetpack: https://github.com/wordpress-mobile/WordPress-Android/pull/20358

**Request:**
`GET /api/v1.1/campaigns`

**Response**
```json
{
    "campaigns": [
        {
            "id": "38404",
            "status": "finished",
            "start_time": "2023-08-02T00:00:00.000Z",
            "duration_days": 1,
            "total_budget": 5,
            "spent_budget": 5,
            "impressions": 6000,
            "clicks": 9,
            "site_name": "Global Climate Change: An Urgent Exam...",
            "text_snippet": "Climate change, specifically global warming, remains one of the most pressing...",
            "target_url": "https://botvik.wordpress.com/2023/06/25/global-climate-change-an-urgent-examination-through-the-lens-of-current-science/",
            "target_urn": "urn:wpcom:post:191174658:44",
            "main_image": {
                "url": "https://i0.wp.com/public-api.wordpress.com/wpcom/v2/wordads/dsp/api/v1/dsp/creatives/79117/image?w=600&zoom=2",
                "mime_type": "image/jpeg",
                "width": 300,
                "height": 250
            }
        }
    ],
    "total_count": 34,
    "skipped": 10
}
```

### Testing
- Smoke test loading and displaying Blaze campaigns using the linked PRs from WordPress and WooCommerce repositories.


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCampaignsRestClient.kt
```diff
@@ -1,98 +1,40 @@
 package org.wordpress.android.fluxc.network.rest.wpcom.blaze
```
JorgeMucientes, line 2: I'd suggest reviewing this file as a completely new file

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WPAndroidDatabase.kt
```diff
@@ -324,6 +325,43 @@ abstract class WPAndroidDatabase : RoomDatabase() {
...
+                            "`title` TEXT NOT NULL, " +
+                            "`imageUrl` TEXT, " +
+                            "`createdAt` TEXT NOT NULL, " +
+                            "`endDate` TEXT, " +
+                            "`uiStatus` TEXT NOT NULL, " +
+                            "`impressions` INTEGER NOT NULL, " +
+                            "`clicks` INTEGER NOT NULL, " +
+                            "`targetUrn` TEXT, " +
+                            "`totalBudget` REAL NOT NULL, " +
+                            "`spentBudget` REAL NOT NULL, " +
```
JorgeMucientes, line 344: The change here is adding these 2 new fields and removed `budgetCents`

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCampaignsRestClient.kt
```diff
@@ -1,98 +1,40 @@
...
-    appContext: Context?,
-    @Named("regular") requestQueue: RequestQueue,
-    accessToken: AccessToken,
-    userAgent: UserAgent,
-    acceptStrategy: AcceptHeaderStrategy.JsonAcceptHeader
-) : BaseWPComRestClient(appContext, dispatcher, requestQueue, accessToken, userAgent, acceptStrategy) {
+    private val wpComNetwork: WPComNetwork
+) {
+    companion object {
+        const val DEFAULT_ITEMS_LIMIT = 25 // Number of items to fetch in a single request
```
hichamboushaba: Just a remark, I think the value in the WordPress app is wrong, as it [expects](https://github.com/wordpress-mobile/WordPress-Android/blob/9795ba93e4436cd3a275b8f0ca032e6c0ec38588/WordPress/src/main/java/org/wordpress/android/ui/blaze/blazecampaigns/campaignlisting/CampaignListingViewModel.kt#L54) 10, and doesn't pass a specific value when fetching (unless I missed it).

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCampaignsRestClient.kt
```diff
@@ -1,98 +1,40 @@
...
-            this,
-            url,
-            mapOf(
-                "page" to page.toString(),
-                "order" to "start_date",
-                "sort" to "desc"
-            ),
-            BlazeCampaignsResponse::class.java
+        siteId: Long,
+        skip: Int,
```
hichamboushaba: np, WordPress (and many famous paginated APIs) [use](https://developer.wordpress.org/rest-api/using-the-rest-api/pagination/) the word `offset` instead of `skip`, and `per_page` instead of `limit`, I know it's a small thing, but WDYT about staying consistent with the naming on our side here and name the argument as `offset`.

fluxc/src/main/java/org/wordpress/android/fluxc/utils/extensions/MapExtensions.kt
```diff
@@ -1,4 +1,11 @@
 package org.wordpress.android.fluxc.utils.extensions
 
 @Suppress("UNCHECKED_CAST")
-fun <K, V : Any> Map<out K, V?>.filterNotNull(): Map<K, V> = filterValues { it != null } as Map<K, V>
+fun <K, V : Any> Map<out K, V?>.filterNotNull(): Map<K, V> =
+    filterValues { it != null } as Map<K, V>
+
+fun <T, K> MutableMap<T, K>.putIfNotNull(vararg pairs: Pair<T, K?>) = apply {
```
hichamboushaba, line 7: np, with this addition, we have a duplicated extension in the WooCommerce plugin code, they just live in different packages `org.wordpress.android.fluxc.utils.extensions` vs `org.wordpress.android.fluxc.utils`, I think we can move this one to use the same package as the old one, and remove the old one from the WooCommerce plugin code, WDYT?

fluxc/src/main/java/org/wordpress/android/fluxc/model/blaze/BlazeCampaignsModel.kt
```diff
@@ -4,9 +4,8 @@ import java.util.Date
 
 data class BlazeCampaignsModel(
```
hichamboushaba, line 5: Is this model used somewhere in the WordPress app? If not, then maybe we can simplify the implementation by getting rid of this model and its storage table 🤔.

hichamboushaba (approved):
Nice work @JorgeMucientes, I tested the changes in the Woo app, and they work well, I left some comments, but nothing blocking, so pre-approving.

But I found an issue with the API itself p1709307106367899-slack-C069RG07W6S, so let's hold on merging this until it's fixed.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCampaignsResponse.kt
```diff
@@ -1,56 +1,114 @@
...
-            campaignId = campaignId ?: 0,
-            title = contentConfig.title,
-            imageUrl = contentConfig.imageUrl,
-            createdAt = BlazeCampaignsUtils.stringToDate(createdAt),
-            endDate = endDate?.let { BlazeCampaignsUtils.stringToDate(it) },
-            uiStatus = uiStatus,
-            budgetCents = budgetCents ?: 0,
-            impressions = campaignStats.impressionsTotal ?: 0L,
-            clicks = campaignStats.clicksTotal ?: 0L,
+            campaignId = id.toInt(),
```
hichamboushaba: One thing that I missed here, are we sure we'll always have Integer `ids`? I suggest making the `campaignId` as String, and avoid this conversion.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCreationRestClient.kt
```diff
@@ -496,13 +496,13 @@ private data class BlazeCampaignCreationNetworkResponse(
         createdAt = Date(), // Set to current date, as the API does not return the actual creation date
         endDate = Date(BlazeCampaignsUtils.stringToDate(startTime).time + durationDays.days.inWholeMilliseconds),
         imageUrl = mainImage.url,
-        budgetCents = (totalBudget * 100).toLong(),
         uiStatus = status,
         // TODO revisit this when the API returns the actual values to confirm the format of IDs
         campaignId = id.substringAfter("campaign-").toIntOrNull() ?: 0,
```
hichamboushaba: If we apply the [change](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2963/files#r1509184747) above, we can get rid of this conversion here as well.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCampaignsRestClient.kt
JorgeMucientes: Yes, I saw that while working on the migration from the WordPress codebase. However, the API's default value when not passing anything is 25. That is the reason I set it to that value to keep consistent with the API behavior when no value is set. I could also remove the default value and set it as a nullable property. Wdyt @hichamboushaba ?

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCampaignsRestClient.kt
JorgeMucientes: Yes, I agree that is a better naming. My question is, should we keep naming consistent with the API or should we add the naming that better suits our codebase? I have mixed feelings here.

Screenshot 2024-03-01 at 17 17 12

fluxc/src/main/java/org/wordpress/android/fluxc/utils/extensions/MapExtensions.kt
JorgeMucientes: Yes, I can do that. However I won't tackle that in this PR as it will generate plenty of filechanges noise just because of imports. I'll tackle that in a different PR

hichamboushaba: Sure, we can handle this later.

> However I won't tackle that in this PR as it will generate plenty of filechanges noise just because of imports

Not necessarily, if we keep the same package name `org.wordpress.android.fluxc.utils` for the new extension in the module `fluxc`, it won't need any import changes.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCampaignsRestClient.kt
hichamboushaba: My 2 cents keeping a better naming that suits our codebase serves us better, the `API` naming is used only in the rest client, while the rest is used across different layers.

But this is a very minor point 🙂

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/blaze/BlazeCampaignsRestClient.kt
hichamboushaba: I think the default value is fine, my point was actually regarding a needed change in the WordPress PR, we need to make sure we pass the `limit` value, instead of using the default one. Sorry if I wasn't clear enough.

fluxc/src/main/java/org/wordpress/android/fluxc/model/blaze/BlazeCampaignsModel.kt
JorgeMucientes: @hichamboushaba  Can you elaborate what do you mean exactly by `simplify the implementation by getting rid of this model and its storage table` ???

fluxc/src/main/java/org/wordpress/android/fluxc/utils/extensions/MapExtensions.kt
JorgeMucientes: @hichamboushaba  hmm, yes I see but this won't work. Because `MapExtensions.kt` already existed before I added the function. And the file exists in `package org.wordpress.android.fluxc.utils.extensions` which is different to `org.wordpress.android.fluxc.utils`. So, I'd need to move `MapExtensions.kt` out of `extensions` package causing multiple import changes as well as changing the approach of having all extension files in the same package. 
Imho is better solution to manually replace the usages of `putIfNotNull` from `MapExt.kt` with the new function from `MapExtensions.kt` that is available for the whole FLuxC codebase. And better to handle that in a different PR. 
Wdyt?

hichamboushaba: Since this is just an extension, it doesn't have to be in the same file `MapExtensions`, it can be part of a file that's in the package `org.wordpress.android.fluxc.utils`, but if you prefer to keep a single file, then changing the package name of one of the two extensions is needed, I don't have a strong opinion on this 🙂.

fluxc/src/main/java/org/wordpress/android/fluxc/model/blaze/BlazeCampaignsModel.kt
hichamboushaba: I mean if the model `BlazeCampaignsModel` (with the pagination information) is not used in the WordPress app, then we can have only a single model `BlazeCampaignModel` and a single DB table, with this we'll get rid of quite a bunch of code.

This is also a minor point, feel free to ignore it.

JorgeMucientes: `BlazeCampaignsModel` is used in both the WooCommerce and WordPress app codebases to handle pagination. So we need it. However I don't think we need to save that model into the DB, and we probably can get rid of the table and the DAO entity. Not sure if you were referring to that. 
I'll check if the refactor is straightforward in both platforms. If it is, I'll add the change. If not I'll leave it as is, don't want to spend mucho more time on this.

hichamboushaba: > BlazeCampaignsModel is used in both the WooCommerce

I don't think it's technically used 😅, we just read the total number of items from it to decide if there are next pages, which means the model itself is not needed, and can be replaced with a simpler return type, but I'm not sure about the WordPress app as I didn't check the code.

But as I said before, this is a minor point, and I agree about not spending much time on it 🙂

AjeshRPai:
Hey @JorgeMucientes 

I have smoke tested the WP PR and everything looks good to me 👍🏼  Thank you for making these changes 👍🏼 

Other than the API issue mentioned by @hichamboushaba above, I don't have any concerns about the changes. Once the API issue is fixed, I think its safe to land these changes. 

 I am approving the PR.

AjeshRPai (approved):

JorgeMucientes:
Thanks for reviewing it, @AjeshRPai. The API problems that Hicham mentioned were already fixed, so we can proceed with merging. 
I'll handle the merge of all 3 PRs

PR #2964, opened 2024-03-01
Allow passing a "source type" when creating Woo Orders.

hichamboushaba:
Part of https://github.com/woocommerce/woocommerce-android/issues/10990

This PR updates the Woo Order store by allowing passing a custom "source type" to attribute the order to (check https://github.com/woocommerce/woocommerce-android/issues/10960 for more details).

Instead of hardcoding the value of the mobile app in FluxC, I opted for making a configurable value, where each client can use their own, even though for now WCAndroid is the only client.

##### Testing
1. Open the example app and sign in.
2. Click on Woo, then Orders.
3. Select a site.
4. Click on Create an order.
5. Enter a product ID.
6. Skip the other steps.
7. When the order is created, check it in wp-admin, and confirm it has a source type set to: "fluxc-example-app" (the Origin will be unknown, as this source type value is not mapped by Core)


0nko (approved):
The test works as expected 👍

PR #2966, opened 2024-03-05
Remove unused stats code

hichamboushaba:
⚠️ Please don't merge until CI of https://github.com/woocommerce/woocommerce-android/pull/11012 is green

While investigating the changes needed to support custom ranges for stats, I found out that we use the `StatsGranularity` as a range, and we have date calculation functions embedded on it and also in other places. While in reality (at least for my understanding and how the iOS app also uses it), the granularity is just a unit, and the API call will need it in addition to a date range, so the date operations should be decoupled from it.

When I tried to make the changes to decouple granularity from date ranges, I was overwhelmed by the amount of code we have in Stats, and I was confused on what's still used and what not, so I decided to start by some cleanup first before moving towards the refactoring.

This PR cleans up the stats code, it removes the following parts:
1. "Top Earners" code, which used a WPCom endpoint, and was replaced later by the functions of `WCLeaderboardsStore` which use wc-analytics REST endpoints. (143a302)
2. Legacy Visitor stats, this was replaced by suspending functions, but used different payloads and DB tables, so removing it is safe. (831ca5 and b23aeb4)
3. Remove unused Order Stats code which a WPCom endpoint and was replaced by "Order Revenue Stats" which uses a wc-analytics REST endpoint (13e423b and b23aeb4)
4. Removes an unused code for "Top Performers" that used `StatsGranularity` as a range and not a unit, this will help with upcoming refactoring (6c099c8)

Sorry for the the size of the PR, but since it's mostly code deletion I think it should be straightforward for the review, and reviewing commit by commit might help also.

##### Testing
All we need here is just to make sure the code still compiles fine, and WCAndroid still compiles fine.


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
```diff
@@ -1997,6 +1997,10 @@ open class WellSqlConfig : DefaultWellConfig {
                     db.execSQL("ALTER TABLE ThemeModel ADD THEME_TYPE TEXT")
                     db.execSQL("ALTER TABLE ThemeModel ADD IS_EXTERNAL_THEME BOOLEAN")
                 }
+                198 -> migrate(version) {
+                    db.execSQL("DROP TABLE IF EXISTS WCOrderStatsModel")
+                    db.execSQL("DROP TABLE IF EXISTS WCVisitorStatsModel")
```
0nko, line 2002: Did you forget to drop the `WCTopEarnerModel`?

0nko (approved):
Awesome cleanup! 🧹  I just left one comment in code.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
hichamboushaba: I don't think `WCTopEarnerModel` [was](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/9427b0bc352a506a8ab3ece1dc859b8f00f130be/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCTopEarnerModel.kt) a table model, so no further cleanup is needed.

PR #2968, opened 2024-03-06
Stats/custom range

hichamboushaba:
This PRs main goal is to decouple granularities from ranges:

### Before
Currently the `StatsGranularity` enum was used both as a unit and also as a range, so when we used `StatsGranularity.YEARS`, we had logic that assumed that we mean the current year, and we used this to calculate the stats range.
This was fine when we used only predefined ranges, but with custom ranges, this will be problematic.

### After
With this PR, we are moving from the above assumption, and now all of the stats functions expect a range and a granularity.

### How
The changes are actually simple, it mainly involved making the `startDate` and `endDate` of the requests non-optional, and removed the code that calculated them from the granularities.

### Other changes
The PR includes some other improvements:
- Simplify the date formatting for the `/visits` endpoint, instead of having different formats depending on the granularity, the API works correctly with the day format, and it's what the iOS app uses.
- Use a fixed quantity of `100` for the revenue stats endpoint (pe5sF9-2ri-p2#comment-3427)
- Removed usage of the `OrderStatsApiUnit` from upper layers, as it's a network unit.
- The `StatsGranularity` we pass is mapped to the matching `OrderStatsApiUnit`, the logic of calculating the correct granularity will be moved to the upper layers (we removed this [logic](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/28e35ae36fe90c3b7bafb4610e71b9cd3f4e2dc4/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/orderstats/OrderStatsRestClient.kt#L46-L76))

### Testing
Check the WCAndroid PR https://github.com/woocommerce/woocommerce-android/pull/11033

### DIFF
--- a/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCStatsTest.kt
+++ b/example/src/androidTest/java/org/wordpress/android/fluxc/mocked/MockedStack_WCStatsTest.kt
@@ -18,7 +18,6 @@ import org.wordpress.android.fluxc.annotations.action.Action
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.module.ResponseMockingInterceptor
 import org.wordpress.android.fluxc.network.rest.wpcom.wc.orderstats.OrderStatsRestClient
-import org.wordpress.android.fluxc.network.rest.wpcom.wc.orderstats.OrderStatsRestClient.OrderStatsApiUnit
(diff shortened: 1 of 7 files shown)


JorgeMucientes (approved):

JorgeMucientes:
Great job @hichamboushaba. Really like how things are way simpler now buy clearly separating the granularity from the actual date ranges. 
I've smoke tested my store stats as well as analytics hub and everything seems to work just fine. 🚢

PR #2972, opened 2024-03-12
Preparations to support order deletion from the app

hichamboushaba:
To support deleting orders from the app, we need to handle deletion in a cancellable way (a way that supports undoing by the user), this PR adds two changes to support this:

1. Update the order list descriptor to support excluding specific orders.
2. Delete the Order summary DB entity when an order is deleted.

##### Testing
Please check https://github.com/woocommerce/woocommerce-android/pull/11068

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/di/WCDatabaseModule.kt
@@ -5,11 +5,13 @@ import dagger.Binds
 import dagger.Module
 import dagger.Provides
 import org.wordpress.android.fluxc.module.DatabaseModule
+import org.wordpress.android.fluxc.persistence.OrderSqlUtils
 import org.wordpress.android.fluxc.persistence.TransactionExecutor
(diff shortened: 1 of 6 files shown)


example/src/test/java/org/wordpress/android/fluxc/wc/order/WCOrderStoreTest.kt
```diff
@@ -71,34 +83,53 @@ class WCOrderStoreTest {
...
-    fun setUp() {
+    private val testCoroutineDispatcher = TestCoroutineDispatcher()
+    private val testCoroutineScope = TestCoroutineScope(testCoroutineDispatcher)
+
+    fun setUp() = setup(initCoroutineEngine(), null)
+
+    fun TestCoroutineDispatcher.setUp() = setup(
+        coroutineEngine = CoroutineEngine(this, AppLogWrapper()),
+        coroutineDispatcher = this
+    )
```
hichamboushaba, line 94: To support test cases that involve delays and cancellations, I added this logic to use the Coroutine's default `TestCoroutineDispatcher`.

0nko (approved):
Tested the associated WCAndroid PR and everything's working well.. :shipit:

PR #2975, opened 2024-03-17
Hack week: Add coroutine calls to fetch installed plugins

0nko:
This PR adds WP.com and WP.org REST clients that support coroutines. That way, we're able to fetch and synchronize the WP.org plugin repository information and avoid a state when we can't show available plugins.

**To test:**
See the woocommerce/woocommerce-android#11081 PR.


fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginCoroutineStore.kt
```diff
@@ -77,6 +87,48 @@ class PluginCoroutineStore
...
+    }
+
+    private suspend fun getInstalledPlugins(site: SiteModel): List<ImmutablePluginModel> {
+        val immutablePlugins: MutableList<ImmutablePluginModel?> = mutableListOf()
+        val sitePlugins = PluginSqlUtils.getSitePlugins(site)
+        sitePlugins.forEach { sitePluginModel ->
+            val slug = sitePluginModel.slug
+            var wpOrgPluginModel = getWPOrgPluginBySlug(slug)
+            if (wpOrgPluginModel == null) {
+                val response = pluginWPOrgCoroutineClient.fetchWpOrgPlugin(slug)
```
hichamboushaba, line 117: @0nko if I understand correctly, this will be called for every plugin in the list, and if the site has a big number of plugins, the wait time will be quite long, WDYT?

hichamboushaba (commented):
Thank @0nko for working on this, I left one concern about the waiting times if the site has a big number of plugins.

And also, while discussing with Gabriel the point about managed plugins in Atomic sites pdfdoF-4OF-p2#comment-5981, I found out that they use the `/wc/v3/system_status` endpoint to fetch plugins, which we already use in the app too in other places, and the advantage of this is that it includes a field `latest_version`, which means that if we switch to it, we can have a unified experience for all types of stores (the endpoint is already implemented using `WooNetwork` which works with all types of connections), and also the screen will be usage by non-admin users as the endpoint is available for shop managers.

I know this was a Hack Week project, and the change to the other endpoint would take time, so IMHO if you agree on the benefits of the other endpoint, we can merge the WCAndroid changes without this FluxC part, but keep the screen under a feature flag, and on the next Hack Week, work on the migration to the other endpoint, WDYT?

fluxc/src/main/java/org/wordpress/android/fluxc/store/PluginCoroutineStore.kt
0nko: I takes a couple of seconds for the first run when the plugins are null initially, but nothing crazy. Subsequent loads are faster.

0nko:
There are changes here that WCAndroid depends on. I guess we can comment out the code that calls the FluxC changes and conserve it for the future. Do you want to close this PR then?

hichamboushaba:
> I guess we can comment out the code that calls the FluxC changes and conserve it for the future.

Will the existing WCAndroid code be useful even if we switch to the `system_status` endpoint?

> Do you want to close this PR then?

All the above was just a suggestion, but please choose the approach that you feel work best, if we want to keep this PR, we just need to discuss this [concern](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2975#discussion_r1528324743) 🙂

0nko:
Yeah, there are some things that are worth saving, like UI and version comparison, for example. 

> All the above was just a suggestion, but please choose the approach that you feel work best, if we want to keep this PR, we just need to discuss this https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2975#discussion_r1528324743 🙂

Actually the simplest approach would be to just expose the `SystemPluginModel` in FluxC that are already being fetched. The WCAndroid changes required would be minimal and with the latest version we could make the feature available.. I'll close this PR and open a new one.

PR #2976, opened 2024-03-18
Hack week: Expose system plugins

0nko:
This PR exposes the `SystemPluginModel` so that we can access the latest version information.

**To test:**
Test the associated WCAndroid PR (woocommerce/woocommerce-android#11081).


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
```diff
@@ -213,7 +214,7 @@ open class WooCommerceStore @Inject constructor(
     }
 
     suspend fun fetchSitePlugins(site: SiteModel): WooResult<List<SitePluginModel>> {
```
hichamboushaba, line 216: np, I would love to have the two functions unified and return the same model which could be specific to Woo. But since this would require changes in WCAndroid, it's just something we can add in the future.

hichamboushaba (approved):
Nice work @0nko, I tested the changes in WCAndroid, and they work well.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WooCommerceStore.kt
0nko: Yeah I know and I agree. I was just trying to get it working with minimum changes.

PR #2977, opened 2024-03-19
Use real requests to fetch Blaze payment methods as well as create a Blaze campaign

JorgeMucientes:
### Description

Replaces usage of mocked responses for Blaze campaign creation flow. Specifically:
- Replaces create campaign mocked responses with the new create campaign endpoint
- Replaces fetching payment methods mocked response with real request

Additionally, it adds a couple of model changes required to support the latest API format.

To test the changes, use this PR: https://github.com/woocommerce/woocommerce-android/pull/11112 

NOTE: We won't be able to test this now, as the API is not ready yet. But we can start merging it as these changes are not breaking changes and are hidden behind a feature flag.


0nko (approved):
Looks good! 👍

PR #2985, opened 2024-04-03
[Better Errors] Detect Invalid Credentials scenario

### LINKED ISSUE #10975: [Better Errors] Differentiate invalid credential errors
# Context

Currently, there is not a specific way to know if an error was due to invalid credentials.

However, the server injects a `shake` pattern when an invalid credential error is found.

For the time being, we will use that pattern as a guess for an invalid credential error message.

## Patern

```
document.querySelector('form').classList.add('shake')
```

- [Source]( https://github.com/WordPress/WordPress/blob/master/wp-login.php#L65-L67)
- [iOS Implementation](https://github.com/woocommerce/woocommerce-ios/pull/12230)

# Task

- [ ] Differentiate invalid credential errors from other errors
- [ ] Do not launch the app password tutorial for invalid credential errors.

ThomazFB:
Fix issue https://github.com/woocommerce/woocommerce-android/issues/10975

# Why

This PR tries to identify invalid credential errors when logging in with site credentials, in order to not navigate the user to the app password tutorial when is not really needed.

# How

When logging in with site credentials there is no way to properly tell if an error message is an invalid credentials error.
However, the server injects a `shake` pattern when an invalid credential error is found.

For the time being, we will use that pattern as a guess for an invalid credential error message.

https://github.com/WordPress/WordPress/blob/master/wp-login.php#L65-L67

---

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
@@ -145,4 +145,11 @@ class NonceRestClient @Inject constructor(
         return HtmlUtils.fastStripHtml(errorHtml)
             .trim(' ', '\n')
     }
+
+    private fun hasInvalidCredentialsPattern(htmlResponse: String) =
+        htmlResponse.contains(INVALID_CREDENTIAL_HTML_PATTERN)
+
+    companion object {
+        const val INVALID_CREDENTIAL_HTML_PATTERN = "document.querySelector('form').classList.add('shake')"
+    }
 }
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
@@ -69,11 +69,18 @@ class NonceRestClient @Inject constructor(
                 // that it's an authentication issue, otherwise we'll assume it's an invalid response
                 val errorMessage = extractErrorMessage(response.data.orEmpty())
 
+                val errorType = if (hasInvalidCredentialsPattern(response.data.orEmpty())) {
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
```diff
@@ -69,11 +69,18 @@ class NonceRestClient @Inject constructor(
...
                 val errorMessage = extractErrorMessage(response.data.orEmpty())
 
+                val errorType = if (hasInvalidCredentialsPattern(response.data.orEmpty())) {
+                    Nonce.CookieNonceErrorType.INVALID_CREDENTIALS
+                } else if (errorMessage != null) {
+                    Nonce.CookieNonceErrorType.NOT_AUTHENTICATED
+                } else {
+                    Nonce.CookieNonceErrorType.INVALID_RESPONSE
+                }
+
```
atorresveiga, line 79: np: Another way to write this could be with a when statement.  However, there is no need to change the code. I'm just sharing the idea.
```
when {
    hasInvalidCredentialsPattern(response.data.orEmpty()) -> 
                 Nonce.CookieNonceErrorType.INVALID_CREDENTIALS
    errorMessage != null -> Nonce.CookieNonceErrorType.NOT_AUTHENTICATED
    else -> Nonce.CookieNonceErrorType.INVALID_RESPONSE
}
```

atorresveiga (approved):
The code looks good! Interesting hack @ThomazFB  :shipit:

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/NonceRestClient.kt
ThomazFB: Oh, I missed your suggestion before merging the PR. 🤦  Muuuuch better indeed! I'll make sure to adjust this in the next FluxC PR for the Better Errors I'll be adding. Thank you, @atorresveiga 🙏

PR #2989, opened 2024-04-10
Add jetpack to plugins

atorresveiga:
Woo PR https://github.com/woocommerce/woocommerce-android/pull/11288

### Why
The session stats of Analytics Hub require the Jetpack plugin to be installed and activated.

### Description
This PR adds Jetpack to the list of plugins of WooCommerceStore so we can check the Jetpack status from Woo.

### Testing
These changes don't need testing


ThomazFB (approved):
Looks good! 😄

PR #2992, opened 2024-04-21
Orders: expose a function to observe orders' count

hichamboushaba:
This small PR just adds a function to observe the orders' count without having to specify some statuses, this is something we'll need for https://github.com/woocommerce/woocommerce-android/pull/11362


0nko (approved):
:shipit:

PR #2995, opened 2024-04-25
Utilities to improve observing products

hichamboushaba:
This PR introduces three changes:

1. Ability to observe products count.
2. Ability to limit number of returned products when observing changes. This should help reducing the changes of getting `SQLiteBlobTooBigExceptions` peaMlT-8i-p2
3. Ability to exclude sample products in the SQL query instead of doing a runtime filtrage, this was needed to support the point 2, and it should also bring some improvements.

Check https://github.com/woocommerce/woocommerce-android/pull/11380 for testing.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductApiResponse.kt
```diff
@@ -173,6 +173,13 @@ data class ProductApiResponse(
             bundledItems = response.bundled_items?.toString() ?: ""
             compositeComponents = response.composite_components?.toString() ?: ""
             metadata = response.metadata?.toString() ?: ""
+            isSampleProduct = response.metadata?.any {
```
0nko, line 176: Great idea! 👏

0nko (approved):
LGTM 👍

PR #2998, opened 2024-04-30
Put params in the body not in query

### LINKED ISSUE #11418: [Backed Receipts] Change API usage to make it work via JP tunnel
New API

```

POST orders/203/receipt body: {"expiration_days": 2}
```

kidinov:
closes: https://github.com/woocommerce/woocommerce-android/issues/11418

As we found out JP tunnel doesn't pass POST requests with query params. That PR moves params to the body.

To test:
* Use composite build with the Woo app
* Login to a store via a WP.com account
* Orders -> Completed Order -> See receipts
* Notice that  it does work now


samiuelson (approved):
🎉  LGTM

PR #3000, opened 2024-04-30
[Crash Fix] Update Woo-specific migration to run only for WooCommerce

thomashorta:
Use `migrateAddOn` for Woo-specific table update.

Otherwise, the `migrate` function causes a crash in any app that uses `FluxC` and doesn't contain the table being updated, which is what is happening on the current version of Jetpack, `24.8-rc-1`, as reported here: p1714495849353259-slack-C0180B5PRJ4

The issue was introduced by https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2995.


oguzkocer (approved):

RenanLukas (approved):
Thanks for the fix! LGTM :shipit:

hichamboushaba:
Thanks @thomashorta for taking care of fixing this, and sorry for introducing the crash, it's been a long time since I worked on any FluxC migrations 🤦.

PR #3007, opened 2024-05-09
Add shipping methods endpoint

atorresveiga:
Part of: https://github.com/woocommerce/woocommerce-android/issues/11391

### Why
As part of the Enhancing Order Shipping Lines pffQ75-m4-p2 we want to enable the shipping method selection.

### Description
This small PR adds the shipping_methods endpoint needed to fetch the store's shipping methods.

### Testing instruction
It is best to test this PR alongside [the Woo PR](https://github.com/woocommerce/woocommerce-android/pull/11478)

### DIFF
--- a/fluxc-processor/src/main/resources/wc-wp-api-endpoints.txt
+++ b/fluxc-processor/src/main/resources/wc-wp-api-endpoints.txt
@@ -117,3 +117,5 @@
 /reports/products
 
 /options
+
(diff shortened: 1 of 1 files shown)


ThomazFB (approved):
Reviewed and test through https://github.com/woocommerce/woocommerce-android/pull/11478#pullrequestreview-2049166599.

PR #3008, opened 2024-05-10
[Dynamic Dashboard] Fetch orders

0nko:
Implements woocommerce/woocommerce-android#11462. This PR adds a way to add to fetch orders using coroutines without deleting the existing data.

**To test:**
Please test the associated [WCAndroid PR](https://github.com/woocommerce/woocommerce-android/pull/11486).

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
@@ -610,7 +610,7 @@ class WCOrderStore @Inject constructor(
         orderBy: OrderBy = OrderBy.DATE,
         sortOrder: SortOrder = SortOrder.DESCENDING,
         statusFilter: String? = null,
-        deleteOldData: Boolean = false
+        deleteOldData: Boolean = page == 1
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -598,6 +602,32 @@ class WCOrderStore @Inject constructor(
...
 
+    @Suppress("SpreadOperator")
+    suspend fun fetchOrders(
+        site: SiteModel,
+        count: Int = NUM_ORDERS_PER_FETCH,
+        page: Int = 1,
+        orderBy: OrderBy = OrderBy.DATE,
+        sortOrder: SortOrder = SortOrder.DESCENDING,
+        statusFilter: String? = null,
+        deleteOldData: Boolean = false
```
hichamboushaba: minor np, I suggest setting this to be `page == 1` just to make it more suitable for any pagination by default, it's not something we need, but it could be more clear for any other usages in the future, WDYT?

hichamboushaba (approved):
Nice work @0nko, I left a minor np, but feel free to ignore it.

PR #3009, opened 2024-05-10
Add log lines to capture the string error message

zwarm:
To assist in troubleshooting a user issue related to publishing a post with media, this PR adds a log line in each of the following:
- MediaStore
- PostStore

The error message the user is experiencing is attached below, which appears to be shown after a media/post upload. This particular message is also the default message if we can't convert the server "string" message to a recognizable error. The intention is to use the next beta to capture the logs, and then remove the log lines before the beta turns into prod.

Alt desc

See conversation p1714119618977609-slack-C0180B5PR for more details of the issue.

If you believe this is a bad idea, then raise a flag and close the PR without approving/merging. There is no companion WP app.


AjeshRPai (approved):
LGTM

PR #3012, opened 2024-05-15
Coupons/fetch reports

### LINKED ISSUE #11518: [Dynamic Dashboard] Integrate the coupons report endpoint for Top 3 coupons card

hichamboushaba:
⚠️ This has a breaking change, please don't merge, I'll handle merging when all PRs are ready.

Closes https://github.com/woocommerce/woocommerce-android/issues/11518

This PR adds the networking implementation needed for the coupons card in the dynamic dashboard.

#### Testing
1. Open the example app.
2. Click on Woo then on Coupons.
3. Select a site.
4. Click on "Fetch the most active coupons"
5. Confirm the 3 most active coupons in the last month are printed to the console.

Also check https://github.com/woocommerce/woocommerce-android/pull/11531 for testing the other changes.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooError.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooError.kt
@@ -38,53 +38,41 @@ enum class WooErrorType {
     RESOURCE_ALREADY_EXISTS
 }
 
-fun WPComGsonNetworkError.toWooError(): WooError {
-    val type = when (type) {
-        TIMEOUT -> WooErrorType.TIMEOUT
-        NO_CONNECTION,
-        SERVER_ERROR,
-        INVALID_SSL_CERTIFICATE,
-        NETWORK_ERROR -> WooErrorType.API_ERROR
-        PARSE_ERROR,
-        CENSORED,
-        INVALID_RESPONSE -> WooErrorType.INVALID_RESPONSE
-        HTTP_AUTH_ERROR,
-        AUTHORIZATION_REQUIRED,
-        NOT_AUTHENTICATED -> WooErrorType.AUTHORIZATION_REQUIRED
-        NOT_FOUND -> WooErrorType.INVALID_ID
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooError.kt
```diff
@@ -32,59 +32,52 @@ enum class WooErrorType {
     INVALID_RESPONSE,
     AUTHORIZATION_REQUIRED,
     INVALID_PARAM,
-    PLUGIN_NOT_ACTIVE,
+    API_NOT_FOUND,
```
hichamboushaba, line 35: I was the one who introduced `PLUGIN_NOT_ACTIVE` during the REST API project, it was to match the value emitted by the Order REST client, but I think that name was not the right choice, the error could happen for any non-existing endpoint, be it because the plugin is not active, or any other reason, so I renamed it to make it more clear, please let me know what you think?

And this is a breaking change for WCAndroid, so let's not merge this until the WCAndroid PR is ready.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/WooError.kt
```diff
@@ -32,59 +32,52 @@ enum class WooErrorType {
...
+    CENSORED,
+    INVALID_RESPONSE -> WooErrorType.INVALID_RESPONSE
+
+    HTTP_AUTH_ERROR,
+    AUTHORIZATION_REQUIRED,
+    NOT_AUTHENTICATED -> WooErrorType.AUTHORIZATION_REQUIRED
+
+    NOT_FOUND -> {
+        when (apiError) {
+            "rest_no_route" -> WooErrorType.API_NOT_FOUND
```
hichamboushaba, line 70: We need this specific differentiation now, the new error type will help identify when the `wc-analytics` endpoint is not available.

0nko (approved):
LGTM 👍

PR #3013, opened 2024-05-15
Shipping method db support

atorresveiga:
### Why
We want to add support to save shipping methods locally to improve the performance of the shipping method selection.

### Description
This PR creates the new `ShippingMethod` table and DAO in our Woo database to cache shipping methods locally and improve the app performance when displaying shipping methods. Additionally, I moved the rest of the clients to FluxC and exposed the shipping methods functionality through the shipping methods store `WCShippingMethodsStore`.

### Testing
I added some unit tests, but it is best to test this PR using the [Woo PR](https://github.com/woocommerce/woocommerce-android/pull/11527)


ThomazFB (approved):
Reviewed through https://github.com/woocommerce/woocommerce-android/pull/11527#pullrequestreview-2059598180.

PR #3014, opened 2024-05-17
Fix shipping methods migration

atorresveiga:
### Description 

This PR reverts the unnecessary changes on the DB v35 schema that caused the DB Migration to fail

### Testing

1. Install Woo version 18.6 and log in into the app
2. Install Woo version 18.7 and check that the app doesn't crash


hichamboushaba (approved):
Works well, thanks for the fix @atorresveiga

PR #3016, opened 2024-05-22
Unit tests db schemas hashes

atorresveiga:
### Description
This PR introduces unit tests that verify the integrity of database schema migrations. Previously generated schemas are hashed and compared to ensure no unintended modifications occur during the migration process. This helps prevent potential crashes due to schema inconsistencies.

Benefits:

- Increased confidence in database migrations.
- Reduced risk of crashes during app updates.
- Improved code maintainability with automated schema verification.

### Testing

- Check that if the hash is different than expected, the test will fail
- Check that if the DB version is updated and the new hash is not added to the hashes map, the test will fail


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/WCAndroidDatabase.kt
```diff
@@ -69,8 +69,10 @@ import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_7_8
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_8_9
 import org.wordpress.android.fluxc.persistence.migrations.MIGRATION_9_10
 
+const val WC_DATABASE_VERSION = 36
```
atorresveiga, line 72: I moved the DB version to a constant for easier reference in unit tests.

plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationSchemasTests.kt
```diff
@@ -0,0 +1,79 @@
...
+            getDBHashKey(33) to "43fd0900518251fbd271b5ffe7a0d89f",
+            getDBHashKey(34) to "878b99bf2a221f0f462015ec1ff9f402",
+            getDBHashKey(35) to "a01f7f90e7751f29876dda9a73c49ccc",
+            getDBHashKey(36) to "2b2e3799733db41ef697c524296acc82",
+        )
+    }
+
+    @Test
+    fun testDBHashes() {
+        for(i in 3..WC_DATABASE_VERSION){
```
atorresveiga: We will iterate from every DB version starting from 3 (first DB schema) and ending in the last version

plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationSchemasTests.kt
```diff
@@ -0,0 +1,79 @@
...
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import org.assertj.core.api.Assertions.assertThat
+import org.json.JSONObject
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class MigrationSchemasTests {
```
kidinov: I'd put that into `MigrationTests`, just to remind people that they actually don't test validity of migration here and should write a test similar to all the rest tests there

plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationSchemasTests.kt
```diff
@@ -0,0 +1,79 @@
...
+            getDBHashKey(32) to "892bffac2c3dc56265f35a4375208c54",
+            getDBHashKey(33) to "43fd0900518251fbd271b5ffe7a0d89f",
+            getDBHashKey(34) to "878b99bf2a221f0f462015ec1ff9f402",
+            getDBHashKey(35) to "a01f7f90e7751f29876dda9a73c49ccc",
+            getDBHashKey(36) to "2b2e3799733db41ef697c524296acc82",
+        )
+    }
+
+    @Test
+    fun testDBHashes() {
```
kidinov: np: shell we at least try to name tests someone readable, like we do in Woo repo?

kidinov (approved):
While that makes it a little bit harder to make a mistake here, it's still more like a reminder and it doesn't actually test if the written migration is valid or not

Ideally on top of that to have an actual "update from a previous version" test as one of the CI steps before release uploading

plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationSchemasTests.kt
atorresveiga: Using spaces is not allowed because of the DEX version, I updated the name of the test following the same pattern as the rest of the test in `MigrationTests`
> `com.android.tools.r8.internal.Tb: Space characters in SimpleName 'this is a test' are not allowed prior to DEX version 040`

plugins/woocommerce/src/androidTest/java/org/wordpress/android/fluxc/persistence/MigrationSchemasTests.kt
atorresveiga: Fixed [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3016/commits/38b7fce9014101dc94530225a4df23fcbf881276)

atorresveiga:
Thanks for the review, @kidinov! This PR is ready for another round 😄

kidinov (approved):
LGTM!

PR #3018, opened 2024-05-24
Keep image url authorisation headers when redirecting

antonis:
## Description
This PR adds a custom `OkHttpClient` intercepting the redirects and keeping the Authorisation headers when the TLD and SLD of the original url and the redirect url are the same. This was needed since a new image resource redirection was failing on wordpress.com for private sites because of an [okhttp limitation](https://github.com/square/okhttp/issues/2937).

## To test
See https://github.com/wordpress-mobile/WordPress-Android/pull/20876

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/module/OkHttpClientModule.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/module/OkHttpClientModule.java
@@ -2,6 +2,7 @@ package org.wordpress.android.fluxc.module;
 
 import org.wordpress.android.fluxc.network.BaseRequest;
 import org.wordpress.android.fluxc.network.MemorizingTrustManager;
+import org.wordpress.android.fluxc.utils.WPUrlUtils;
 import org.wordpress.android.util.AppLog;
 import org.wordpress.android.util.AppLog.T;
 
@@ -21,8 +22,11 @@ import dagger.Module;
 import dagger.Provides;
 import dagger.multibindings.Multibinds;
 import okhttp3.CookieJar;
+import okhttp3.HttpUrl;
 import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
 import okhttp3.internal.tls.OkHostnameVerifier;
 
 @Module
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/module/OkHttpClientModule.java
```diff
@@ -46,6 +50,36 @@ public static OkHttpClient provideNoRedirectsOkHttpClientBuilder(
...
+            if (response.isRedirect()) {
+                String location = response.header("Location");
+                if (location != null && !location.isEmpty()) {
+                    Request.Builder newBuilder = originalRequest.newBuilder().url(location);
+
+                    // Remove the authorization header if the hosts' TLD and SLD are not the same
+                    String originalHost = originalRequest.url().host();
+                    HttpUrl redirectHttpUrl = HttpUrl.parse(location);
+                    String redirectHost = (redirectHttpUrl != null) ? redirectHttpUrl.host() : "";
+                    if (!WPUrlUtils.tldAndSldAreEqual(originalHost, redirectHost)) {
```
antonis: [The default OkHttp implementation](https://github.com/square/okhttp/blob/master/okhttp/src/main/kotlin/okhttp3/internal/-UtilJvm.kt#L151) checks the full host string along with the port and scheme. The implementation used here makes a less strict check using only TLD and SLD to make sure that both urls are under the same domain (in our case wordpress.com).

jarvislin (commented):
Thanks for handling the auth issue for the redirection! The logic looks correct, I'll approve this PR after I test another PR https://github.com/wordpress-mobile/WordPress-Android/pull/20876

jarvislin (approved):
I just tested and it works as expected 👍

antonis:
Thank you for the prompt review @jarvislin 🙇 
I've made a minor improvement in order to make the code testable with https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3018/commits/206463b4e06683915198e1862ab06fe09fac1def. The logic should remain the same.

fluxc/src/main/java/org/wordpress/android/fluxc/network/CustomRedirectInterceptor.kt
```diff
@@ -0,0 +1,50 @@
...
+        if (response.isRedirect) {
+            val newRequest = getRedirectRequest(originalRequest, response)
+            if (newRequest != null) {
+                return chain.proceed(newRequest)
+            }
+        }
+        return response
+    }
+
+    fun getRedirectRequest(originalRequest: Request, redirectResponse: Response): Request? {
```
antonis, line 22: Extracted a separate function that constructs the new request since it was not accessible in the `fun intercept(chain: Chain): Response`  for testing.

jarvislin (approved):
Nice, it looks better after the refactor!

PR #3022, opened 2024-05-29
[Woo Min/Max Quantities] Add Support to New REST API

toupper:
⚠️  Needed to close: https://github.com/woocommerce/woocommerce-android/pull/11611

This PR adds support to the new REST API for the Min/Max Quantities extension additions (Check docs [here](https://woocommerce.com/document/minmax-quantities/#section-6)). It replaces the old approach based on the products/variations metadata to add the values as json fields.

Specifically we:

- Add new columns in the Product and Product Variations tables of the wellSQL DB
- Migrate DB version
- Add support for the new values in the API Response classes.
- Remove metadata support.


atorresveiga (approved):
Great work @toupper! The code looks nice and it is working as expected. I didn't face any issue with the DB migration. Thanks for cleaning up the unnecessary code.

LGTM! :shipit:

PR #3023, opened 2024-05-29
Issue/11558 fetch product sales reports

JorgeMucientes:
This PR adds the endpoints needed to fetch product sales stats for products and variation products. These endpoints are needed to display the new dashboard product stock card: 

Screenshot 2024-05-30 at 23 29 48

The changes can be tested using [this PR](https://github.com/woocommerce/woocommerce-android/pull/11631)

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCLeaderboardsStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCLeaderboardsStore.kt
@@ -175,7 +175,7 @@ class WCLeaderboardsStore @Inject constructor(
         productVariationIds: List<Long>
     ): WooResult<Array<ReportsProductApiResponse>> =
         coroutineEngine.withDefaultContext(API, this, "fetchProductVariationsSalesReport") {
-            reportsRestClient.fetchProductSalesReport(
+            reportsRestClient.fetchProductVariationsSalesReport(
                 site,
                 startDate,
                 endDate,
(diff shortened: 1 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCLeaderboardsStore.kt
```diff
@@ -151,12 +153,45 @@ class WCLeaderboardsStore @Inject constructor(
         }
     }
 
+    suspend fun fetchProductSalesReport(
```
hichamboushaba: very minor np, having these part of the `WCLeaderbordsStore` could make them less discoverable, I wish we had a single `WCProductReportsStore` that included all of the functions in addition to the stock related ones, it would have been more aligned with what we do on the other parts of the library.

I'm just sharing a thought here, no need to make any changes now.

hichamboushaba (approved):
Nice work :shipit: 

I left on np thought, but it's very minor and can be ignored

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCLeaderboardsStore.kt
JorgeMucientes: I had some doubts on where to place those new functions, but yeah I think it makes sense to have them altogether inside a `WCProductReportsStore` class. Refactor applied 👍🏼

hichamboushaba: Thank you @JorgeMucientes

PR #3025, opened 2024-05-29
Remove temporary log lines used to capture string error message

zwarm:
This PR reverts the log line added in #3009. These were temporary log messages used to investigate a user issue. They are no longer needed.

See conversation p1714119618977609-slack-C0180B5PR for more details of the issue.


AjeshRPai (approved):
LGTM

PR #3027, opened 2024-05-31
[Bug] Add null check if response is null.

notandyvee:
[Issue](https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/3024)

## What

This PR fixes a bug where fetching the revision history of a post can possibly be null. I'm not clear on how it's possible. But forced a null `revisionsResponseToRevisionsModel` and can reproduce the crash. The simplest solution is to null check. Outside of that, I could not reproduce the crash organically.

## Test

- [ ] Make sure builds pass
- [ ] Use [this PR in WP-Android](https://github.com/wordpress-mobile/WordPress-Android/pull/20926) to test the change itself.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/post/PostRestClient.java
@@ -31,6 +31,8 @@ import org.wordpress.android.fluxc.model.revisions.Diff;
 import org.wordpress.android.fluxc.model.revisions.DiffOperations;
 import org.wordpress.android.fluxc.model.revisions.RevisionModel;
 import org.wordpress.android.fluxc.model.revisions.RevisionsModel;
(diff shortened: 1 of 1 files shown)


irfano (changes requested):
May I suggest a different approach? If the response is null, `onResponse()` in the listener of `fetchRevisions()` can return `BaseNetworkError(INVALID_RESPONSE)`. This way, the user can understand that there is an error instead of seeing an empty revision list. 
You can see a similar solution in [fetchSites()](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/trunk/fluxc/src/main/java/org/wordpress/android/fluxc/network/xmlrpc/site/SiteXMLRPCClient.kt#L71-L78) method.

notandyvee:
Very useful @irfano . Thank you for pointing out a good error. It wasn't obvious in the file, at least that I could see. The base error case is a nice compromise.

irfano (approved):
Thank you! LGTM! 👍🏻

PR #3031, opened 2024-06-10
[Voice to Content] Endpoint adjustments

zwarm:
Addresses the following issues
https://github.com/Automattic/wordpress-mobile/issues/111
https://github.com/Automattic/wordpress-mobile/issues/76

This PR updates the following voice to content endpoints

1. Extracts the `upgrade-url` from the ai-assistant-feature endpoint and adds it to the JetpackAIAssistantFeature model
2. Refactors the `JetpackAIQueryDto` to allow null in choices. This is because gson.toJson doesn't enforce Kotlin's nullability rule, so it is possible for choices to be null. If there is no choices list, then we don't have any content - this is an error.

Test using the companion WP PR (tbd)


pantstamp (approved):
Tested with companion PR and looks good! Thank you @zwarm!

PR #3032, opened 2024-06-13
Use site name as blaze campaign title

JorgeMucientes:
Minor change to use `site_name` field from Blaze API response as the title of the campaign. This matches the web behavior as well as iOS behavior. 

Additionally I noticed we were ordering the campaigns by the start date of the campaign instead of the date they were created which is the default order in which the API returns the campaigns, and the order in which they are displayed in the web and in iOS

⚠️ Please do not merge until this part is clarified: p1718294996958399-slack-C03L1NF1EA3
I'd like to re use this PR to apply that change if we agree on it.

**Edit: Merging to get this fixes into prod asap. Will add any needed changes that come up from the linked discussion in a different PR**


hafizrahman (approved):
LGTM for the current changes 👍🏼

PR #3039, opened 2024-06-26
Issue/11802 update jetpack ai query impl

JorgeMucientes:
Part of: https://github.com/woocommerce/woocommerce-android/issues/11802

Adds new functions to use the endpoint `/jetpack-ai-query` for asking questions. The new functions can be tested using FluxC sample app: 

https://github.com/wordpress-mobile/WordPress-FluxC-Android/assets/2663464/69b3990c-ef88-426c-8a34-22af12364a34

Changes can also be tested as part of [this PR](https://github.com/woocommerce/woocommerce-android/pull/11819) where we started using the new endpoint for AI features in WooCommerce app. 

For context. There was already an existing method `JetpackAIRestClient` to send queries to `/jetpack-ai-query` endpoint. This method was intended to be used for `voice-to-content` feature in Jetpack app. But after multiple testing I wasn't able to re use the existing implementation to make it work for our purpose of asking questions to Jetpack AI, so I've provided an additional implementation to do this.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/JetpackAIFragment.kt
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/JetpackAIFragment.kt
@@ -130,7 +130,7 @@ class JetpackAIFragment : StoreSelectingFragment() {
 
     private fun setJetpackAIQueryQuestionButton() {
         jetpack_ai_query_question.setOnClickListener {
-            siteStore.sites[0].let {
+           selectedSite?.let {
                 lifecycleScope.launch {
                     val question = "What is WooCommerce?"
                     prependToLog("Loading Jetpack AI Query... $question")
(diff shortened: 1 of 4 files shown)


example/src/main/java/org/wordpress/android/fluxc/example/JetpackAIFragment.kt
```diff
@@ -115,7 +116,36 @@ class JetpackAIFragment : StoreSelectingFragment() {
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    private fun setJetpackAIQueryQuestionButton() {
+        jetpack_ai_query_question.setOnClickListener {
+            siteStore.sites[0].let {
```
hafizrahman:
```suggestion
           selectedSite?.let {
```

This stumped me for a bit about why my test site didn't work. Also consider updating other functions in this file that also uses `siteStore.sites[0]`

hafizrahman (approved):
Overall works correctly. Sent a suggestion to update the Example app code. Feel free to merge after taking a look at that.

example/src/main/java/org/wordpress/android/fluxc/example/JetpackAIFragment.kt
JorgeMucientes: Great catch!!

PR #3041, opened 2024-06-27
[GLA Analytics] Add networking support for Google Listings & Ads analytics

### LINKED ISSUE #11810: [GLA Analytics] Add networking support for Google Listings & Ads analytics
We need to add support in the networking layer for the following endpoints:

## Connection

Request to check if the Google Ads account is connected:

* Endpoint: `GET /wc/gla/ads/connection`
* Expected response: `status` field has value `connected`
* Sample response:

```
{
    "id": 1234567890,
    "currency": "USD",
    "symbol": "$",
    "status": "connected"
}
```

## Analytics

Request for campaign analytics:

* Endpoint: `GET /wc/gla/ads/reports/programs`
* Supported arguments:
   * `after`: Date string (default 7 days ago)
   * `before`: Date string (default now)
   * `ids`: Limit result to items with specified product or campaign ids
   * `fields`: Limit totals to the provided fields (`sales`, `spend`, `clicks`, `impressions`, and/or `conversions`)
   * `order`: `asc` or `desc` (default)
   * `orderby`: String (one of the requested fields)
   * `per_page`: Integer (default 200, min 1, max 1000)
   * `next_page`: String (token to retrieve the next page)
   * `interval`: `day` (default), `week`, `month`, `quarter`, `year`
* For this milestone, we can use the following arguments:
   * `before` and `after` to request the selected date range
   * `fields` to limit the response to `sales` and `spend` (we will only use these two fields in M1, but we should support requests for any of the supported fields, which we will use in M2)
   * `orderby` to have the response ordered by `sales`
* Note: If the Google Ads account is connected but there are no campaigns, the API can return `null` values (e.g. the `campaigns` field is `null` rather than an empty array).
* Sample response:

```
{
    "campaigns": [
        {
            "id": 123,
            "name": "One",
            "status": "enabled",
            "subtotals": {
                "sales": 11,
                "spend": 63.14
            }
        },
        {
            "id": 234,
            "name": "Two",
            "status": "enabled",
            "subtotals": {
                "sales": 0,
                "spend": 9.87
            }
        }
    ],
    "intervals": [
        {
            "interval": "2021-09",
            "subtotals": {
                "sales": 0,
                "spend": 38
            }
        },
        {
            "interval": "2021-10",
            "subtotals": {
                "sales": 11,
                "spend": 35.01
            }
        }
    ],
    "totals": {
        "sales": 11,
        "spend": 73.01
    }
}
```

Sample response with no campaigns (and with `data` envelope):
```
{
    "data": {
        "products": null,
        "campaigns": null,
        "intervals": null,
        "totals": {
            "sales": 0,
            "spend": 0
        },
        "next_page": null
    }
}
```

ThomazFB:
Summary
==========
Fix issue https://github.com/woocommerce/woocommerce-android/issues/11810 by introducing the network layer for the Google Ads Programs endpoint, alongside all required models and mappers.

## Request specs

Request for campaign analytics:

* Endpoint: `GET /wc/gla/ads/reports/programs`
* Supported arguments:
   * `after`: Date string (default 7 days ago)
   * `before`: Date string (default now)
   * `ids`: Limit result to items with specified product or campaign ids
   * `fields`: Limit totals to the provided fields (`sales`, `spend`, `clicks`, `impressions`, and/or `conversions`)
   * `order`: `asc` or `desc` (default)
   * `orderby`: String (one of the requested fields)
   * `per_page`: Integer (default 200, min 1, max 1000)
   * `next_page`: String (token to retrieve the next page)
   * `interval`: `day` (default), `week`, `month`, `quarter`, `year`
* For this milestone, we can use the following arguments:
   * `before` and `after` to request the selected date range
   * `fields` to limit the response to `sales` and `spend` (we will only use these two fields in M1, but we should support requests for any of the supported fields, which we will use in M2)
   * `orderby` to have the response ordered by `sales`
* Note: If the Google Ads account is connected but there are no campaigns, the API can return `null` values (e.g. the `campaigns` field is `null` rather than an empty array).
* 

Update release notes:


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/google/WCGoogleAdsCampaign.kt
```diff
@@ -12,6 +14,21 @@ data class WCGoogleAdsCampaign(
...
+                } catch (e: IllegalArgumentException) {
+                    AppLog.w(
+                        AppLog.T.API,
+                        "Unknown campaign status returned: `$value`, defaulting to DISABLED " +
+                            e.message
+                    )
+                    DISABLED
+                }
+            }
+        }
```
ThomazFB, line 32: @hafizrahman I decided to extend upon the implementation of the `Status` enum by bringing your implementation of the String parsing to the enum itself and also reusing it on both mappers introduced so far. Hope that's ok!

PR #3043, opened 2024-07-01
[Google for Woo] Store and REST client functionality to fetch campaign lists.

hafizrahman:
This PR adds the functionality to fetch Google Ads Campaigns list.

To test:
1. You need to prepare a test site with the Google Listings & Ads plugin setup and activated with existing campaigns. Warning that this might be a bit time consuming if it's your first time (about 30 minutes). Setup guide is available here pe5sF9-2Qo-p2
2. Run Example app, and login to the test site,
3. Go to Woo, then select the site,
4. Scroll down to the end and tap the **Google For Woo: Fetch Campaigns List** button,
5. It should list the existing campaigns if they exist in your test site,
6. Switch to a different site with no Google Listings & Ads plugin
7. Test the same button and it should show the error message.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/google/WCGoogleAdsCampaign.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/google/WCGoogleAdsCampaign.kt
@@ -6,8 +6,8 @@ data class WCGoogleAdsCampaign(
     val status: Status?,
     val type: String?,
     val amount: Double?,
-    val country: String?,
-    val targetedLocations: List<String>?
+    val countryISOCode: String?,
+    val targetedCountryISOCodes: List<String>?
 ) {
     enum class Status {
         ENABLED,
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/google/WCGoogleAdsCampaign.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/google/WCGoogleAdsCampaign.kt
@@ -1,0 +1,17 @@
+package org.wordpress.android.fluxc.model.google
+
+data class WCGoogleAdsCampaign(
+    val id: Long?,
+    val name: String?,
+    val status: Status?,
+    val type: String?,
+    val amount: Double?,
+    val country: String?,
+    val targetedLocations: List<String>?
+) {
+    enum class Status {
+        ENABLED,
+        DISABLED,
+        REMOVED
+    }
+}
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
@@ -36,22 +36,26 @@ class WCGoogleStore @Inject constructor(
             }
         }
 
-    /**
-     * Fetches the Google Ads campaigns.
-     *
-     * @return WooResult<List<WCGoogleAdsCampaign>> a list of Google Ads campaigns. Optionally,
-     * passes error, too.
-     */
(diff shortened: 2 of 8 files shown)


hafizrahman:
Setting back to draft as I need to add a bit more

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/google/WCGoogleAdsCampaign.kt
```diff
@@ -0,0 +1,17 @@
+package org.wordpress.android.fluxc.model.google
+
+data class WCGoogleAdsCampaign(
+    val id: Long?,
+    val name: String?,
+    val status: Status?,
+    val type: String?,
+    val amount: Double?,
+    val country: String?,
+    val targetedLocations: List<String>?
```
JorgeMucientes: Minor NP, but I'd indicate in the naming of the field if these are country ISO codes or the names of the countries or whatever format the API returns. I think it might help quickly grasp what values are expected there :)

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
```diff
@@ -15,7 +17,8 @@ import javax.inject.Singleton
 @Singleton
 class WCGoogleStore @Inject constructor(
```
JorgeMucientes, line 18: Didn't notice in the first review, but I'm thinking that it might make sense to name this call `WCGoogleAdsStore`, wdyt?

PS: And same for the rest client.

JorgeMucientes (approved):

JorgeMucientes:
Nice job, @hafizrahman! It works well, and the code looks good. I just left a couple of minor NPs. Feel free to address them 🙂.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
hafizrahman: Thanks for the feedback! So just like the plugin's name "Google Listings & Ads", Google Ads is actually just one part of the features that the plugin offers. Google Ads specifically refers to the paid campaign feature. There is also Google Listings, which is a free feature to list in-store products for free and can make them appear in Google Maps or Search (although idk what's the details yet here).

The Ads part and Listings part use different endpoints in the API, etc, so having an umbrella name like this for the store and REST API I think makes sense, in case we want to support all the features in the future.

Finally, they're also going to rename the plugin to Google for WooCommerce pretty soon, and I feel the name WCGoogleStore luckily makes even more sense that way.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/google/WCGoogleAdsCampaign.kt
hafizrahman: Good point! I'll double check and update it.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
JorgeMucientes: Makes sense! Thanks for providing the context Hafiz 👍🏼

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/google/WCGoogleAdsCampaign.kt
hafizrahman: Added https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3043/commits/f70a60ca712444f3467c22a26121af6cf7828622

PR #3052, opened 2024-07-16
[Woo] Add store and REST functionality to fetch Google Ads impressions and clicks

hafizrahman:
This PR updates `WCGoogleStore` with a function to fetch impressions and clicks, with the matching REST client function.

Technically this hits the same endpoint as the existing `fetchAllPrograms()` function. However, I found its `restClient.fetchAllPrograms()` functionality to only support one field at a time, while for my work I need to get multiple fields at once (impressions and clicks). Another reason for adding separate function is because the existing function is being used in a separate, ongoing project, and I don't want to introduce breaking changes.

To test, please use the WooCommerce PR: https://github.com/woocommerce/woocommerce-android/pull/12030

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
@@ -3,6 +3,7 @@ package org.wordpress.android.fluxc.store
 import org.wordpress.android.fluxc.model.SiteModel
 import org.wordpress.android.fluxc.model.google.WCGoogleAdsCampaign
 import org.wordpress.android.fluxc.model.google.WCGoogleAdsCampaignMapper
+import org.wordpress.android.fluxc.model.google.WCGoogleAdsCardStats
 import org.wordpress.android.fluxc.model.google.WCGoogleAdsPrograms
 import org.wordpress.android.fluxc.model.google.WCGoogleAdsProgramsMapper
 import org.wordpress.android.fluxc.network.BaseRequest.GenericErrorType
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
```diff
@@ -42,28 +42,29 @@ class WCGoogleStore @Inject constructor(
...
+        coroutineEngine.withDefaultContext(API, this, "fetchGoogleAdsCampaigns") {
+            val response = restClient.fetchGoogleAdsCampaigns(site, excludeRemovedCampaigns)
+            when {
+                response.isError -> WooResult(response.error)
+                response.result != null -> {
+                    val campaigns = response.result.map { campaignMapper.mapToModel(it) }
+                    WooResult(campaigns)
+                }
+
+                else -> WooResult(emptyList())
```
hafizrahman, line 66: Just a formatting update

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
```diff
@@ -88,6 +89,31 @@ suspend fun fetchGoogleAdsCampaigns(
             }
         }
 
+    suspend fun fetchImpressionsAndClicks(
+        site: SiteModel,
+        startDate: String,
+        endDate: String
+    ): WooResult<Pair<Double, Double>> =
```
hichamboushaba: np, I would vote againt returning a `Pair` here, it's ambiguous and we can't know which value is the clicks and which is the impressions, creating a simple `data class` to hold the values will be a better approach IMO.

hichamboushaba (approved):
Nice work @hafizrahman, I left just one np comment, but it's not blocking.

PS: This is just a code review, as I couldn't test the functionality (more on this [here](https://github.com/woocommerce/woocommerce-android/pull/12013#issuecomment-2233819149)), but I think a code review is enough for this, the functionality will be reviewed in the other PR, we'll just need to delay merging until the other PR is approved.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCGoogleStore.kt
hafizrahman: Sounds good! I'll update it, thank you!

PR #3054, opened 2024-07-18
Update WCOrderStore to support updating payment method

malinajirka:
This PR enables the client app to update not just the order status but also the `payment_method` and `payment_method_title`. This is used by the Woo app to set the payment method when an order is marked as paid with cash.

To test:
Automated tests + [testing the PR in Woo](https://github.com/woocommerce/woocommerce-android/pull/12067) should be enough.

[More info can be found on the PR in Woo.](https://github.com/woocommerce/woocommerce-android/pull/12067)

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
@@ -662,7 +662,13 @@ class WCOrderStore @Inject constructor(
                 // Ensure the code gets executed even when the VM dies - eg. when the client app marks an order as
                 // completed and navigates to a different screen.
                 val remoteUpdateResult: OnOrderChanged = withContext(NonCancellable) {
-                    val remotePayload = wcOrderRestClient.updateOrderStatus(orderModel, site, newStatus.statusKey)
+                    val remotePayload = wcOrderRestClient.updateOrderStatusAndPaymentMethod(
+                        orderModel,
+                        site,
+                        newStatus.statusKey,
+                        paymentMethodId,
+                        paymentMethodTitle
+                    )
                     if (remotePayload.isError) {
                         revertOrderStatus(remotePayload)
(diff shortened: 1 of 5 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -632,15 +632,26 @@ class WCOrderStore @Inject constructor(
         orderId: Long,
         site: SiteModel,
         newStatus: WCOrderStatusModel
+    ): Flow<UpdateOrderResult> =
```
malinajirka, line 635: 📓 I decided to create a new method instead of renaming the original one to keep the client code more readable -> clearly splitting which calling sites want to update both status and Payment Method and which ones are updating only the status.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt
```diff
@@ -632,15 +632,26 @@ class WCOrderStore @Inject constructor(
...
+
+    suspend fun updateOrderStatusAndPaymentMethod(
+        orderId: Long,
+        site: SiteModel,
+        newStatus: WCOrderStatusModel,
+        newPaymentMethodId: String?,
+        newPaymentMethodTitle: String?,
     ): Flow<UpdateOrderResult> {
-        return coroutineEngine.flowWithDefaultContext(API, this, "updateOrderStatus") {
+        return coroutineEngine.flowWithDefaultContext(API, this, "updateOrderStatusAndPaymentMethod") {
```
malinajirka, line 645: 📓 This new identifier will be now printed for both `updateOrderStatus` and `updateOrderStatusAndPaymentMethod` => since this is used for internal logging only, this change is imo fine.

AnirudhBhat (approved):
Thanks for working on this @malinajirka LGTM :shipit:

PR #3056, opened 2024-07-21
Include productId and customerId on order list id

atorresveiga:
### Description
This pull request introduces changes to the ListDescriptorUniqueIdentifier calculation to include both productId and customerId fields. This improvement will:

- Improve cache utilization: By incorporating `productId` and `customerId`, the identifier will now uniquely represent lists based on these filters. This allows us to determine if cached data can be reused for specific product-customer combinations, potentially reducing API calls.
- Enhance list differentiation: The updated identifier will clearly distinguish between lists with and without `productId` and `customerId` filters.

### Testing
It is better to test this PR using the [Woo PR](https://github.com/woocommerce/woocommerce-android/pull/12094)


ThomazFB (approved):
Looks good! Reviewed through https://github.com/woocommerce/woocommerce-android/pull/12094

PR #3057, opened 2024-07-22
[Migrate] Kotlin Android Extensions to ViewBinding

ParaskP7:
Depends On:
 - [x] [ADC#32](https://github.com/Automattic/android-dependency-catalog/pull/32)
 - [x] Update ADC to `1.17.0`

### Description

This PR is about migrating from the deprecated `kotlin-android-extensions` plugin, replacing Kotlin synthetics with ViewBinding and using `kotlin-parcelize` instead.

FYI: The `kotlin-android-extensions` plugin has been deprecated for years now and it is now preventing us from updating to a newer Kotlin version (ie. [Kotlin 1.9.25](https://github.com/JetBrains/kotlin/releases/tag/v1.9.25)).

### To Test

- Also, if you want to be thorough about reviewing the `kotlinx.parcelize.Parcelize` import change, you could quickly smoke test, either the [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android) and/or [WCAndroid](https://github.com/woocommerce/woocommerce-android), with this version of `FluxC`, or via composite builds, and see if it works as expected.


wzieba (approved):
Looks good, great work @ParaskP7 🙌 . Thanks for improving some log messages along the way.

btw: during the review, I started to wonder how often the `example` app in FluxC is used nowadays. Maybe it's something we could discuss internally what's the future for this module.

Not merging as I understand we need to update versions catalog 👍

ParaskP7:
> Looks good, great work @ParaskP7 🙌 . Thanks for improving some log messages along the way.

Awesome, thanks for the quick review @wzieba ! 🙇 ❤️ 

> btw: during the review, I started to wonder how often the example app in FluxC is used nowadays.

Hmmm, to my understanding, to some extend, but to be honest, I am actually not sure. I bet sometimes it might help to use this `example` app, to speed-up development and testing, while other times, using included builds might be preferred... 🤷 

We could ask JP/WPAndroid and WCAndroid devs this question directly and get a quick feedback on it. 🤔 

> Maybe it's something we could discuss internally what's the future for this module.

👍 💯 

> Not merging as I understand we need to update versions catalog 👍

Yes, thanks you, I am in the process of creating the new [1.17.0](https://github.com/Automattic/android-dependency-catalog/releases/tag/1.17.0) release, will then push an update commit and merge it myself. 👍

PR #3059, opened 2024-07-23
[Build] Update Kotlin to 1.9.25 (++)

ParaskP7:
Depends On:
 - [x] [ADC#34](https://github.com/Automattic/android-dependency-catalog/pull/34)
 - [x] Update ADC to `1.18.0`

### Description

This PR updates Kotlin to [1.9.25](https://github.com/JetBrains/kotlin/releases/tag/v1.9.25).

This update isn't that necessary, but should be considered anyway as it will do the following:
- Align this library, to that Kotlin version, which is already used by most of this library's clients and libraries.
- Unblock the usage of the [Dependency Analysis](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) Gradle plugin for this project.
- Prepares this project for the Kotlin 2+ major update.

FYI: Since updating Kotlin alone wasn't enough, due to various build failures, this update also incorporates the below 4 additional updates:
- `detekt` -> [1.23.6](https://github.com/detekt/detekt/releases/tag/v1.23.6)
- `androidx-room` -> [2.6.1](https://developer.android.com/jetpack/androidx/releases/room#2.6.1)
- `dagger` -> [2.51.1](https://github.com/google/dagger/releases/tag/dagger-2.51)
- `kotlinx-coroutines` -> [1.8.1](https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.8.1)

### Test

- In addition to that, please be thorough about reviewing this `Kotlin` and 4 extra updates, by quickly smoke test both, the [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android) and [WCAndroid](https://github.com/woocommerce/woocommerce-android), with this version of `FluxC`, or via composite builds, and see if everything is working as expected.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -60,7 +60,7 @@ allprojects {
         debug = false
     }
 
-    tasks.withType(KotlinCompile).all {
+    tasks.withType(KotlinCompile).configureEach {
         kotlinOptions {
             jvmTarget = JavaVersion.VERSION_1_8
         }
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
 plugins {
     // Those declarations are just a workaround for a false-positive Kotlin Gradle Plugin warning
     // https://youtrack.jetbrains.com/issue/KT-46200
@@ -56,6 +58,12 @@ allprojects {
         ignoreFailures = false
         parallel = true
         debug = false
+    }
+
+    tasks.withType(KotlinCompile).all {
+        kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8
+        }
     }
 }
 
--- a/fluxc/schemas/org.wordpress.android.fluxc.persistence.WPAndroidDatabase/27.json
+++ b/fluxc/schemas/org.wordpress.android.fluxc.persistence.WPAndroidDatabase/27.json
@@ -83,10 +83,10 @@
           }
         ],
         "primaryKey": {
-          "autoGenerate": false,
           "columnNames": [
             "localSiteId"
-          ]
+          ],
+          "autoGenerate": false
         },
         "indices": [],
         "foreignKeys": []
@@ -139,10 +139,10 @@
           }
         ],
         "primaryKey": {
-          "autoGenerate": true,
           "columnNames": [
             "id"
-          ]
+          ],
+          "autoGenerate": true
         },
         "indices": [
           {
@@ -181,10 +181,10 @@
           }
         ],
         "primaryKey": {
-          "autoGenerate": true,
           "columnNames": [
             "id"
-          ]
+          ],
+          "autoGenerate": true
         },
         "indices": [],
         "foreignKeys": [
@@ -237,10 +237,10 @@
           }
         ],
         "primaryKey": {
-          "autoGenerate": true,
           "columnNames": [
             "id"
-          ]
+          ],
+          "autoGenerate": true
         },
         "indices": [],
         "foreignKeys": [
@@ -377,10 +377,10 @@
           }
         ],
         "primaryKey": {
-          "autoGenerate": true,
           "columnNames": [
             "id"
-          ]
+          ],
+          "autoGenerate": true
         },
         "indices": [],
         "foreignKeys": []
@@ -415,11 +415,11 @@
           }
         ],
         "primaryKey": {
-          "autoGenerate": false,
           "columnNames": [
             "siteLocalId",
             "type"
-          ]
+          ],
+          "autoGenerate": false
         },
         "indices": [],
         "foreignKeys": []
@@ -490,10 +490,10 @@
           }
         ],
         "primaryKey": {
-          "autoGenerate": false,
           "columnNames": [
             "date"
-          ]
+          ],
+          "autoGenerate": false
         },
         "indices": [],
         "foreignKeys": []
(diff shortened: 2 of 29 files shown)


fluxc/schemas/org.wordpress.android.fluxc.persistence.WPAndroidDatabase/27.json
wzieba, line 1: Why are these updates necessary? And why in this specific migration files? Looking at e.g. `26.json`, the `autoGenerate` key is below `columnNames`

ParaskP7: This was done automatically (as usual), I just committed the changes. I guess, most probably due to the `room` update ([2.6.1](https://developer.android.com/jetpack/androidx/releases/room#2.6.1)), the structure of those `*.json` files is now updated too... 🤷

wzieba: > I guess, most probably due to the room update ([2.6.1](https://developer.android.com/jetpack/androidx/releases/room#2.6.1)), the structure of those *.json files is now updated too... 🤷

But in this case, I'd expect all other `*.json` file to be updated as well, not only very specific `27.json`.

I've reverted the ordering in this file and re-run build, and it wasn't changed back.

I'm not very concerned about this particular change, as it's only ordering, but more about a precedence of editing old database schema, which should never be edited. Taking all of this into account, can we revert this change?

ParaskP7: > But in this case, I'd expect all other *.json file to be updated as well, not only very specific 27.json.

Well, I am not sure what I would expect to be honest... 🤷 

> I've reverted the ordering in this file and re-run build, and it wasn't changed back.

Yea, I tried that myself just now too, after reverting this change, and couldn't make this change back... 🤷 

PS: This behavior was consistent for me, because it happened twice to me, when I was investigating how to update Kotlin to `1.9.25`, and then, once more, when I actually applied my solution (post investigation). So, I personally saw that happening automatically, and thus, wanted to commit that change so that other won't have to afterwards, thinking that if I don't commit it now, it will anyway pop-up later for someone else. 🤔 

> I'm not very concerned about this particular change, as it's only ordering...

Exactly my thoughts! 👍 

> ...but more about a precedence of editing old database schema, which should never be edited.

This is a bit debatable, but I understand the concern and I find it valid nevertheless. 👍 

FYI: When I did some research in the past on these updates, I figured out that those are not mandatory and will not fail the migration and as such upgrading the app. It is just a good practice to have this file committed in the version control system and have it in the codebase. For more info check the [documentation](https://developer.android.com/reference/android/arch/persistence/room/Database#exportSchema()).

> Taking all of this into account, can we revert this change?

Done: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3059/commits/ce00973b295ca8058dd1bdc0d42cd1e48431d88c

build.gradle
```diff
@@ -57,6 +59,12 @@ allprojects {
         parallel = true
         debug = false
     }
+
+    tasks.withType(KotlinCompile).all {
```
wzieba:
```suggestion
    tasks.withType(KotlinCompile).configureEach {
```

WDYT about configuring this task lazily? The impact here will be not noticeable, but I think it's a good hygiene to use more performant alternative.

fluxc/src/main/java/org/wordpress/android/fluxc/tools/CoroutineEngine.kt
```diff
@@ -44,12 +44,14 @@ open class CoroutineEngine
...
-                    .onEach { appLog.d(tag, "${caller.javaClass.simpleName}: $loggedMessage OnEvent: $it") }
-                    .onCompletion { appLog.d(tag, "${caller.javaClass.simpleName}: $loggedMessage Completed") }
+    ): Flow<RESULT_TYPE> {
+        val safeContext = context.minusKey(Job) // Remove Job from context to make it safe for flows
+        return flow { block() }
+            .flowOn(safeContext)
+            .onStart { appLog.d(tag, "${caller.javaClass.simpleName}: $loggedMessage Started") }
+            .onEach { appLog.d(tag, "${caller.javaClass.simpleName}: $loggedMessage OnEvent: $it") }
+            .onCompletion { appLog.d(tag, "${caller.javaClass.simpleName}: $loggedMessage Completed") }
+    }
```
wzieba: I've prepared a suggestion in #3062, let's discuss it there

ParaskP7: Thanks for that @wzieba , much appreciated, let me check on it and we shall proceed accordingly with it. 🙇

build.gradle
ParaskP7: This is now done: da917d92a406d669f657acc95f6d413dbee57c84

fluxc/src/main/java/org/wordpress/android/fluxc/tools/CoroutineEngine.kt
ParaskP7: This is now done: 77e2dfb156dc0025642db0796fd7d0289d02c020

wzieba (approved):
Looks good, great work 🙌 !

PR #3062, opened 2024-07-24
Do not alter context in `CoroutineEngine#flowWithDefaultContext`

wzieba:
This PR is a review suggestion for #3059 .

I thought it'll be easier to propose a change than to discuss it without code.

### Description

`CoroutineEngine#flowWithDefaultContext` doesn't communicate that it'll change context, and it's not something that API consumers can predict without going into implementation details.

I think it'll be better to not alter `CoroutineContext` in this method. For the failing test, we'll use `EmptyCoroutineContext` which is default context for `TestCoroutineScope`, but doesn't contain job.

Anyway, `TestCoroutineScope().coroutineContext` shouldn't be used in the first place as

> Accessing this property in general code is not recommended for any purposes except accessing the Job instance for advanced usages.

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/tools/CoroutineEngine.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/tools/CoroutineEngine.kt
@@ -45,9 +45,8 @@ open class CoroutineEngine
         loggedMessage: String,
         block: suspend FlowCollector<RESULT_TYPE>.() -> Unit
     ): Flow<RESULT_TYPE> {
-        val safeContext = context.minusKey(Job) // Remove Job from context to make it safe for flows
         return flow { block() }
-            .flowOn(safeContext)
+            .flowOn(context)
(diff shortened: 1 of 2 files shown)


ParaskP7 (approved):
👋 @wzieba and thanks for proposing this change! 🥇 

> `CoroutineEngine#flowWithDefaultContext` doesn't communicate that it'll change context, and it's not something that API consumers can predict without going into implementation details.

I agree, I was trying to be smart about it, but this change is not worth-it, maybe even dangerous... 🫣

> I think it'll be better to not alter `CoroutineContext` in this method. For the failing test, we'll use `EmptyCoroutineContext` which is default context for `TestCoroutineScope`, but doesn't contain job.
> 
> Anyway, `TestCoroutineScope().coroutineContext` shouldn't be used in the first place as... Accessing this property in general code is not recommended for any purposes except accessing the Job instance for advanced usages.

Totally! 💯

PR #3066, opened 2024-07-24
Update the site duplication logic to not throw when IDC occurs

hichamboushaba:
### Issue
We have a logic that detects duplicate sites when inserting them to the database, this logic is causing an [issue](https://github.com/woocommerce/woocommerce-android/issues/12037) in the WCAndroid app where users with duplicate websites under their WordPress.com account can't login to them, the duplication of WordPress.com sites can happen during an "Identity Crisis" (**IDC**) (PCYsg-Jfw-p2).

This PR fixes the issue by relaxing the condition to allow inserting duplicate websites coming from the WPCom API to allow the login in the client apps.

### Rationale of the change of the PR
Looking at the historic of the logic mentioned above, I believe it was added due to special circumstances in the WordPress app to deal with duplication that could happen when an XMLRPC website gets added later as Jetpack website or the other way around, but for some reason it also checks for duplicates when both websites are coming from the WordPress.com API, and I didn't find an explanation of why this was added.
This logic to prevent adding duplicates is not available in any other client (as far I was able to check), the login works to those duplicated websites work well with Calypso and the iOS apps:

| Calypso | Jetpack iOS app | Woo iOS app |
| --- | --- | --- |
|  | IMG_0115 | IMG_0116 |

Given this, I think the Android apps should behave the same, as the backend allows those duplicate websites for a reason (to preserve the user's data after the IDC), we shouldn't lock the users out of signing in to those websites.

#### Example app
1. DM and I'll share with you the credentials of an account with IDC.
2. Login using the example app.
3. Go to the Sites screen, and tap on Log Sites.
4. Confirm the site with IDC is listed.

#### WCAndroid
https://github.com/woocommerce/woocommerce-android/pull/12143

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.kt
@@ -107,7 +107,7 @@ class SiteSqlUtils
      * 5. Exists in the DB, originally an XML-RPC site, and matches by XMLRPC_URL -> UPDATE
      * 6. Not matching any previous cases -> INSERT
      */
-    @Suppress("LongMethod", "ReturnCount")
+    @Suppress("LongMethod", "ReturnCount", "ComplexMethod")
     @Throws(DuplicateSiteException::class)
     fun insertOrUpdateSite(site: SiteModel?): Int {
         if (site == null) {
@@ -176,8 +176,9 @@ class SiteSqlUtils
             if (!siteResult.isEmpty()) {
                 AppLog.d(DB, "Site found using XML-RPC url: " + site.xmlRpcUrl)
                 // Four possibilities here:
-                // 1. DB site is WP.com, new site is WP.com:
-                // Something really weird is happening, this should have been caught earlier --> DuplicateSiteException
+                // 1. DB site is WP.com, new site is WP.com with the same siteId:
+                // The site could be having an "Identity Crisis", while this should be fixed on the site itself,
+                // it shouldn't block sign-in -> proceed
                 // 2. DB site is WP.com, new site is XML-RPC:
                 // It looks like an existing Jetpack-connected site over the REST API was added again as an XML-RPC
                 // Wed don't allow this --> DuplicateSiteException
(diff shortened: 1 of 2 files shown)


hichamboushaba:
@aforcier before moving forward with this PR, can I get your opinion on the changes and the explanation I wrote above in the description? does it make sense for you or am I missing some cases here?

Sorry for pinging you directly, I saw you edited some parts of this logic in the past, and I know it's been a long time ago, but if you still remember something about the reasons, it would be really helpful to have more confidence in the suggested change.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.kt
```diff
@@ -172,12 +172,14 @@ class SiteSqlUtils
...
-                    .endWhere().asModel
-            if (!siteResult.isEmpty()) {
+                    .endWhere()
+                    .asModel
+            if (siteResult.isNotEmpty()) {
                 AppLog.d(DB, "Site found using XML-RPC url: " + site.xmlRpcUrl)
                 // Four possibilities here:
-                // 1. DB site is WP.com, new site is WP.com:
-                // Something really weird is happening, this should have been caught earlier --> DuplicateSiteException
+                // 1. DB site is WP.com, new site is WP.com with different siteIds:
```
hichamboushaba, line 180: We can safely here assume that the websites have different site IDs, because the case of having same siteIds is already handled [above](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/cf23c7cd8cca570d42d697e27b822a7200ebd6ff/fluxc/src/main/java/org/wordpress/android/fluxc/persistence/SiteSqlUtils.kt#L143-L152).

Also the API will never return duplicate websites that have the same ID (unless there is some bug that causes it), so the logic above considers that the website to be inserted is the same as the one we have in the DB, and then updates it.

hichamboushaba:
@nbradbury sorry for pinging you multiple times these days, this is one additional PR to fix an issue with the login, please read the PR description, and tell me if it makes sense to you.
IMO, there is no need to update FluxC right now in WPAndroid; the changes should be transparent and will be brought in using other FluxC updates, but please let me know if you prefer to have an explicit PR in WPAndroid and I can create it.

JorgeMucientes (approved):

JorgeMucientes:
The rationale on the changes makes sense to me and the code changes look good and work as expected. Nice job!

PR #3070, opened 2024-08-02
[Woo] Metadata refactoring 1

hichamboushaba:
This PR prepares to update the storage layer to allow saving the product's metadata in a separate SQL table.

Before starting on the main change, I found out that our usage of the metadata wasn't optimal:
- We expose the raw JSON in some places.
- We expose the DB entity `OrderMetaDataEntity` instead of the common `WCMetaData` in other places.
- We expose `WCMetaData` in some other places.

This PR aims to improve the usage in the orders models by exposing the `WCMetaData` for all (or at least the majority in case I missed some cases) the usages.

And below is a list of the changes in detail:
- Update the WCMetaData model to expose a new model, `WCMetaDataValue` instead of `Any` for the values. The `WCMetaDataValue` allows representing the different cases of the value, and this differentiation allows the code to identify Json values without resorting to regex.
- Update the orders store to expose `WCMetaData` instead of `OrderMetaDataEntity`.
- Update the insertion logic of orders to align with the above change.
- Update `OrderEntity` to expose a `List<WCMetaData>` instead of raw json.

The **next PR** will focus on products, and the changes will be:
- Update the DB table `OrderMetaDataEntity` to be a common table for all meta data across all objects (like what WordPress Core does), this will now be easier since this is not used anymore by the upper layers.
- Save the product metadata to that table.
- Remove `metadata` completely from `WCProductModel` and update all usages to align with this.

### Testing
Please refer to https://github.com/woocommerce/woocommerce-android/pull/12214 for the testing steps.


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCMetaData.kt
```diff
@@ -1,28 +1,76 @@
...
 import com.google.gson.annotations.SerializedName
 
 data class WCMetaData(
     @SerializedName(ID) val id: Long,
-    @SerializedName(KEY) val key: String?,
-    @SerializedName(VALUE) val value: Any,
-    @SerializedName(DISPLAY_KEY) val displayKey: String?,
-    @SerializedName(DISPLAY_VALUE) val displayValue: Any?
+    @SerializedName(KEY) val key: String,
+    @SerializedName(VALUE) val value: JsonElement,
```
hichamboushaba, line 12: Feel free to share your thoughts on using `JsonElement` here, personally I think it made the code easier to read, but I would love to hear your opinion on this.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCMetaData.kt
```diff
@@ -1,28 +1,76 @@
...
+            displayKey?.let { key -> it.addProperty(DISPLAY_KEY, key) }
+            displayValue?.let { value -> it.add(DISPLAY_VALUE, value) }
+        }
+    }
+
     companion object {
-        const val ID = "id"
+        private const val ID = "id"
         const val KEY = "key"
         const val VALUE = "value"
```
hichamboushaba, line 57: My goal is to make those `private` as well in the next PR, upper layers shouldn't depend on the json keys of the object.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCMetaData.kt
```diff
@@ -1,30 +1,92 @@
...
-    @SerializedName(VALUE) val value: Any,
-    @SerializedName(DISPLAY_KEY) val displayKey: String?,
-    @SerializedName(DISPLAY_VALUE) val displayValue: Any?
+    @SerializedName(ID)
+    val id: Long,
+    @SerializedName(KEY)
+    val key: String,
+    @JsonAdapter(WCMetaDataValue.WCMetaDataValueJsonAdapter::class)
+    @SerializedName(VALUE)
+    val value: WCMetaDataValue,
```
hichamboushaba, line 15: On the first attempt of this PR, I used just `JsonElement` to represent the `value` property, but I wasn't satisfied with it as it adds a direct dependency on `Gson` for our domain model here, and also there was an issue with it as `toString` wasn't behaving as expected, it returned String values wrapped in double quotes.

With this second attempt, I added a `sealed class` to represent the different possibilities that the `value` can have, and even though it still uses `JsonElement` internally, it's not exposed to the upper layers, which would make replacing it with a different library easier in the future.

Please feel free to share you thoughts.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -124,10 +125,14 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
     val attributeList: Array<ProductAttribute>
         get() = Gson().fromJson(attributes, Array<ProductAttribute>::class.java) ?: emptyArray()
 
+    val parsedMetaData: List<WCMetaData>
+        get() = runCatching { JsonParser().parse(metadata).asJsonArray }
+            .getOrNull()
+            ?.mapNotNull { element -> (element as? JsonObject)?.let { WCMetaData.fromJson(it) } }
+            ?: emptyList()
```
hichamboushaba, line 132: This is just a temporary change to prepare for the next PR where we'll remove the metadata completely from the `WCProductModel`.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCMetaData.kt
```diff
@@ -2,29 +2,85 @@ package org.wordpress.android.fluxc.model
...
+                key = json[KEY].asString,
+                value = WCMetaDataValue.fromJsonElement(json[VALUE]),
+                displayKey = json[DISPLAY_KEY]?.asString,
+                displayValue = json[DISPLAY_VALUE]?.let { WCMetaDataValue.fromJsonElement(it) }
+            )
+        }
+
         fun addAsMetadata(metadata: JsonArray, key: String, value: String) {
             val item = JsonObject().also {
+                it.addProperty(ID, 0)
```
hichamboushaba, line 83: I intend to remove `addAsMetadata` completely on the next PR, but before doing so, we need to have this default value here to avoid NPEs.
I asked about the need of this logic [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2883), depending on the answer I'll try to choose the right approach in the next PR.

hafizrahman:
👋🏼 removing myself because I'm back to iOS now.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderMetaDataEntity.kt
```diff
@@ -39,6 +38,6 @@ data class OrderMetaDataEntity(
     fun toDomainModel() = WCMetaData(
         id = id,
         key = key,
-        value = JsonParser().parse(value)
+        value = value
```
JorgeMucientes: I was becoming a bit crazy about this 😅 , because the FluxC changeset [added to PR](https://github.com/woocommerce/woocommerce-android/pull/12214/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7) didn't contain this change which fixes a crash.
I was experiencing a crash when I manually adding custom fields to an order and the values contained some blank spaces:
```
com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 8 path $
	at com.google.gson.JsonParser.parseReader(JsonParser.java:76)
	at com.google.gson.JsonParser.parseString(JsonParser.java:51)
	at com.google.gson.JsonParser.parse(JsonParser.java:115)
	at org.wordpress.android.fluxc.persistence.entity.OrderMetaDataEntity.toDomainModel(OrderMetaDataEntity.kt:42)
	at org.wordpress.android.fluxc.store.WCOrderStore.getOrderMetadata(WCOrderStore.kt:438)
	at org.wordpress.android.fluxc.store.WCOrderStore$getOrderMetadata$1.invokeSuspend(Unknown Source:17)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
```

The problem is fixed with the new way of handling raw strings👍🏼

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCMetaData.kt
```diff
@@ -2,29 +2,85 @@ package org.wordpress.android.fluxc.model
...
+    val displayKey: String? = null,
+    @JsonAdapter(WCMetaDataValue.WCMetaDataValueJsonAdapter::class)
+    @SerializedName(DISPLAY_VALUE)
+    val displayValue: WCMetaDataValue? = null
 ) {
+    constructor(id: Long, key: String, value: String) : this(id, key, WCMetaDataValue.fromRawString(value))
+
+    /**
+     * Verify if the Metadata key is not null or a internal store attribute
+     * @return false if the `key` starts with the `_` character
```
JorgeMucientes, line 26: Is this event possible? I think the backend already strips any starting `_` from the key names. 

Ok nevermind I see the `_` prefix is used for internal attributes

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderMetaData.kt
```diff
@@ -3,34 +3,18 @@ package org.wordpress.android.fluxc.network.rest.wpcom.wc.order
...
+    operator fun invoke(orderDto: OrderDto): List<WCMetaData> {
         return parseMetaDataJSON(orderDto.meta_data)
-            ?.asSequence()
-            ?.filter { it.isDisplayableAttribute || it.key in WCMetaData.SUPPORTED_KEYS }
-            ?.map { it.asOrderMetaDataEntity(orderDto.id, localSiteId) }
-            ?.filter { it.value.isNotEmpty() && it.value.matches(jsonRegex).not() }
-            ?.toList()
+            ?.filter {
+                (it.isDisplayable || it.key in WCMetaData.SUPPORTED_KEYS)
+                        && !it.isJson
```
JorgeMucientes, line 16: Not sure the logic to check if `!it.json` is working as expected. I've manually added a custom field with a json value and I can see it displayed in the custom fields section for the order: 
Screenshot 2024-08-06 at 12 26 45

JorgeMucientes (commented):
Excellent job @hichamboushaba. Great refactor.
I tested the custom fields feature throughly and just found an issue. Custom fields with json values are being displayed. Aside from that everything worked great!

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/persistence/entity/OrderMetaDataEntity.kt
hichamboushaba: Sorry about this 🤦, and thanks for checking the changeset to confirm 🙏

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderMetaData.kt
hichamboushaba: Thanks for mentioning this @JorgeMucientes, this showed a behavior that I wasn't aware of in the wp-admin `Custom Fields` section, and it'll allow us to update our logic to align with it.

But for our task here, I think the logic is working as expected, the value entered in `wp-admin` is not a valid Json, it's a `String` value:
Screenshot 2024-08-07 at 10 23 52

And that's why it doesn't get filtered with the `isJson` property, if you add a properly formatted JSON like below (which can be done only using the API):
Screenshot 2024-08-07 at 10 27 17

Then it gets filtered out correctly.

WDYT about this? IMO we can keep this as it's working just wp-admin, where the real JSON value is not shown, but the String wrapped JSON is shown.

JorgeMucientes: Oh, I see. I was expecting JSON values could manually be added from wp-admin as well. Fair enough I think its fine to show manually added jsons as strings. Just a warning on that. Before these changes were added, manually added JSONs from wp-admin where being filtered/invisible. However, iOS currently behaves like this PR, manually added json values are displayed. 
So lets keep this as is 👍🏼

JorgeMucientes (approved):

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderMetaData.kt
hichamboushaba: > Just a warning on that. Before these changes were added, manually added JSONs from wp-admin where being filtered/invisible

Yes, I'm aware of this, and I think it was just a bug given that we were parsing the values as `Any`.

PR #3071, opened 2024-08-05
Add new Blaze notification types to FluxC

JorgeMucientes:
Part of: https://github.com/woocommerce/woocommerce-android/issues/12191

Adds new push notification types for Blaze campaign status changes. 

Ideally, we could've added a single `Blaze` type. However, the backend code: D137233-code which was deployed months ago to support these Blaze notifications in Jetpack app, added 4 different Blaze types. So, we need to stick to that. 

The changes should be tested alongside WooCommerce app changes: https://github.com/woocommerce/woocommerce-android/pull/12219


irfano (approved):
LGTM! 👍🏻

PR #3072, opened 2024-08-05
Do not strip payment intent metadata on orders

malinajirka:


malinajirka:
This isn't needed for the Stripe SDK migration after all.

PR #3074, opened 2024-08-09
Woo/meta data refactor 3

hichamboushaba:
First of all, sorry for the size of the PR, fixing the tests caused more changes than I anticipated, and it's not easy to split this work further.

This PR adds the third part of the metadata refactor for Woo parts, the main change of the PR is to handle saving the product's metadata in the separate table `MetaData`, and this happens at every product persist call.

This required some other changes:
- 59b9d2a updated the `WCProductModel` table with two columns to store the bundle min/max sizes, these values were saved as virtual meta (by virtual I mean it's not real metadata coming from the API) data previously, and this was getting on our way to get rid of the `metadata` column.
- ad057bd removes the logic to calculate the `addons` from the WCProductModel, and moves it to the addons DTO itself.

I'll add additional comments below for places where I consider it necessary for understanding.

### Testing
- Smoke test the Products screen in the example app.
- Check https://github.com/woocommerce/woocommerce-android/pull/12252


example/src/test/java/org/wordpress/android/fluxc/wc/product/WCProductStoreTest.kt
```diff
@@ -89,6 +89,23 @@ class WCProductStoreTest {
         )
         WellSql.init(config)
         config.reset()
+
+        val roomDb = Room.inMemoryDatabaseBuilder(appContext, WCAndroidDatabase::class.java)
+            .allowMainThreadQueries()
+            .build()
```
hichamboushaba, line 95: I added this initialization because it's needed for the MetaData table.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCMetaData.kt
```diff
@@ -116,6 +104,10 @@ data class WCMetaData(
                 SESSION_PAGE_VIEWS
             )
     }
+
+    object AddOnsMetadataKeys {
+        const val ADDONS_METADATA_KEY = "_product_addons"
+    }
```
hichamboushaba, line 117: I moved this object here instead of having it on WCProductModel.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/WCProductModel.kt
```diff
@@ -108,11 +105,17 @@ data class WCProductModel(@PrimaryKey @Column private var id: Int = 0) : Identif
     @Column var length = ""
     @Column var width = ""
     @Column var height = ""
+
+    /**
+     * This holds just the subscription keys, for the rest of product's metadata please check [ProductWithMetaData]
+     */
     @Column var metadata = ""
```
hichamboushaba, line 112: **Important:** My goal is still to get rid of this completely, but given the impact on product subscriptions and the big changes it would require, I decided to keep it for now, I'll prioritize the project work first, and if after finishing I still have time, I'll come back to this part to remove the column completely.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/addons/RemoteAddonDto.kt
```diff
@@ -70,4 +72,17 @@ data class RemoteAddonDto(
...
+        @Suppress("SwallowedException", "TooGenericExceptionCaught")
+        fun fromMetaDataValue(value: WCMetaDataValue): List<RemoteAddonDto>? {
+            return try {
+                Gson().run {
+                    fromJson(value.jsonValue, Array<RemoteAddonDto>::class.java)
+                }.toList()
+            } catch (ex: Exception) {
+                null
+            }
+        }
```
hichamboushaba, line 86: This is a replication of the logic that we had on `WCProductModel` above.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductDtoMapper.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.wordpress.android.fluxc.model.WCProductModel
+import org.wordpress.android.fluxc.network.utils.getString
+import javax.inject.Inject
+
+class ProductDtoMapper @Inject constructor(
+    private val gson: Gson,
+    private val stripProductMetaData: StripProductMetaData
+) {
+    @Suppress("LongMethod", "ComplexMethod")
+    fun mapToModel(localSiteId: LocalOrRemoteId.LocalId, dto: ProductDto): ProductWithMetaData {
```
hichamboushaba, line 17: This function is almost the same as `ProductApiResponse#asProductModel` that it replaced, but with some small changes that I'll mention below.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductDtoMapper.kt
```diff
@@ -0,0 +1,146 @@
...
+            maxAllowedQuantity = dto.max_quantity?.let { it.ifEmpty { "0" } }?.toInt() ?: -1
+            groupOfQuantity = dto.group_of_quantity?.toInt() ?: -1
+
+            combineVariationQuantities = dto.combine_variations?.let {
+                it == "yes"
+            } ?: false
+
+            // Save only the subscription data, the rest of the metadata will be saved separately
+            metadata = metaData.filter { it.key in WCProductModel.SubscriptionMetadataKeys.ALL_KEYS }
+                .let { gson.toJson(it) }
```
hichamboushaba, line 114: Now we save to the `metadata` column only the subscription keys.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/product/ProductDtoMapper.kt
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private fun WCProductModel.applyBundledProductChanges(dto: ProductApiResponse) = apply {
+        if (this.type == CoreProductType.BUNDLE.value) {
+            if (dto.bundle_stock_status !in CoreProductStockStatus.ALL_VALUES) {
+                specialStockStatus = dto.bundle_stock_status ?: ""
+            }
+
+            bundleMaxSize = dto.bundle_max_size?.toFloatOrNull()
+            bundleMinSize = dto.bundle_min_size?.toFloatOrNull()
```
hichamboushaba, line 142: these are new columns on the WCProductModel, and we set them directly now instead of using the metadata (as mentioned in the PR description).

JorgeMucientes (commented):
Excellent refactor @hichamboushaba 👏🏼 . 
I've tested: 
 - Product Addons feature following steps from [this PR](https://github.com/woocommerce/woocommerce-android/pull/12252) and everything worked as expected
 - Tested subscriptions and same, everything worked well
 - Smoked tested the product section from sample app and everything worked well. 

Everything worked well on that part

### To double check

While testing, I noticed that the meta data that is not directly tied to the subscription product metadata, is missing. If I open `View custom fields` from order details, I can only see `is_vat_exmpt` field. However, my product contains several other custom fields that are missing and are not displayed. Is that expected at this point and I missed something? Because from what I see [these changes were merged ](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3070). So, I'd expect custom fields to show like in that PR.

hichamboushaba:
Thanks for the review @JorgeMucientes 🙏 

> While testing, I noticed that the meta data that is not directly tied to the subscription product metadata, is missing. If I open View custom fields from order details, I can only see is_vat_exmpt field. However, my product contains several other custom fields that are missing and are not displayed. Is that expected at this point and I missed something?

Currently, we don't offer any screen to show the metadata linked to a given product, the "View custom fields" button in the app lists only the metadata linked to the given order, so what you faced is expected.
With the project we are working on now, we'll add the screen to list the product metadata as well with the ability to edit it.

JorgeMucientes:
Hey @hichamboushaba 

> Currently, we don't offer any screen to show the metadata linked to a given product, the "View custom fields" button in the app lists only the metadata linked to the given order, so what you faced is expected.

I see. I mixed up adding custom fields to products and to orders. I thought that an order that contained products with custom fields would reflect the custom fields for the product (not only for the order itself). Thanks for clarifying! 

Good to go :shipit:

JorgeMucientes (approved):

PR #3075, opened 2024-08-09
[Dependency Analysis] Remove Unused Dependencies

ParaskP7:
Project Thread: paaHJt-6YV-p2
Depends On:
- [x] [ADC#35](https://github.com/Automattic/android-dependency-catalog/pull/35)
- [x] Update ADC to `1.21.0`

### Description

Based on the `unused` related advises generated by the [Dependency Analysis](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) Gradle plugin, this PR removes all unused dependencies from this project.

FYI: As part of this change, the [Dependency Analysis](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) Gradle plugin got also updated to its latest [1.33.0](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CHANGELOG.md#version-1330) version.

### Testing Steps

1. Tooling:
    - Run the `./gradlew buildHealth` task and verify that there is no `unused` related advise remaining within both reports, JSON and txt included:
        - `build-health-report.json` -> Search for `unusedCount`
        - `build-health-report.txt` -> Search for `unused`
    - Check the manually triggered [scheduled build](https://buildkite.com/automattic/wordpress-fluxc-android/builds/3804) and verify that everything works as expected with this newer version of the [Dependency Analysis](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) Gradle plugin ([1.33.0](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CHANGELOG.md#version-1330)).
3. Testing:
    - Also, if you want to be thorough about reviewing the changes, you could quickly smoke test, either the [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android) and/or [WCAndroid](https://github.com/woocommerce/woocommerce-android), with this version of `FluxC`, or via composite builds, and see if it works as expected.

### DIFF
--- a/fluxc-processor/build.gradle
+++ b/fluxc-processor/build.gradle
@@ -1,7 +1,5 @@
 plugins {
     id "java"
-    alias(sharedLibs.plugins.kotlin.jvm)
-    alias(sharedLibs.plugins.kotlin.kapt)
     alias(sharedLibs.plugins.automattic.publishToS3)
 }
 
@@ -16,23 +14,8 @@ java {
 dependencies {
     implementation fluxcAnnotationsProjectDependency
     implementation sharedLibs.google.autoService
-    kapt sharedLibs.google.autoService
+    annotationProcessor sharedLibs.google.autoService
     implementation sharedLibs.squareup.javapoet
-}
-
-dependencyAnalysis {
-    issues {
-        onUnusedDependencies {
-            exclude("org.jetbrains.kotlin:kotlin-stdlib")
-            exclude("com.google.auto.service:auto-service")
-        }
-        onUnusedAnnotationProcessors {
-            exclude("com.google.auto.service:auto-service")
-        }
-        onRedundantPlugins {
-            exclude("kotlin-kapt")
-        }
-    }
 }
 
 project.afterEvaluate {
--- a/fluxc-processor/build.gradle
+++ b/fluxc-processor/build.gradle
@@ -20,6 +20,21 @@ dependencies {
     implementation sharedLibs.squareup.javapoet
 }
 
+dependencyAnalysis {
+    issues {
+        onUnusedDependencies {
+            exclude("org.jetbrains.kotlin:kotlin-stdlib")
+            exclude("com.google.auto.service:auto-service")
+        }
+        onUnusedAnnotationProcessors {
+            exclude("com.google.auto.service:auto-service")
+        }
+        onRedundantPlugins {
+            exclude("kotlin-kapt")
+        }
+    }
+}
+
 project.afterEvaluate {
     publishing {
         publications {
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -107,8 +107,6 @@ dependencies {
     implementation sharedLibs.androidx.recyclerview
     implementation sharedLibs.androidx.constraintlayout
     implementation sharedLibs.androidx.annotation
-    implementation sharedLibs.androidx.arch.core.common
-    implementation sharedLibs.androidx.arch.core.runtime
     implementation sharedLibs.androidx.lifecycle.runtime.ktx
 
     implementation(sharedLibs.wordpress.utils) {
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -119,9 +119,8 @@ dependencies {
 
     implementation sharedLibs.google.gson
 
-    implementation sharedLibs.androidx.room.runtime
-    kapt sharedLibs.androidx.room.compiler
-    implementation sharedLibs.androidx.room.ktx
+    testImplementation sharedLibs.androidx.room.runtime
+    androidTestImplementation sharedLibs.androidx.room.runtime
     androidTestImplementation sharedLibs.androidx.room.testing
 
     // Dagger
--- a/example/build.gradle
+++ b/example/build.gradle
@@ -142,7 +142,6 @@ dependencies {
 
     androidTestImplementation sharedLibs.assertj.core
     androidTestImplementation sharedLibs.androidx.arch.core.testing
-    kaptAndroidTest sharedLibs.google.dagger.compiler
     androidTestCompileOnly sharedLibs.glassfish.javax.annotation
     // Test orchestrator
     androidTestImplementation sharedLibs.androidx.test.runner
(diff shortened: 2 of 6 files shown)


fluxc-processor/build.gradle
```diff
@@ -20,6 +20,21 @@ dependencies {
     implementation sharedLibs.squareup.javapoet
 }
 
+dependencyAnalysis {
+    issues {
+        onUnusedDependencies {
+            exclude("org.jetbrains.kotlin:kotlin-stdlib")
```
wzieba: It's true that stdlib is not used in this module: can we indeed exclude adding this by Kotlin Gradle plugin by introducing `gradle.properties` file with `kotlin.stdlib.default.dependency=false` ?

wzieba (approved):

fluxc-processor/build.gradle
ParaskP7: Interesting point @wzieba thanks! 👍 

Thoughts:
1. Hmmm, do you think it will work, have you tried it? 🤔 
2. I wonder if it is better to just keep it as is, and exclude it on the Dependency Analysis level, than changing `gradle.properties` for the whole project for that matter, just because of `fluxc-processor` and the fact that this module is legacy. Although this change doesn't seem risky, I think it is better to leave this alone, not sure... 🤔

If you really think this is a good move, I'll go ahead and make this change anyway. 👍

ParaskP7: UPDATE: As discussed during our Android Core catch-up, I am going to try and replace Kotlin `kapt` with Java `annotationProcessor` and completely remove the unused `s3` plugins from this and the `fluxc-annotations` modules.

wzieba: Thanks @ParaskP7 ! My one request would be to verify if performance wasn't affected much (maybe gradle-profiler with change of a code that requires `fluxc-annotations` to regenerate code?)

ParaskP7: That's a good idea @wzieba , will try gradle-profiler and see how it goes, wise me luck! 🍀

ParaskP7: Okay @wzieba , this is now done!

1. 5893fa42d560d5f08beb7f221bd81085c220d969

I couldn't benchmark incremental build as there was no `apply-x-change-to` for pure `.txt` file (ie. targeting `wc-wp-api-endpoints.txt`) that Gradle Profiler can provide us, nor a custom scenario for that matter. The closer we could get is `apply-property-resource-change-to` but this adds/removes property entries (ie. `xyz=123`). 🤔 

Also, I couldn't think of any other better place to update the code and test an incremental build as we don't want to verify compilation, but rather code generation. 🤷 

However, I did benchmark at least normal builds (via `fluxc-processor.scenarios`) and everything seems to be okay, see diff below:

<details>
    <summary>fluxc-processor.scenarios</summary>

```txt
fluxc-processor-with-kapt {
    title = "FluxC Processor with KAPT"
    tasks = ["assembleDebug"]
    gradle-args = ["--rerun-tasks"]
    # Caused by: java.lang.IllegalArgumentException: Can only modify Java or Kotlin source files
    # apply-abi-change-to = ["fluxc-processor/src/main/resources/wc-wp-api-endpoints.txt"]
    # Caused by: java.lang.IllegalArgumentException: not a valid name: org_acme_some=_5b4c6680_3d59_411a_9da0_4b684648a239_fluxc_processor_with_kapt_2ed5a0f4_WARM_UP_1
    # apply-property-resource-change-to = ["fluxc-processor/src/main/resources/wc-wp-api-endpoints.txt"]
}

fluxc-processor-with-annotation-processor {
    title = "FluxC Processor with Annotation Processor"
    tasks = ["assembleDebug"]
    gradle-args = ["--rerun-tasks"]
}
```

</details>

image


2. fc25de28672848a2d918c2f4f4cab8b6b39775b7 + f1e4b62e257b720f127d1b95f63d105fc42c3051

In addition to the above, the unnecessary `publishToS3` configuration got removed from both, the `fluxc-annotations` and `fluxc-processor` modules to simply the overall configuration, plus win some CI time back for us.

PS: Before merging (or reverting) this let's wait to get a response back from @oguzkocer (see interal thread: p1723556208848519-slack-C02KLTL3MKM).

ParaskP7: UPDATE: Having the build succeed with only publishing `fluxc` and `plugins:woocommerce`, I then tried using the `<PR>-<SHA>` combination on the WCAndroid project, this actually quickly failed with the below:

```

* What went wrong:
Execution failed for task ':WooCommerce:dataBindingMergeDependencyArtifactsJalapenoDebug'.
> Could not resolve all files for configuration ':WooCommerce:jalapenoDebugCompileClasspath'.
   > Could not find FluxC:fluxc-annotations:unspecified.
     Required by:
         project :WooCommerce > org.wordpress:fluxc:3075-f1e4b62e257b720f127d1b95f63d105fc42c3051
         project :WooCommerce > org.wordpress.fluxc.plugins:woocommerce:3075-f1e4b62e257b720f127d1b95f63d105fc42c3051
```

It seems that the published artifacts for `fluxc-annotations` and `fluxc-processor` are actually needed, I just don't understand enough about the `s3` configuration to derive the why...

ParaskP7: UPDATE: Actually, maybe this (ffa0b2c315cace54c8dd1d104ee35ba69e44c104) was the configuration I was missing to make the above working. 😅 Testing...

ParaskP7: UPDATE: Nope, the above didn't help and the WCAndroid build is still failing.

FYI: I also noticed that the `import org.wordpress.android.fluxc.annotations.action.Action` is the only `fluxc-annotations` specific import being used directly on the WCAndroid project, just like on the JP/WPAndroid project.

ParaskP7: FYI: Having discussing it with @oguzkocer and @wzieba I am now dropping those 3 commits that were removing the seemingly unnecessary publishing to `s3` of the `fluxc-annotations` and `fluxc-processors` artifacts.

PS: For more info see: p1723556208848519-slack-C02KLTL3MKM

wzieba (approved):

PR #3076, opened 2024-08-09
Fix the 2FA login on the example app

hichamboushaba:
This PR fixes the 2FA login on the example app, this is done by listening for the event `OnTwoFactorAuthStarted` that was added when we added support for passkeys.

Disclaimer: I didn't handle any edge cases, and I'm not very familiar with this part. But given this is just the example app we use for testing, I think the risks are low.

#### Testing
1. Enable 2FA on a WordPress.com account (regular WordPress.com account).
2. Open the example app.
3. Attempt login, and confirm it works without issues.


ParaskP7 (approved):
Works like a charm, thanks for handling this @hichamboushaba ! 🙇 x 🙇 ^ 🙇

PR #3078, opened 2024-08-12
[Login] Fix the logic of triggering an SMS OTP 

hichamboushaba:
As identified in https://github.com/woocommerce/woocommerce-android/issues/12304, we have two issues with SMS 2FA. This PR fixes one of them: fix the broken button to trigger sending an SMS notification, the fix is to indicate that our client supports 2fa by passing the argument `wpcom_supports_2fa` with the request.

For the second part, I will create a separate PR as involves some changes to the login library, and this fix is the more urgent one as it should unblock the users when attempting to login.

### Testing
Please check https://github.com/woocommerce/woocommerce-android/pull/12306


JorgeMucientes (approved):
Great job at finding the culprit of this issue. Tested it and 2fa via SMS now works as expected. 

:shipit:

PR #3079, opened 2024-08-13
[Login] Revert to using the /token endpoint for the authentication

hichamboushaba:
This PR reverts to using the `/token` endpoint for the authentication using WordPress.com instead of the `/wp-login.php?action=login-endpoint`.

The new endpoint is missing a logic to trigger the SMS OTP automatically, and instead defaults to sending a mobile push notification which our apps don't support yet.
You can find more details on this decision here: p1723548882803229/1712217177.026859-slack-C6H8C3G23

### Testing
Check https://github.com/woocommerce/woocommerce-android/pull/12319


ThomazFB (approved):
Tested and approved from https://github.com/woocommerce/woocommerce-android/pull/12319

PR #3082, opened 2024-08-16
[Custom Fields] Add store for CRUD operations

hichamboushaba:
Closes: https://github.com/woocommerce/woocommerce-android/issues/12357 Closes: https://github.com/woocommerce/woocommerce-android/issues/12356

This PR adds a Store to handle the metadata (Custom Fields) CRUD operations, and to allow observing the DB.

#### Testing
Check https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3084


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/metadata/MetaDataParentItemType.kt
```diff
@@ -0,0 +1,6 @@
+package org.wordpress.android.fluxc.model.metadata
+
+enum class MetaDataParentItemType {
+    ORDER,
+    PRODUCT
+}
```
hichamboushaba, line 6: I moved this enum from the DB entity to here, I did this because this will be used for actual business logic

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/metadata/UpdateMetadataRequest.kt
```diff
@@ -0,0 +1,17 @@
...
+    val insertedMetadata: List<WCMetaData> = emptyList(),
+    val updatedMetadata: List<WCMetaData> = emptyList(),
+    val deletedMetadata: List<WCMetaData> = emptyList(),
+) {
+    init {
+        // The ID of inserted metadata is ignored, so to ensure that there is no data loss here,
+        // we require that all inserted metadata have an ID of 0.
+        require(insertedMetadata.all { it.id == 0L }) {
+            "Inserted metadata must have an ID of 0"
+        }
```
hichamboushaba, line 15: This is just a safety check to ensure the field is used correctly (it's aimed for us the developers), the alternative to this was to create a separate class to represent the new items (without an ID), feel free to share your thoughts.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/StripOrderMetaData.kt
```diff
@@ -1,28 +1,14 @@
...
 import javax.inject.Inject
 
-class StripOrderMetaData @Inject internal constructor(private val gson: Gson) {
-    private val type = object : TypeToken<List<WCMetaData>>() {}.type
-
-    operator fun invoke(orderDto: OrderDto): List<WCMetaData> {
-        return parseMetaDataJSON(orderDto.meta_data)
-            ?.filter {
+class StripOrderMetaData @Inject internal constructor() {
+    operator fun invoke(metaData: List<WCMetaData>): List<WCMetaData> {
```
hichamboushaba, line 7: I refactored this function to make using on the new Store easier.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/metadata/UpdateMetadataRequest.kt
JorgeMucientes: Hmm, I understand we want to reuse the `WCMetaData` model here, but I don't fully get why the `id` has to be 0, even if it's ignored in the app.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/metadata/MetaDataRestClient.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.wordpress.android.fluxc.model.metadata.WCMetaData
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooNetwork
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.WooPayload
+import org.wordpress.android.fluxc.utils.toWooPayload
+import javax.inject.Inject
+
+internal class MetaDataRestClient @Inject internal constructor(
+    private val wooNetwork: WooNetwork
+) {
+    suspend fun refreshMetaData(
```
JorgeMucientes: Minor np, is there a reason to call this `refresh` instead of the more common naming `fetchMetaData`. You are just running a `GET` request here, and "refresh" sounds more like you are updating some resource. Wdyt?

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/MetaDataStore.kt
```diff
@@ -0,0 +1,131 @@
...
+                parentItemType = request.parentItemType,
+                metaDataList = it
+            )
+        }
+
+        @Suppress("RedundantUnitExpression")
+        return result.asWooResult { Unit }
+    }
+
+    suspend fun refreshMetaData(
```
JorgeMucientes, line 43: Related to the comment above, in this case, I'd say it makes sense to call it `refresh` since you're updating the database after fetching the new metadata.

JorgeMucientes (approved):
Once again excellent refactor and implementation @hichamboushaba. 

Everything works as expected for CRUD operations over metada. I just left a minor nitpick and a question about something I didn't fully understand. 

Approving :shipit:

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/metadata/UpdateMetadataRequest.kt
hichamboushaba: If a non-zero `id` is supplied, and the one who provided it might be thinking that it would be used as the actual `id` of the resource, so this could lead to some misunderstanding or even we might consider it a data loss since the provided `id` will be lost after executing the request, that's why I added this check.

I know you are AFK today, and I'm going to merge this, but please feel free to continue the discussion, and I can adjust if needed later.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/metadata/MetaDataRestClient.kt
hichamboushaba: Good call, I renamed the RestClient function to `fetch`.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/MetaDataStore.kt
hichamboushaba: Yes, I'm keeping `refresh` just for the `Store` function.

PR #3084, opened 2024-08-16
[Custom Fields] Screen on the example app to test the feature

hichamboushaba:
This PR just adds a screen to the example app to test the changes added in #3082. The PR seems big, but since the code all goes to the example app, I don't think you need to spend too much time reviewing the code.

I opted to add Compose to the example app, if you have any concerns please share them.

### Steps
1. Open the example app.
2. Sign in.
3. Tap on Woo, then select a store.
4. Tap on Orders.
5. Tap on "View Order Custom Fields"
6. Enter the ID of an existing order.
7. On the next screen, confirm the screen loads the custom fields correctly.
8. Make some changes.
9. Save.
10. Confirm the flow works as expected.
11. Repeat 4-10 for Products.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -7,9 +7,11 @@ config:
 complexity:
   LongParameterList:
     ignoreDefaultParameters: true
-    ignoreAnnotated: ['Inject']
+    ignoreAnnotated: ['Inject', 'Composable']
   TooManyFunctions:
     active: false
+  LongMethod:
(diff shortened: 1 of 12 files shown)


JorgeMucientes (approved):
Nice one @hichamboushaba 👏🏼 
It's good to see Compose code in the FluxC example app, thanks for adding it! I was able to test custom fields editing easily with the UI you've added and everything worked well!

:shipit:

PR #3087, opened 2024-08-22
[Blaze] Fix date formatting on the ad forecast endpoint

hichamboushaba:
This applies the same fix we did for the campaign creation request in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3085 to the forecast endpoint too.

I think the testing done on the previous PR is enough to confirm the solution, if you prefer having some explicit testing steps, I can share them.


irfano (approved):
I was able to reproduce the issue and confirm that this fix addresses it. Great job! 👍🏻
I believe this could be worth including as a beta fix, but beta fixes for FluxC are cumbersome and risky. So, I'm merging this into `trunk` for a regular release.

PR #3089, opened 2024-09-04
 Fix order pagination break

atorresveiga:
Part of: https://github.com/woocommerce/woocommerce-android/issues/12491
### Description
Previously, the order list relied on the refresh event to update when changes were detected, leading to inconsistent pagination. This PR addresses the issue by using the invalidate event instead. This ensures that the list is updated only when necessary, improving performance and ensuring a smooth pagination experience.

### Testing 
It is better to test this PR using the [Woo PR](https://github.com/woocommerce/woocommerce-android/pull/12544)

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/InsertOrder.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/InsertOrder.kt
@@ -52,7 +52,7 @@ class InsertOrder @Inject internal constructor(
         )
         // Re-fetch the list only when at least one of the inserted orders has changed
         if (orderChanged) {
-            dispatcher.dispatch(ListActionBuilder.newListRequiresRefreshAction(listTypeIdentifier))
+            dispatcher.dispatch(ListActionBuilder.newListDataInvalidatedAction(listTypeIdentifier))
         }
     }
 }
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/InsertOrder.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/InsertOrder.kt
@@ -30,7 +30,7 @@ class InsertOrder @Inject internal constructor(
                     order,
                     OrdersDaoDecorator.ListUpdateStrategy.SUPPRESS
                 )
-                if (result != OrdersDaoDecorator.UpdateOrderResult.UNCHANGED) {
+                if (result == OrdersDaoDecorator.UpdateOrderResult.UPDATED) {
                     orderChanged = true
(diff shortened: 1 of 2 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/InsertOrder.kt
```diff
@@ -52,7 +52,7 @@ class InsertOrder @Inject internal constructor(
         )
         // Re-fetch the list only when at least one of the inserted orders has changed
         if (orderChanged) {
-            dispatcher.dispatch(ListActionBuilder.newListRequiresRefreshAction(listTypeIdentifier))
+            dispatcher.dispatch(ListActionBuilder.newListDataInvalidatedAction(listTypeIdentifier))
```
malinajirka, line 55: I still couldn't understand how come the order is placed to the correct list when a filter is applied even though we just invalidate the list. The [answer is here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/3c50bc8b12a654742fb6626d23c424f6690206e4/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCOrderStore.kt#L693) - we manually invoke list refresh when an order status is changed - seems like Jirka 5 months ago was smarter than now :P.

malinajirka (approved):
Great job @atorresveiga! It seems these changes work like a charm - I've tested the app on both tablet (OnePlus Android 14) and phone (Pixel 8, Android 15) and I found one minor issue I wouldn't worry about, or at least I wouldn't block this PR on it.

1. Apply `Status: Completed` filter.
2. Refund an order
3. Notice the order status changes to `Refunded` but the order remains in the `Status: Completed` list.

I've tested the following:
- Order Creation - new order is added.
- Changing order status - status gets updated.
- Changing order status with filter applied - status updated and order is placed to the correct filtered list.
- Order Refund - status update but order is not placed to the correct list (the issue mentioned above).
- Collect Payment for order - status updated and order is placed to the correct filtered list.
- Pagination after fresh install - works as expected.
- Pagination when data located in the local database - works as expected.
- Pagination after fresh install, list of ids loaded but order models not loaded, one of the orders change on the backend - works as expected.
- Pagination when data located in the local database and order status changed on the backend - works as expected.



I discovered one more issue that is unrelated to these changes - it's quite interesting as it doesn't seem to be related to the ListStore component - or at least there are no ListAction events fired.
1. Scroll down in the list to at least page 2.
2. Open an order on a phone
3. Press back
4. Notice the list displays loading/shimmering effect on the rows even though they were correctly displayed just a moment ago. 

https://github.com/user-attachments/assets/c70de72b-0e69-4ad9-a364-760feed1324f

PR #3090, opened 2024-09-06
Fix load more

atorresveiga:
Part of: https://github.com/woocommerce/woocommerce-android/issues/12491

This PR fixes an issue that prevented loading more orders (stopped the pagination) when scrolling the orders list after a certain time. 

On the previous version of the orders list screen, we always refreshed the orders list when entering the screen. This behavior ensured that we always had fresh data when displaying the screen. 

The order pagination saves the current list state in the local DB. Because the list state is saved in the local DB, there could be times when the app saved an intermediary state, for example, Fetching the first page, and the app closed before the final state was saved (can-load-more or fetched). To prevent inconsistent states on the `getListState` function, a timeout was added to check for more than 1 minute states. If the state was older than 1 minute, the function returned a `need-to-refresh state.`

> On the previous version of the orders list screen, we always refreshed the orders list when entering the screen. 

We changed the always refreshed orders list with the background tasks project. Now, we rely more on cached data on the local DB, meaning that final states (can-load-more, fetched) can live longer than 1 minute. 

This PR updates the `getListState` logic so the timeout does not affect final states. The plan is that on Woo, we will implement the logic to expire these final states.

### Testing
It is better to test this PR with the [Woo PR](https://github.com/woocommerce/woocommerce-android/pull/12545)

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/store/ListStore.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/store/ListStore.kt
@@ -351,14 +351,17 @@ class ListStore @Inject constructor(
     /**
      * A helper function that returns the [ListState] for the given [ListDescriptor].
      */
-    private fun getListState(listDescriptor: ListDescriptor): ListState {
+     fun getListState(listDescriptor: ListDescriptor): ListState {
         val listModel = listSqlUtils.getList(listDescriptor)
-        return if (listModel != null && !isListStateOutdated(listModel)) {
-            requireNotNull(ListState.values().firstOrNull { it.value == listModel.stateDbValue }) {
+        val currentState = listModel?.let {
+            requireNotNull(ListState.entries.firstOrNull { it.value == listModel.stateDbValue }) {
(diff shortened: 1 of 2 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/store/ListStore.kt
```diff
@@ -351,14 +351,17 @@ class ListStore @Inject constructor(
     /**
      * A helper function that returns the [ListState] for the given [ListDescriptor].
      */
-    private fun getListState(listDescriptor: ListDescriptor): ListState {
+     fun getListState(listDescriptor: ListDescriptor): ListState {
```
malinajirka, line 354: ⚠️⚠️⚠️ @atorresveiga  This code is used by the WPAndroid app as well - have we made sure it works there as expected?

atorresveiga: Thanks for pointing this out @malinajirka. I tested the changes on the WPAndroid app and the pagination keeps working as expected

PR #3096, opened 2024-09-18
Product Password using site credentials

hichamboushaba:
This PR adds support for fetching and updating the product password using WooCommerce REST API.

#### Testing
To be tested with the PR: https://github.com/woocommerce/woocommerce-android/pull/12642


fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
```diff
@@ -2024,6 +2024,35 @@ open class WellSqlConfig : DefaultWellConfig {
...
+                        PURCHASE_NOTE TEXT NOT NULL,MENU_ORDER INTEGER,CATEGORIES TEXT NOT NULL,TAGS TEXT NOT NULL,
+                        IMAGES TEXT NOT NULL,ATTRIBUTES TEXT NOT NULL,VARIATIONS TEXT NOT NULL,DOWNLOADS TEXT NOT NULL,
+                        RELATED_IDS TEXT NOT NULL,CROSS_SELL_IDS TEXT NOT NULL,UPSELL_IDS TEXT NOT NULL,
+                        GROUPED_PRODUCT_IDS TEXT NOT NULL,WEIGHT TEXT NOT NULL,LENGTH TEXT NOT NULL,WIDTH TEXT NOT NULL,
+                        HEIGHT TEXT NOT NULL,METADATA TEXT NOT NULL,BUNDLED_ITEMS TEXT NOT NULL,
+                        COMPOSITE_COMPONENTS TEXT NOT NULL,SPECIAL_STOCK_STATUS TEXT NOT NULL,BUNDLE_MIN_SIZE REAL,
+                        BUNDLE_MAX_SIZE REAL,MIN_ALLOWED_QUANTITY INTEGER,MAX_ALLOWED_QUANTITY INTEGER,
+                        GROUP_OF_QUANTITY INTEGER,COMBINE_VARIATION_QUANTITIES INTEGER,PASSWORD TEXT,IS_SAMPLE_PRODUCT INTEGER
+                        )
+                    """.trimIndent())
```
JorgeMucientes, line 2054: hmm, curious what is the reason of recreatiing the whole table instead of adding just the new column with something like:
```SQL
db.execSQL("ALTER TABLE WCProductModel ADD PASSWORD TEXT")
```

Are there any other changes I'm missing?

JorgeMucientes (approved):
Changes look good and work as expected based on the tests I conducted from the [companion PR](https://github.com/woocommerce/woocommerce-android/pull/12642). 

Just left a non blocking question about recreating the products table.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
hichamboushaba: Just to avoid any data loss, if we keep the previous cache, it will have the `PASSWORD` set to `null`, so if we happen to push some updates to the product, we could cause some data loss.

WDYT?

JorgeMucientes: Hmm, I see. I had not contemplated we could override server values with our cached values. Good point, it seems safer to clear the local db.

PR #3098, opened 2024-09-19
Adds missing provider method for new Objective DAO

JorgeMucientes:
###Description

Just fixing some missing provider method to be able to inject the new `BlazeObjectivesDao`. 

`BlazeObjectivesDao` is not used in WooCommerce `trunk` code yet. So there's no risk of any feature breaking. However this changes are needed to compile WooCommerce app when targeting the latest FluxC `trunk` changeset. So it is mandatory to merge this PR before the next cde freeze process is triggered. 

Nothing to test. 

**Extra context**: I merged [this PR earlier today](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3095). Everything compiled and worked as expected. However, I applied a last minute refactor to FluxC PR, and missed adding the changes added here. While FluxC sample app would still compile, WooCommerce app won't if it targets the lastest FluxC `trunk`.


ThomazFB (approved):
Looks good!

PR #3101, opened 2024-09-24
Revert "[ANR fix] Move UserAgent initialization to a background thread"

hichamboushaba:
In WCAndroid, we are getting some crashes related to WebView usage from multiple processes since 19.7, which was the version where we introduced the change of #3067.

The changes of the linked PR shouldn't cause these crashes, but we want to experiment with reverting them to confirm if there are any unwanted side effects that they caused.

For more information on the crashes, check peaMlT-Tk-p2

### Testing
I think green CI should be enough for this, ~~and we don't need a PR to update FluxC on WCAndroid either, it will be done by the release manager during the code freeze.~~ scratch this, I'll create a PR to update the reference on WCAndroid to make sure a release note about the change is included, I'll do it when this PR is merged.

### DIFF
--- a/example/src/main/java/org/wordpress/android/fluxc/example/di/AppConfigModule.java
+++ b/example/src/main/java/org/wordpress/android/fluxc/example/di/AppConfigModule.java
@@ -15,8 +15,6 @@ import org.wordpress.android.util.AppLog.T;
 
 import java.lang.reflect.Field;
 
-import javax.inject.Singleton;
-
 import dagger.Module;
 import dagger.Provides;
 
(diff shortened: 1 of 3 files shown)


example/src/main/java/org/wordpress/android/fluxc/example/di/AppConfigModule.java
```diff
@@ -43,7 +41,6 @@ public AppSecrets provideAppSecrets() {
     }
 
     @Provides
-    @Singleton
```
hichamboushaba, line 46: I'm leaving the revert as is (which means reverting this part also), if we find out that the revert fixes the issue, I'll create a separate PR to make the UserAgent singleton for the example app, otherwise if we need to bring back the changes, reverting this PR will be easier if we keep all the changes.

kidinov (approved):
LGTM!

PR #3102, opened 2024-09-25
[Custom Fields] Store Json values

### LINKED ISSUE #12696: [Custom Fields] Keep JSON values

hichamboushaba:
Closes https://github.com/woocommerce/woocommerce-android/issues/12696

This PR adds logic to store the JSON custom fields, this is needed because we will allow merchants to see those values in read-only mode in the app.

### Testing
1. You'll need to perform an API request to your website to add some json values:
```
POST {site}/wc/v3/orders/{orderId}

{
    "meta_data": [
        {
            "key": "json_object",
            "value": {
                "Key": "Value"
            }
        },
        {
            "key": "json_array",
            "value": [
                {
                    "Key": "Value"
                }
            ]
        }
    ]
}
```
2. Open the example app and sign in if needed.
3. Tap on Woo, then select the store.
4. Tap on Orders then select the store.
5. Tap on "View custom fields"
6. Confirm that the json values are shown, and not editable from the app.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductMetaData.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductMetaData.kt
@@ -7,7 +7,7 @@ import javax.inject.Inject
 
 class StripProductMetaData @Inject internal constructor() {
     operator fun invoke(metaData: List<WCMetaData>): List<WCMetaData> {
-        return metaData.filter { !it.isJson || SUPPORTED_KEYS.contains(it.key) }
+        return metaData.filter { it.isDisplayable || SUPPORTED_KEYS.contains(it.key) }
     }
 
     companion object {
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/metadata/WCMetaData.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/metadata/WCMetaData.kt
@@ -65,12 +65,6 @@ data class WCMetaData(
         private const val DISPLAY_KEY = "display_key"
         private const val DISPLAY_VALUE = "display_value"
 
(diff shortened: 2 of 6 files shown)


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/StripProductMetaData.kt
```diff
@@ -7,7 +7,7 @@ import javax.inject.Inject
 
 class StripProductMetaData @Inject internal constructor() {
     operator fun invoke(metaData: List<WCMetaData>): List<WCMetaData> {
-        return metaData.filter { !it.isJson || SUPPORTED_KEYS.contains(it.key) }
+        return metaData.filter { it.isDisplayable || SUPPORTED_KEYS.contains(it.key) }
```
hichamboushaba, line 10: Another change of this is that we are not going to cache non displayable items, this is done in both platforms to avoid taking up the phone device for unused stuff.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/metadata/WCMetaData.kt
```diff
@@ -65,12 +65,6 @@ data class WCMetaData(
         private const val DISPLAY_KEY = "display_key"
         private const val DISPLAY_VALUE = "display_value"
 
-        val SUPPORTED_KEYS: Set<String> = buildSet {
-            add(SubscriptionMetadataKeys.SUBSCRIPTION_RENEWAL)
-            add(BundleMetadataKeys.BUNDLED_ITEM_ID)
-            addAll(OrderAttributionInfoKeys.ALL_KEYS)
-        }
```
hichamboushaba, line 72: I moved these to a better location below.

JorgeMucientes (approved):
Nice job @hichamboushaba changes look good and everything worked as expected in my test:

PR #3105, opened 2024-10-03
Make objective field nullable

JorgeMucientes:
Makes objective field nullable to avoid sending `objective = ""` values when attempting to create a Blaze campaign from Woo mobile. Sending `objective = ""` will make the API request fail.


irfano (approved):
LGTM! 👍🏻

PR #3106, opened 2024-10-08
[Woo] Extract subscriptions outside of WCProductModel table

hichamboushaba:
⚠️ Please don't merge yet, this has some breaking changes.

This PR has two main changes:

- It removes the `METADATA` column from the `WCProductModel` table.
- It adds the ability to update the metadata separately when adding or updating a product.

##### Testing
Testing steps are provided in https://github.com/woocommerce/woocommerce-android/pull/12766 and https://github.com/woocommerce/woocommerce-android/pull/12767


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/metadata/MetadataChanges.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import com.google.gson.JsonArray
+import com.google.gson.JsonNull
+import com.google.gson.JsonObject
+
+data class MetadataChanges(
+    val insertedMetadata: List<WCMetaData> = emptyList(),
+    val updatedMetadata: List<WCMetaData> = emptyList(),
+    val deletedMetadataIds: List<Long> = emptyList(),
+) {
```
hichamboushaba, line 11: I extracted this bit to a separate class, this class can now be used in other places when needed (now it's used in WCProductStore, but can be used for orders as well), while the `UpdateMetadataRequest` will still be used for the MetadataStore operations.

My goal of extracting this is to not have to deal with unneeded information for the other usages (I mean `parentItemId` and `parentItemType`).

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/metadata/WCMetaDataValue.kt
```diff
@@ -78,6 +78,16 @@ sealed class WCMetaDataValue {
...
     companion object {
+        operator fun invoke(value: String?): WCMetaDataValue {
+            if (value == null) return StringValue(null)
+
+            return runCatching { JsonParser().parse(value) }
+                .getOrElse { JsonPrimitive(value) }
+                .let { fromJsonElement(it) }
+        }
+        operator fun invoke(value: Number): WCMetaDataValue = NumberValue(value)
+        operator fun invoke(value: Boolean): WCMetaDataValue = BooleanValue(value)
```
hichamboushaba, line 89: Just a nicer way to instantiate the values instead of the `fromRawString` version.

fluxc/src/main/java/org/wordpress/android/fluxc/persistence/WellSqlConfig.kt
```diff
@@ -2053,6 +2053,66 @@ open class WellSqlConfig : DefaultWellConfig {
                         )
                     """.trimIndent())
                 }
+
+                204 -> migrateAddOn(ADDON_WOOCOMMERCE, version) {
+                    // Drop the column METADATA from WCProductModel
+                    // Since SQLite version varies accross devices, we can't rely on the support of DROP COLUMN,
```
JorgeMucientes: Well spotted as most Android versions won't support alter table -> drop column operation. It seems this was added in SQLite `3.35.0` and it was in [Android API 34](https://developer.android.com/reference/android/database/sqlite/package-summary) when SQLite was updated to a version > `3.35.0`


Super minor np, typo in `across` word

JorgeMucientes (approved):
Well done LGTM!

PR #3109, opened 2024-11-15
[Media] Migrate WPCom media upload to `/wp/v2` endpoints

hichamboushaba:
Part of https://github.com/woocommerce/woocommerce-android/issues/12836

This PR updates the handling of the media operations (upload and fetch) to the user the `/wp/v2` endpoints instead of the legacy `/v1.1` ones, this is needed to fix upload problems for sites with XMLRPC issues.

**Note:** Since WCAndroid is the only client app using FluxC now, the change is applied without any conditions.

For testing, please check: https://github.com/woocommerce/woocommerce-android/pull/12938

### DIFF
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
@@ -216,7 +216,7 @@ abstract class BaseWPV2MediaRestClient constructor(
             params["offset"] = offset.toString()
         }
         if (mimeType != null) {
-            params["mime_type"] = mimeType.value
+            params["media_type"] = mimeType.value
         }
         val response = executeGetGsonRequest(
             site,
--- a/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
+++ b/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
@@ -12,11 +12,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onStart
-import okhttp3.Call
-import okhttp3.Callback
-import okhttp3.OkHttpClient
-import okhttp3.Request
-import okhttp3.Response
+import okhttp3.*
 import org.json.JSONException
 import org.json.JSONObject
 import org.wordpress.android.fluxc.Dispatcher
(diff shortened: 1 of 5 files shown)


fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
```diff
@@ -107,25 +106,15 @@ abstract class BaseWPV2MediaRestClient constructor(
...
         return callbackFlow {
             val url = WPAPI.media.getFullUrl(site)
             val body = WPRestUploadRequestBody(media) { media, progress ->
-                if (!isClosedForSend) {
-                    val payload = ProgressPayload(media, progress, false, null)
-                    try {
-                        trySend(payload).isSuccess
-                    } catch (e: CancellationException) {
-                        // Do nothing (the flow has been cancelled)
-                    }
```
hichamboushaba, line 127: I removed these parts about catching the CancellationException, I was the one who added these, but they are not useful, as the exception is automatically ignored by the Coroutine's exception handler.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
```diff
@@ -107,25 +106,15 @@ abstract class BaseWPV2MediaRestClient constructor(
...
-                // Do nothing (the flow has been cancelled)
-            }
+            trySendBlocking(payload)
+            close()
         }
 
         return callbackFlow {
             val url = WPAPI.media.getFullUrl(site)
             val body = WPRestUploadRequestBody(media) { media, progress ->
-                if (!isClosedForSend) {
```
hichamboushaba, line 121: I also removed all statements of `isClosedForSend`, it's marked as delicate API, and it's not needed, as now we use `trySend` instead of the old `offer` API.

fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpapi/media/BaseWPV2MediaRestClient.kt
```diff
@@ -234,7 +216,7 @@ abstract class BaseWPV2MediaRestClient constructor(
             params["offset"] = offset.toString()
         }
         if (mimeType != null) {
-            params["mime_type"] = mimeType.value
+            params["media_type"] = mimeType.value
```
hichamboushaba, line 219: Working on this PR identified a bug with our previous implementation, when I worked on this in the first time, I based my implementation on the WPCom one and then used the `mime_type` argument, but it turns out it's not working the same between the two APIs:
- In WPCom, there is only `mime_type`, and supports both a fully qualified type, or just the base type (`image/jpeg` is fully qualified, while `image` is the base type)
- In WordPress Core, we [have](https://developer.wordpress.org/rest-api/reference/media/) two arguments, `media_type` which supports the base type, and `mime_type` which requires the fully qualified type, so when we passed the base type here, the filtering wasn't working, which meant that in the WordPress Media Library screen we would list non-image files.

I believe the same issue is happening on iOS as well ([here](https://github.com/woocommerce/woocommerce-ios/blame/6448eef3dfc36dcdde49ce021fab45cbff42b4c5/Networking/Networking/Remote/MediaRemote.swift#L95)), I'll create an issue for them to fix it.

JorgeMucientes (approved):
Great job @hichamboushaba I tested it on the Woo app with XML RPC disabled/enabled and everything worked well

PR #3110, opened 2024-11-18
Updates Authenticator file to use API 1.3 for sign up links

JorgeMucientes:
Updates Authenticator.java file to use latest API version when sending login and signup links. 

To be tested in this PR: https://github.com/woocommerce/woocommerce-android/pull/12935


hichamboushaba (approved):
Looks good, thanks @JorgeMucientes

PR #3111, opened 2024-11-21
[Woo] Add the product global unique id property

toupper:
With this PR we add support for the product's global unique identifier. Specifically we:

- Add new columns in the Product and Product Variations tables of the well SQL DB
- Migrate DB version
- Add support for the new value in the API Response classes.
- Add new value on the update requests.


ThomazFB (approved):
Looks good, César! Congratulation on expanding your Android horizons far into the Database area 😄

PR #3112, opened 2024-11-22
Fix issue of deserializing `/wp/v2/media` response when having an empty `media_details.sizes` property

hichamboushaba:
Part of fixing https://github.com/woocommerce/woocommerce-android/issues/12989

This PR adds a fix to the issue of parsing the `/wp/v2/media` response when `media_details.sizes` contains an array instead of the expected object type.

To test: https://github.com/woocommerce/woocommerce-android/pull/12990


JorgeMucientes (approved):
Tested it in the linked PR for testing and everything worked as expected.

PR #3113, opened 2024-11-25
Order form is dropping second last names unexpectedly

### LINKED ISSUE #11236: Order form is dropping second last names unexpectedly
**Describe the Bug**  
When creating a user that has more than one last name, which is common in many countries, and that user is loaded onto a new order, only one of the last names is pulled. If you edit the user, only the last name was saved. The question is whether this is by design or a bug.

For additional information, reference this ticket: [https://a8c.zendesk.com/agent/tickets/7895292](https://a8c.zendesk.com/agent/tickets/7895292) and Slack [discussion](https://a8c.slack.com/archives/C06NC969ZF0/p1711115213546839).  It's not clear why this issue has not been encountered before, suggesting it may not have been reported by customers.

**To Reproduce**  
1. Create a user with more than one last name.
2. Load that user onto a new order.
3. Only one of the last names appears.
4. Edit the user.
5. Only the last name was saved.

**Screenshots**

Screenshot_20240322-094247
Screenshot_20240322-094308
Screenshot_20240322-094329

**Expected behavior**  
When creating users with more than one last name and loading those users onto new orders, or editing them, all last names should accurately be saved and displayed.

- [ ] This bug occurs with the latest version of WooCommerce Android.  
- [ ] This bug happens consistently using the steps above.

AnirudhBhat:
closes [#11236](https://github.com/woocommerce/woocommerce-android/issues/11236)

This PR fixes the logic to retrieve first name and last name given customer full name.

**Before this PR**, we were considering first word to be the first name and last word to be the last name. This caused problem when there were multiple words in customers name. 
`Full Name` : `Firstname and a very long last name`
`firstname` : `Firstname`
 `lastname` : `name`

**After this PR**, we are considering first word to be the first name and all other words as last names
`Full Name` : `Firstname and a very long last name`
`firstname` : `Firstname`
`lastname` : `and a very long last name`

https://github.com/user-attachments/assets/a4f62ddf-13d3-4e0d-ae1c-d5143a024c02

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
@@ -286,9 +286,22 @@ class WCCustomerMapperTest {
         val result = mapper.mapToModel(site, customerDTO)
 
         // then
-        with(result) {
-            assertEquals("firstname", firstName)
-        }
+        assertEquals("firstname", result.firstName)
+    }
+
+    @Test
+    fun `given customer name as null, then first name returns empty string`() {
+        // given
+        val siteId = 23
+        val site = SiteModel().apply { id = siteId }
+
+        val customerDTO = CustomerFromAnalyticsDTO(name = null)
+
+        // when
+        val result = mapper.mapToModel(site, customerDTO)
+
+        // then
+        assertEquals("", result.firstName)
     }
 
     @Test
--- a/example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
@@ -305,6 +305,21 @@ class WCCustomerMapperTest {
     }
 
     @Test
+    fun `given customer name with no space, then first name returns full string`() {
+        // given
+        val siteId = 23
+        val site = SiteModel().apply { id = siteId }
+
+        val customerDTO = CustomerFromAnalyticsDTO(name = "firstnamelastname")
+
+        // when
+        val result = mapper.mapToModel(site, customerDTO)
+
+        // then
+        assertEquals("firstnamelastname", result.firstName)
+    }
+
+    @Test
     fun `given customer name, then last name is properly assigned`() {
         // given
         val siteId = 23
--- a/example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
@@ -348,4 +348,19 @@ class WCCustomerMapperTest {
         // then
         assertEquals("", result.lastName)
     }
+
+    @Test
+    fun `given customer name has no space, then last name returns empty string`() {
+        // given
+        val siteId = 23
+        val site = SiteModel().apply { id = siteId }
+
+        val customerDTO = CustomerFromAnalyticsDTO(name = "firstnamelastname")
+
+        // when
+        val result = mapper.mapToModel(site, customerDTO)
+
+        // then
+        assertEquals("", result.lastName)
+    }
 }
--- a/example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
@@ -280,7 +280,7 @@ class WCCustomerMapperTest {
         val siteId = 23
         val site = SiteModel().apply { id = siteId }
 
-        val customerDTO = CustomerFromAnalyticsDTO(name = "firstname with a very long last name")
+        val customerDTO = CustomerFromAnalyticsDTO(name = "firstname lastname")
 
         // when
         val result = mapper.mapToModel(site, customerDTO)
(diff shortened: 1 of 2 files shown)


example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
```diff
@@ -273,4 +273,124 @@ class WCCustomerMapperTest {
         // THEN
         assertThat(result.isPayingCustomer).isEqualTo(false)
     }
+
+    @Test
+    fun `given customer name, then first name is properly assigned`() {
```
kidinov, line 278: np: in all tests, maybe assert both first and last names, because they kinda related and deducted from one string?

example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
```diff
@@ -273,4 +273,124 @@ class WCCustomerMapperTest {
...
+        // given
```
kidinov: np: the comments are upper case in the other tests

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/customer/WCCustomerMapper.kt
```diff
@@ -117,6 +117,7 @@ class WCCustomerMapper @Inject constructor() {
         }
     }
 
-    private fun String?.firstNameFromName() = this?.split(" ")?.firstOrNull() ?: ""
-    private fun String?.lastNameFromName() = this?.split(" ")?.lastOrNull() ?: ""
+    private fun String?.firstNameFromName() = this?.substringBefore(" ") ?: ""
+
+    private fun String?.lastNameFromName() = this?.substringAfter(" ", "") ?: ""
```
kidinov: Does this handle extra spaces correctly? I wrote a test and it doesn't pass 

```
    @Test
    fun `given customer name has multiple spaces, then last name returns proper string`() {
        // GIVEN
        val siteId = 23
        val site = SiteModel().apply { id = siteId }

        val customerDTO = CustomerFromAnalyticsDTO(name = "firstname   and a very long last name")

        // WHEN
        val result = mapper.mapToModel(site, customerDTO)

        // THEN
        assertEquals("and a very long last name", result.lastName)
        assertEquals("firstname", result.firstName)
    }
```

kidinov (commented):
@AnirudhBhat, thanks for the improvement! I left a couple of nitpicks and one more potential bug. Please take a look!

example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
AnirudhBhat: Done: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3113/commits/8f338d2c450d468df5b1de37069c838d27dd1b23

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/customer/WCCustomerMapper.kt
AnirudhBhat: Good catch! I've fixed this
1. Multiple in-between spaces
2. Leading spaces
3. Trailing spaces

https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3113/commits/3b5294ea5210ea7edd5185f67fba409c483cbf89

example/src/test/java/org/wordpress/android/fluxc/model/customer/WCCustomerMapperTest.kt
AnirudhBhat: I agree, my reasoning for splitting the test cases is mainly to understand why the test fails in future. There must be only one reason for the failure of tests IMHO. If we combine both, we need to see if the test failed because of logical error in the extraction of first name or last name which could be avoided if the test asserts just one thing.

kidinov: In my opinion, the one reason here - "if correct logic used in splitting the string". The current implementation, which consists of two private extensions, should not affect the test's logic. Currently, the test does not validate whether the splitter is functioning correctly; instead, it only checks that the first word is designated as the first name.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/model/customer/WCCustomerMapper.kt
```diff
@@ -117,7 +117,11 @@ class WCCustomerMapper @Inject constructor() {
         }
     }
 
-    private fun String?.firstNameFromName() = this?.substringBefore(" ") ?: ""
+    // Please refer WCCustomerMapperTest file which serves as documentation of how this function behaves.
+    private fun String?.firstNameFromName(): String =
```
kidinov, line 121: Wouldn't it do the trick in both cases? 

`.substringBefore(" ").trim() ?: ""`

You implementation works, but both (maybe) slow and not easy to read

AnirudhBhat: This wouldn't work as it fails for names that has leading spaces, multiple spaces in between.

For example: ` First Name`:
**.substringBefore(" "):** Returns "" because the first substring before a space is empty due to leading spaces.
**.trim():** Has no effect since the result is already an empty string.
Result: **""**

kidinov: Considering that we run this code 6 times per 1 customer mapped I'd propose something simple like that:

```
    private fun String?.firstNameFromName(): String =
        this?.trim()?.substringBefore(' ')?.trim().orEmpty()

    private fun String?.lastNameFromName(): String =
        this?.trim()?.substringAfter(' ', "")?.trim().orEmpty()
```

It won't remove spaces inside of the first and last name, but IMO, its not what we even want and we should leave it up to the users and the way they want to write their names

AnirudhBhat: Done: https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3113/commits/2aca0010300896a4a2dc56e4bbc939d20c8f189a

kidinov (approved):
LGTM!

PR #3114, opened 2024-11-27
[Woo POS][non-simple products] Remove empty filters condition from deleting stale products from database

AnirudhBhat:
### Please review the corresponding [WooCommerce PR](https://github.com/woocommerce/woocommerce-android/pull/13018) after reviewing and merging this one

This PR removes empty filter condition which was one among few conditions we were checking before removing stale products from the database before inserting new ones fetched from the remote.

This was done since for POS, when we fetch products, there always will be filter applied for products (simple + variable) unless we support all types of products in POS. Since from POS, we won't be having an empty filter, the stale products in the database weren't being removed.

### DIFF
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
@@ -1604,8 +1604,7 @@ class WCProductStore @Inject constructor(
                     if (forceRefresh &&
                         offset == 0 &&
                         includedProductIds.isEmpty() &&
-                        excludedProductIds.isEmpty() &&
-                        filterOptions.isEmpty()
+                        excludedProductIds.isEmpty()
                     ) {
                         productStorageHelper.deleteProductsForSite(site)
                     }


plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
```diff
@@ -1604,8 +1604,7 @@ class WCProductStore @Inject constructor(
                     if (forceRefresh &&
                         offset == 0 &&
                         includedProductIds.isEmpty() &&
-                        excludedProductIds.isEmpty() &&
-                        filterOptions.isEmpty()
+                        excludedProductIds.isEmpty()
```
samiuelson, line 1607: ❓ Does it affect the Product screen in store management?

samiuelson (approved):
LGTM. I left one question.

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/store/WCProductStore.kt
AnirudhBhat: Yes and no. 

Yes, because we have removed `filterOptions`.isEmpty() condition that was a requirement earlier before we deleted products. 

No, because AFAIU, there will not be any behavioural changes in the products screen because of this. When we apply any filter on the products screen and pull to refresh, we delete and then upsert products now instead of just upserting product.

PR #3119, opened 2024-12-12
Woo: Part one for orders/batch endpoint

hafizrahman:
Part of https://github.com/woocommerce/woocommerce-android/issues/13115

Working on adding the endpoint for `/wc/v3/orders/batch` but in parts. Documentation in pe5sF9-3kb-p2

In this first part I'm focusing on the API response class and making sure it's correct first.

Actual API call and store functions, etc, will be added in a later PR.

### Discussion:

According to the documentation https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#batch-update-orders

The endpoint will return an object of three fields: "create", "update", and "delete". This is because the batch order action can handle all three actions in one go.

However, for our current purpose, we only want to deal with "update" batch action. This is why `BatchOrderApiResponse` now only has an "update" field.

Furthermore, the "update" field returns an array that contains either (and can be a mix of):
- success case, where it will return a complete Order object
- failure case, where it will return an object with "id" and "error" fields.

Please check `orders-batch.json` for an example of this.

Because of these two differences, I created `OrderResponse` to support either `Success` or `Error`.

I think this architecture makes sense. Later on the handling for the response can check whether it's of type `OrderResponse.Success` or `OrderResponse.Error,` and act accordingly (e.g: for updating successful case in the DB, or displaying success/error state). But, let me know if there's a way to improve this.

### To test:

1. Ensure `testDeserializeBatchOrderResponse()` passes.

### DIFF
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
@@ -1,10 +1,12 @@
 package org.wordpress.android.fluxc.wc.order
 
 import com.google.gson.Gson
+import com.google.gson.GsonBuilder
 import com.google.gson.reflect.TypeToken
 import org.junit.Test
 import org.wordpress.android.fluxc.UnitTestUtils
 import org.wordpress.android.fluxc.model.order.ShippingLine
+import org.wordpress.android.fluxc.network.rest.wpcom.wc.order.BatchOrderApiResponse
 import kotlin.test.assertEquals
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
@@ -142,5 +144,32 @@ class OrderEntityTest {
 
         assertEquals("Flat Rate Shipping", shippingLinesList[0].methodTitle)
         assertEquals("Local Pickup Shipping", shippingLinesList[1].methodTitle)
+    }
+
+    @Test
+    fun testDeserializeBatchOrderResponse() {
+        val testGson = GsonBuilder()
+            .registerTypeAdapter(
+                BatchOrderApiResponse.OrderResponse::class.java,
+                BatchOrderApiResponse.OrderResponseDeserializer())
+            .create()
+
+        val batchOrderJson = UnitTestUtils.getStringFromResourceFile(
+            this.javaClass, "wc/orders-batch.json"
+        )
+
+        val response = testGson.fromJson(batchOrderJson, BatchOrderApiResponse::class.java)
+        val orders = response.update
+
+        assertEquals(2, orders.size)
+
+        val firstOrder = orders[0] as BatchOrderApiResponse.OrderResponse.Success
+        assertEquals(1032L, firstOrder.order.id)
+        assertEquals("224.00", firstOrder.order.total)
+
+        val secondOrder = orders[1] as BatchOrderApiResponse.OrderResponse.Error
+        assertEquals(525L, secondOrder.id)
+        assertEquals("woocommerce_rest_shop_order_invalid_id", secondOrder.error.code)
+        assertEquals(400, secondOrder.error.data.status)
     }
 }
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
@@ -149,9 +149,6 @@ class OrderEntityTest {
     @Test
     fun testDeserializeBatchOrderResponse() {
         val testGson = GsonBuilder()
-            .registerTypeAdapter(
-                BatchOrderApiResponse.OrderResponse::class.java,
-                BatchOrderApiResponse.OrderResponseDeserializer())
             .create()
 
         val batchOrderJson = UnitTestUtils.getStringFromResourceFile(
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
@@ -145,28 +145,4 @@ class OrderEntityTest {
         assertEquals("Flat Rate Shipping", shippingLinesList[0].methodTitle)
         assertEquals("Local Pickup Shipping", shippingLinesList[1].methodTitle)
     }
-
-    @Test
-    fun testDeserializeBatchOrderResponse() {
-        val testGson = GsonBuilder()
-            .create()
-
-        val batchOrderJson = UnitTestUtils.getStringFromResourceFile(
-            this.javaClass, "wc/orders-batch.json"
-        )
-
-        val response = testGson.fromJson(batchOrderJson, BatchOrderApiResponse::class.java)
-        val orders = response.update
-
-        assertEquals(2, orders.size)
-
-        val firstOrder = orders[0] as BatchOrderApiResponse.OrderResponse.Success
-        assertEquals(1032L, firstOrder.order.id)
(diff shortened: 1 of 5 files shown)


example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
```diff
@@ -143,4 +145,31 @@ class OrderEntityTest {
         assertEquals("Flat Rate Shipping", shippingLinesList[0].methodTitle)
         assertEquals("Local Pickup Shipping", shippingLinesList[1].methodTitle)
     }
+
+    @Test
+    fun testDeserializeBatchOrderResponse() {
+        val testGson = GsonBuilder()
+            .registerTypeAdapter(
+                BatchOrderApiResponse.OrderResponse::class.java,
+                BatchOrderApiResponse.OrderResponseDeserializer())
```
hichamboushaba: Suggestion: Instead of having to register the type adapter manually, we can use the `@TypeAdapter` annotation directly, this is something that recent (relatively 😄) versions of Gson added, this patch would show how it can be used:

```patch
Index: plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/BatchOrderApiResponse.kt
IDEA additional info:
<+>UTF-8
diff --git a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/BatchOrderApiResponse.kt b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/BatchOrderApiResponse.kt
--- a/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/BatchOrderApiResponse.kt	(revision 3c423f4b9646625298b72fbc52b981b47a7484c6)
+++ b/plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/BatchOrderApiResponse.kt	(date 1734005033319)
@@ -3,12 +3,14 @@
 import com.google.gson.JsonDeserializationContext
 import com.google.gson.JsonDeserializer
 import com.google.gson.JsonElement
 import java.lang.reflect.Type
 import org.wordpress.android.fluxc.network.Response
 
 data class BatchOrderApiResponse(
     val update: List<OrderResponse>
 ) : Response {
+    @JsonAdapter(OrderResponseDeserializer::class)
     sealed class OrderResponse {
         data class Success(
             val order: OrderDto
@@ -30,7 +32,7 @@
         val status: Int
     )
 
-    class OrderResponseDeserializer : JsonDeserializer<OrderResponse> {
+    private class OrderResponseDeserializer : JsonDeserializer<OrderResponse> {
         override fun deserialize(
             json: JsonElement,
             typeOfT: Type,
Index: example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
IDEA additional info:
<+>UTF-8
diff --git a/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt b/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
--- a/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt	(revision 3c423f4b9646625298b72fbc52b981b47a7484c6)
+++ b/example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt	(date 1734004691032)
@@ -149,9 +149,6 @@
     @Test
     fun testDeserializeBatchOrderResponse() {
         val testGson = GsonBuilder()
-            .registerTypeAdapter(
-                BatchOrderApiResponse.OrderResponse::class.java,
-                BatchOrderApiResponse.OrderResponseDeserializer())
             .create()
 
         val batchOrderJson = UnitTestUtils.getStringFromResourceFile(
```

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/BatchOrderApiResponse.kt
```diff
@@ -0,0 +1,53 @@
+package org.wordpress.android.fluxc.network.rest.wpcom.wc.order
+
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
+import com.google.gson.JsonElement
+import java.lang.reflect.Type
+import org.wordpress.android.fluxc.network.Response
+
+data class BatchOrderApiResponse(
```
hichamboushaba, line 10: minor, I would make this `internal`, I know we haven't been doing it often in the past, but I think it's still something we should consider for new features, as it would reduce leaking the networking models to the upper layers.

example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
```diff
@@ -143,4 +145,31 @@ class OrderEntityTest {
         assertEquals("Flat Rate Shipping", shippingLinesList[0].methodTitle)
         assertEquals("Local Pickup Shipping", shippingLinesList[1].methodTitle)
     }
+
+    @Test
+    fun testDeserializeBatchOrderResponse() {
```
hichamboushaba: Minor, I think this test would better be in a separate file, `OrderEntityTest` is originally about `OrderEntity` logic, which doesn't apply here, WDYT?

hichamboushaba (approved):
Nice work @hafizrahman 👏, I left some comments, but nothing is blocking, so I'm pre-approving.

example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
hafizrahman: Makes sense, moved in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3119/commits/6e2cdef91ec341072f91c5568125115ba42ef6e3 👍🏼

example/src/test/java/org/wordpress/android/fluxc/wc/order/OrderEntityTest.kt
hafizrahman: Thanks for mentioning that! Applied in https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3119/commits/254cf5cc46aaa810fad33c172580151105ffdde3

plugins/woocommerce/src/main/kotlin/org/wordpress/android/fluxc/network/rest/wpcom/wc/order/BatchOrderApiResponse.kt
hafizrahman: This makes sense. However when I do so after the last commit, [6e2cdef](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/3119/commits/6e2cdef91ec341072f91c5568125115ba42ef6e3), the unit test is unable to see this class, even though the package location is now the same. Still see the same issue after adding `@VisibleForTesting`, too.

Not sure yet why this happens, do you know? I'll merge this PR for now to continue work and add this change later.

hichamboushaba: This is because the test is defined in another module (`example`), I think it would be better to move the test to the same module [`plugins/woocommerce`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/tree/trunk/plugins/woocommerce/src/test/)

